PR #11, opened 2016-01-20
KFSPTS-4771: Added ability to send warning emails when XML-loaded PDP payments reference inactive vendors.

cah292:
It is possible to send PDP payments to inactive vendors via the payment XML files; however, that does not necessarily mean the payments should be blocked. For now, this feature just sends a warning email to the customer and a pre-defined list of other recipients, to warn them of cases where payments were issued to inactive vendors.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
@@ -6,13 +6,17 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.HashSet;
 import java.util.List;
+
+import javax.mail.MessagingException;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.pdp.businessobject.Batch;
 import org.kuali.kfs.pdp.businessobject.CustomerProfile;
 import org.kuali.kfs.pdp.businessobject.LoadPaymentStatus;
+import org.kuali.kfs.pdp.businessobject.PaymentDetail;
 import org.kuali.kfs.pdp.businessobject.PaymentFileLoad;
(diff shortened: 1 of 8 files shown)


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -184,5 +200,101 @@ private PaymentFileLoad getCustomerProfileFromUnparsableFile(String incomingFile
...
+     * Checks whether any of the batch load's payment groups reference inactive vendors,
+     * and sends warning emails appropriately if so.
+     * 
+     * @param paymentGroups The payment groups that were loaded.
+     * @param customer The customer's profile.
+     */
+    private void checkForInactiveVendors(List<PaymentGroup> paymentGroups, CustomerProfile customer) {
+        final int MESSAGE_START_SIZE = 300;
+        StringBuilder inactiveVendorsMessage = new StringBuilder(MESSAGE_START_SIZE);
+        VendorService vendorService = SpringContext.getBean(VendorService.class);
```
bhutchinson: Is it possible to inject this service (and the other two services that are used below)? That would be consistent with the Kuali Financials coding standards (see: https://kualico.atlassian.net/wiki/x/WoAZ#JavaCodingStandards-InjectAllServices).

cah292: Yes, that should be possible, just some Spring bean edits. I will make that change for consistency.

nkimble:
How is the KFS system parameter that holds the email addresses going to be populated?  
Is there going to be SQL run to pre-populate that value or are the users going to have to manually create that parameter?

cah292:
I was planning to populate the initial parameter value with SQL. However, I'm still awaiting word from the functionals on what address(es) they want to use, so I don't have the final SQL ready yet. (I've been testing with a different value locally.)

cah292:
Due to some issues that arose with the interactive rebase, I'm going to close this pull request without merging and then create a new one. I may delete and recreate the relevant code branch as well.

PR #12, opened 2016-01-20
KFSPTS-4771: Added ability to send warning emails when XML-loaded PDP payments reference inactive vendors.

cah292:
This is an updated version of today's earlier KFSPTS-4771 pull request, which now implements Bryan's suggestions and contains only the files that were actually updated.


nkimble:
Code looks good but the new Parameter needs to be setup in the KFS-DEV database before the code is merged into the develop branch to ensure all pieces are in place for this issue.

cah292:
The parameter has been added to DEV with a temporary value, as per Cathy's comment on the KFSPTS-4990 sub-task. If everything looks good, then this can be merged to develop.

nkimble:
Parameter is setup in dev and looks good.

PR #13, opened 2016-01-27
KFSPTS-3959: Added new Organization Global document.

cah292:
This change adds a new Organization Global document for performing bulk org updates from a single maintenance document. It currently allows bulk updates to just the following fields: Org Manager, Physical Campus Code, and all address fields (Address Lines 1 and 2, City, State, Zip, and Country).

Due to the updatable org fields being optional on the global doc, some additional validation was added to help prevent incorrect bulk changes. For example, when changing zip/postal code or country code, both must be specified to prevent users from populating invalid zip-and-country combinations. (As with the regular Organization doc, the city and state fields are uneditable and will be auto-updated based on zip and country.)

Some new SQL and a new workflow doc type XML file are required for this change. I have already applied the SQL to DEV, so it is safe to merge this code if the review doesn't find any issues. I will ingest the new doc type after the code has been merged.


db483:
Looks good.

PR #14, opened 2016-01-29
[KFSPTS-4390] Closed account errors that occur in poster process should go to continuation account

kyleoliveira:


src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
```diff
@@ -0,0 +1,1306 @@
...
+                String warningMessage = MessageFormat.format(warningMessagePattern, errorFieldName, subAccountBOLabel, subAccountValue, accountBOLabel, accountValue);
+                reportWriterService.writeError(et, new Message(warningMessage, Message.TYPE_WARNING));
+                subAccountOK = false;
+            }
+
+            if (subAccountOK) {
+                IndirectCostRecoveryGenerationMetadata metadata = new IndirectCostRecoveryGenerationMetadata(a21SubAccount.getIndirectCostRecoveryTypeCode(),
+                        a21SubAccount.getFinancialIcrSeriesIdentifier());
+
+                List<CuIndirectCostRecoveryAccountDistributionMetadata> icrAccountList = metadata.getAccountLists();
```
db483: The PosterServiceImpl has already been customized for Cornell in edu.cornell.kfs.gl.batch.service.impl.CuPosterServiceImpl. Is it possible to override the retrieveSubAccountIndirectCostRecoveryMetadata method in CuPosterServiceImpl instead of bringing in the whole class?

src/main/java/org/kuali/kfs/gl/batch/service/impl/IndirectCostRecoveryGenerationMetadata.java
```diff
@@ -0,0 +1,71 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.gl.batch.service.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import edu.cornell.kfs.gl.batch.service.impl.CuIndirectCostRecoveryAccountDistributionMetadata;
+
+public class IndirectCostRecoveryGenerationMetadata {
```
db483: Can we extend this class and add Cornell customizations? Then call the extended class in retrieveSubAccountIndirectCostRecoveryMetadata and retrieveAccountIndirectCostRecoveryMetadata methods that we can override in CuPosterServiceImpl?

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
```diff
@@ -0,0 +1,1306 @@
...
+    }
+
+
+    protected IndirectCostRecoveryGenerationMetadata retrieveAccountIndirectCostRecoveryMetadata(ExpenditureTransaction et) throws Exception {
+        Account account = et.getAccount();
+
+        IndirectCostRecoveryGenerationMetadata metadata = new IndirectCostRecoveryGenerationMetadata(account.getAcctIndirectCostRcvyTypeCd(),
+                account.getFinancialIcrSeriesIdentifier());
+
+        List<CuIndirectCostRecoveryAccountDistributionMetadata> icrAccountList = metadata.getAccountLists();
```
db483: See comment on retrieveSubAccountIndirectCostRecoveryMetadata.

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
kyleoliveira: `generateTransactionsBySymbol` is private, so we'd at least need to include the class for that, right?

src/main/java/org/kuali/kfs/gl/batch/service/impl/IndirectCostRecoveryGenerationMetadata.java
kyleoliveira: `List<IndirectCostRecoveryAccountDistributionMetadata> accountLists` is private. Would I create a new private class and override the the getters/setters then? I didn't think that would be the best course of action because accountLists is used in various class methods

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
db483: true about generateTransactionsBySymbol, missed this one

src/main/java/org/kuali/kfs/gl/batch/service/impl/IndirectCostRecoveryGenerationMetadata.java
db483: good point

kyleoliveira:
I think I can move the two protected methods (`IndirectCostRecoveryGenerationMetadata retrieveAccountIndirectCostRecoveryMetadata(ExpenditureTransaction et)` and `IndirectCostRecoveryGenerationMetadata retrieveSubAccountIndirectCostRecoveryMetadata(ExpenditureTransaction et)`) into the CU class, if that's worthwhile.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuIndirectCostRecoveryAccountDistributionMetadata.java
```diff
@@ -0,0 +1,75 @@
...
+      super(icrAccount);
+
+      // Do what super does if the ICRA is null or not closed
+      if(ObjectUtils.isNotNull(icrAccount) && 
+         ObjectUtils.isNotNull(icrAccount.getIndirectCostRecoveryAccount()) && 
+         icrAccount.getIndirectCostRecoveryAccount().isClosed()) {
+        Account replacementICRA = icrAccount.getIndirectCostRecoveryAccount();
+        int i = 0;
+  
+        while((i < 10) && ((replacementICRA != null) && replacementICRA.isClosed())) {
```
db483: We usually use ObjectUtils.isNotNull to check for nulls.
10 looks like a magic number, maybe add a comment on why 10 was chosen?

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
bhutchinson: I guess my question is did you really have to change that method? Looks like the only change was to the type from IndirectCostRecoveryAccountDistributionMetadata to CuIndirectCostRecoveryAccountDistributionMetadata. Is that really necessary? If the passed in collection contains CuIndirectCostRecoveryAccountDistributionMetadata objects wouldn't the correct version of the method be called even if the type is specified as IndirectCostRecoveryAccountDistributionMetadata?

kyleoliveira: That seems reasonable. I think we'll still need to include the overridden version of IndirectCostRecoveryAccountDistributionMetadata, but this should make it so we don't need to override PosterServiceImpl.
I'll give it a shot.

kyleoliveira:
Consider this PR on-hold until at least Monday. It's not going to be ready to go back into DEV today at this point, and I'll be able to spend more time on development work then. I'll bug @db483 when I've got updated stuff to look at.

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
bhutchinson: Great, thanks!

kyleoliveira: I was able to do this :)

kyleoliveira:
@db483 This is ready for review again.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuIndirectCostRecoveryAccountDistributionMetadata.java
```diff
@@ -0,0 +1,76 @@
...
+
+      // Do what super does if the ICRA is null or not closed
+      if(ObjectUtils.isNotNull(icrAccount) && 
+         ObjectUtils.isNotNull(icrAccount.getIndirectCostRecoveryAccount()) && 
+         icrAccount.getIndirectCostRecoveryAccount().isClosed()) {
+        Account replacementICRA = icrAccount.getIndirectCostRecoveryAccount();
+        int i = 0;
+
+        // Search 10 levels or less to find an open Account, as we do in other processes
+        while((i < 10) && ((replacementICRA != null) && replacementICRA.isClosed())) {
```
db483: Should use ObjectUtils.isNotNull to check if replacementICRA is not null

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuIndirectCostRecoveryAccountDistributionMetadata.java
```diff
@@ -0,0 +1,76 @@
...
+            // If the ICR Account is still closed after searching this far down, raise an exception.
+            String errorText = SpringContext.getBean(ConfigurationService.class)
+                                            .getPropertyValueAsString(CUKFSKeyConstants.ERROR_ICRACCOUNT_CONTINUATION_ACCOUNT_CLOSED);
+            Object[] args = {
+                icrAccount.getIndirectCostRecoveryFinCoaCode(),
+                icrAccount.getIndirectCostRecoveryAccountNumber()
+              };
+            errorText = MessageFormat.format(errorText, args);
+  
+            throw new UnsupportedOperationException(errorText);
```
db483: As I have mentioned to you on Friday, I believe on hipchat, it does not seem right to throw this exception if an open ICR account is not found. If this was Dave's requirement then this is fine, otherwise I wonder if an error message should be written to the error report instead? I know you mentioned it is highly unlikely for this situation to occur but should this occur the poster job would fail with an exception and I am not sure this is the desired behavior.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuIndirectCostRecoveryAccountDistributionMetadata.java
kyleoliveira: Good catch, will fix

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
```diff
@@ -0,0 +1,1308 @@
...
+
+                            reportOriginEntryGenerated = reportOriginEntryGenerated + 2;
+                        }
+                    }
+                    // Delete expenditure record
+                    expenditureTransactionDao.delete(et);
+                    reportExpendTranDeleted++;
+
+                }
+                catch (RuntimeException re) {
```
kyleoliveira: I had to include the foundation's PosterServiceImpl class because of this line. In order to pass an Exception subclass up when no suitable ICR Account is found (which then writes adds that error info for the line to the log and continues as seen in the Exception catch), we need to subclass Exception rather than RuntimeException. In order to throw something that is not a RuntimeException within a method, you must either catch it and handle it or declare that your method throws it. However, if a superclass doesn't declare that it throws this exception, your subclass can't either. So I also needed to change retrieveSubAccountIndirectCostRecoveryMetadata and retrieveAccountIndirectCostRecoveryMetadata so that they declared the throw because each of those methods calls the CuIndirectCostRecoveryAccountDistributionMetadata constructor when inserting a new item in the list. That constructor must be used in order for all this to work, unfortunately (can't just use the stock IndirectCostRecoveryAccountDistributionMetadata).

Alternatively, I could have only changed retrieveSubAccountIndirectCostRecoveryMetadata and retrieveAccountIndirectCostRecoveryMetadata to add the throw declaration, and changed the constructor call in a pair of overridden methods in CuPosterImpl. I did not do this because it seems like it would make it more difficult to hunt down what actually changed in the class. As it is, you should be able to do a diff and see the difference fairly quickly.

src/main/java/org/kuali/kfs/gl/batch/service/impl/PosterServiceImpl.java
kyleoliveira: Just kidding - alas, it didn't work. See my comment on PosterServiceImpl above.

kyleoliveira: See my comment on PosterServiceImpl above. I still had to include it.

kyleoliveira:
I believe this is ready for review again.

PR #15, opened 2016-02-03
KFSPTS-4343: Add changes so that job does not use doc search service that is restricted to 500 result entries, and also ignores documents whose KFS doc headers have ENROUTE status but whose corresponding KEW route headers have a different status.

cah292:
This fixes our auto-disapprove batch job to not limit the number of affected documents to 500 (or whatever the KEW doc search limit is), and to not try updating documents whose KFS doc headers have a stale ENROUTE status (which can happen when recalling to action list or taking certain Doc Operation screen actions).

Daniela coded up the portion that removes the 500-document-limit restriction several months ago. I coded up the part that filters out the stale ENROUTE statuses.

Because of some difficulties with getting the pull request as one commit (due to complications from rebasing the branch from develop prior to the interactive rebase), I deleted and recreated the branch and then reapplied all the changes. If something seems amiss, please let me know.


kyleoliveira:
Looks ok to me.

PR #16, opened 2016-02-03
KFSPTS-5021: create new maintenance table for recurring wires/ACH

bhutchinson:


src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/IncomingWireAchMapping.xml
```diff
@@ -0,0 +1,205 @@
...
+    	<property name="label" value="Short Description"/>
+    	<property name="shortLabel" value="Short Desc"/>
+    	<property name="maxLength" value="30"/>
+    	<property name="forceUppercase" value="true"/>
+        <property name="required" value="true"/>
+    	<property name="control">
+            <bean parent="TextControlDefinition" p:size="35"/>
+        </property>
+    	<property name="validationPattern">
+      		<bean parent="AnyCharacterValidationPattern">
```
cah292: Any particular reason why this file needs to make beans extending from "AnyCharacterValidationPattern" directly? KFS should already have a child bean of "AnyCharacterValidationPattern" that is configured to allow for whitespace, and the interior max-length check seems redundant since the attribute definition already defines one.

cah292:
I did have one quick question pertaining to why a certain choice of bean setup was used, but other than that it looks good to me.

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
```diff
@@ -733,8 +733,53 @@
     </reference-descriptor>
 </class-descriptor>
 
-</descriptor-repository>
 
+  <class-descriptor class="edu.cornell.kfs.fp.businessobject.IncomingWireAchMapping" table="FP_INCOMING_WIRE_ACH_MAP_TC">
+    <field-descriptor name="shortDescription" column="SHORT_DESC" jdbc-type="VARCHAR" primarykey="true" index="true" />
```
cah292: Is the "shortDescription" going to be unique enough for PK use? Will users or automated processes end up having trouble fitting an appropriate description into this column that doesn't clash with others?

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/IncomingWireAchMapping.xml
bhutchinson: no particular reason, probably from cut-n-paste; I'll change it

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
bhutchinson: my understanding is yes it will be unique enough, and if they do have trouble, we'll work it out when the time comes

cah292:
Looks good to me.

PR #17, opened 2016-02-08
KFSPTS-4528: Updated the YEJV document to allow for generating CB-period GL pending entries.

cah292:
This commit changes the YEJV document so that, when necessary, it can create "C & G Balance Forward" GL pending entries (which will have a "CB" fiscal period). It is mostly code and logic copied from BalanceServiceImpl, BalanceDaoOjb and BalanceForwardRuleHelper that has been adapted for creating GL pending entries instead of origin entries.

As noted on the in-code comments, there is one "if" block that is meant to replicate existing selection behavior in KFS, even though the existing behavior could contain a bug. (The functionals requested that the selection logic remain the same for now.) I will be creating a separate JIRA ticket to look into that problem further, and if changes do indeed need to be made as a result of that, then they will be made for a different story JIRA.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/document/service/impl/YearEndGeneralLedgerPendingEntriesServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/document/service/impl/YearEndGeneralLedgerPendingEntriesServiceImpl.java
@@ -10,13 +10,16 @@ import java.util.Map;
 import java.util.Set;
 
 import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.businessobject.ObjectCode;
 import org.kuali.kfs.coa.businessobject.OffsetDefinition;
 import org.kuali.kfs.coa.service.ObjectTypeService;
 import org.kuali.kfs.coa.service.OffsetDefinitionService;
+import org.kuali.kfs.coa.service.SubFundGroupService;
 import org.kuali.kfs.gl.GeneralLedgerConstants;
 import org.kuali.kfs.gl.batch.BalanceForwardStep;
 import org.kuali.kfs.gl.batch.NominalActivityClosingStep;
+import org.kuali.kfs.gl.businessobject.OriginEntryFull;
 import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.KFSKeyConstants;
 import org.kuali.kfs.sys.businessobject.AccountingLine;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/fp/document/service/impl/YearEndGeneralLedgerPendingEntriesServiceImpl.java
```diff
@@ -622,6 +662,97 @@ public GeneralLedgerPendingEntry generateBBCashOffset(AccountingDocumentBase doc
...
+        entry.setFinancialSubObjectCode(originEntry.getFinancialSubObjectCode());
+        entry.setFinancialBalanceTypeCode(originEntry.getFinancialBalanceTypeCode());
+        entry.setFinancialObjectTypeCode(originEntry.getFinancialObjectTypeCode());
+        
+        // Continue with regular setup.
+        entry.setUniversityFiscalPeriodCode(KFSConstants.PERIOD_CODE_CG_BEGINNING_BALANCE);
+        entry.setFinancialDocumentTypeCode(document.getFinancialSystemDocumentHeader().getWorkflowDocument().getDocumentTypeName());
+        entry.setFinancialSystemOriginationCode(homeOriginationService.getHomeOrigination().getFinSystemHomeOriginationCode());
+        entry.setDocumentNumber(new StringBuilder(KFSConstants.BALANCE_TYPE_ACTUAL).append(postable.getAccountNumber()).toString());
+        entry.setTransactionLedgerEntrySequenceNumber(new Integer(sequenceNumber.intValue()));
```
db483: sequenceNumber passed in is an Integer. Is it new Integer(sequenceNumber.intValue()) needed?

db483:
The code looks good, I just had one question about sequenceNumber. If we can clarify that, I will then merge the code. Thank you.

src/main/java/edu/cornell/kfs/fp/document/service/impl/YearEndGeneralLedgerPendingEntriesServiceImpl.java
cah292: You're right, it shouldn't be necessary since Integer objects are immutable anyway. In this case, though, I was copying it to follow the setup of similar GL entry generation methods in the same class, which do use the intValue() call. I'll see if I can tweak this to not use the intValue() call.

cah292:
I've updated the code to avoid the unneeded intValue() call.

db483:
Looking good.

PR #18, opened 2016-02-11
KFSPTS-3877 Move Account notes to object level; add code to add autom…

db483:
…atic notes when an Account is created/edited

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
@@ -8,15 +8,23 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.log4j.Logger;
 import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
 import org.kuali.kfs.coa.document.KualiAccountMaintainableImpl;
+import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.rice.core.api.datetime.DateTimeService;
 import org.kuali.rice.kns.document.MaintenanceDocument;
 import org.kuali.rice.kns.maintenance.Maintainable;
+import org.kuali.rice.krad.bo.Note;
+import org.kuali.rice.krad.document.Document;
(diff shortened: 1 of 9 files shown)


src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
```diff
@@ -68,11 +81,27 @@ public void saveBusinessObject() {
...
+        
+		// KFSPTS-3877 save xml again so that object id is persisted, this way
+		// notes on the doc will be retrieved form the DB instead of xml doc
+		// content and will display any future notes on account edits as well
+        try {
+            Document document = SpringContext.getBean(DocumentService.class).getByDocumentHeaderId(getDocumentNumber());
+            Account xmlaccount = (Account) ((MaintenanceDocument) document).getNewMaintainableObject().getBusinessObject();
+            if (ObjectUtils.isNull(xmlaccount.getObjectId()) &&( KFSConstants.MAINTENANCE_NEW_ACTION.equals(getMaintenanceAction()) || KFSConstants.MAINTENANCE_COPY_ACTION.equals(getMaintenanceAction()))) {
+                ((MaintenanceDocument) document).getNewMaintainableObject().setBusinessObject(account);
+                SpringContext.getBean(DocumentService.class).saveDocument(document);
```
cah292: I think the maintenance document will also be saved after performing the doRouteStatusChange() logic that calls the saveBusinessObject() method (but uses a saving method that does not update the maintainable XML). Have you verified that this extra save does not cause the calling code to run into an optimistic locking exception when it tries to do its own document persistence?

cah292:
Looks good in general, though I did have one question about the extra save that was added to properly update the maintainable XML.

src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
db483: I have tested in my local environment and I did not run into any exception. I have followed an example from VendorMaintainableImpl (line 166) with this one. The problem is that the objectId is not persisted in the XML and that is why the notes don't display from the database but rather from the XML doc content. That is causing any notes added on account edit not to be displayed on the doc that originally created the account. Please let me know if you can think of a better way to do this and I am happy to change it. Thank you very much!

cah292: Given that you haven't seen any errors in local testing and that PVEN docs already do something similar, it looks good to me.

PR #19, opened 2016-02-19
KFSPTS-3877 Move Account notes to object level; add code to add autom…

db483:
…atic notes when an Account is created/edited

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
@@ -8,15 +8,25 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang.StringUtils;
+import org.apache.log4j.Logger;
 import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
 import org.kuali.kfs.coa.document.KualiAccountMaintainableImpl;
+import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.rice.core.api.datetime.DateTimeService;
+import org.kuali.rice.kew.api.document.DocumentStatus;
 import org.kuali.rice.kns.document.MaintenanceDocument;
 import org.kuali.rice.kns.maintenance.Maintainable;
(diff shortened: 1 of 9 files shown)


db483:
Chad, there has been a request to update the code so that the document description would be included in the note that is automatically added when an account is edited. I had to revert previous merge, changed the code and created a new pull request.

src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
```diff
@@ -153,5 +184,65 @@ public List getSections(org.kuali.rice.kns.document.MaintenanceDocument document
...
+    }
+    
+    @Override
+    public void processAfterNew(MaintenanceDocument document, Map<String, String[]> requestParameters) {
+    	document.getNotes().add(getNewBoNote(CUKFSConstants.AccountCreateAndUpdateNotePrefixes.ADD));
+    	super.processAfterNew(document, requestParameters);
+    }
+    
+    @Override
+	public void processAfterPost(MaintenanceDocument document, Map<String, String[]> requestParameters) {
```
db483: I added this override so that the text for the note that is automatically added when an account is edited is updated to include the document description. The doc description cannot be changed after doc is routed so this customized code is only called when doc is in status Initiated or Saved.

src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
```diff
@@ -153,5 +184,65 @@ public List getSections(org.kuali.rice.kns.document.MaintenanceDocument document
...
+	}
+    
+    /**
+     * creates a new bo note and sets the timestamp.
+     *
+     * @return a newly created note
+     */
+    protected Note getNewBoNote(String prefix){
+        Note newBoNote = new Note();
+        newBoNote.setNoteText(prefix + " account document ID " + getDocumentNumber());
```
cah292: Since this " account document ID " String is used in multiple places and is also used for note comparisons, I would suggest moving it to its own constant.

cah292:
I did have one suggestion related to a certain String value used in multiple places, but other than that it looks good to me.

src/main/java/edu/cornell/kfs/coa/document/CUAccountMaintainableImpl.java
db483: Makes sense, I will make the change. Thank you Chad.

cah292:
Code looks good.

PR #20, opened 2016-02-19
KFSPTS-5169: Added principal name to Payee ACH Account document and lookup, as well as ACH Payee lookup; also reintroduced proper restrictions on Payee ACH Account copy and inquiry actions.

cah292:
This adds principal name to ACH Payee and Payee ACH Account. For the former, this property is added to the lookup. For the latter, this property is added to the lookup, inquiry and maintenance document, and has its value derived according to the payee type and number.

Because of the principal name being derived instead of stored on the Payee ACH Account, its lookup was customized accordingly to convert the principal name criteria into equivalent employee-or-entity criteria.

Also, the Payee ACH Account lookup has been set up to make use of a base KFS lookupable helper class that restricts inquiry and copy actions to those with editing permissions. Some additional KIM permissions will get added to DEV/TEST/PROD so that these restrictions are enforced consistently throughout the other KFS screens related to Payee ACH Account.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/businessobject/lookup/CuACHPayeeLookupableHelperServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/businessobject/lookup/CuACHPayeeLookupableHelperServiceImpl.java
@@ -1,17 +1,33 @@
 package edu.cornell.kfs.pdp.businessobject.lookup;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.fp.businessobject.DisbursementPayee;
 import org.kuali.kfs.pdp.PdpConstants;
+import org.kuali.kfs.pdp.PdpKeyConstants;
+import org.kuali.kfs.pdp.businessobject.ACHPayee;
 import org.kuali.kfs.pdp.businessobject.lookup.ACHPayeeLookupableHelperServiceImpl;
 import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.vnd.businessobject.VendorDetail;
+import org.kuali.rice.kim.api.identity.Person;
 import org.kuali.rice.kim.impl.KIMPropertyConstants;
+import org.kuali.rice.krad.bo.BusinessObject;
+import org.kuali.rice.krad.lookup.CollectionIncomplete;
+import org.kuali.rice.krad.util.BeanPropertyComparator;
+import org.kuali.rice.krad.util.GlobalVariables;
+
+import edu.cornell.kfs.pdp.businessobject.CuACHPayee;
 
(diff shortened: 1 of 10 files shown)


src/main/java/edu/cornell/kfs/pdp/businessobject/lookup/CuACHPayeeLookupableHelperServiceImpl.java
```diff
@@ -31,8 +47,99 @@
...
 
         return personFieldValues;
     }
 
+    /**
+     * Overridden to only search for employee or entity payees when principal name is specified.
+     * 
+     * @see org.kuali.kfs.pdp.businessobject.lookup.ACHPayeeLookupableHelperServiceImpl#getSearchResults(java.util.Map)
+     */
+    @SuppressWarnings({"deprecation", "rawtypes", "unchecked"})
```
cah292: This method is just a tweaked copy of a superclass method, and the superclass method contains code that brings up these warnings.

src/main/java/edu/cornell/kfs/pdp/businessobject/lookup/CuPayeeACHAccountLookupableHelperServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private PersonService personService;
+    private CriteriaLookupService criteriaLookupService;
+
+    /**
+     * Overridden to perform custom searching when a principal name is specified on the search screen.
+     * 
+     * @see org.kuali.rice.kns.lookup.KualiLookupableHelperServiceImpl#getSearchResultsHelper(java.util.Map, boolean)
+     */
+    @SuppressWarnings("unchecked")
```
cah292: Suppressing "unchecked" warnings here is not ideal, but was needed because of this method's reliance on base Rice KNS code that does not have consistent use of generics.

src/main/java/edu/cornell/kfs/pdp/businessobject/lookup/CuPayeeACHAccountLookupableHelperServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+                    if (StringUtils.isNotBlank(person.getEmployeeId())) {
+                        employeeIds.add(person.getEmployeeId());
+                    }
+                }
+                
+                // Create a map without blank values and with all encrypted values decrypted, similar to the ancestor class's logic.
+                Map<String,String> finalFieldValues = new HashMap<String,String>();
+                for (Map.Entry<String,String> entry : fieldValues.entrySet()) {
+                    // Only add non-blank values.
+                    if (StringUtils.isBlank(entry.getValue())) {
```
cah292: I could have structured this "if" block to not need all these "else if" or "else" parts, but I did it this way instead in order to avoid CheckStyle complaints about nested "if" depth.

kyleoliveira:
Looks fine to me.

PR #27, opened 2016-02-25
KFSPTS-5202: remove slow test of dubious benefit

bhutchinson:


cah292:
Seems okay for the test to be removed.

PR #31, opened 2016-02-25
KFSPTS-5169: Added principal name to Payee ACH Account document, lookup and inquiry, as well as ACH Payee lookup.

cah292:
Functional users discovered some issues with the changes from the original KFSPTS-5169 pull request (#20), so those changes have been reverted and this new pull request has been created to implement the correct functionality.

This should be almost the same as the #20 pull request, except that the new lookupable helper service now extends from a different class. This was necessary to prevent usage of potentially-buggy code in the base KFS lookupable helper for Payee ACH Accounts.


kyleoliveira:
Looks fine to me.

PR #32, opened 2016-03-02

cah292:
This feature adds the ability to create Payee ACH Account BOs based on .csv file input data, with the intent of receiving such files from Workday. File entries representing ACH Accounts not in KFS yet will create and route PAAT documents as needed to create Employee and/or Entity accounts for a given employee. (Notification emails will be sent as well.) Existing ACH Accounts in KFS will not be updated by this job; that will be a future enhancement.


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,691 @@
...
+                        // If a values finder is defined, use the label from the matching key/value pair instead.
+                        if (attDefinition.getControl() != null && StringUtils.isNotBlank(attDefinition.getControl().getValuesFinderClass())) {
+                            KeyValuesFinder valuesFinder = (KeyValuesFinder) Class.forName(attDefinition.getControl().getValuesFinderClass()).newInstance();
+                            String key = replacement;
+                            replacement = valuesFinder.getKeyLabel(key);
+                            // If the key is in the label, then remove it from the label.
+                            if (attDefinition.getControl().getIncludeKeyInLabel() != null
+                                    && attDefinition.getControl().getIncludeKeyInLabel().booleanValue()) {
+                                // Check for key-and-dash or key-in-parentheses, and remove them if found.
+                                // (Former can come from BO values finders, latter is only for custom values finders that append the keys as such.)
```
cah292: The code section below these comments has been set up in a manner not using more "if" blocks, to avoid some CheckStyle complaints about the nested "if" block depth being too deep.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,691 @@
...
+            maintainable.setMaintenanceAction(KFSConstants.MAINTENANCE_NEW_ACTION);
+            PayeeACHAccount achAccount = (PayeeACHAccount) maintainable.getDataObject();
+            
+            // Setup payee ID and type.
+            if (PayeeIdTypeCodes.ENTITY.equals(payeeType)) {
+                achAccount.setPayeeIdNumber(payee.getEntityId());
+            } else if (PayeeIdTypeCodes.EMPLOYEE.equals(payeeType)) {
+                achAccount.setPayeeIdNumber(payee.getEmployeeId());
+            } else {
+                throw new WorkflowException("Unexpected payee ID type for automated PAAT creation: " + payeeType);
```
kyleoliveira: Since you provide the doc type abbreviation as a variable above, would it make sense to include it in the string creation here. I highly doubt it would ever change, but referencing that variable would minimize places that would need to be updated.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,691 @@
...
+                        }
+                        // Because of the way that Matcher.appendReplacement() works, escape the special replacement characters accordingly.
+                        if (replacement.indexOf('\\') != -1) {
+                            replacement = replacement.replace("\\", "\\\\");
+                        }
+                        if (replacement.indexOf('$') != -1) {
+                            replacement = replacement.replace("$", "\\$");
+                        }
+                        
+                    } catch (ClassNotFoundException e) {
```
kyleoliveira: Looks like since we're developing for [Java 7 and above, you can simplify these catch statements into one](http://stackoverflow.com/questions/3495926/can-i-catch-multiple-java-exceptions-in-the-same-catch-clause):
```java
} catch (ClassNotFoundException | IllegalAccessException | InstantiationException | RuntimeException e) {
  replacement = KFSConstants.EMPTY_STRING;
}
```

kyleoliveira:
Done reviewing. Waiting for comments or an update to the commit before I merge.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: Good catch, I'll update the text accordingly.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: Thanks for pointing this out. I'll get this updated.

cah292:
Code has now been updated based on feedback, and is ready to be reviewed again.

kyleoliveira:
Ok, looks good.

PR #35, opened 2016-03-02

kyleoliveira:
Implements processing and viewing for the Type 50 record for our US Bank PCard files. More information (including an example upload file) can be found in KFSPTS-4030 (which was the only JIRA for Level3 PCard data originally).

Level3 information is already included in the upload files, this simply adds the ability to read, store, and display it. We're also only dealing with the Type 50 data in this JIRA. After processing a pcard flat file, find a PCDO doc and click on the Transaction Reference Number on its inquiry page. This will show the Procurement Card Transaction Detail Inquiry, which should have the Type 50 (Purchasing Data) information below it.

See the JIRA for the associated SQL file.

If you wish to test this out without tampering with what's in devlocal right now, I can explain how to launch it in the Sandbox environment.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/ProcurementCardFlatInputFileType.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/ProcurementCardFlatInputFileType.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.sql.Date;
 import java.util.ArrayList;
+import java.util.List;
 
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.fp.businessobject.ProcurementCardTransaction;
@@ -34,9 +35,11 @@ import org.kuali.rice.core.api.datetime.DateTimeService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 import org.kuali.rice.coreservice.framework.parameter.ParameterService;
 import org.kuali.rice.krad.service.SequenceAccessorService;
+import org.kuali.rice.krad.util.ObjectUtils;
 
 import edu.cornell.kfs.fp.batch.service.ProcurementCardErrorEmailService;
 import edu.cornell.kfs.fp.businessobject.ProcurementCardTransactionExtendedAttribute;
+import edu.cornell.kfs.fp.businessobject.USBankRecordFieldUtils;
 import edu.cornell.kfs.sys.CUKFSParameterKeyConstants;
 
 public class ProcurementCardFlatInputFileType extends BatchInputFileTypeBase {
@@ -64,6 +67,15 @@ public class ProcurementCardFlatInputFileType extends BatchInputFileTypeBase {
     private DateTimeService dateTimeService;
     private ParameterService parameterService;
     private ProcurementCardErrorEmailService procurementCardErrorEmailService;
+    
+    // USBank record ID's
+    // If/when we create dedicated classes for representing these records, 
+    // these lines can move there as we've done with the addendum records.
+    public static final String FILE_HEADER_RECORD_ID = "01";
+    public static final String CARDHOLDER_DATA_HEADER_RECORD_ID = "02";
+    public static final String TRANSACTION_INFORMATION_RECORD_ID = "05";
+    public static final String CARDHOLDER_TRAILER_RECORD_ID = "95";
+    public static final String FILE_TRAILER_RECORD_ID = "98";
 
     /**
      * @see org.kuali.kfs.sys.batch.BatchInputFileType#getFileTypeIdentifer()
@@ -176,7 +188,12 @@ public class ProcurementCardFlatInputFileType extends BatchInputFileTypeBase {
         	
             while ((fileLine=bufferedFileReader.readLine()) != null) {
             	ProcurementCardTransaction theTransaction = generateProcurementCardTransaction(fileLine);
+                
             	if (theTransaction!=null){
+                    if(ObjectUtils.isNotNull(theTransaction.getExtension())) {
+                      lineCount = ((ProcurementCardTransactionExtendedAttribute)theTransaction.getExtension()).addAddendumLines(bufferedFileReader, lineCount);
+                    }
+
             		transactions.add(theTransaction);
             	}
             	lineCount++;
@@ -231,24 +248,34 @@ public class ProcurementCardFlatInputFileType extends BatchInputFileTypeBase {
         
     }
 
+    /**
+     * Parses a line into a ProcurementCardTransaction.
+     * 
+     * The data in the ProcurementCardTransaction comes from USBank record types 02 and 05. 
+     * Other USBank record types (01, 95, 98) are either skipped or used for validation.
+     * 
+     * @param line The current line
+     * @return A completed transaction or null if there are additional lines in the transaction
+     * @throws Exception
+     */
     private ProcurementCardTransaction generateProcurementCardTransaction(String line) throws Exception {
 	        
-        String recordId = extractNormalizedString(line, 0, 2);
+        String recordId = USBankRecordFieldUtils.extractNormalizedString(line, 0, 2);
         if (recordId == null) {
         	throw new Exception("Unable to determine record Id necessary in order to parse line " + lineCount);
         }
-        if (recordId.equals("01")){ //file header record
-        	headerTransactionCount = Integer.parseInt(extractNormalizedString(line, 67, 75));
+        if (recordId.equals(FILE_HEADER_RECORD_ID)){ //file header record
+        	headerTransactionCount = Integer.parseInt(USBankRecordFieldUtils.extractNormalizedString(line, 67, 75));
         }
-        if (recordId.equals("98")) { //file footer record
+        if (recordId.equals(FILE_TRAILER_RECORD_ID)) { //file footer record
         		        	
-        	footerTransactionCount = Integer.parseInt(extractNormalizedString(line, 21, 29));
-        	fileFooterCredits = fileFooterCredits.add( new KualiDecimal(extractNormalizedString(line,44,59)));
-        	fileFooterDebits = fileFooterDebits.add( new KualiDecimal(extractNormalizedString(line,29,44)));
+        	footerTransactionCount = Integer.parseInt(USBankRecordFieldUtils.extractNormalizedString(line, 21, 29));
+        	fileFooterCredits = fileFooterCredits.add( new KualiDecimal(USBankRecordFieldUtils.extractNormalizedString(line,44,59)));
+        	fileFooterDebits = fileFooterDebits.add( new KualiDecimal(USBankRecordFieldUtils.extractNormalizedString(line,29,44)));
         	if (totalDebits.compareTo(fileFooterDebits.abs()) != 0) {
         		StringBuffer sb = new StringBuffer();
         		sb.append("Debits in file footer do not match the sum of debits in transaction.");
-        		sb.append(lineCountMessage());
+        		sb.append(USBankRecordFieldUtils.lineCountMessage(lineCount));
         		sb.append(" Debit value given in file footer: ");
         		sb.append(fileFooterDebits.abs());
         		sb.append(" Debit value generated by summing transactions: ");
@@ -258,7 +285,7 @@ public class ProcurementCardFlatInputFileType extends BatchInputFileTypeBase {
         	if (totalCredits.compareTo(fileFooterCredits.abs()) != 0) {
         		StringBuffer sb = new StringBuffer();
         		sb.append("Credits in file footer do not match the sum of credits in transaction.");
-        		sb.append(lineCountMessage());
+        		sb.append(USBankRecordFieldUtils.lineCountMessage(lineCount));
         		sb.append(" Credit value given in file footer: ");
         		sb.append(fileFooterCredits.abs());
         		sb.append(" Credit value generated by summing transactions: ");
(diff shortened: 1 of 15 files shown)


src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardCreateDocumentServiceImpl.java
```diff
@@ -161,6 +167,55 @@ protected ProcurementCardDocument createProcurementCardDocument(List transaction
...
+
+    protected void createProcurementCardTransactionDetailExtension(ProcurementCardTransaction transaction, ProcurementCardTransactionDetail transactionDetail) {
+      ProcurementCardTransactionDetailExtendedAttribute detailExtension;
+      if (ObjectUtils.isNull(transactionDetail.getExtension())) {
+          detailExtension = new ProcurementCardTransactionDetailExtendedAttribute();
+      } else {
+          detailExtension = (ProcurementCardTransactionDetailExtendedAttribute) transactionDetail.getExtension();
+      }
+
+      if (ObjectUtils.isNotNull(transaction.getExtension())) {
```
bhutchinson: couldn't this be combined with the else above?

kyleoliveira: Yeah, probably.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataDetail.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.LinkedHashMap;
+
+import org.kuali.kfs.sys.KFSPropertyConstants;
+
+public class PurchasingDataDetail extends PurchasingDataBase {
+    private static final long serialVersionUID = 1L;
+    private String documentNumber;
+    private Integer financialDocumentTransactionLineNumber;
+
+    public PurchasingDataDetail() {
```
bhutchinson: could just remove this and let Java give us a default constructor

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataRecord.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package edu.cornell.kfs.fp.businessobject;
+
+public class PurchasingDataRecord extends PurchasingDataBase
+{
+    private static final long serialVersionUID = 1L;
+    private Integer transactionSequenceRowNumber;
+
+    public PurchasingDataRecord()
```
bhutchinson: another unnecessary constructor

src/main/java/edu/cornell/kfs/fp/businessobject/USBankAddendumRecord.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * @param recordType the recordType to set
+     */
+    public void setRecordId(String recordId) {
+      this.recordId = recordId;
+    }
+
+    public USBankAddendumRecord() {
```
bhutchinson: and again

src/main/java/edu/cornell/kfs/fp/businessobject/USBankAddendumRecord.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Extracts a substring less any ending whitespace for a given beginning and ending position.
+     * 
+     * @param line Superstring
+     * @param begin Beginning index
+     * @param end Ending index
+     * @return The trimmed substring
+     * @throws StringIndexOutOfBoundsException May occur when taking the actual substring if the provided bounds are beyond the limits of the superstring
+     */
+    protected String extractNormalizedString(String line, int begin, int end) throws StringIndexOutOfBoundsException {
```
bhutchinson: it looks like this same code is in multiple classes; I wonder if it could be extracted into a service or utility class to eliminate the duplication

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/PurchasingDataRecord.xml
```diff
@@ -0,0 +1,295 @@
...
+  </bean>
+
+  <bean id="PurchasingDataRecord-accountNumber" parent="PurchasingDataRecord-accountNumber-parentBean" />
+  <bean id="PurchasingDataRecord-accountNumber-parentBean" abstract="true" parent="AttributeDefinition">
+    <property name="name" value="accountNumber" />
+    <property name="label" value="Account Number" />
+    <property name="forceUppercase" value="false" />
+    <property name="maxLength" value="16" />
+    <property name="control">
+      <bean parent="TextControlDefinition" p:size="15"/>
```
bhutchinson: do we want a control size smaller than the maxLength of the field?

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/PurchasingDataRecord.xml
```diff
@@ -0,0 +1,295 @@
...
+  </bean>
+
+  <bean id="PurchasingDataRecord-tsysTranCode" parent="PurchasingDataRecord-tsysTranCode-parentBean" />
+  <bean id="PurchasingDataRecord-tsysTranCode-parentBean" abstract="true" parent="AttributeDefinition">
+    <property name="name" value="tsysTranCode" />
+    <property name="label" value="TSYS Tran Code" />
+    <property name="forceUppercase" value="false" />
+    <property name="maxLength" value="4" />
+    <property name="control">
+      <bean parent="TextControlDefinition" p:size="15"/>
```
bhutchinson: do we want a control size so much larger than the maxLength of the field

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataDetail.java
kyleoliveira: Ok

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataRecord.java
kyleoliveira: Ok

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/PurchasingDataRecord.xml
```diff
@@ -0,0 +1,295 @@
...
+  </bean>
+
+  <bean id="PurchasingDataRecord-merchantOrderNumber" parent="PurchasingDataRecord-merchantOrderNumber-parentBean" />
+  <bean id="PurchasingDataRecord-merchantOrderNumber-parentBean" abstract="true" parent="AttributeDefinition">
+    <property name="name" value="merchantOrderNumber" />
+    <property name="label" value="Merchant Order Number" />
+    <property name="forceUppercase" value="false" />
+    <property name="maxLength" value="12" />
+    <property name="control">
+      <bean parent="TextControlDefinition" p:size="12"/>
```
bhutchinson: a common pattern I have seen in base code is that the control size for text controls is two larger than the maxLength for the field (so it would be 14 in this case); not sure if we've been following that pattern too, or if not, if we want to

src/main/java/edu/cornell/kfs/fp/batch/ProcurementCardFlatInputFileType.java
```diff
@@ -176,7 +178,15 @@ public Object parse(byte[] fileByteContent) throws ParseException {
...
             	ProcurementCardTransaction theTransaction = generateProcurementCardTransaction(fileLine);
+                
             	if (theTransaction!=null){
+            	  
+            	    // Addendum record lines may follow after a record
+                    if(ObjectUtils.isNotNull(theTransaction.getExtension())) {
+                      lineCount = ((ProcurementCardTransactionExtendedAttribute)theTransaction.getExtension()).addAddendumLines(bufferedFileReader, lineCount);
+                    }
+                    
+                    // Add the transaction
```
bhutchinson: I'm not a fan of comments like this, I don't feel it's adding any value.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankAddendumRecord.java
kyleoliveira: That was the goal with this (move all the record parsing to their own classes and base it off of this) but I got sidetracked somewhere...
I think a utility class would be ideal.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankAddendumRecord.java
kyleoliveira: Will clean.

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/PurchasingDataRecord.xml
kyleoliveira: Good catch. I'll fix it up so they match.

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/PurchasingDataRecord.xml
kyleoliveira: I'm not opposed. I'll go through and update these accordingly, at least to be consistent with base code.

src/main/java/edu/cornell/kfs/fp/batch/ProcurementCardFlatInputFileType.java
kyleoliveira: I'll remove them. They were mostly for planning/outlining purposes.

kyleoliveira:
I've added a util class that should refactor the utilities for parsing the PCard. I'm going to test this out in the SB environment before I squash the commit that contains that work.

Feel free to comment on what I've fixed up so far though.

kyleoliveira:
Ok, still works. Please review what's up now.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardCreateDocumentServiceImpl.java
bhutchinson: yeah, sorry, that was me reading too quickly and thinking the objects being checked for null/not null where the same when they're not; my bad

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,194 @@
...
+  public static Date extractDate(String line, int begin, int end, int lineCount) throws Exception {
+    Date theDate;
+    try {
+        String sub = line.substring(begin, end);
+        String year = sub.substring(0,4);
+        String month = sub.substring(4,6);
+        String day = sub.substring(6,8);
+        theDate = Date.valueOf(year+"-"+month+"-"+day);
+    }
+    catch (Exception e) {
```
bhutchinson: can we be more specific here than just Exception?

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,194 @@
...
+   * Extracts a cycle Date from a substring less any ending whitespace for a given beginning and ending position.
+   * 
+   * @param line Superstring
+   * @param begin Beginning index
+   * @param end Ending index
+   * @param lineCount The current line number
+   * @return The cycle Date parsed from the trimmed substring
+   * @throws Exception When unable to parse the cycle Date
+   */
+  @SuppressWarnings("deprecation")
```
bhutchinson: what is generating a deprecation warning and can we avoid it rather than just suppressing it?

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,194 @@
...
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+/**
+ * Provides methods for parsing USBank addendum records.
+ * In the long-run, it may be wise/useful to turn this into a subclass of 
+ * org.kuali.kfs.sys.businessobject.BusinessObjectStringParserFieldUtils and 
+ * provide subclasses that assist in parsing, but we won't for now.
+ *
+ */
+public class USBankRecordFieldUtils {
```
bhutchinson: Seems like this could be more generic, but I'm fine with leaving it specific for now and making it more generic later if/when that becomes beneficial.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: I was keeping it the same as the method it replaced, but we can probably throw it as an IOException. What do you think?

bhutchinson: I think we should be as specific as possible with exceptions, both catching and throwing. So I think if we can use IOException here, that would be preferable to Exception.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: The Date#setDate method is deprecated in favor of Calendar.set(Calendar.DAY_OF_MONTH, int date).

I didn't want to mess with it due to scope creep, but I can look into it if you think it should really be handled here.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: Ok good.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: Yeah, I generally agree with/shoot for that. I'll revise the Exceptions thrown by the methods in this class.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
bhutchinson: If you can make the change in less than an hour, I'd say do it. I know we get a lot of warnings now, but I'd rather pay down that tech debt than add to it. org.kuali.kfs.sys.util.KfsDateUtils.newDate() uses Calendar to construct a java.sql.Date in case that's a helpful reference.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
bhutchinson: Great, thanks!

kyleoliveira:
New version pushed up. I'll test this in SB tomorrow to make sure nothing broke, so do not merge yet, but please let me know if there's anything else that needs reworking.

kyleoliveira:
Tested in the SB environment and the changes seem to check out. If you're comfortable with the code as it is now, we should be good for merge. There is SQL to be applied, attached to the JIRA.

bhutchinson:
I think this is fine. I'm still somewhat conflicted about the Exception catching and throwing in USBankRecordFieldUtils, but I think that's a larger conversation and I think this is ok for now.

I'll defer to @jhulslander on the merge since this PR is assigned to him.

jhulslander:
I think is OK

PR #40, opened 2016-03-07
KFSPTS-5237 updated YEBT document to remove org routing, and putting …

jhulslander:
…in account routing


cah292:
Looks good to me.

PR #57, opened 2016-03-14
KFSPTS-5204:Update document header label on Sub-Account Global edoc.

nkimble:
This change will need to be ingested as is progresses through each of the environments.


cah292:
Change looks good, will wait for unit tests to finish before merging.

bhutchinson:
retest this please

PR #58, opened 2016-03-14
KFSPTS-5237 updated YEBT document to remove org routing, and putting …

jhulslander:
…in account routing.  Also added labor year end labor doc workflow

Note the workflow document.  I updated the XML you created for a parent task.  It didn't look like it was added to source control.  To fix the parallel versus sequential routing, I just needed to add in the activation type.  I am not sure if this idea or not, so please let me know what you think.


cah292:
Yes, it looks like the Rice code is still using the doc type route node sequential/parallel indicators even for KIM-based routing. Your updated doc type appears to be correct, and the unit tests have also passed.

This doc type change should fix all of the year-end LD documents. We just need a similar update via a separate story to fix the year-end FP documents.

PR #59, opened 2016-03-15
KFSPTS-5214: Added ability to configure Receipt Processing output directory via a config property.

cah292:
This change simply allows for one of the Receipt Processing base directory paths to be configurable via a property.


jhulslander:
This looks like, is there puppet related changes that need to be in place before we merge this in?

cah292:
Yes, I created a DevOps ticket and I believe Chad Bower has the puppet changes in place for the unit tests and DEV at the moment.

jhulslander:
Sounds good, I merged it in

PR #60, opened 2016-03-15
KFSPTS-4905: Updated Scrubber so that closed Cost Sharing accounts will post to their continuation accounts.

cah292:
This change allows for closed Cost Sharing accounts to post to their continuation accounts when the Scrubber runs. This is similar to the recent feature for allowing ICR accounts to post to continuation accounts, except that this feature is in the Scrubber instead of the Poster (due to Cost Sharing being processed at the Scrubber phase).

Note that, due to limitations of the KFS-delivered ScrubberProcessImpl, the new CU-specific subclass needs to be in the same package as the superclass in order to access a required internal class.

### DIFF
--- a/src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
+++ b/src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
@@ -1,0 +1,312 @@
+package org.kuali.kfs.gl.batch.service.impl;
+
+import java.io.IOException;
+import java.text.MessageFormat;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.kuali.kfs.coa.businessobject.A21SubAccount;
+import org.kuali.kfs.coa.businessobject.Account;
+import org.kuali.kfs.coa.businessobject.OffsetDefinition;
+import org.kuali.kfs.gl.GeneralLedgerConstants;
+import org.kuali.kfs.gl.batch.ScrubberStep;
+import org.kuali.kfs.gl.businessobject.OriginEntryFull;
+import org.kuali.kfs.gl.businessobject.OriginEntryInformation;
+import org.kuali.kfs.gl.businessobject.Transaction;
+import org.kuali.kfs.gl.service.ScrubberReportData;
+import org.kuali.kfs.gl.service.impl.StringHelper;
+import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.KFSKeyConstants;
+import org.kuali.kfs.sys.Message;
+import org.kuali.kfs.sys.businessobject.SystemOptions;
+import org.kuali.kfs.sys.exception.InvalidFlexibleOffsetException;
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+import edu.cornell.kfs.sys.CUKFSKeyConstants;
+
+/**
+ * Cornell's implementation of ScrubberProcess that extends from the KFS-delivered ScrubberProcessImpl.
+ * 
+ * NOTE: Because ScrubberProcessImpl declares an internal TransactionError class with default (package-private)
+ * visibility, this subclass has to be in the same package as ScrubberProcessImpl in order to use that
+ * internal class. If future versions of KFS update that class with a higher visibility, then move
+ * this subclass into a more appropriate package.
+ */
+public class CuScrubberProcessImpl extends ScrubberProcessImpl {
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(CuScrubberProcessImpl.class);
+
+    private static final int CONTINUATION_ACCOUNT_DEPTH_LIMIT = 10;
+
+    /**
+     * Overridden to update cost share source account entries so that they will post to the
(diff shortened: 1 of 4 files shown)


src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
```diff
@@ -0,0 +1,312 @@
...
+
+    /**
+     * Overridden to update cost share source account entries so that they will post to the
+     * continuation account if the cost share account is closed.
+     * 
+     * @see org.kuali.kfs.gl.batch.service.impl.ScrubberProcessImpl#generateCostShareEntries(
+     *         org.kuali.kfs.gl.businessobject.OriginEntryInformation, org.kuali.kfs.gl.service.ScrubberReportData)
+     */
+    @Override
+    protected TransactionError generateCostShareEntries(OriginEntryInformation scrubbedEntry, ScrubberReportData scrubberReport) {
```
cah292: This is mostly the same as the method from the superclass, except for the portion that calls a new method for doing the Cost Share and continuation account setup.

src/main/resources/edu/cornell/kfs/gl/cu-spring-gl.xml
```diff
@@ -209,6 +209,39 @@
        </property>
 	</bean>
 
+	<bean id="scrubberProcess-parentBean" class="org.kuali.kfs.gl.batch.service.impl.CuScrubberProcessImpl" abstract="true">
```
cah292: Similar to what we're doing with the Poster, I'm overriding the base/parent Scrubber bean so that the numerous Scrubber child beans can all be overridden by updating just one bean.

src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
```diff
@@ -0,0 +1,312 @@
...
+            for (int i = 0; i < CONTINUATION_ACCOUNT_DEPTH_LIMIT && ObjectUtils.isNotNull(continuationAccount) && continuationAccount.isClosed(); i++) {
+                continuationAccount = accountingCycleCachingService.getAccount(
+                        continuationAccount.getContinuationFinChrtOfAcctCd(), continuationAccount.getContinuationAccountNumber());
+            }
+            if (ObjectUtils.isNull(continuationAccount) || costShareAccount == continuationAccount || continuationAccount.isClosed()) {
+                // Could not find a valid Cost Share continuation account; return an error.
+                return new TransactionError(costShareSourceAccountEntry, new Message(MessageFormat.format(
+                        configurationService.getPropertyValueAsString(CUKFSKeyConstants.ERROR_CSACCOUNT_CONTINUATION_ACCOUNT_CLOSED),
+                                scrubbedEntryA21SubAccount.getCostShareChartOfAccountCode(), scrubbedEntryA21SubAccount.getCostShareSourceAccountNumber()),
+                        Message.TYPE_FATAL));
```
nkimble: Is this error misleading? There are two other conditions that could cause it besides the account being closed?

src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            createOutputEntry(costShareSourceAccountOffsetEntry, OUTPUT_GLE_FILE_ps);
+            scrubberReport.incrementCostShareEntryGenerated();
+
+            scrubCostShareAmount = KualiDecimal.ZERO;
+        } catch (IOException ioe) {
+            LOG.error("generateCostShareEntries() Stopped: " + ioe.getMessage());
+            throw new RuntimeException("generateCostShareEntries() Stopped: " + ioe.getMessage(), ioe);
+        }
```
nkimble: If this exception is thrown, will there be enough output to the log(s) to troubleshoot and fix the root cause of the exception?

cah292: This portion of the code is a copy from the superclass's version of the method. Also, note that the RuntimeException is given the IOException as an argument, so whatever logs the RuntimeException output should still be able to see the original exception.

nkimble:
This is a change to a critical path batch job.  As such, I would like to request that other team members also review and comment on this change before it is merged.

src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
cah292: We want to indicate an error/failure if the Cost Share account is closed and a valid open continuation account could not be found after searching up to a depth of 10. These three conditions should cover confirming such a situation (no continuation account found, exact same Cost Share account was found for some odd reason, or deepest continuation account within limit is closed). Unless there was some other concern you had?

cah292: Just for further clarification, there's a prior "if" check to determine whether the Cost Share account is closed. If it is, then the continuation processing and the three-condition "if" should cover reporting the appropriate error/failure if necessary.

bhutchinson:
Looks fine to me.

nkimble:
Thanks for the clarification Chad. Merging

PR #62, opened 2016-03-15
KFSPTS-5248: Removed now-unused OrgAssignedValidation class from codebase.

cah292:
Now that our year-end docs are no longer performing org-based routing, we can safely remove a custom CU-specific class pertaining to that particular routing.


jhulslander:
+1

PR #63, opened 2016-03-16

db483:


src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -0,0 +1,378 @@
...
+
+		if (StringUtils.isBlank(accountNumber)) {
+			GlobalVariables.getMessageMap().putError(KFSPropertyConstants.ICR_ACCOUNT_NUMBER, KFSKeyConstants.ERROR_REQUIRED, getDDAttributeLabel(KFSPropertyConstants.ICR_ACCOUNT_NUMBER));
+			success &= false;
+		}
+
+		if (StringUtils.isNotBlank(chartOfAccountsCode) && StringUtils.isNotBlank(accountNumber)) {
+			Map<String, String> chartAccountMap = new HashMap<String, String>();
+			chartAccountMap.put(KFSPropertyConstants.CHART_OF_ACCOUNTS_CODE, chartOfAccountsCode);
+			if (SpringContext.getBean(BusinessObjectService.class).countMatching(Chart.class, chartAccountMap) < 1) {
```
cah292: Since the BO service is used several times in this method, would it be better to declare a local variable for it?

src/main/java/edu/cornell/kfs/coa/businessobject/A21SubAccountChange.java
```diff
@@ -0,0 +1,119 @@
...
+    protected String costShareChartOfAccountCode;
+    protected String costShareSourceAccountNumber;
+    protected String costShareSourceSubAccountNumber;
+    
+    protected Chart costShareChartOfAccount;
+    protected Account costShareAccount;
+    protected SubAccount costShareSourceSubAccount;
+    protected IndirectCostRecoveryType indirectCostRecoveryType;
+	
+	public A21SubAccountChange() {
```
bhutchinson: technically, we could leave this out and it would be provided by Java by default

src/main/java/edu/cornell/kfs/coa/businessobject/SubAccountGlobal.java
```diff
@@ -38,28 +62,130 @@ public SubAccountGlobal() {
...
-		if(inactivate){
-			for(SubAccountGlobalDetail subAccountGlobalDetail : subAccountGlobalDetails){
-				subAccountGlobalDetail.refreshReferenceObject("subAccount");
-				SubAccount subAccount = subAccountGlobalDetail.getSubAccount();
+
+		for (SubAccountGlobalDetail subAccountGlobalDetail : subAccountGlobalDetails) {
+			subAccountGlobalDetail.refreshReferenceObject("subAccount");
+			SubAccount subAccount = subAccountGlobalDetail.getSubAccount();
+			
+			// update active flag
```
bhutchinson: These types of comments don't really seem to add anything as the code is pretty self-explanatory. My preference would be for them to be removed (if you need to make code changes from this review; if not, they can stay and this note can be for future reference).

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalRule.java
```diff
@@ -93,4 +148,298 @@ public boolean checkSubAccountDetail(SubAccountGlobalDetail dtl) {
...
+        }
+        
+        if (!allFieldsSet && StringUtils.isNotBlank(a21.getCostShareSourceAccountNumber())) {
+        	success &= checkEmptyBOField(KFSPropertyConstants.A21_SUB_ACCOUNT + "." + KFSPropertyConstants.COST_SHARE_SOURCE_CHART_OF_ACCOUNTS_CODE, a21.getCostShareChartOfAccountCode(),  getDisplayName(KFSPropertyConstants.A21_SUB_ACCOUNT + "." + KFSPropertyConstants.COST_SHARE_SOURCE_CHART_OF_ACCOUNTS_CODE));
+        }
+
+        a21.refreshReferenceObject(KFSPropertyConstants.COST_SHARE_SOURCE_ACCOUNT);
+        // existence test on Cost Share Account
+        if (allFieldsSet) {
+            if (ObjectUtils.isNull(a21.getCostShareAccount())) {
```
cah292: I see there are various places where the business rules are assuming the referenced objects exist if the FK references are valid. Is that necessarily the case here and in other places in these business rules? Are the reference objects guaranteed to be refreshed or reloaded at this stage?

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
db483: Good point, I will add the change.

src/main/java/edu/cornell/kfs/coa/businessobject/A21SubAccountChange.java
db483: Good point, I will remove it.

src/main/java/edu/cornell/kfs/coa/businessobject/SubAccountGlobal.java
db483: I will clean it up.

src/main/java/edu/cornell/kfs/coa/businessobject/SubAccountGlobal.java
```diff
@@ -38,28 +62,130 @@ public SubAccountGlobal() {
...
+				for (IndirectCostRecoveryAccountChange icrAccount : indirectCostRecoveryAccounts) {
+					subAccountGlobalDetail.refreshReferenceObject("subAccount");
+					found = false;
+					if (ObjectUtils.isNotNull(a21IcrAccounts) && a21IcrAccounts.size() > 0) {
+						for (A21IndirectCostRecoveryAccount a21ICRAccount : a21IcrAccounts) {
+							if (icrAccount.getIndirectCostRecoveryFinCoaCode().equalsIgnoreCase(a21ICRAccount.getIndirectCostRecoveryFinCoaCode()) && icrAccount.getIndirectCostRecoveryAccountNumber().equalsIgnoreCase(a21ICRAccount.getIndirectCostRecoveryAccountNumber())) {
+								if ((icrAccount.getAccountLinePercent().compareTo(a21ICRAccount.getAccountLinePercent()) == 0) && (icrAccount.isActive() != a21ICRAccount.isActive())) {
+									a21ICRAccount.setActive(icrAccount.isActive());
+									found = true;
+									break;
```
cah292: Not sure if using a "break" statement in a "for" loop is ideal, although a more iterative version of this would require use of an index variable or an Iterator.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalRule.java
db483: I am not sure I understand the concern on this particular line.  
the cost share account was refreshed on line 351:
a21.refreshReferenceObject(KFSPropertyConstants.COST_SHARE_SOURCE_ACCOUNT); 

Is it the a21 object that is the concern?

src/main/java/edu/cornell/kfs/coa/businessobject/SubAccountGlobal.java
db483: As far as I understand the requirements they only want one ICR account updated by one ICR entry on the global doc so if one match is found there is no need to look further, that is why I used the break. What would be your suggested change here?

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalRule.java
cah292: Sorry, wrong example; this particular one looks fine.

src/main/java/edu/cornell/kfs/coa/businessobject/SubAccountGlobal.java
cah292: The thing that concerned me was the use of a "break" statement to terminate the loop early, as opposed to a more iterative-friendly approach. However, doing the latter would either require using an index variable to loop through the collection items (only doable for a list), or using an Iterator.

cah292:
Finished review, awaiting recommended changes before merging.

cah292:
Daniela, another pull request was recently merged that contained updates to some of the files modified by this one, and now this pull request is no longer in a state to be auto-merged. Please rebase as appropriate.

db483:
Hi Chad, I have updated the PR. Please review the updated code at your convenience. Thank you.

bhutchinson:
There are two failing tests. These tests failed because of a SciQuest outage (see https://usertest.sciquest.com/apps/Router/POXMLImport detailing an outage between 03/18/2016 08:00 AM EST and 03/20/2016 12:00 PM EST) and I'm fine with the PR being merged with failing tests in this instance.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -0,0 +1,392 @@
...
+	/**
+	 * @see org.kuali.rice.kns.maintenance.rules.MaintenanceDocumentRuleBase#processCustomSaveDocumentBusinessRules(org.kuali.rice.kns.document.MaintenanceDocument)
+	 */
+	@Override
+	protected boolean processCustomSaveDocumentBusinessRules(MaintenanceDocument document) {
+		boolean success = super.processCustomSaveDocumentBusinessRules(document);
+		SubAccountGlobal subAccountGlobal = (SubAccountGlobal) getNewBo();
+		
+		if( ObjectUtils.isNotNull(subAccountGlobal.getIndirectCostRecoveryAccounts()) && subAccountGlobal.getIndirectCostRecoveryAccounts().size() >0){
+			validateIndirectCostRecoveryAccounts(subAccountGlobal.getIndirectCostRecoveryAccounts());
```
cah292: Just wanted to confirm if the "save" operation is intended to succeed even if validation errors happen (likewise with the other updated business rules class in this changeset). I believe other global documents have similar functionality.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalRule.java
```diff
@@ -93,4 +148,308 @@ public boolean checkSubAccountDetail(SubAccountGlobalDetail dtl) {
...
+        subAccountGlobalDetail.refreshReferenceObject(KFSPropertyConstants.ACCOUNT);
+        if (ObjectUtils.isNotNull(subAccountGlobalDetail.getAccount())) {
+        	Account account = subAccountGlobalDetail.getAccount();
+        	account.refreshReferenceObject(KFSPropertyConstants.SUB_FUND_GROUP); 
+            if (ObjectUtils.isNotNull(account.getSubFundGroup())) {
+
+                // compare them, exit if the account isn't for contracts and grants
+                if (!getSubFundGroupService().isForContractsAndGrants(account.getSubFundGroup())) {
+
+                    if (checkCgCostSharingIsEmpty(subAccountGlobalDetail) == false) {
```
cah292: It would be better to use "!" instead of "== false" here and in other places in this same file. Not sure if it's worth changing at this point unless other updates are needed.

cah292:
Finished second review of changes. Found one piece of functionality that I wanted to confirm the intent of, plus some other minor suggested boolean expression improvements. Unless either of these is serious enough to justify another update, I'm fine with proceeding with the merge.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalRule.java
db483: This is a good point. I copied this code from SubAccountRule class and did not think of changing it.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
db483: Chad, that is correct, the intention here was for the doc to be saved even if errors occurred. Validation errors are still displayed but don't stop the doc from being saved. I checked on the other Global docs and as you mentioned the functionality was similar.

cah292:
Given that the boolean expression segments were copies of those from an existing KFS class, I will defer to others on whether they think this is something worth correcting before the merge. If other developers are fine with it as-is, then I will proceed with merging.

bhutchinson:
I'm fine with proceeding with the merge since this is a copy of pre-existing code. However, I would like a JIRA created in the Technical Work queue to address changing these at some point in the future.

PR #66, opened 2016-03-16
KFSPTS-5332: Updated PREQ creation process to copy the doc header explanation from the PO.

cah292:
This is a quick change to copy the PO's document explanation over to the PREQ's document explanation.


db483:
Looks good.

PR #68, opened 2016-03-17
KFSPTS-4670: Added auto-PK-populate field on Security Model document collection, to fix an issue preventing multiple definitions from being added to the model.

cah292:
This is a reimplementation of an earlier fix for allowing multiple security definition usage on the Security Model document. The old fix was following our project conventions for introducing a new module configuration child bean in each module; however, doing so on the SEC module was causing it to overwrite some of our overrides from our other customized modules. This new fix avoids the issue by overriding an intermediate SEC module configuration parent bean already provided by KFS.


bhutchinson:
Looks fine to me.

PR #69, opened 2016-03-21

db483:
Hi Chad, After testing in DEV I found a couple of issues with some reference objects not being refreshed and fixed that. I have also changed the order for the ICR rate and type to match the Sub-Account doc.  Please review at your convenience. Thank you.


cah292:
Looks good to me.

PR #70, opened 2016-03-23
KFSPTS-5408: Updated 1099/1042S transaction detail output and transaction override input files to use a tab-delimited format instead, to make it easier to modify such files via Microsoft Excel.

cah292:
The existing 1099/1042S transaction detail output file and transaction override input file use a fixed-width format that is not flexible enough for what our tax managers need. This change updates both to use a simpler tab-delimited file format instead (and with the input file no longer needing all the fields of the generated output file). This will allow our tax managers to import/export the data using Microsoft Excel, which supports the use of tab-delimited text files.


src/main/java/edu/cornell/kfs/tax/batch/TransactionOverrideCsvBatchInputFileType.java
```diff
@@ -0,0 +1,443 @@
...
+            valid = false;
+        } else if (box1099.length() > CUTaxConstants.TAX_1099_MAX_BUCKET_LENGTH && !NULL_STRING.equalsIgnoreCase(box1099)) {
+            LOG.error("Found a line with a 1099 box number that is too long. Line number: " + Integer.toString(lineNumber));
+            valid = false;
+        } else {
+            // Validation succeeded; configure override of a 1099 tax box.
+            transOverride.setTaxType(CUTaxConstants.TAX_TYPE_1099);
+            transOverride.setBoxNumber(NULL_STRING.equalsIgnoreCase(box1099) ? CUTaxConstants.TAX_1099_UNKNOWN_BOX_KEY : box1099);
+        }
+
```
nkimble: This method seems to imply that the users can mix 1099 and 1042S overrides in the same file.  I thought that those overrides were tax form specific.  If they are, should this method somehow deal with that to let them know?  Maybe with a parameter specifying which tax form type the caller believes the data should be from?

cah292: The TransactionOverride objects do have a tax type field to indicate which type of tax processing the override is for. However, the current transaction detail table (which the transaction detail output file sources from) uses separate columns for 1099 and 1042S box numbers, so that's reflected in the output file. I went with preserving the column separation in the input file so that, if the tax managers just wanted to modify the output file to create the input one, they wouldn't have to add a new column.

There are several other places in the code that rely on the transaction detail table having separate 1099 and 1042S box number columns, so I do not want to change those at this time. If you think that the raw transaction detail output file should consolidate those columns and add a new tax-type indicator column instead, let me know and I can check if the tax managers are okay with that change.

I would prefer not to add a new parameter for controlling what tax type the uploaded overrides are for, since I think that would be overkill.

nkimble: We want to minimize the labor intensity of the process for the users. Your implementation and explanation indicates that.

nkimble:
Merging. Looks good.

PR #71, opened 2016-03-23

db483:
Hi Chad,

I had to revert and add a pre-rule for the SubAccount Global doc. The only change is the addition of the SubAccountGlobalPreRule. 

Thank you.


cah292:
Changes look good.

PR #65, opened 2016-03-16
KFSPTS-5032: Clear payee on DV copy when payee type is invalid for pa…

bhutchinson:
…yment reason

PR #72, opened 2016-03-23
KFSPTS-5032: Clear payee on DV copy when payee type is invalid for pa…

bhutchinson:
…yment reason, set tripAssociationStatusCode to the false code on a copied DV (and initialize to the false code on a new DV) instead of null to avoid unnecessary service calls to the travel system.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1217,6 +1217,13 @@
         <scope>provided</scope>
     </dependency>
 
+    <dependency>
+        <groupId>org.easymock</groupId>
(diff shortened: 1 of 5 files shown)


bhutchinson:
This is mostly the same as PR #65 which I had to manually revert, with the addition that the tripAssociationStatusCode is now defaulted to the false value instead of null, and it is set to the false value on a copied DV.

jhulslander:
works for me

PR #73, opened 2016-03-28
KFSPTS-5322 excluding inactive ICR records from GL Poster Job

jhulslander:
This is a cu customization of PostExpenditureTransaction.hasExclusionByAccount where we only want active rules, the KualiCo code doesn't consider the active/inactive flag.  I also created a new unit test for this functionality.  It requires the spring context as it needs to query the database.

### DIFF
--- a/src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPostExpenditureTransaction.java
+++ b/src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPostExpenditureTransaction.java
@@ -1,0 +1,49 @@
+package edu.cornell.kfs.gl.batch.service.impl;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.kuali.kfs.coa.businessobject.Account;
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryExclusionAccount;
+import org.kuali.kfs.coa.businessobject.ObjectCode;
+import org.kuali.kfs.gl.batch.service.impl.PostExpenditureTransaction;
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.krad.service.BusinessObjectService;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+public class CuPostExpenditureTransaction extends PostExpenditureTransaction {
+	
+	private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(CuPostExpenditureTransaction.class);
+	
+	/**
+     * Determines if there's an exclusion by account record for the given account and object code
+     * @param account the account to check
+     * @param objectCode the object code to check
+     * @return true if the given account and object code have an exclusion by account record, false otherwise
+     */
+	@Override
+    public boolean hasExclusionByAccount(Account account, ObjectCode objectCode) {
+    	if (LOG.isDebugEnabled()) {
+            LOG.debug("CuPostExpenditureTransaction/hasExclusionByAccount() started, looking up accountL " + account.getAccountNumber() + " and objectCode " + objectCode.getCode());
+        }
+        Map<String, Object> keys = new HashMap<String, Object>();
+        keys.put(KFSPropertyConstants.CHART_OF_ACCOUNTS_CODE, account.getChartOfAccountsCode());
+        keys.put(KFSPropertyConstants.ACCOUNT_NUMBER, account.getAccountNumber());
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPostExpenditureTransaction.java
```diff
@@ -0,0 +1,53 @@
...
+import org.kuali.kfs.gl.batch.service.impl.PostExpenditureTransaction;
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.krad.service.BusinessObjectService;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+public class CuPostExpenditureTransaction extends PostExpenditureTransaction {
+	
+	private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(CuPostExpenditureTransaction.class);
+	public static final String ACTIVE_INDICATOR_FIELD = "ACTV_IND";
```
cah292: Any particular reason why the table column name is being referenced here instead of the property name? I would recommend using the existing equivalent Rice or KFS property name constant, unless it only works when the table column is specified.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPostExpenditureTransaction.java
```diff
@@ -0,0 +1,53 @@
...
+	
+	public CuPostExpenditureTransaction() {
+		super();
```
cah292: If all this subclass constructor does is call a public default constructor on the superclass, then you may not need this constructor.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPostExpenditureTransaction.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+	
+	/**
+     * Determines if there's an exclusion by account record for the given account and object code
+     * @param account the account to check
+     * @param objectCode the object code to check
+     * @return true if the given account and object code have an exclusion by account record, false otherwise
+     */
+    public boolean hasExclusionByAccount(Account account, ObjectCode objectCode) {
```
cah292: If you're overriding a superclass method, I would suggest adding the "@Override" annotation for clarity.

src/main/resources/edu/cornell/kfs/gl/cu-spring-gl.xml
```diff
@@ -338,5 +338,22 @@
 		<property name="persistenceStructureService">
 			<ref bean="persistenceStructureService" />
 		</property>
-	</bean> 	 		     
+	</bean> 	 
+	
+	<bean id="glPostExpenditureTransaction" parent="glPostExpenditureTransaction-parentBean" />
+	<bean id="glPostExpenditureTransaction-parentBean" class="edu.cornell.kfs.gl.batch.service.impl.CuPostExpenditureTransaction" abstract="true">
```
cah292: You don't need to override the parent bean in this case. You just need the "glPostExpenditureTransaction" override and then set the "class" on it.

src/test/resources/org/kuali/kfs/gl/spring-gl.xml
```diff
@@ -1157,7 +1157,7 @@
 	</bean>
 
 	<bean id="glPostExpenditureTransaction" parent="glPostExpenditureTransaction-parentBean" />
-	<bean id="glPostExpenditureTransaction-parentBean" class="org.kuali.kfs.gl.batch.service.impl.PostExpenditureTransaction" abstract="true">
+	<bean id="glPostExpenditureTransaction-parentBean" class="edu.cornell.kfs.gl.batch.service.impl.CuPostExpenditureTransaction" abstract="true">
```
cah292: Do the unit tests load the "cu-spring-gl.xml" file? If so, then you may not need this change.

cah292:
Finished initial round of reviewing. I did find a couple minor items that should be addressed, but other than that it looks good. Awaiting Jay's feedback and updates before proceeding further.

jhulslander:
Thanks Chad, I will make adjustments as noted.  Concerning use of ACTIVE_INDICATOR_FIELD = "ACTV_IND"; ... I tied using the rice parameter to get to the field and had a weird error.  I circumvented the error by using the database field name.  I will dig a bit deeper tomorrow.

jhulslander:
test this please

jhulslander:
@cah292 I implemented your recommendations.  Please review when you're ready

cah292:
The revised changes look good. Once I receive confirmation of the unit tests passing, I will proceed with the merge.

PR #74, opened 2016-03-29
KFSPTS-5214: Changed the Receipt Processing directory path to one that is relative to the KFS staging directory.

cah292:
This updates the Receipt Processing Service to use a different input/output file directory path. The new path is relative to the KFS staging directory.

The original version of this enhancement was going to use a property to control the whole path. However, after further discussion, it was decided to use this other path instead, and the old enhancement version has been reverted.


nkimble:
Checks have passed. Change looks fine.

PR #75, opened 2016-03-31
KFSPTS-5465: Updated Tax Processing Job to automatically mask certain confidential or potentially-sensitive output fields when a particular config property has been set.

cah292:
This change will cause the Tax Processing Job to automatically mask/scrub confidential and certain potentially-sensitive fields in the output files, if a new "tax.output.scrubbed" config property has been set accordingly. This will allow developers to easily run the batch job locally without worrying about confidential data getting onto their systems.

This pull request was previously going to just add some duplicate output config files with the confidential parts "scrubbed" out, but based on Nancy's feedback, the other solution mentioned above is now being proposed instead.

### DIFF
--- a/src/main/java/edu/cornell/kfs/tax/CUTaxConstants.java
+++ b/src/main/java/edu/cornell/kfs/tax/CUTaxConstants.java
@@ -31,6 +31,9 @@ public final class CUTaxConstants {
     public static final String TAX_1099_UNKNOWN_BOX_KEY = "???";
     public static final String TAX_1042S_UNKNOWN_BOX_KEY = "????";
     public static final String ANY_OR_NONE_PAYMENT_REASON = "*";
+    public static final String MASKED_VALUE_9_CHARS = "XXXXXXXXX";
+    public static final String MASKED_VALUE_11_CHARS = "XXXXXXXXXXX";
+    public static final String MASKED_VALUE_19_CHARS = "XXXXXXXXXXXXXXXXXXX";
     public static final int INSERT_BATCH_SIZE = 500;
     public static final int TAX_1099_MAX_BUCKET_LENGTH = 3;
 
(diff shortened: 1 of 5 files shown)


nkimble:
I really appreciate the need for this type of configuration change but I have some concerns with this proposed implementation.  These code changes duplicate and slightly modify three existing output XML files increasing the maintenance burden in a process that already is complicated by the shear amount of data that must be manipulated and formatted.

Could a different approach be taken where the data that needs to be scrubbed is replaced with neutral values at the point where the data is either stored in the tables used in the tax processing OR updated to neutral values in the tax processing tables after the processing is completed but prior to the data being retrieved for output OR updated to neutral values during the actual output process in a way that the XML does not have to be duplicated?

cah292:
Yes, that makes sense. I will work on changing the solution so that it can scrub automatically based on a config property, which will cut down on maintenance in the long run. Thanks for pointing this out!

cah292:
This pull request has been updated to instead rely on a new config property to control whether the confidential or potentially-sensitive sections should be auto-"scrubbed" by the batch job. Please review again when you get a chance.

cah292:
Did some more analysis and had a discussion with Bryan concerning the unit test that failed. It seems that the failed test is related to some changes made for KFSPTS-5322, where the test is expecting this one ICR Exclusion Account to be inactive but it is actually active. The problematic ICR Exclusion Account is indeed active in UTEST and DEV, but it is inactive in TEST. Not sure if users made some data changes for their DEV validation that resulted in this issue.

Since this unit test failure appears to be a data-related one and not a consequence of these particular code changes, I think it is safe to merge the pull request if there are no complaints about the proposed changes.

nkimble:
Code changes look fine and reason for unit test failure is acceptable for the PR to occur.

PR #77, opened 2016-04-07
KFSPTS-4490: Added checkbox on REQS notes to indicate whether they should be copied to the PO.

cah292:
This feature allows users to specify whether REQS notes without attachments should copy over to the PO, by means of a new extended attribute for KRAD Note objects. (REQS notes with attachments will always copy to the PO if they were created before routing termination, and the UI will prevent editing of that flag on such lines, including addLines that specify an attachment.) Daniela originally started the development of the feature, and I finished up the remainder.

Some new SQL is required for this change, and is attached to the story JIRA. In addition, a new KRAD override file has been added, which will require updating a KFS config property for KRAD overrides. This change has already been made for the config file used by the PR-triggered unit tests.

There were some places where OJB's persistence functionality would ordinarily prevent successful saving of the new extended attribute in certain cases, such as when generating a "copied from document" note or copying IWNT attachments to the new REQS document. Some workarounds have been added to such areas accordingly.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/IWantDocumentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/IWantDocumentServiceImpl.java
@@ -10,6 +10,7 @@ import java.util.Map;
 
 import javax.mail.MessagingException;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.fp.document.DisbursementVoucherDocument;
 import org.kuali.kfs.fp.document.web.struts.DisbursementVoucherForm;
@@ -35,7 +36,6 @@ import org.kuali.kfs.vnd.businessobject.VendorAddress;
 import org.kuali.kfs.vnd.document.service.VendorService;
 import org.kuali.kfs.vnd.service.PhoneNumberService;
 import org.kuali.rice.core.api.mail.MailMessage;
-import org.kuali.rice.core.api.util.ConcreteKeyValue;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 import org.kuali.rice.coreservice.framework.parameter.ParameterService;
 import org.kuali.rice.kew.api.KewApiConstants;
@@ -49,7 +49,6 @@ import org.kuali.rice.kim.api.identity.entity.Entity;
 import org.kuali.rice.kim.api.identity.principal.Principal;
 import org.kuali.rice.kim.api.services.KimApiServiceLocator;
 import org.kuali.rice.krad.bo.Attachment;
-import org.kuali.rice.krad.bo.DocumentHeader;
 import org.kuali.rice.krad.bo.Note;
 import org.kuali.rice.krad.bo.PersistableBusinessObject;
 import org.kuali.rice.krad.exception.InvalidAddressException;
@@ -64,6 +63,7 @@ import org.kuali.rice.krad.util.KRADConstants;
 import org.kuali.rice.krad.util.ObjectUtils;
 
 import edu.cornell.kfs.fp.document.CuDisbursementVoucherDocument;
+import edu.cornell.kfs.krad.businessobject.NoteExtendedAttribute;
 import edu.cornell.kfs.module.purap.CUPurapConstants;
 import edu.cornell.kfs.module.purap.businessobject.IWantAccount;
 import edu.cornell.kfs.module.purap.businessobject.IWantDocUserOptions;
@@ -324,7 +324,7 @@ public class IWantDocumentServiceImpl implements IWantDocumentService {
         purapService.saveDocumentNoValidation(requisitionDocument);
 
         // copy attachments from I Want document
-        copyIWantDocAttachments(requisitionDocument, iWantDocument, false);
+        copyIWantDocAttachments(requisitionDocument, iWantDocument, false, true);
 
         // set up deliver to section
         setUpDeliverToSectionOfReqDoc(requisitionDocument, iWantDocument, purapService);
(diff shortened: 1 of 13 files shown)


src/main/resources/edu/cornell/kfs/module/purap/document/datadictionary/RequisitionDocument.xml
```diff
@@ -25,6 +25,15 @@
    	<ref bean="RequisitionDocument-vendorEmailAddress"/>
    </list>
    </property>
+   <property name="webScriptFiles" ref="reqsWebScriptFiles" />
```
cah292: Because the parent bean uses a factory to create the list, I was not able to use <list merge="true"> directly, and instead had to extend the factory and add it to its "sourceList" property.

cah292:
retest this please

cah292:
test this please

src/main/webapp/scripts/module/purap/reqsNotes.js
```diff
@@ -0,0 +1,16 @@
...
+    var copyToPOElem = document.getElementById("newNote.extension.copyNoteIndicator");
+    var forceCopyYesLabel = document.getElementById("newNoteForceCopyYesLabel");
+    if (copyToPOElem) {
+        copyToPOElem.checked = "checked";
+        copyToPOElem.style.display = "none";
+        if (forceCopyYesLabel) {
+            forceCopyYesLabel.style.display = "inline";
+        }
+    }
+}
```
bhutchinson: There's probably no way to accomplish this functionality without using JavaScript is there?

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/IWantDocumentServiceImpl.java
```diff
@@ -558,12 +558,17 @@ private void setUpVendorSectionForReqDoc(RequisitionDocument requisitionDocument
...
+     * 
      * @param document
      * @param iWantDocument
      * @param copyConfidentialAttachments
+     * @param extCopyNoteIndicator
      * @throws Exception
      */
-    private void copyIWantDocAttachments(AccountingDocument document, IWantDocument iWantDocument, boolean copyConfidentialAttachments) throws Exception {
+    private void copyIWantDocAttachments(AccountingDocument document, IWantDocument iWantDocument, boolean copyConfidentialAttachments,
+            Boolean extCopyNoteIndicator) throws Exception {
```
bhutchinson: Is there a reason this can't be a primitive boolean and callers just pass in true or false?

cah292: I suppose it could be reduced to just a primitive boolean for the current use cases. I had left it as a Boolean in case we needed the ability to create extension attributes with copyNoteIndicator set to false; however, if we don't expect to be needing that at all, then I can change it to be just a primitive.

src/main/webapp/scripts/module/purap/reqsNotes.js
cah292: Not that I'm aware of, unless we want to forcibly overwrite the copy-to-PO flag on the server after the form gets submitted. As per the user story, the functionals want notes with attachments to copy to the PO, so this seemed like the most user-friendly way of doing that.

I may need to verify with the functionals whether attachments created after REQS finalization should still copy to the PO if the document did not become an APO.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/IWantDocumentServiceImpl.java
bhutchinson: Unless we know we'll need that future case, I'd change it to the primitive which is simpler.

src/main/webapp/scripts/module/purap/reqsNotes.js
bhutchinson: Generally, I would prefer to do things on the server and not depend on JavaScript, but if that doesn't make sense in this case, I'm ok with it as is.

cah292:
The code has been updated based on feedback from the review.

I do have one pending question to the functionals about how this should behave on certain finalized REQS docs, so please hold off on merging until I get a response to that.

db483:
Hi Chad, 

I wonder if your question may be already answered in the KFS Change Request Note to PO.docx file attached to KFSPTS-4490 jira. In the Assumptions section it says: 

"After the REQ becomes final, this note would be subject to the same warning that the note was added after the REQ was finalized and will not pass forward to the PO. "

cah292:
Hi Daniela,

Thanks for pointing that out. However, based on Cathy's and Karen's comments on the ticket as well as the current code, it looks like attachments added to finalized REQS may still copy over to the PO if it did not become an APO and the PO did not get generated yet. I wanted to confirm with the functionals on whether we need to preserve that behavior.

jhulslander:
test this please

bhutchinson:
retest this please

cah292:
As per discussion with Bryan and notes on the JIRA, I have updated the feature to follow the existing behavior of still allowing copy-to-PO if the PO has not been created yet (which is the case if the REQS did not generate an APO). Let me know when you are ready for me to modify the unit test config file so that the tests can work properly here.

bhutchinson:
Chad, go ahead and modify the unit test config file. If the tests pass, I'll merge this. Thanks.

cah292:
Unit test config has been updated. Note that DEV config will need updating after the merge, though.

cah292:
retest this please

bhutchinson:
good point about dev, thanks

cbower4545:
I added it to puppet for dev. I'll try to get the elb work done tonight so
it comes up clean tomorrow.

bhutchinson:
Thanks Chad

PR #80, opened 2016-04-12
KFSPTS-5340: refactor unit tests to eliminate spring dependency where…

bhutchinson:
… possible


src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardErrorEmailServiceImplTest.java
```diff
@@ -1,40 +1,132 @@
...
-		TestUtils.setSystemParameter(ProcurementCardDocument.class, "PCARD_UPLOAD_ERROR_EMAIL_ADDR", "fake@cornell.edu");
-
 		ArrayList<String> errorMessages = new ArrayList<String>();
 		errorMessages.add("TEST ERROR FOR PROCUREMENT CARD");
 		procurementCardErrorEmailService.sendErrorEmail(errorMessages);
-
-		
 	}
     
+	private class MockParameterService implements ParameterService {
```
cah292: Would it make more sense to define some of these "mock" services in their own separate files instead, to make them reusable across multiple tests?

cah292:
Looks good overall, though I did have a question about whether the "mock" services should be moved to their own files. I also noticed a removed test that only tested coverage and did not assert anything; just wanted to verify that that test was indeed not adding any value.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardErrorEmailServiceImplTest.java
bhutchinson: I did create two versions of the "mock" ParameterService. Let me see if I can generalize that.

bhutchinson:
I generalized the MockParameterService since that was used twice. I left the other "mock" services as is since they were only used once; we can refactor and generalize them in the future if we want them to be reused. The test I removed didn't assert anything and the method it called is empty, so I'm comfortable removing it.

bhutchinson:
retest this please

cah292:
Changes look good, and the unit tests have passed. Merging.

PR #81, opened 2016-04-13
KFSPTS-4278 Change attachments folder; remove rsmartTypes.xsd reference

db483:
Batch job for this feature already exists but it is not yet set up to run through Tidal. The service used by the batch job is currently used by a manual DV batch upload link in KFS.
The only code changes were:
- create a separate folder for attachment files: /staging/fp/disbursementVoucher/attachment. This is to keep the files clean: data files and attachments separate, and also to keep this consistent with other similar KFS batch jobs.
- remove reference to rsmartTypes.xsd as it was not used, file did not even exist in our project and it was confusing for our customers

### DIFF
--- a/src/main/resources/com/rsmart/kuali/kfs/fp/spring-fp.xml
+++ b/src/main/resources/com/rsmart/kuali/kfs/fp/spring-fp.xml
@@ -100,7 +100,7 @@
 		<property name="vendorService" ref="vendorService"/>
 		<property name="batchFeedHelperService" ref="batchFeedHelperService"/>
 		<property name="attachmentsPath">
-	       <value>${staging.directory}/fp/disbursementVoucher</value>
+	       <value>${staging.directory}/fp/disbursementVoucher/attachments</value>
 	    </property>
 	</bean>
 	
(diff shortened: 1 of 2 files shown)


src/main/resources/com/rsmart/kuali/kfs/fp/spring-fp.xml
```diff
@@ -100,7 +100,7 @@
 		<property name="vendorService" ref="vendorService"/>
 		<property name="batchFeedHelperService" ref="batchFeedHelperService"/>
 		<property name="attachmentsPath">
-	       <value>${staging.directory}/fp/disbursementVoucher</value>
+	       <value>${staging.directory}/fp/disbursementVoucher/attachment</value>
```
bhutchinson: Should this be attachments (plural) to be consistent with our main attachments directory?

db483: That would make sense. I will update it. I was trying to be consistent with the existing vendor batch job that reads the attachments from: staging/vnd/vendorBatch/attachment forlder. See VendorBatchServiceImpl line 551.

bhutchinson: Thanks, I like consistency, but I should know better with KFS. :-)

db483:
I changed the folder name to 'attachments'.

bhutchinson:
Thanks Daniela

bhutchinson:
retest this please

bhutchinson:
retest this please

PR #82, opened 2016-04-14
KFSPTS-5409 mapping payment reasons to 1099 tax box

jhulslander:
Please don't merge this yet, I need to some more testing in regards to the "mapped to no boxes" configuration.  I would like to get some feedback on this code as I complete testing.

### DIFF
--- a/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
+++ b/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
@@ -18,6 +18,7 @@ import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.fp.document.DisbursementVoucherConstants;
+import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.rice.core.api.CoreApiServiceLocator;
 import org.kuali.rice.core.api.encryption.EncryptionService;
 import org.kuali.rice.krad.util.KRADConstants;
@@ -28,6 +29,7 @@ import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.DerivedValuesRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.TransactionDetailRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.VendorAddressRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.VendorRow;
+import edu.cornell.kfs.tax.service.PaymentReason1099BoxService;
 
 /**
  * Default TransactionRowProcessor implementation for handling 1099 tax processing.
@@ -238,6 +240,8 @@ public class TransactionRow1099Processor extends TransactionRowProcessor<Transac
     private OriginSpecificStats dvStats;
     private OriginSpecificStats pdpStats;
     private OriginSpecificStats currentStats;
+
+	private PaymentReason1099BoxService paymentReason1099BoxService;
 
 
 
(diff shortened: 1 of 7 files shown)


src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
```diff
@@ -1127,6 +1131,14 @@ private void processExclusionsAndAmounts(ResultSet rs, TransactionDetailRow deta
                         .append(paymentDateP.value).append(';')
                         .append(docNumberP.value).append(';')
                         .append(docLineNumberP.value).toString());
+        
+        //Determine if there is override of tax box due to the payment reason
+        if (getPaymentReason1099BoxService().isPaymentReasonMappedTo1099Box(paymentReasonCodeP.value)) {
+			String mappedBox = getPaymentReason1099BoxService().getPaymentReason1099Box(paymentReasonCodeP.value);
+			LOG.info("Overriding to tax box " + mappedBox + "  because of payment reason " + paymentReasonCodeP.value + " for vendor " + vendorNameForOutput);
```
bhutchinson: Should this be debug level, or is this something that we'll want to generally see in our logs?

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
```diff
@@ -1146,44 +1156,52 @@ private void processExclusionsAndAmounts(ResultSet rs, TransactionDetailRow deta
...
-            }
-            
-        } else {
-            // Otherwise, do not update amounts.
+        //If the payment reason has been marked to go to NO tax boxes, then don't update the tax boxes.
+        if (getPaymentReason1099BoxService().isPaymentReasonMappedToNo1099Box(paymentReasonCodeP.value)) {
+        	// Do not update amounts.
             rs.updateString(detailRow.form1099Box.index, CUTaxConstants.TAX_1099_UNKNOWN_BOX_KEY);
             numNoBoxDeterminedRows++;
+            LOG.info("Overriding to NO tax box because of payment reason " + paymentReasonCodeP.value + " for vendor " + vendorNameForOutput);
```
bhutchinson: same question on logging

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+	public String getPaymentReason1099Box(String paymentReasonCode) {
+		return convertCollectionToMap(getPaumentReasonTo1099BoxMappings()).get(paymentReasonCode);
+	}
+
+	@Override
+	public boolean isPaymentReasonMappedToNo1099Box(String paymentReasonCode) {	
+		return getPaumentReasonToNo1099Boxes().contains(paymentReasonCode);
+	}
+	
+	public Collection<String> getPaumentReasonTo1099BoxMappings() {
```
bhutchinson: method name misspelling

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+		return getParameterService().getParameterValuesAsString(CUTaxConstants.TAX_NAMESPACE, CUTaxConstants.TAX_1099_PARM_DETAIL, 
+				Tax1099ParameterNames.PAYMENT_REASON_TO_TAX_BOX);
+	}
+	
+	public Collection<String> getPaumentReasonToNo1099Boxes() {
```
bhutchinson: another misspelling I think

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
jhulslander: Good question.  I thought this would be helpful in the log.  I will check with Cathy

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+	public void setParameterService(ParameterService parameterService) {
+		this.parameterService = parameterService;
+	}
+	
+	/**
+	 * The format of each collection entry must be "R=2" where R is payment reason code, and 2 is a 1099 box
+	 * @param collectionOfStrings
+	 * @return
+	 */
+	public Map<String,String> convertCollectionToMap(Collection<String> collectionOfStrings) {
```
bhutchinson: Is this the only place we have code like this? It seems like something that would/could be a useful utility method to be used elsewhere.

bhutchinson:
I added a few comments - was there anything else in particular you wanted feedback on?

jhulslander:
This is my first time looking at the 1099 code, I was mostly concerned my approach was valid in relation to other 1099 work.

bhutchinson:
Ok, thanks; I can take a look some more tomorrow, and it might also be helpful to ask ChadH to take a look since he has more experience with the 1099 code than I do.

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
jhulslander: There is other conversion utilities, but they aren't sharable at this point.  Do you think I should refactor the tax conversion methods into a more utility setup?

jhulslander:
Thanks, after I make the updates you mentioned, I will ask Chad to take a look

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
bhutchinson: I guess if it's only used here and there's not something else that could be re-used, it's fine. We can refactor it to a more general use utility setup when there is other code that could use it too.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
```diff
@@ -1127,6 +1131,14 @@ private void processExclusionsAndAmounts(ResultSet rs, TransactionDetailRow deta
                         .append(paymentDateP.value).append(';')
                         .append(docNumberP.value).append(';')
                         .append(docLineNumberP.value).toString());
+        
+        //Determine if there is override of tax box due to the payment reason
+        if (getPaymentReason1099BoxService().isPaymentReasonMappedTo1099Box(paymentReasonCodeP.value)) {
+			String mappedBox = getPaymentReason1099BoxService().getPaymentReason1099Box(paymentReasonCodeP.value);
+			LOG.info("Overriding to tax box " + mappedBox + "  because of payment reason " + paymentReasonCodeP.value + " for vendor " + vendorNameForOutput);
+			overrideTaxBox = summary.getBoxNumberConstant(mappedBox);
```
cah292: You should verify whether the tax managers want payment-reason-based tax box overrides to ignore other exclusions that may have been discovered for the given transaction. Your current setup appears to be doing so because of its re-use of the overrideTaxBox variable. If the tax managers do not want this behavior, then I would suggest using a different variable for tracking such overrides. Also, I believe payment reason overrides should *not* overwrite any of those derived from the summary.transactionOverrides field just a couple lines earlier.

src/main/java/edu/cornell/kfs/tax/service/impl/PaymentReason1099BoxServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.coreservice.framework.parameter.ParameterService;
+
+import edu.cornell.kfs.tax.CUTaxConstants;
+import edu.cornell.kfs.tax.CUTaxConstants.Tax1099ParameterNames;
+import edu.cornell.kfs.tax.service.PaymentReason1099BoxService;
+
+public class PaymentReason1099BoxServiceImpl implements PaymentReason1099BoxService, Serializable {
```
cah292: Using a service is one way of finding the parameter values. If you want to explore other options, you can try adding new fields to the Transaction1099Summary class, which already sets up various parameter-based immutable collections and maps based on parameter values. (Such "summary" objects are discarded after the batch job finishes, so they will be recreated with fresh parameter values on the next run.)

jhulslander:
I am going to close the PR

PR #83, opened 2016-04-15
KFSPTS-4278 Change attachments folder; remove rsmartTypes.xsd referen…

db483:
…ce, always remove .done file


db483:
Added another change so that .done files are always removed, even when errors occur during processing.

bhutchinson:
retest this please

PR #84, opened 2016-04-15
KFSPTS-5185: Updated Account Reversion Global document to allow changing active status of reversions and adding new reversions, and made some other misc fixes to the document.

cah292:
This change makes the following updates to the Account Reversion Global document:

[1] Added the ability to create new Account Reversions, not just update existing ones. (If creating new ones, all fields become required to ensure that the new reversions are populated accordingly.)

[2] Added a new property for bulk changes of active/inactive status on Account Reversions; the story ticket has the needed SQL for the new property.

[3] Fixed some miscellaneous issues with the existing document, such as business rule problems, references to Org Reversion, and JavaScript-based UI label updates.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/businessobject/AccountReversionGlobal.java
+++ b/src/main/java/edu/cornell/kfs/coa/businessobject/AccountReversionGlobal.java
@@ -32,18 +32,19 @@ import org.kuali.rice.krad.bo.GlobalBusinessObject;
 import org.kuali.rice.krad.bo.GlobalBusinessObjectDetail;
 import org.kuali.rice.krad.bo.PersistableBusinessObject;
 import org.kuali.rice.krad.bo.PersistableBusinessObjectBase;
-import org.kuali.rice.krad.service.BusinessObjectService;
 import org.kuali.rice.krad.service.PersistenceStructureService;
+import org.kuali.rice.krad.util.ObjectUtils;
 
 import edu.cornell.kfs.coa.service.AccountReversionService;
 
 /**
- * The representation of a Global Organization Reversion. A Global Organization Reversion is made up of three sections: 1. The
- * University Fiscal Year and Chart of Accounts code for the Organizations going through reversion, with some account information.
- * 2. A list of the appropriate Object Reversion Details 3. A list of Organizations to apply the Organization Reversion to
+ * The representation of a Global Account Reversion. A Global Account Reversion is made up of three sections: 1. The
+ * University Fiscal Year for the Accounts going through reversion, with some account information.
+ * 2. A list of the appropriate Object Reversion Details 3. A list of Accounts to apply the Account Reversion to
  */
 public class AccountReversionGlobal extends PersistableBusinessObjectBase implements GlobalBusinessObject {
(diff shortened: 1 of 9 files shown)


src/main/java/edu/cornell/kfs/coa/businessobject/AccountReversionGlobal.java
```diff
@@ -272,7 +274,19 @@ public void setCashReversionFinancialChartOfAccountsCode(String cashReversionFin
         this.cashReversionFinancialChartOfAccountsCode = cashReversionFinancialChartOfAccountsCode;
     }
 
-    /**
+    public Boolean isReversionActiveIndicator() {
```
jhulslander: do we need both isReversionActiveIndicator and getReversionActiveIndicator

cah292: I think getReversionActiveIndicator is the only one strictly needed since the property is a Boolean object and not a boolean primitive. There is another property on the same class that also has this duplication, though, so I added both for consistency with that property. If you think it's too redundant, I can remove the extra method.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/AccountReversionGlobalRule.java
```diff
@@ -159,17 +162,14 @@ public boolean processCustomAddCollectionLineBusinessRules(MaintenanceDocument d
...
         else if (line instanceof AccountReversionGlobalAccount) {
             AccountReversionGlobalAccount acct = (AccountReversionGlobalAccount) line;
-            if (!checkEmptyValue(acct.getChartOfAccountsCode())) {
-                GlobalVariables.getMessageMap().putError("chartOfAccountsCode", KFSKeyConstants.ERROR_REQUIRED, "Chart of Accounts Code");
-            }
-            if (!checkEmptyValue(acct.getAccountNumber())) {
-                GlobalVariables.getMessageMap().putError("accountNumber", KFSKeyConstants.ERROR_REQUIRED, "Account Number");
+            if (!checkEmptyValue(acct.getChartOfAccountsCode()) || !checkEmptyValue(acct.getAccountNumber())) {
+                // Skip Most validation if chart or account are empty. The default required-field checking will populate the error map accordingly.
+                success = false;
```
jhulslander: Since a rule failed, don't we want to give the user some information?

cah292: Both the chart code and account number are already configured as required fields, so the other business rules processing will already add appropriate error messages to the screen. This check is just needed to help skip some other validation that should be ignored if they are not both specified.

src/main/java/edu/cornell/kfs/coa/businessobject/AccountReversionGlobal.java
jhulslander: Consistency is better, let's leave as is

src/main/java/edu/cornell/kfs/coa/document/validation/impl/AccountReversionGlobalRule.java
jhulslander: perfect

jhulslander:
This looks good, I will merge

PR #85, opened 2016-04-15
KFSPTS-5409 mapping payment reasons to 1099 tax box

jhulslander:

@bhutchinson since you also reviewed the original PR with Chad H, I am pinging you in as well


cah292:
This looks good overall. My only concern is that the parameter maps on the service are going to keep getting reconstructed for every transaction row, which could be thousands of times per job run. If this happens to be only a very minor increase to the job run time, then I am okay with that.

Will wait for Bryan's feedback before merging.

jhulslander:
Also, I have completed my local testing, and everything seems OK.  Unless code review concerns prop up, I am ready for this to be merged. 

@cah292 Are we OK with testing run times with my current design in kfs-dev, or should I reevaluate the solutions?

cah292:
As long as we do not expect either parameter map to contain a large number of items that could slow down map creation, then I think this setup is okay due to Rice ParameterService caching of the values collections.

jhulslander:
It's a very small starting list, so I think we should be OK

cah292:
I believe this is okay to move forward to DEV, due to the number of possible payment reasons being quite small.

PR #86, opened 2016-04-18
KFSPTS-4490: Added checkbox on REQS notes to indicate whether they should be copied to the PO.

cah292:
This is an updated version of the feature originally committed for PR #77 but was eventually reverted. This version is more explicit about always ensuring that notes with attachments will copy over to the PO, regardless of the state of the new copy-to-PO checkbox on the form or behind-the-scenes.


cah292:
Looks like the PR was merged before the unit tests finished running. Fortunately, as per build 357 of the Jenkins kuali-test-develop-pr job, all the unit tests for this change passed.

PR #87, opened 2016-04-18
KFSPTS-5483: Added new parameter to control the Scrubber's generation of plant indebtedness entries based on document type.

cah292:
Users need to be able to limit generation of plant indebtedness entries based on document type, similar to how an existing parameter is used for limiting capitalization entry generation. This feature adds a new parameter to control the inclusion/exclusion of doc types for plant indebtedness processing by the Scrubber. It is very similar to the related parameter for capitalization entry generation.

This has been implemented to limit the amount of code needing overrides, while also following similar logic from the Scrubber's capitalization entry doc-type-based handling.

### DIFF
--- a/src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
+++ b/src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
@@ -20,10 +20,14 @@ import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.KFSKeyConstants;
 import org.kuali.kfs.sys.Message;
 import org.kuali.kfs.sys.businessobject.SystemOptions;
+import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.kfs.sys.exception.InvalidFlexibleOffsetException;
+import org.kuali.rice.core.api.parameter.ParameterEvaluator;
+import org.kuali.rice.core.api.parameter.ParameterEvaluatorService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 import org.kuali.rice.krad.util.ObjectUtils;
 
+import edu.cornell.kfs.gl.CuGeneralLedgerConstants;
 import edu.cornell.kfs.sys.CUKFSKeyConstants;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/kuali/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
```diff
@@ -309,4 +313,26 @@ protected TransactionError setupEntryWithPotentialContinuation(OriginEntryFull c
...
+     * Overridden to also restrict plant indebtedness based on document type,
+     * by means of a custom PLANT_INDEBTEDNESS_DOCUMENT_TYPES parameter.
+     * 
+     * @see org.kuali.kfs.gl.batch.service.impl.ScrubberProcessImpl#processPlantIndebtedness(
+     * org.kuali.kfs.gl.businessobject.OriginEntryInformation, org.kuali.kfs.gl.service.ScrubberReportData)
+     */
+    @Override
+    protected String processPlantIndebtedness(OriginEntryInformation scrubbedEntry, ScrubberReportData scrubberReport) {
+        // Make sure the entry was from a document that supports plant indebtedness, similar to the logic from the processCapitalization() method.
+        ParameterEvaluator documentTypeCodes = (parameterService.getParameterValueAsBoolean(ScrubberStep.class,
```
db483: Code looks good but I was wondering if readability could be improved a bit by extracting the plant indebtedness indicator into a boolean variable. The documentTypeCodes variable name is also a bit misleading since it is an evaluator rather than a collection of codes.

The if condition for returning null is also a bit unclear. null is returned if plant indebtedness indicator is false, scrubbed value is null or scrubbed entry doc type is not in the PLANT_INDEBTEDNESS_DOC_TYPE_CODES param. Checking if documentTypeCodes is null is hiding a bit the logic behind the decision and at first read without checking how the value is built makes you think that maybe the PLANT_INDEBTEDNESS_DOC_TYPE_CODES param is null/empty/does not exist and hence the docTypesCodes collection is null.

cah292: Some of the variable naming was based on that from the similar capitalization entry processing. However, I will work on modifying this to make it more readable. Thanks for the suggestions.

cah292:
retest this please

cah292:
Changes have been made based on initial review. Please review this again.

PR #89, opened 2016-04-20

bhutchinson:


src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -21,4 +51,47 @@
     public static final String NON_EMPLOYEE_TRAVEL = "Non-Employee Travel";
 
     public static final String USE_OVERSIZED_CHECK_STUB_TEXT_QUESTION_ID = "UseOversizedCheckStubTextQuestion";
+
+    public static final String ACH_INCOME_FILE_DATE_FORMAT = "yymmddHHmm";
+    public static final String ADVACE_DEPOSIT_DEFAULT_CAMPUS_CODE = "IT";
```
cah292: Minor typo in variable name.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -21,4 +51,47 @@
...
+    public static final String ADVANCE_DEPOSIT_NOTE_FILE_PREFIX = "advancedeposit_note_";
+    public static final String BATCH_DETAIL_TYPE = "Batch";
+    public static final String IU_FROM_EMAIL_ADDRESS_PARM_NM = "prod_control@cornell.edu";
```
cah292: Should this variable be renamed, given that we modified the IU code for CU use, and that the value is an actual email address instead of the name of an email parameter?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        }
+
+        for (String documentId : documentIdList) {
+            try {
+                AdvanceDepositDocument advanceDocument = (AdvanceDepositDocument) documentService.getByDocumentHeaderId(documentId);
+                if (LOG.isInfoEnabled()) {
+                    LOG.info("Routing advance deposit document # " + documentId + ".");
+                }
+                documentService.prepareWorkflowDocument(advanceDocument);
+                workflowDocumentService.route(advanceDocument.getDocumentHeader().getWorkflowDocument(), "document routed by achIncome batch job", null);
```
cah292: Any particular reason why the documents are being routed using WorkflowDocumentService instead of DocumentService?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+
+        Person systemUser = getPersonService().getPersonByPrincipalName(KFSConstants.SYSTEM_USER);
+        String principalName = systemUser.getPrincipalName();
+
+        DocumentSearchCriteria.Builder criteria = DocumentSearchCriteria.Builder.create();
+        criteria.setDocumentTypeName(KFSConstants.FinancialDocumentTypeCodes.ADVANCE_DEPOSIT);
+        criteria.setDocumentStatuses(routeStatuses);
+        criteria.setInitiatorPrincipalName(principalName);
+
+        DocumentSearchResults results = KewApiServiceLocator.getWorkflowDocumentService().documentSearch(systemUser.getPrincipalId(), criteria.build());
```
cah292: This call might not retrieve all eligible documents if there are more than the configured doc search result limit. Iteration or alternative retrieval means may be needed to get all of the documents, depending on how many documents the job is expected to create per run.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
nkimble: Should the actual email address value be in a properties file instead of hard coded as a constant string to prevent recompilation if we need to change the address in the future?

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
bhutchinson: Thanks, I'll fix that.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
bhutchinson: Good catch. Either a properties file or a parameter. A parameter is going to be easier to change. We do have some email properties (and parameters I believe) already. Let me look and see what we're doing elsewhere.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: No idea. Let me investigate and see if I can determine a reason. If not do you think using the DocumentService would be better here.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: I'm not sure how many we are expecting, but I can ask Cathy. Do we have an example somewhere of using iteration or alternative retrieval means to bypass the search result limit?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        try {
+            os = new BufferedWriter(new FileWriter(attachmentFile));
+
+            for (AchIncomeNote achIncomeNote : transaction.getNotes()) {
+                os.append(achIncomeNote.getNoteText());
+                os.append("\n");
+            }
+            os.close();
+
+            Note note = documentService.createNoteFromDocument(document, "");
```
cah292: I would advise against creating a note with empty text, given that document notes entered on the UI require text.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
cah292: I think using DocumentService would be cleaner if possible, since it should handle the workflow document preparation.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Ok thanks, I'll work with Cathy to come up with some appropriate text.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Ok, thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
cah292: Not sure of any easy iterative means to get the desired documents with that criteria. Some of the 1099/1042S code does have an iterative solution to get around the doc search limit; however, it relies on knowing all the desired document IDs in advance, and may need tweaking in the future if we decide to change the limit (due to how it was implemented). If you're just interested in the document IDs, then perhaps you could perform a query or BO/CriteriaLookup service call involving the AD document table and the KFS doc header table.

bhutchinson: Ok, thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+    }
+
+    private void createNotes(AchIncomeTransaction transaction, AdvanceDepositDocument document) {
+
+        String attachmentFile = attachmentsDirectory + "/" + CuFPConstants.ADVANCE_DEPOSIT_NOTE_FILE_PREFIX
+                + document.getDocumentNumber() + "_" + new SimpleDateFormat("yyyyMMdd_HHmmss").format(dateTimeService.getCurrentDate()) + ".txt";
+        BufferedWriter os = null;
+
+        try {
+            os = new BufferedWriter(new FileWriter(attachmentFile));
```
cah292: Is it strictly necessary for the ACH notes to be written to an actual file before being converted into a note attachment file? Would it make more sense to use a byte array (and byte array streams) instead?

bhutchinson: Writing to a file is probably not necessary; using a byte array (and byte array streams) probably makes more sense, thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        String account = parameterService.getParameterValueAsString(GenerateAdvanceDepositDocumentsStep.class, CuFPParameterConstants.AdvanceDepositDocument.ACCOUNT);
+
+        IncomingWireAchMapping matchingIncomingWireAchMapping = null;
+        Collection<IncomingWireAchMapping> incomingWireAchMappings = businessObjectService.findAll(IncomingWireAchMapping.class);
+
+        for (IncomingWireAchMapping mapping : incomingWireAchMappings) {
+            List<AchIncomeNote> notes = transaction.getNotes();
+            if (doNotesMatch(mapping, notes)) {
+                matchingIncomingWireAchMapping = mapping;
+                break;
```
cah292: Is the use of "break" okay for a "for (item : items)" loop, or is it not a recommended practice?

bhutchinson: I think it's ok; do you think it's bad practice?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+            chart = matchingIncomingWireAchMapping.getChartOfAccountsCode();
+            objectCode = matchingIncomingWireAchMapping.getFinancialObjectCode();
+            account = matchingIncomingWireAchMapping.getAccountNumber();
+        }
+
+        setupSourceAccountingLine(transaction, advanceDepositDocument, chart, objectCode, account);
+    }
+
+    protected boolean doNotesMatch(IncomingWireAchMapping mapping, List<AchIncomeNote> notes) {
+        final String regex = ".*" + mapping.getShortDescription() + ".*";
```
cah292: If the ACH mapping description could potentially contain special regex characters and is not meant for having the characters treated as such, then it would be good to escape them before searching. Alternatively, since this code appears to just be checking if the description is present in the note, perhaps the regex matching could be replaced with a String indexOf() search.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
cah292: I was often taught to try to avoid using "break" and "continue" within loops, but perhaps it's okay for this kind of loop. I think it's fine for now; perhaps we will learn more about this as we work through the "Clean Code" book.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Good suggestions, thanks. Let me try the String indexOf() search first since that would be simpler than the regex, and if not, I'll see about escaping any special characters before the search.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: I think there can be good general rules like this sometimes, but sometimes following the rules doesn't make the code simpler, faster or easier to understand. I don't remember if this one comes up specifically in clean code, but there are probably general principles that it discusses that are related and would be good for us to discuss as a team.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        } catch (InvalidAddressException e) {
+            LOG.error("sendErrorEmail() Invalid email address. Message not sent", e);
+        } catch (MessagingException me) {
+            throw new RuntimeException("Could not send mail", me);
+        }
+
+    }
+
+    private String composeAchIncomeSummaryEmailBody(FlatFileInformation flatFileInformation) {
+        StringBuffer body = new StringBuffer();
```
cah292: Unless you need to use StringBuffer for thread safety or other reasons, I would suggest using StringBuilder instead for efficiency.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+            LOG.error("Error while validating the Ach Income file" + message);
+            GlobalVariables.getMessageMap().putError(KFSConstants.GLOBAL_ERRORS, KFSKeyConstants.ERROR_CUSTOM, message);
+            return false;
+        }
+
+        return true;
+    }
+
+    private String validatePayerName(List<AchIncomeFileTransaction> achIncomeFileTransaction) {
+        StringBuffer payerMessage = new StringBuffer();
```
cah292: Another area where StringBuffer could be replaced with StringBuilder.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        InputStream fileContents;
+        byte[] fileByteContent;
+        try {
+            fileContents = new FileInputStream(fileName);
+        } catch (FileNotFoundException e1) {
+            LOG.error("Batch file not found [" + fileName + "]. " + e1.getMessage());
+            throw new RuntimeException("Batch File not found [" + fileName + "]. " + e1.getMessage());
+        }
+        try {
+            fileByteContent = IOUtils.toByteArray(fileContents);
```
cah292: IOUtils.toByteArray() may not close the stream when done, so I would advise adding a "finally" block to close the stream. I think there are several places in KFS that don't add the explicit close for this type of byte loading. IOUtils does have a closeQuietly() method to help with closing the stream and ignoring exceptions.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        List<AchIncomeTransaction> achIncomeTransactions = new ArrayList();
+        List<AchIncomeFile> achIncomeFiles = (List<AchIncomeFile>) parsedObject;
+        for (AchIncomeFile achIncomeFile : achIncomeFiles) {
+            setAchTransactions(achIncomeFile, achIncomeTransactions);
+        }
+
+        saveTransaction(achIncomeTransactions);
+    }
+
+    private void saveTransaction(List achIncomeFileTransactions) {
```
cah292: Method argument is missing generics.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+    private KualiDecimal achTotalPostedTransactionAmount;
+    private KualiDecimal achTotalSkippedTransactionAmount;
+    private KualiDecimal wiredTotalPostedTransactionAmount;
+    private KualiDecimal wiredTotalSkippedTransactionAmount;
+
+    /**
+     * @see AdvanceDepositService#createDocuments()
+     */
+    public void createDocuments() {
+        List<AchIncomeTransaction> transactions = (List<AchIncomeTransaction>) businessObjectService.findMatchingOrderBy(AchIncomeTransaction.class, new HashMap(), ACH_TRANSACTION_SEQUENCE_NUMBER, true);
```
cah292: The HashMap is not using generics, I would recommend adding it unless the method being invoked doesn't work well with generics.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+
+    private void saveTransaction(List achIncomeFileTransactions) {
+        businessObjectService.save(achIncomeFileTransactions);
+    }
+
+    /**
+     * Calls businessObjectService to remove all the ach Income transaction rows from the transaction load table.
+     */
+    public void cleanTransactionsTable() {
+        businessObjectService.deleteMatching(AchIncomeTransaction.class, new HashMap());
```
cah292: Another HashMap without generics.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        }
+
+        return payerName;
+    }
+
+    private List<AchIncomeNote> createNotes(AchIncomeFileTransaction achIncomeFileTransaction) {
+        StringBuffer notes = addPaymentInformationToNotes(achIncomeFileTransaction);
+        List<String> achNotes = addRMRAndNTELinesToNotes(achIncomeFileTransaction, notes);
+
+        List<AchIncomeNote> achIncomeNotes = new ArrayList<>();
```
cah292: ArrayList has the brackets for the generics but no type specified; can this code compile?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+
+        if (payerName.length() > CuFPConstants.AchIncomeFileTransactionPayerOrPayeeName.ACH_TRN_PAYR_NM_DB_SIZE) {
+            payerName = payerName.substring(0, CuFPConstants.AchIncomeFileTransactionPayerOrPayeeName.ACH_TRN_PAYR_NM_DB_SIZE);
+        }
+
+        return payerName;
+    }
+
+    private List<AchIncomeNote> createNotes(AchIncomeFileTransaction achIncomeFileTransaction) {
+        StringBuffer notes = addPaymentInformationToNotes(achIncomeFileTransaction);
```
cah292: Another use of StringBuffer where StringBuilder would likely be preferable.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+            achIncomeNote.setNoteLineNumber(noteLineNumber);
+            achIncomeNote.setNoteText(noteText);
+            achIncomeNotes.add(achIncomeNote);
+            noteLineNumber++;
+        }
+
+        return achIncomeNotes;
+    }
+
+    private StringBuffer addPaymentInformationToNotes(AchIncomeFileTransaction achIncomeFileTransaction) {
```
cah292: I would recommend creating and returning a StringBuilder instead unless StringBuffer is strictly necessary.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        if (CuFPConstants.AchIncomeFileTransaction.TRANS_PAYMENT_METHOD_ACH.equals(achIncomeFileTransaction.getPaymentMethodCode())) {
+            notes.append("ACH PAYMENT INFORMATION: ");
+        }
+
+        if (CuFPConstants.AchIncomeFileTransaction.TRANS_PAYMENT_METHOD_FWT.equals(achIncomeFileTransaction.getPaymentMethodCode())) {
+            notes.append("WIRE PAYMENT INFORMATION: ");
+        }
+        notes.append("\n");
+        notes.append("CREDIT: ");
+        notes.append("$" + getFormattedAmount("##,##,###.00", achIncomeFileTransaction.getTransactionAmount()) + " " + achIncomeFileTransaction.getCreditDebitIndicator() + "\n");
```
cah292: Would it be better to use "append" instead of "+" for building the string here and on other code lines below, given that it's using a builder/buffer anyway?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+        if (StringUtils.isBlank(payerNoteText) && achIncomeFileTransaction.getPremiumAdminsContact() != null) {
+            payerNoteText = achIncomeFileTransaction.getPremiumAdminsContact().getName();
+        }
+
+        if (StringUtils.isBlank(receiverNoteText) && achIncomeFileTransaction.getPremiumReceiverName() != null) {
+            receiverNoteText = achIncomeFileTransaction.getPremiumReceiverName().getName();
+        }
+
+        if (StringUtils.isNotBlank(payerNoteText)) {
+            notes.append("RECEIVER: ");
```
cah292: The labeling seems a little confusing; the payer note text populates the "RECEIVER" section, while the receiver note text populates the "ORIGINATOR" section. Is this intentional?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+            notes.append("ORIGINATOR: ");
+            notes.append(receiverNoteText);
+            notes.append("\n");
+        }
+
+        return notes;
+    }
+
+    private List<String> addRMRAndNTELinesToNotes(AchIncomeFileTransaction achIncomeFileTransaction, StringBuffer notes) {
+        List<String> achNotes = new ArrayList();
```
cah292: ArrayList missing generics.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+
+        boolean header = true;
+        Integer lineCount = 1;
+
+        for (AchIncomeFileTransactionOpenItemReference openItemReference : achIncomeFileTransaction.getOpenItemReferences()) {
+            int leftNoteSize = MAX_NOTE_SIZE - notes.length();
+            if ((notes.length() >= MAX_NOTE_SIZE) || (leftNoteSize <= 80)) {
```
cah292: Should this "80" number be moved to a variable or constant?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1202 @@
...
+                } else {
+                    wiredTotalPostedTransactions++;
+                    wiredTotalPostedTransactionAmount = wiredTotalPostedTransactionAmount.add(transaction.getTransactionAmount());
+                }
+            }
+        }
+    }
+
+    private String getEmailMessageText(AchIncomeFile achIncomeFile, String payerMessage) {
+        StringBuffer message = new StringBuffer();
```
cah292: Another place where StringBuffer could be replaced with StringBuilder.

src/main/java/edu/cornell/kfs/fp/businessobject/AchIncomeFileGroup.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+public class AchIncomeFileGroup extends TransientBusinessObjectBase {
+    private String groupControlNumber;
+    private String groupFunctionIdentifierCode;
+    private List<AchIncomeFileTransactionSet> transactionSets;
+    private AchIncomeFileGroupTrailer groupTrailer;
+
+    public AchIncomeFileGroup() {
+        this.transactionSets = new ArrayList<>();
```
cah292: Another ArrayList missing the type for the generics.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
cah292: Never mind, it looks like Java 7 and above can use "<>" for generics.

src/main/java/edu/cornell/kfs/fp/businessobject/AchIncomeFileGroup.java
cah292: Never mind, it looks like Java 7 and above can use "<>" for generics.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
```diff
@@ -0,0 +1,647 @@
...
+            accountNumber.setAccessible(true);
+            try {
+                accountNumber.set(sourceAccountingLine, account);
+            } catch (IllegalAccessException e) {
+                throw new RuntimeException("huh, how did this happen?", e);
+            }
+        }
+    }
+
+    private class MockParameterService implements ParameterService {
```
cah292: It might be a good idea to override the new MockParameterServiceImpl so that you don't have to define all the null-returning methods here.

cah292:
Finished review for first version of the pull request. Will wait for updates before reexamining.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Ok, thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Good catch.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: yup

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: yup, and another one I'll fix

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: k, thx

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: yes, that's the preferred way for Java 7 and above; it leverages what is defined with the type

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: sure

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: yes, it probably would be better; another change added to my list :-) thanks

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Not sure since I didn't actually write the original code :-), but it does seem confusing. I'll look further and switch it if that makes sense, thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: good catch

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: probably a constant would be good

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: yup

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
bhutchinson: yeah, let me look into that; I think when I first wrote this test, I hadn't rebased to pull in the new MockParameterServiceImpl yet

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: good suggestion, thanks

bhutchinson:
Thanks Chad! I'll work on updating the code and this PR today.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
bhutchinson: Turns out this was intended to be a parameter; I'm fixing it.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: The DocumentService routeDocument validates and persists the document as well as calling WorkflowDocumentService.route(). Not sure if the code was intentionally avoiding those calls. I'll try it with the DocumentService.route and see what happens.

bhutchinson: Using the DocumentService seems to work, so let's go with that.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Yeah, this code was a bit mixed up; I fixed it.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: Talked to Cathy and it looks like we're talking less than 100 transactions a day, maybe around 34 avg / day with 100 being peak, so I think we're ok here.

bhutchinson:
I believe I have addressed all the issues raised in the code review and this PR is ready to review again.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1200 @@
...
+            LOG.info("Advance deposit to Route: " + documentIdList);
+        }
+
+        for (String documentId : documentIdList) {
+            try {
+                AdvanceDepositDocument advanceDocument = (AdvanceDepositDocument) documentService.getByDocumentHeaderId(documentId);
+                if (LOG.isInfoEnabled()) {
+                    LOG.info("Routing advance deposit document # " + documentId + ".");
+                }
+                documentService.prepareWorkflowDocument(advanceDocument);
```
cah292: This line may be redundant, since DocumentService.routeDocument already calls DocumentService.prepareWorkflowDocument

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -0,0 +1,1200 @@
...
+        List<DocumentStatus> routeStatuses = new ArrayList<DocumentStatus>();
+        routeStatuses.add(DocumentStatus.fromCode(statusCode));
+
+        Person systemUser = getPersonService().getPersonByPrincipalName(KFSConstants.SYSTEM_USER);
+        String principalName = systemUser.getPrincipalName();
+
+        DocumentSearchCriteria.Builder criteria = DocumentSearchCriteria.Builder.create();
+        criteria.setDocumentTypeName(KFSConstants.FinancialDocumentTypeCodes.ADVANCE_DEPOSIT);
+        criteria.setDocumentStatuses(routeStatuses);
+        criteria.setInitiatorPrincipalName(principalName);
```
cah292: Do we have any other processes that generate Advance Deposit documents via the system user? If so, this search could end up retrieving them.

cah292:
Finished second code review. Looks good overall. I did find one possibly redundant line, but I wouldn't consider it a show-stopper unless it's interfering with the document creation or routing. I also have one question regarding the AD document retrieval, which I would like to get feedback on.

Will wait for feedback on these two items before merging.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: I checked with Cathy and also Kymber from the product team and they didn't know of any. Also, FWIW, I had to explicitly give the system user permissions to route the AD documents, so if there was something it would just be saving the documents and not routing them which seems odd.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: You're correct, I can remove the redundant call. Thanks.

bhutchinson:
I removed the redundant line and updated the PR, and commented with an answer to the question. Hopefully the unit test suite will run (and pass) soon and this can get a final review and merge. Thanks.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
cah292: It looks like the appending is correct now but the conditions are still backwards. The receiver section is outputting the receiver note text but is checking for non-blank payer note text, and vice versa for the originator/payer output.

cah292:
Found one more minor issue in the code related to printing receiver/originator content. If this can get fixed quickly, I should still have time to review and merge before the end of the day.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
bhutchinson: 'doh, missed that; let me fix it, thanks

bhutchinson:
fixed that minor issue, thanks!

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -21,4 +51,46 @@
     public static final String NON_EMPLOYEE_TRAVEL = "Non-Employee Travel";
 
     public static final String USE_OVERSIZED_CHECK_STUB_TEXT_QUESTION_ID = "UseOversizedCheckStubTextQuestion";
+
+    public static final String ACH_INCOME_FILE_DATE_FORMAT = "yymmddHHmm";
```
cah292: Is this a valid date format? Should it be "yyMMddHHmm" instead? Sorry for not noticing this earlier.

bhutchinson: no worries; good question, let me investigate and fix if necessary

cah292:
So sorry, but noticed a potential date pattern issue at the last minute that I had not spotted earlier. Once it's fixed or confirmed that it's valid, I will do a (hopefully) final review before merging. I might still have time for the final review today if this can be addressed promptly.

bhutchinson:
Fixed the date format issue, thanks for catching that!

cah292:
Code looks good now (YES!) and unit tests have passed. Merging.

bhutchinson:
Thanks Chad!

PR #92, opened 2016-04-25

kyleoliveira:
See #56 . This got into TEST and it was discovered that some PCard docs have different Line Item Detail Indicator values that previously expected. I was instructed to remove value-checking validation from that field. I also made the Item Sequence Number and Line Item Detail Indicator not required.

Otherwise, this is identical to what was previously accepted.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/businessobject/ProcurementCardTransactionDetailExtendedAttribute.java
+++ b/src/main/java/edu/cornell/kfs/fp/businessobject/ProcurementCardTransactionDetailExtendedAttribute.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2006 The Kuali Foundation
+ * 
+ * Licensed under the Educational Community License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.opensource.org/licenses/ecl2.php
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package edu.cornell.kfs.fp.businessobject;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.rice.krad.bo.PersistableBusinessObjectExtensionBase;
+
+/**
+ * This class is used to represent a procurement card transaction detail business object.
+ */
+public class ProcurementCardTransactionDetailExtendedAttribute extends PersistableBusinessObjectExtensionBase {
+
+    private static final long serialVersionUID = 1L;
+    private String documentNumber;
+    private Integer financialDocumentTransactionLineNumber;
+    
+    // list of pcard data
+    private List<PurchasingDataDetail> purchasingDataDetails;
+    
+    public ProcurementCardTransactionDetailExtendedAttribute() {
+        super();
+        purchasingDataDetails = new ArrayList<PurchasingDataDetail>();
+    }
+    
+    public String getDocumentNumber() {
+        return documentNumber;
+    }
+    
+    public void setDocumentNumber(String documentNumber) {
+        this.documentNumber = documentNumber;
+    }
+    
+    public Integer getFinancialDocumentTransactionLineNumber() {
+        return financialDocumentTransactionLineNumber;
+    }
+    
+    public void setFinancialDocumentTransactionLineNumber(Integer financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = financialDocumentTransactionLineNumber;
+    }
+
+    public void setFinancialDocumentTransactionLineNumber(String financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = Integer.valueOf(financialDocumentTransactionLineNumber);
+    }
+    
+    public List<PurchasingDataDetail> getPurchasingDataDetails() {
+        return purchasingDataDetails;
+    }
+    
+    public void setPurchasingDataDetails(List<PurchasingDataDetail> purchasingDataDetails) {
+        this.purchasingDataDetails = purchasingDataDetails;
+    }
+
+    /**
+     * @see org.kuali.rice.krad.bo.BusinessObjectBase#toStringMapper()
+     */
+    @SuppressWarnings("rawtypes")
+    protected LinkedHashMap toStringMapper_RICE20_REFACTORME() {
+        LinkedHashMap<String, String> m = new LinkedHashMap<String, String>();
+        m.put(KFSPropertyConstants.DOCUMENT_NUMBER, this.documentNumber);
+        if (this.financialDocumentTransactionLineNumber != null) {
+            m.put("financialDocumentTransactionLineNumber", this.financialDocumentTransactionLineNumber.toString());
+        }
+        return m;
+    }
+    
+}
--- a/src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
+++ b/src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
@@ -1,0 +1,500 @@
+/*
+ * Copyright 2016 The Kuali Foundation.
+ * 
+ * Licensed under the Educational Community License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ * http://www.opensource.org/licenses/ecl2.php
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.cornell.kfs.fp.businessobject;
+
+import java.sql.Date;
+import java.text.ParseException;
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.core.api.datetime.DateTimeService;
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+public class PurchasingDataBase extends USBankAddendumRecord {
+    private static final long serialVersionUID = 1L;
+    private Integer purchasingDataId;
+    private String accountNumber;
+    private String tsysTranCode;
+    private String itemCommodityCode;
+    private String merchantOrderNumber;
+    private KualiDecimal discountAmount;
+    private KualiDecimal freightShippingAmount;
+    private KualiDecimal dutyAmount;
+    private String destinationPostalZipCode;
+    private String shipFromPostalZipCode;
+    private String destinationCountryCode;
+    private String uniqueVATInvoice;
+    private Date orderDate;
+    private String itemDescriptor;
+    private Integer quantity;
+    private String unitOfMeasure;
+    private KualiDecimal unitCost;
+    private String typeOfSupply;
+    private String serviceIdentifier;
+    private String messageIdentifier;
+    private Integer itemSequenceNumber;
+    private Integer lineItemDetailIndicator;
+    
+
+    public PurchasingDataBase() {
+      super();
+
+      this.discountAmount = new KualiDecimal(0);
+      this.freightShippingAmount = new KualiDecimal(0);
+      this.dutyAmount = new KualiDecimal(0);
+      this.unitCost = new KualiDecimal(0);
+    }
+
+    /**
+     * @return the purchasingDataId
+     */
+    public Integer getPurchasingDataId() {
+      return purchasingDataId;
+    }
+
+    /**
+     * @param purchasingDataId the purchasingDataId to set
+     */
+    public void setPurchasingDataId(Integer purchasingDataId) {
+      this.purchasingDataId = purchasingDataId;
+    }
+    
+    /**
+     * @return the accountNumber
+     */
+    public String getAccountNumber() {
+      return accountNumber;
+    }
+
+    /**
+     * @param accountNumber the accountNumber to set
+     */
+    public void setAccountNumber(String accountNumber) {
+      this.accountNumber = accountNumber;
+    }
+
+    /**
+     * @return the tsysTranCode
+     */
+    public String getTsysTranCode() {
+      return tsysTranCode;
+    }
+
+    /**
+     * @param tsysTranCode the tsysTranCode to set
+     */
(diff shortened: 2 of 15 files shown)


cah292:
Most of the unit tests failed due to an inability to connect to the UTEST database: "IO Error: Unknown host specified". Once UTEST is working again, the tests will need to be re-run.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+    /**
+     * @param dutyAmount the dutyAmount to set
+     */
+    public void setDutyAmount(KualiDecimal dutyAmount) {
+      this.dutyAmount = dutyAmount;
+    }
+
+
+    /**
+     * @param freightShippingAmount the freightShippingAmount to set
```
cah292: Minor mismatch between Javadoc and method, probably not worth fixing unless other items come up in the review.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+
+    /**
+     * @param orderDate the orderDate to set
+     */
+    public void setOrderDate(Date orderDate) {
+      this.orderDate = orderDate;
+    }
+    
+
+    public void setOrderDate(String orderDate) {
```
cah292: This setter will only overwrite the orderDate if the String is non-blank. Is this intentional, or should it allow for setting the orderDate to null for a blank String?

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+     */
+    public void setQuantity(Integer quantity) {
+      this.quantity = quantity;
+    }
+
+    /**
+     * @param quantity the quantity to set
+     */
+    public void setQuantity(String quantity) {
+      this.quantity = Integer.parseInt(quantity);
```
cah292: Implicit conversion of primitive int to an Integer object; you can use Integer.valueOf() to get an explicit Integer object instead. Also, this setter is not consistent with some of the others that have special handling for blank values; is this intentional?

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+    public void setUnitCost(KualiDecimal unitCost) {
+      this.unitCost = unitCost;
+    }
+
+
+    /**
+     * @param unitCost the unitCost to set
+     */
+    public void setUnitCost(String unitCost) {
+      this.unitCost = new KualiDecimal(unitCost);
```
cah292: Another setter that does not follow the pattern of some of the others for explicitly handling blank values.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+    public void setItemSequenceNumber(Integer itemSequenceNumber) {
+      this.itemSequenceNumber = itemSequenceNumber;
+    }
+
+
+    /**
+     * @param itemSequenceNumber the itemSequenceNumber to set
+     */
+    public void setItemSequenceNumber(String itemSequenceNumber) {
+      this.itemSequenceNumber = Integer.parseInt(itemSequenceNumber);
```
cah292: Another implicit int-to-Integer conversion.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+    public void setLineItemDetailIndicator(Integer lineItemDetailIndicator) {
+      this.lineItemDetailIndicator = lineItemDetailIndicator;  
+    }
+
+
+    /**
+     * @param lineItemDetailIndicator the lineItemDetailIndicator to set
+     */
+    public void setLineItemDetailIndicator(String lineItemDetailIndicator) {
+      this.setLineItemDetailIndicator(Integer.parseInt(lineItemDetailIndicator));
```
cah292: Another implicit int-to-Integer conversion, and also calls the Integer setter instead of setting the value directly like the other methods. Not sure if this is intentional.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
```diff
@@ -0,0 +1,525 @@
...
+     * Parses a supposed Type 50 record addendum line
+     * 
+     * @param line The current line
+     * @param lineCount The current line number
+     * @throws IOException 
+     * @throws Exception When there is a string parsing error or a missing required field
+     */
+    public void parse(String line, int lineCount) throws IOException {
+      // Lines with a * indicate data that the documentation states will be available in the future
+      this.setRecordId(USBankRecordFieldUtils.extractNormalizedString(line, 0, 2, true, lineCount));
```
cah292: The line start/end indexes all appear to be hard-coded. If feasible and if potential format changes are anticipated in the future, I would recommend defining these indexes through some other means.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataDetail.java
```diff
@@ -0,0 +1,84 @@
...
+    public Integer getFinancialDocumentTransactionLineNumber() {
+        return financialDocumentTransactionLineNumber;
+    }
+
+    public void setFinancialDocumentTransactionLineNumber(Integer financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = financialDocumentTransactionLineNumber;
+    }
+
+    public void setFinancialDocumentTransactionLineNumber(String financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = Integer.parseInt(financialDocumentTransactionLineNumber);
```
cah292: Another implicit int-to-Integer conversion.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
kyleoliveira: Good point. I'll change that to valueOf().

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,206 @@
...
+   * @throws IOException When unable to parse the date
+   */
+  public static Date extractDate(String line, int begin, int end, int lineCount) throws IOException {
+    Date theDate;
+    try {
+      String sub = line.substring(begin, end);
+      String year = sub.substring(0,4);
+      String month = sub.substring(4,6);
+      String day = sub.substring(6,8);
+      theDate = Date.valueOf(year+"-"+month+"-"+day);
```
cah292: I would suggest either using the DateTimeService or BatchDateFormatter to convert to a SQL Date value. With the latter, you can construct an instance that checks for a specific date pattern.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,206 @@
...
+   * @param end Ending index
+   * @param lineCount The current line number
+   * @return The KualiDecimal parsed from the trimmed substring
+   * @throws IOException When unable to parse the KualiDecimal
+   */
+  public static KualiDecimal extractDecimal(String line, int begin, int end, int lineCount) throws IOException {
+    KualiDecimal theDecimal;
+    try {
+      String sanitized = line.substring(begin, end);
+      sanitized = StringUtils.remove(sanitized, '-');
```
cah292: Is this method expected to just return the absolute value of a given number, since it seems to be removing the negative sign? The same goes for the extractDecimalWithCents method.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,206 @@
...
+   * @return The cycle Date parsed from the trimmed substring
+   * @throws IOException When unable to parse the cycle Date
+   */
+  public static Date extractCycleDate(String line, int begin, int end, int lineCount) throws IOException {
+    String day;
+    Calendar cal = Calendar.getInstance();
+    Date theDate;
+    try {
+      day = line.substring(begin, end);
+      theDate = KfsDateUtils.newDate(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH), Integer.parseInt(day));
```
cah292: Another implicit int-to-Integer conversion. Also, is the expectation that the line just contains the day and that the month and year should be the current ones?

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
```diff
@@ -0,0 +1,206 @@
...
+  /**
+   * Converts a given string to either a Debit code or a Credit code
+   * @param val String to convert. Should either be "+" or "-"
+   * @return "D" if Debit, "C" if Credit
+   * @throws IOException When unable to determine if the provided string is equivalent to the two codes
+   */
+  public static String convertDebitCreditCode(String val) throws IOException {
+    switch(val) {
+      case "+":
+        return "D";
```
cah292: I would advise checking to see if KFS already has constants defined for using "D" and "C" as Debit and Credit, respectively. Also, is "+" supposed to map to Debit in this case? Should it be Credit instead? (And vice versa for "-")

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
kyleoliveira: Could you elaborate? I don't understand what you mean.

src/main/java/edu/cornell/kfs/fp/businessobject/ProcurementCardTransactionDetailExtendedAttribute.java
```diff
@@ -0,0 +1,84 @@
...
+    public Integer getFinancialDocumentTransactionLineNumber() {
+        return financialDocumentTransactionLineNumber;
+    }
+    
+    public void setFinancialDocumentTransactionLineNumber(Integer financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = financialDocumentTransactionLineNumber;
+    }
+
+    public void setFinancialDocumentTransactionLineNumber(String financialDocumentTransactionLineNumber) {
+        this.financialDocumentTransactionLineNumber = Integer.parseInt(financialDocumentTransactionLineNumber);
```
cah292: One more implicit int-to-Integer conversion.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
kyleoliveira: Do you have any suggestions for other means?

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
cah292: Sorry, wrong line. I'll add the comment to the correct line.

cah292: Mismatch between parameter names in Javadoc and those in the actual method.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
cah292: Apart from either defining lots of constants or reading the values from some file or parameter, I'm not sure of an easy way to avoid all these hard-coded amounts. These will likely be okay for now, but a more flexible approach may be needed in the future if we anticipate changes to the formatting.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
kyleoliveira: Good point, I'll handle that situation.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: These were already implemented as-is, I only moved them from the ProcurementCardFlatInputFileType class to this class.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: This was already implemented as-is, I only moved it from the ProcurementCardFlatInputFileType class to this class.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
kyleoliveira: This was already implemented as-is, I only moved it from the ProcurementCardFlatInputFileType class to this class.

cah292:
Finished reviewing; will wait for feedback on my analyses.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
cah292: Ok, sounds good.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
cah292: Ok, sounds good.

src/main/java/edu/cornell/kfs/fp/businessobject/USBankRecordFieldUtils.java
cah292: Ok, sounds good.

src/main/java/edu/cornell/kfs/fp/businessobject/PurchasingDataBase.java
kyleoliveira: Given that the other record types are handled in this way and that the format does not change very often, I'm going to leave this as-is for now.

kyleoliveira:
@cah292 I've updated the commit based on your suggestions. Let me know if it's ready to go.

cah292:
Looks like there's still a problem with the UTEST database. I'll follow up with DevOps so that we can get unit tests running again on the PRs.

cah292:
Code looks good overall. I did notice a minor Javadoc method typo, but it's not worth reworking at this time. And that one method for conversion to a SQL Date should be fine as-is for now, given that it's simply a relocated existing method from a different class.

Once UTEST has been sorted out so that the unit tests can run and pass, I will perform the merge.

kyleoliveira:
Excellent, thank you.

cah292:
Code looks good and unit tests have passed. Merging.

PR #97, opened 2016-05-03

kyleoliveira:
Resubmit of #92. The date parsing for PurchasingDataBase#orderDate was wrong. 

The only update here is that PurchasingDataBase#parse now uses extractDate for parsing orderDate rather than extractNormalizedString and I removed the setter that takes a String since it isn't used anymore.


cah292:
Code looks good, and tests have passed. Merging.

PR #99, opened 2016-05-05
KFSPTS-5543 fixing PurchaseOrderTransmissionMethodDataRulesServiceImp…

jhulslander:
…l and its test cases


bhutchinson:
talked to Jay and he's going to try to go one level deeper with mock services to add a little more value; PR will get recreated when he's done

PR #101, opened 2016-05-06
KFSPTS-5543 fixing PurchaseOrderTransmissionMethodDataRulesServiceImp…

jhulslander:
…l and its test cases

I create a local CuPostalCodeValidationServiceImpl implementation of PostalCodeValidationServiceImpl from KFS, so I could make a getter and setter for the state service.  I also needed to make "protected ZipcodeValidationPattern getZipcodeValidatePattern()", so I could over rirde it in the test case so it doesn't call rice services to get the zip code regluar expression from a file.  Using a static string in the test case is fine, which is what is in the config file.  This was all needed to avoid using spring in the test case.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/PurchaseOrderTransmissionMethodDataRulesServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/PurchaseOrderTransmissionMethodDataRulesServiceImpl.java
@@ -1,7 +1,6 @@
 package edu.cornell.kfs.module.purap.document.service.impl;
 
 import org.apache.commons.lang.StringUtils;
-import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.kfs.sys.service.PostalCodeValidationService;
 import org.kuali.kfs.vnd.service.PhoneNumberService;
 
@@ -16,6 +15,8 @@ import edu.cornell.kfs.module.purap.document.service.PurchaseOrderTransmissionMe
  */
 public class PurchaseOrderTransmissionMethodDataRulesServiceImpl implements PurchaseOrderTransmissionMethodDataRulesService{
 	
+	protected PostalCodeValidationService postalCodeValidationService;
+	protected PhoneNumberService phoneNumberService;
 	
 	/**
 	 * Returns false when faxNumber is null OR blank OR not is not in the PhoneNumberService.isValidPhoneNumber format;
@@ -23,7 +24,7 @@ public class PurchaseOrderTransmissionMethodDataRulesServiceImpl implements Purc
 	 */
 	public boolean isFaxNumberValid(String faxNumber) {
 		boolean dataIsValid = true; 
-		if ( (faxNumber == null) || (StringUtils.isEmpty(faxNumber)) || (!SpringContext.getBean(PhoneNumberService.class).isValidPhoneNumber(faxNumber)) ) {
+		if ( (faxNumber == null) || (StringUtils.isEmpty(faxNumber)) || (!getPhoneNumberService().isValidPhoneNumber(faxNumber)) ) {
 			dataIsValid &= false;
 	    }
 		return dataIsValid;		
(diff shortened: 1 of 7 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/service/impl/PurchaseOrderTransmissionMethodDataRulesServiceImpl.java
```diff
@@ -16,14 +15,16 @@
  */
 public class PurchaseOrderTransmissionMethodDataRulesServiceImpl implements PurchaseOrderTransmissionMethodDataRulesService{
 	
+	private PostalCodeValidationService postalCodeValidationService;
```
bhutchinson: we might want to make these protected (I know we've done that a lot in base code to facilitate being able to override)

jhulslander: The getters and setters are public, shouldn't that provide the needed capability to override?

jhulslander:
retest this pelase

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/PurchaseOrderTransmissionMethodDataRulesServiceImpl.java
bhutchinson: In theory but not necessarily in practice. I'd have to dig up the details, but I know IU submitted contributions to change some of them to protected, and I think we generally did that on the product team. I think it may have to do with spring, beanutils, reflection, etc. Some things don't actually use the setters & getters.

jhulslander: OK, I will change it

bhutchinson:
retest this please

bhutchinson:
retest this please

src/main/java/edu/cornell/kfs/sys/service/impl/CuPostalCodeValidationServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.KFSKeyConstants;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.kfs.sys.service.impl.PostalCodeValidationServiceImpl;
+import org.kuali.rice.kns.datadictionary.validation.fieldlevel.ZipcodeValidationPattern;
+import org.kuali.rice.krad.util.GlobalVariables;
+import org.kuali.rice.location.api.state.State;
+import org.kuali.rice.location.api.state.StateService;
+
+public class CuPostalCodeValidationServiceImpl extends PostalCodeValidationServiceImpl {
```
bhutchinson: Since we already have an overlay to locally customize PostalCodeValidationServiceImpl, I would just modify that rather than creating a new class. We should look to contribute our customizations to this class back to base code.

src/main/java/edu/cornell/kfs/sys/service/impl/CuPostalCodeValidationServiceImpl.java
```diff
@@ -0,0 +1,77 @@
...
+        
+        return valid;
+    }
+	
+	/**
+	 * Pulling getZipcodeValidatePattern() out for testability.  Lower down, ZipcodeValidationPattern loads params from a config file.
+	 * @return
+	 */
+	protected ZipcodeValidationPattern getZipcodeValidatePattern() {
+		ZipcodeValidationPattern zipPattern = new ZipcodeValidationPattern();
```
bhutchinson: Not a huge deal, but I would combine these two lines into one and just return new ZipcodeValidationPattern();

src/test/java/edu/cornell/kfs/module/purap/document/service/impl/PurchaseOrderTransmissionMethodDataRulesServiceImplTest.java
```diff
@@ -1,104 +1,183 @@
...
+		bo.setCode(STATE);
+		bo.setName("New York");
+		bo.setCountryCode(COUNTRY);
+		return StateBo.to(bo);
+	}
+	
+	private class testablePhoneNumberServiceImpl extends PhoneNumberServiceImpl{
+		@Override
+		protected String[] parseFormats() {
+			//This is the string that is in our parameter table
```
bhutchinson: I'm not a huge fan of this comment (and the one like it below) as I'm not sure it adds a lot of value.

bhutchinson:
retest this please

jhulslander:
retest this please

jhulslander:
This is ready for review again.  I believe I have made the requested changes

src/test/resources/org/kuali/kfs/sys/spring-sys.xml
```diff
@@ -800,7 +800,9 @@
 
 	<bean id="postalCodeValidationService" parent="postalCodeValidationService-parentBean"/>
 	
-    <bean id="postalCodeValidationService-parentBean" class="org.kuali.kfs.sys.service.impl.PostalCodeValidationServiceImpl" abstract="true" />
+    <bean id="postalCodeValidationService-parentBean" class="org.kuali.kfs.sys.service.impl.PostalCodeValidationServiceImpl" abstract="true">
```
bhutchinson: Do you need this in the non-test spring-sys.xml too so we don't get NPEs?

jhulslander: Good catch, just uploaded the udpate

jhulslander:
retest this please

PR #103, opened 2016-05-10
KFSPTS-5636: Added Favorite Account feature to the IWantDocument, and performed updates to make Favorite Accounts more flexible across different accounting documents.

cah292:
This updates our Favorite Accounts feature to make it available on the IWantDocument.

Some major changes have been made to make this feature flexible enough for both IWant and regular accounting documents. For example, some new "builder" classes have been introduced to simplify the addition of favorite-account-based lines from the Struts action classes. Some use of Java generics has also been introduced to expand the feature beyond REQS and PO documents.

Since IWantDocument is not a standard PURAP document and does not use the standard accounting line groups from other documents, the IWNT has its own custom JSP-related means of adding the Favorite Accounts drop-down to the form. Also, the IWNT account lines now implement the GLPE Detail Source interface to simplify the favorite-account-based line construction, while minimizing the amount of extra methods needed on those objects. The IWNT line's don't implement the accounting line interface because of the massive number of new unused methods that the line would need to add; if you think this code would work better by explicitly requiring accounting lines instead of just the superinterface of GLPE Detail Source, let me know.

I've updated one of the existing Favorite Accounts unit tests to clean it up a little and to add testing for IWNT. In addition, I've added a new unit test class to validate the new "builder" classes.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/businessobject/IWantAccount.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/businessobject/IWantAccount.java
@@ -8,13 +8,15 @@ import org.kuali.kfs.coa.businessobject.ObjectCode;
 import org.kuali.kfs.coa.businessobject.ProjectCode;
 import org.kuali.kfs.coa.businessobject.SubAccount;
 import org.kuali.kfs.coa.businessobject.SubObjectCode;
-import edu.cornell.kfs.module.purap.CUPurapConstants;
+import org.kuali.kfs.sys.businessobject.GeneralLedgerPendingEntrySourceDetail;
 import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.kfs.sys.service.UniversityDateService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 import org.kuali.rice.krad.bo.PersistableBusinessObjectBase;
 
-public class IWantAccount extends PersistableBusinessObjectBase {
+import edu.cornell.kfs.module.purap.CUPurapConstants;
+
+public class IWantAccount extends PersistableBusinessObjectBase implements GeneralLedgerPendingEntrySourceDetail {
 
 	private static final long serialVersionUID = 1L;
 	private String documentNumber;
@@ -207,6 +209,46 @@ public class IWantAccount extends PersistableBusinessObjectBase {
     }
 
     /**
+     * Returns the amount-or-percent value if configured to use amounts, zero otherwise.
+     * 
+     * @see org.kuali.kfs.sys.businessobject.GeneralLedgerPendingEntrySourceDetail#getAmount()
+     */
+    @Override
+    public KualiDecimal getAmount() {
+        return CUPurapConstants.AMOUNT.equals(getUseAmountOrPercent()) ? getAmountOrPercent() : KualiDecimal.ZERO;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getBalanceTypeCode() {
+        return null;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getFinancialDocumentLineDescription() {
+        return null;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getReferenceNumber() {
+        return null;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getReferenceOriginCode() {
+        return null;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getReferenceTypeCode() {
+        return null;
+    }
+
+	/**
      * Helper method for copying an account.
      * 
      * @param oldAccount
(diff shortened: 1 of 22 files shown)


src/main/java/edu/cornell/kfs/module/purap/businessobject/IWantAccount.java
```diff
@@ -207,6 +209,46 @@ public void setUseAmountOrPercent(String useAmountOrPercent) {
...
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getBalanceTypeCode() {
+        return null;
+    }
+
+    // Only here to comply with the GeneralLedgerPendingEntrySourceDetail interface.
+    @Override
+    public String getFinancialDocumentLineDescription() {
+        return null;
```
jhulslander: Should this throw a Run time exception or something of the like instead of just returning NULL?

cah292: I guess these could throw an exception if needed. However, depending on where the IWNT account is used, there could be code that calls these getters, though that may not be the case for Favorite Accounts and the IWNT document purposes. Not sure whether it's good practice for value retrieval methods to throw exceptions when unused.

jhulslander: I'm not strong on the opinion of throwing a run time exception. I just don't like returning a blanket null.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
```diff
@@ -617,6 +625,13 @@ public ActionForward route(ActionMapping mapping, ActionForm form, HttpServletRe
             added &= addNewItem(iWantDocForm, iWantDocument, item);
         }
         
+        if (added && iWantDocument.getFavoriteAccountLineIdentifier() != null) {
+            added &= addNewFavoriteAccount(iWantDocument);
+            if (added) {
+                iWantDocument.setFavoriteAccountLineIdentifier(null);
+            }
+        }
+
```
jhulslander: This block is called three times here, do we have an opportunity to extract it to a function to eliminate duplication?

src/main/java/edu/cornell/kfs/module/purap/util/PurApFavoriteAccountLineBuilderForIWantDocument.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Returns the static path to the document's Favorite Account Line ID property,
+     * starting with the document itself.
+     * 
+     * @see edu.cornell.kfs.sys.util.FavoriteAccountLineBuilderBase#getErrorPropertyName()
+     */
+    @Override
+    public String getErrorPropertyName() {
+        return "document.favoriteAccountLineIdentifier";
```
jhulslander: should this be a static final string somewhere?

src/main/java/edu/cornell/kfs/sys/util/FavoriteAccountLineBuilderBase.java
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Base class for helper objects that generate accounting line instances for a given Favorite Account object.
+ *
+ * @param <E> The element type of the list that the new accounting line will be inserted into.
+ * @param <T> The actual type of the accounting line to be inserted.
+ */
+public abstract class FavoriteAccountLineBuilderBase<E extends GeneralLedgerPendingEntrySourceDetail,T extends E> {
+
+    private UserProcurementProfileValidationService userProcurementProfileValidationService;
+    private UserFavoriteAccountService userFavoriteAccountService;
```
jhulslander: I noticed on another PR that we should start making services protected member variables instead of private.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
cah292: Yes, that makes sense. I'll work on moving this to a new function.

jhulslander:
This is good for the most part, just a few small considerations.  One additional thought on CuUserFavoriteAccountServiceImplTest, would it be worthwhile spending some time to eliminate spring from the test, and make it more of a microservice test? I don't know, just bringing up the idea.

src/main/java/edu/cornell/kfs/module/purap/util/PurApFavoriteAccountLineBuilderForIWantDocument.java
cah292: Thanks, I'll look for a good spot to add a constant for this.

src/main/java/edu/cornell/kfs/sys/util/FavoriteAccountLineBuilderBase.java
cah292: Good point, I'll get these changed over to "protected".

cah292:
retest this please

cah292:
I've updated this PR based on feedback. Please review again when you get a chance.

As for CuUserFavoriteAccountServiceImplTest, the original changes I had made were geared for adding IWant into its coverage and for improving the assertions being used. I think a rewrite of the test to avoid Spring may be beyond the scope of this story, since it could take a fair amount of work to get the test to not use DB-derived data. Perhaps we can address that test for a different story.

jhulslander:
Looks good now.  I agree about CuUserFavoriteAccountServiceImplTest, and I wil add a Jira to the tech sprint if it doesn't exist.  As for my comments about the NULL returns, that shouldn't stop this from going forward.  It might be worth talking about that as a group sometime though.

PR #104, opened 2016-05-10

bhutchinson:


bhutchinson:
@cah292, this is very similar to the prior PR #87 for this work, with the addition that it will now remove .done files in the event of a failure, and it has a new corresponding unit test method to validate that functionality.

cah292:
Code looks good, and unit tests have passed. Merging.

PR #106, opened 2016-05-16

cah292:
This is a newer version of the code from PR #32, which has been modified to use the proper header names from the Workday ACH files and to make a couple other tweaks. In addition, a new microtest class has been added to perform some validation of the new feature. (Portions of the new test are based on one of Bryan's recent test case additions, to help with adding, moving and deleting test files as needed.)

There are some references to email subject/body parameter strings for updates to existing ACH accounts. However, this feature is only designed for additions of new ACH accounts at this time, so those update-related params are not in active use and will not be added to our environments yet. (The email parms for added ACH accounts will still be used.)

I'm new to the process of using EasyMock to convert Spring-related tests into microtests, so please let me know if you believe there are areas in the new test that need improvement.


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,681 @@
...
+                PayeeIdTypeCodes.EMPLOYEE, payee.getEmployeeId(), getDirectDepositTransactionType());
+        
+        // Add or update Entity ID account.
+        if (ObjectUtils.isNull(entityAccount)) {
+            addACHAccount(payee, achDetail, PayeeIdTypeCodes.ENTITY);
+            LOG.info("Created new ACH Account of Entity type for payee " + payee.getPrincipalName());
+        } else {
+            updateACHAccountIfNecessary(payee, achDetail, entityAccount);
+        }
+
```
nkimble: If-clause logs info message for add action. For consistency, should that same type of message be logged for update actions as well?

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,681 @@
...
+        if (ObjectUtils.isNull(employeeAccount)) {
+            addACHAccount(payee, achDetail, PayeeIdTypeCodes.EMPLOYEE);
+            LOG.info("Created new ACH Account of Employee type for payee " + payee.getPrincipalName());
+        } else {
+            updateACHAccountIfNecessary(payee, achDetail, employeeAccount);
+        }
+        
+        return true;
+    }
+
```
nkimble: If-clause logs info message for add action. For consistency, should that same type of message be logged for update actions as well?

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,681 @@
...
+        if (!StringUtils.equals(achDetail.getBankRoutingNumber(), achAccount.getBankRoutingNumber())
+                || !StringUtils.equals(achDetail.getBankAccountNumber(), achAccount.getBankAccountNumber())) {
+            LOG.info("Account information from input file for payee " + achDetail.getNetID() + " and payee type '"
+                    + achAccount.getPayeeIdentifierTypeCode() + "' does not match what is in KFS, but will be left as-is.");
+            /*
+             * For a future enhancement, we will modify this method to actually update the Payee ACH Account
+             * via a maintenance document, and possibly use more fields for change comparison.
+             */
+        }
+    }
```
nkimble: If-clause info message is inferring that processing would have happening if the data had been correct. There is no corresponding Log.info message for the else-clause to actually indicate that functionality is NOT currently implemented.  Won't this be confusing should the if-clause condition be met for some rows and not for others?

kyleoliveira:
retest this please

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: The update action may not actually perform an update (even if we did intend to implement the update feature now), but I will see if I can log some sort of message to indicate an attempt was made.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: The update action may not actually perform an update (even if we did intend to implement the update feature now), but I will see if I can log some sort of message to indicate an attempt was made.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: Good point, I'll add appropriate logging for when the data matches.

src/main/java/edu/cornell/kfs/pdp/CUPdpParameterConstants.java
```diff
@@ -4,7 +4,18 @@
 
     public static final String WARNING_INACTIVE_VENDOR_TO_EMAIL_ADDRESSES = "WARNING_INACTIVE_VENDOR_TO_EMAIL_ADDRESSES";
 
+    public static final String ACH_PERSONAL_CHECKING_TRANSACTION_CODE = "ACH_PERSONAL_CHECKING_TRANSACTION_CODE";
+    public static final String ACH_PERSONAL_SAVINGS_TRANSACTION_CODE = "ACH_PERSONAL_SAVINGS_TRANSACTION_CODE";
+    public static final String ACH_DIRECT_DEPOSIT_TRANSACTION_TYPE = "ACH_DIRECT_DEPOSIT_TRANSACTION_TYPE";
+    public static final String ACH_SECURE_FIELDS_ALLOWED_IN_ACCOUNT_EMAILS = "ACH_SECURE_FIELDS_ALLOWED_IN_ACCOUNT_EMAILS";
```
nkimble: In my opinion, it is not prudent to have this type of override ability as a user configurable application parameter; especially when that data can be sent out in emails.  The risks and ramifications of this type of data being accidentally exposed should this parameter be set incorrectly are too high. 
Was this an implementation decision or an end user requirement?  
Is there possibly another alternative?

cah292: The users wanted to be able to include the bank name in notification emails, where I added the ability to specify property placeholders in an email message parameter. I also added checks of DD AttributeSecurity to disallow secure fields in the final message; unfortunately, our customizations are masking bank routing number and bank name. Some sort of override was necessary for this strategy to allow for bank name inclusion.

The easiest alternative would be to add a new config property to replace the secure fields parameter, to prevent it from being altered at runtime. Another alternative would be to define a whitelist of specific placeholders that map to BO properties, but that would be a larger effort. I will at least look into implementing the former if you believe that would be enough to alleviate the security concerns.

bhutchinson: Is there a reason this needs to be configurable? Is the only field that is masked that they want to include the bank name? Are we expecting that to change in the future?

cah292: Yes, the bank name is the only masked field that users want included in the generated emails, and I do not anticipate users needing any other masked ones in there. We could hard-code a bank name property constant if necessary, if we're fine with adjusting it accordingly if the property name ends up changing with the upgrade.

kyleoliveira:
retest this please

cah292:
retest this please

cah292:
Thanks for the input, Nancy and Bryan. I've updated the feature so that it has better logging for the accounts that exist in KFS, and also relies on a hard-coded bank name constant to discern whether to include the potentially-sensitive field in the email body. I've also updated the unit test class so that the email formatting microtest will check for proper replacement of non-bank-name sensitive placeholders.

Please review again when you get a chance.

nkimble:
Changes look fine. Unit tests passed last night when code review changes were committed. Merging.

PR #91, opened 2016-04-21
KFSPTS-5643: Require positive approval for PREQs whose POs' amounts exceed a given limit, and display a prompt to the FO or FO delegate for such documents.

cah292:
This feature enforces positive approval of PREQ docs when their POs have a total amount matching or exceeding the "DEFAULT_POS_APRVL_LMT" parameter value already used for PREQs. Also, it forces a yes/no prompt to be displayed on such documents when a FO or FO delegate goes to approve them, to inform the users of why the positive approval is necessary.

PR #108, opened 2016-05-19
KFSPTS-5643: Require positive approval for PREQs whose POs' amounts equal or exceed a given limit, if the PO has quantity-type items.

cah292:
This is a newer version of the code from PR #91, which has removed the prompt to the FO/delegate users and now has some unit tests to go along with it.

I spread out the logic of the new PREQ service functionality to several methods to simplify testing the new logic via unit tests. And in the unit test itself, I added a forced update to a particular dependent parameter to ensure that its value would work with the item amounts from the associated fixtures. I think the change will get rolled back after the test runs; if you think that is not the case or if you have suggestions on improving that test, please let me know. (Due to it being in a Spring-dependent test, a mock service may not be suitable.)

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
@@ -9,6 +9,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.module.purap.PurapConstants;
 import org.kuali.kfs.module.purap.PurapConstants.ItemTypeCodes;
@@ -17,8 +18,10 @@ import org.kuali.kfs.module.purap.PurapParameterConstants;
 import org.kuali.kfs.module.purap.businessobject.PaymentRequestItem;
 import org.kuali.kfs.module.purap.businessobject.PurApAccountingLine;
 import org.kuali.kfs.module.purap.businessobject.PurApItem;
+import org.kuali.kfs.module.purap.businessobject.PurchaseOrderItem;
 import org.kuali.kfs.module.purap.document.AccountsPayableDocument;
 import org.kuali.kfs.module.purap.document.PaymentRequestDocument;
+import org.kuali.kfs.module.purap.document.PurchaseOrderDocument;
 import org.kuali.kfs.module.purap.document.service.impl.PaymentRequestServiceImpl;
 import org.kuali.kfs.sys.businessobject.Bank;
 import org.kuali.kfs.sys.context.SpringContext;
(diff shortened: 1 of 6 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -415,4 +418,65 @@ public String getPaymentRequestNoteTargetObjectId(String documentNumber) {
...
+    /**
+     * This implementation just checks the first item since CU does not allow a mix of
+     * quantity and non-quantity items on the same document.
+     * 
+     * @see edu.cornell.kfs.module.purap.document.service.CuPaymentRequestService#purchaseOrderForPaymentRequestHasQuantityItems(
+     * org.kuali.kfs.module.purap.document.PaymentRequestDocument)
+     */
+    @Override
+    public boolean purchaseOrderForPaymentRequestHasQuantityItems(PaymentRequestDocument document) {
+        return !((PurchaseOrderItem) document.getPurchaseOrderDocument().getItems().get(0)).isNoQtyItem();
```
db483: Are we sure there is at least one item on the PO doc?

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -415,4 +418,65 @@ public String getPaymentRequestNoteTargetObjectId(String documentNumber) {
...
+    /**
+     * Overridden to also require that if the associated PO has quantity items, then its amount must be within auto-approval limits.
+     * 
+     * @see org.kuali.kfs.module.purap.document.service.impl.PaymentRequestServiceImpl#isEligibleForAutoApproval(
+     * org.kuali.kfs.module.purap.document.PaymentRequestDocument, org.kuali.rice.core.api.util.type.KualiDecimal)
+     */
+    @Override
+    protected boolean isEligibleForAutoApproval(PaymentRequestDocument document, KualiDecimal defaultMinimumLimit) {
+        return purchaseOrderForPaymentRequestIsNonQuantityOrWithinAutoApproveAmountLimit(document)
+                && super.isEligibleForAutoApproval(document, defaultMinimumLimit);
```
db483: do you need to call super method here?

db483:
Code looks good, I had only one question about the items list on the PO doc. Is there always at least one item in that list? If not maybe the method should check if there are any items before retrieving the first item in the list.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
cah292: I know that REQS are required to have at least one item, so I believe the POs will also have an item. However, I'll work on adding a safeguard in there in case the list does end up being empty for some odd reason.

db483: Thank you Chad, once that is done I will merge the code in.

cah292:
retest this please

cah292:
I've updated that one method to add a check for a null/empty items list. Please review again when you get a chance.

db483:
Thank you Chad, looks good. I will merge it as soon as the unit tests finish.

PR #110, opened 2016-05-20
[KFSPTS-5731] Increase vendorOwnershipCategoryCode maxLenth to four c…

nkimble:
…haracters and remove the exactLength validation pattern restriction.


cah292:
Vendor field change looks good, and unit tests have passed. Merging.

PR #111, opened 2016-05-20

bhutchinson:


bhutchinson:
@cah292, this is very similar to the prior PR #104 (and #87 for that matter) for this work, just the batch job name has been changed to match the functional request. There were also some changes to the MockParameterServiceImpl because of how I had to revert the last commit for this since subsequent PRs had leveraged changes in that class.

cah292:
Changes appear to be good, and unit tests have passed. Merging.

PR #112, opened 2016-05-20
KFSPTS-5520: fix startup errors with upgraded code running against st…

bhutchinson:
…andalone rice 2.5.8 without cynergy


bhutchinson:
@jhulslander FYI, as with KFSPTS-5252, I don't expect the unit tests to pass for this since there are db updates that would be needed for this code to run that haven't been run against UTEST.

jhulslander:
well, that was a lot of changes.

PR #113, opened 2016-05-24
KFSPTS-5727: Added ability to control whether an Asset Transfer GLPE entry should have its sub-account forcibly cleared out, based on its account number and a new parameter.

cah292:
There are situations where Asset Transfer documents will create GLPE entries for plant accounts, but will often assign them the sub-account from the account that actually purchased the asset. This results in invalid account + sub-account pairs that require correction via GLCP. The proposed solution is to make the Asset Transfers check for plant accounts on the generated GLPE entries (based on exact-match or wildcard naming schemes from a parameter), and clear out their defined sub-accounts.

Because the default ParameterEvaluator object and service implementations do not allow for wildcard or regex matching (and there may not be an easy way to have the service return different instances without significant work), this solution does its own custom processing to check for wildcard/regex-based matches while also factoring the parameter's ALLOW/DISALLOW flag into account.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
@@ -1,0 +1,109 @@
+package edu.cornell.kfs.module.cam.document.service.impl;
+
+import java.util.regex.PatternSyntaxException;
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.coa.businessobject.Account;
+import org.kuali.kfs.module.cam.businessobject.Asset;
+import org.kuali.kfs.module.cam.businessobject.AssetGlpeSourceDetail;
+import org.kuali.kfs.module.cam.businessobject.AssetPayment;
+import org.kuali.kfs.module.cam.document.AssetTransferDocument;
+import org.kuali.kfs.module.cam.document.service.impl.AssetTransferServiceImpl;
+import org.kuali.rice.coreservice.api.parameter.EvaluationOperator;
+import org.kuali.rice.coreservice.api.parameter.Parameter;
+import org.kuali.rice.coreservice.framework.parameter.ParameterService;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+import edu.cornell.kfs.module.cam.CuCamsConstants;
+
+public class CuAssetTransferServiceImpl extends AssetTransferServiceImpl {
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(CuAssetTransferServiceImpl.class);
+
+    protected ParameterService parameterService;
+
+    /**
+     * Overridden to forcibly set the sub-account number on the generated postable object to null, if the account number
+     * is flagged as being one that should not have a sub-account number defined on related source details.
+     * The "ASSET_TRANSFER_PLANT_ACCOUNTS_FOR_SUB_ACCOUNTS" parameter governs whether such clearing behavior should be applied;
+     * see shouldPreserveSubAccount() for details.
+     * 
+     * @see org.kuali.kfs.module.cam.document.service.impl.AssetTransferServiceImpl#createAssetGlpePostable(
+     * org.kuali.kfs.module.cam.document.AssetTransferDocument, org.kuali.kfs.coa.businessobject.Account,
+     * org.kuali.kfs.module.cam.businessobject.AssetPayment, boolean, org.kuali.kfs.module.cam.document.service.impl.AssetTransferServiceImpl.AmountCategory)
+     */
+    @Override
+    protected AssetGlpeSourceDetail createAssetGlpePostable(
+            AssetTransferDocument document, Account plantAccount, AssetPayment assetPayment, boolean isSource, AmountCategory amountCategory) {
+        AssetGlpeSourceDetail postable = super.createAssetGlpePostable(document, plantAccount, assetPayment, isSource, amountCategory);
+        clearSubAccountIfNecessary(postable);
+        return postable;
+    }
+
+    /**
+     * Clears out the sub-account number on the source detail if the account number matches a particular pattern.
+     * In the event that a badly-formatted pattern causes an exception while matching, the source detail will be left as-is.
+     * 
+     * This was separated into its own method for unit testing convenience. 
+     * 
+     * @param postable The GLPE source detail whose sub-account number may need to be cleared based on account number.
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+     * @return true if the parameter is null or if its ALLOW/DISALLOW operator lines up with the match/no-match account check result, false otherwise.
+     * @throws PatternSyntaxException if the parameter exists and has a non-empty value but contains one or more invalid search expressions.
+     */
+    protected boolean shouldPreserveSubAccount(AssetGlpeSourceDetail postable) {
+        Parameter accountPatternsParameter = parameterService.getParameter(
+                Asset.class, CuCamsConstants.Parameters.ASSET_TRANSFER_PLANT_ACCOUNTS_FOR_SUB_ACCOUNTS);
+        if (ObjectUtils.isNull(accountPatternsParameter)) {
+            return true;
+        }
+        return EvaluationOperator.ALLOW.equals(accountPatternsParameter.getEvaluationOperator()) == accountMatchesAtLeastOnePattern(
```
bhutchinson: This line made me pause and spend a fair amount of time trying to understand what it was doing. It might be beneficial to pull this logic out into another method with a name that explains what's going on here just to make it more readable/understandable.

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+     * Clears out the sub-account number on the source detail if the account number matches a particular pattern.
+     * In the event that a badly-formatted pattern causes an exception while matching, the source detail will be left as-is.
+     * 
+     * This was separated into its own method for unit testing convenience. 
+     * 
+     * @param postable The GLPE source detail whose sub-account number may need to be cleared based on account number.
+     */
+    protected void clearSubAccountIfNecessary(AssetGlpeSourceDetail postable) {
+        try {
+            if (!shouldPreserveSubAccount(postable)) {
```
bhutchinson: I'm generally not a fan of negative logic since it seems to require more thought to process. Would it be possible to invert this to something like shouldClearSubAccount()?

src/test/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImplTest.java
```diff
@@ -0,0 +1,251 @@
...
+    private static final String TEST_PATTERN_REGULAR_EXACT_MATCH = TEST_ACCOUNT;
+    private static final String TEST_PATTERN_PLANT_EXACT_MATCH = TEST_PLANT_ACCOUNT;
+    private static final String TEST_PATTERN_WILDCARD_BOTH_MATCH = "G10*";
+    private static final String TEST_PATTERN_NO_MATCH = "NONE";
+    private static final String TEST_MULTI_PATTERN = "*XXXX;*CAPT";
+    private static final String TEST_MULTI_PATTERN_REGULAR_MATCH = "*XXXX;G104*";
+    private static final String TEST_MULTI_PATTERN_NO_MATCH = "*XXXX;NONE";
+    private static final String TEST_BAD_PATTERN = "[*CAPT}";
+
+    // These flags are for readability convenience.
```
bhutchinson: Not sure this and some of the other comments in this test are adding much value.

bhutchinson:
In general, this looks good. I did add a couple comments/questions mostly related to readability/understandability of the code.

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
cah292: Ok, I'll get this separated out for clarity.

src/test/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImplTest.java
cah292: Ok, I'll review and remove comments as needed.

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
cah292: I could invert it, but that would ultimately require changing the "==" to "!=" in the line that I'm separating out for clarity. I will see if that other line can be moved to a method that can be named accordingly for the flipped logic.

bhutchinson: Ok, thanks; if it becomes a lot of work or ends up making things too confusing elsewhere, don't worry about it.

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
bhutchinson: thanks

src/test/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImplTest.java
bhutchinson: thanks

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetTransferServiceImpl.java
cah292: I'm trying to determine if reversing this to ultimately change the other line to a "!=" would be too confusing, since finding a match in whitelist mode or not finding a match in blacklist mode currently means the entry should be left as-is. I could alternatively reverse the interpretation of the parameter or make it just a blacklist, but that would be inconsistent with similar KFS parameters that I was trying to copy the style of.

cah292:
retest this please

cah292:
Code has been updated based on feedback. I did work on reversing the logic in a few places to minimize NOT-operator usage; let me know if it seems too confusing now. As for the whitelist-state-vs-parameter-match line, the existing method was already quite small to begin with, so I added some local variables to add to the clarity instead of creating another method. If you would rather have the separate method for that check, please let me know.

bhutchinson:
Looks good, thanks.

PR #114, opened 2016-05-25
KFSPTS-5922: get unit test suite to run with upgraded code

bhutchinson:


src/main/resources/edu/cornell/kfs/module/purap/cu-ojb-purap.xml
```diff
@@ -3265,10 +3265,10 @@
     <field-descriptor name="preqBankCountryCode" column="PREQ_BNK_CNTRY_CD" jdbc-type="VARCHAR" />
     <field-descriptor name="preqAttentionLineText" column="PREQ_ATTN_LN_TXT" jdbc-type="VARCHAR" />
     <field-descriptor name="preqAdditionalWireText" column="PREQ_ADDL_WIRE_TXT" jdbc-type="VARCHAR" />
-    <field-descriptor name="preqPayeeAccountNumber" column="PREQ_PAYEE_ACCT_NBR" jdbc-type="VARCHAR" conversion="org.kuali.rice.core.framework.persistence.ojb.conversion.OjbKualiEncryptDecryptFieldConversion"/>
+    <field-descriptor name="preqPayeeAccountNumber" column="PREQ_PAYEE_ACCT_NBR" jdbc-type="VARCHAR" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbKualiEncryptDecryptFieldConversion"/>
```
jhulslander: I don't think it is worth stopping this from going in, but based on the conversation on slack, is there something different we should do with encryption?

jhulslander:
I concentrated on the added or changed files, it looks fine to me.  My only concern is around did we actually modify any of those test package files that were removed as many test package files were just copied.

src/main/resources/edu/cornell/kfs/module/purap/cu-ojb-purap.xml
bhutchinson: Short term, I added cynergy-customizations as a test dependency in the pom, overrode the encryptionService with ours and set the encryption key in our properties file. This made the tests pass. Long term, we will need to put cynergy back.

bhutchinson:
None of those files were necessary to make the tests pass, so I don't think we need them, even if they were changed, thanks.

PR #115, opened 2016-05-25
KFSPTS-5486: Add Vendor Type name to the Payee Type name on DVs.

cah292:
This adds a new Payee Type Suffix property to the Payee Detail Extension object so that when a vendor payee is selected, its vendor type will be displayed to the right of the regular Payee Type value. The suffix is expected to be auto-refreshable prior to routing or ad hoc completion, but after that it is expected to remain static even if the vendor's type is changed later. This allows for directly displaying the vendor's type on the DV, and for showing what that value was as of the time that it started routing.

I also updated one of the existing microtest classes to add some microtests for this. I did have to modify some existing parts and tests slightly because it previously had incorrect usage of vendor ID values.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/businessobject/CuDisbursementVoucherPayeeDetail.java
+++ b/src/main/java/edu/cornell/kfs/fp/businessobject/CuDisbursementVoucherPayeeDetail.java
@@ -1,8 +1,10 @@
 package edu.cornell.kfs.fp.businessobject;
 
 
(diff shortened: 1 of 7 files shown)


cah292:
retest this please

cah292:
I had forgotten to apply the SQL to UTEST prior to creating the PR. This next run of the tests should work.

db483:
Code looks good.

PR #116, opened 2016-05-26
KFSPTS-5484: Added Budget Start/Stop/Total fields from our EZRA integration to the Award extended attribute.

cah292:
This feature adds Budget Start, Budget Stop, and Budget Total fields to our Award extension objects, which can be populated via the corresponding fields from our EZRA integration tables and can also be edited via the Award maintenance document. Our EZRA-related OJB mappings in KFS have been updated to add the missing field references, but no SQL is required on the EZRA end because those fields already exist in the relevant table; we only need SQL for adding the fields to the AwardExtendedAttribute table.

The only extra business rule needed at this time is to make sure the Budget End is after the Budget Start. The new fields are not configured as required ones, to avoid potentially breaking older Award maintenance docs or objects.

### DIFF
--- a/src/test/java/edu/cornell/kfs/module/ezra/service/ezraUpdateAwardImplTest.java
+++ b/src/test/java/edu/cornell/kfs/module/ezra/service/ezraUpdateAwardImplTest.java
@@ -1,26 +1,26 @@
 package edu.cornell.kfs.module.ezra.service;
 
+import static org.junit.Assert.assertNotEquals;
 import static org.kuali.kfs.sys.fixture.UserNameFixture.ccs1;
 
+import java.sql.Date;
+import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.sql.Date;
-import java.util.Calendar;
-
-import edu.cornell.kfs.module.cg.businessobject.CuAward;
-import edu.cornell.kfs.module.ezra.businessobject.EzraProposalAward;
-import edu.cornell.kfs.module.ezra.service.EzraService;
-
-import org.kuali.rice.core.api.datetime.DateTimeService;
-import org.kuali.rice.krad.service.BusinessObjectService;
 
 import org.kuali.kfs.module.cg.businessobject.Award;
 import org.kuali.kfs.sys.ConfigureContext;
 import org.kuali.kfs.sys.context.KualiTestBase;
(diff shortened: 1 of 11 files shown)


src/test/java/edu/cornell/kfs/module/ezra/service/ezraUpdateAwardImplTest.java
```diff
@@ -41,10 +41,17 @@ protected void setUp() throws Exception {
     public void testUpdateAwards () {
         Map fields = new HashMap();
         String awardProjectTitle = "ezraUpdateTest";
-        String awardProjectTitleAfter = ""; 
-        java.sql.Date today = SpringContext.getBean(DateTimeService.class).getCurrentSqlDate();
+        String awardProjectTitleAfter = "";
+        Date today = SpringContext.getBean(DateTimeService.class).getCurrentSqlDate();
```
jhulslander: Can we create  a getDateTimeService() function, to support micro testing in the future?

jhulslander:
This looks pretty good, I only had one note for a possible change.  Concerning the rule about beginning date must be before ending date, what happens if data from exra that breaks that rule?

src/test/java/edu/cornell/kfs/module/ezra/service/ezraUpdateAwardImplTest.java
cah292: We could create a getDateTimeService() function, but that might be beyond the scope of this user story. This unit test case already makes heavy use of SpringContext.getBean() (see the setUp() function above), and also relies on the EzraService performing actual data access to EZRA. Some significant refactoring would be necessary to turn this unit testing class into a micro testing one.

In the case of the DateTimeService above, I was simply editing an existing SpringContext.getBean() call to shorten the fully-qualified Date classname, to avoid confusion among the adjacent variables that are not fully qualified but still refer to the same class.

cah292:
With regards to the begin-date-before-end-date rule, getting bad data from EZRA may not prevent the document from routing. The EzraService appears to be using the KEW WorkflowDocument directly to perform the routing actions, thus possibly skipping business rule validation for EZRA-triggered updates. (Perhaps it's because EZRA is intended to be the System of Record for this CG data, as per the CG-14-01 OpDoc?)

If we think this validation behavior needs to be changed, I would suggest creating a separate user story for that.

PR #117, opened 2016-06-01
KFSPTS-5411: Updated 1099 processing to allow for specifying mappings between doc types and 1099 boxes.

cah292:
This updates the 1099 tax file generation process so that specific document types will be explicitly mapped to a given 1099 tax box. (A new parameter controls these mappings.) This takes precedence over (and has been implemented in a similar manner as) the new payment-reason-based 1099 box mappings from KFSPTS-5409.

### DIFF
--- a/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
+++ b/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
@@ -29,6 +29,7 @@ import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.DerivedValuesRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.TransactionDetailRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.VendorAddressRow;
 import edu.cornell.kfs.tax.dataaccess.impl.TaxTableRow.VendorRow;
+import edu.cornell.kfs.tax.service.DocumentType1099BoxService;
 import edu.cornell.kfs.tax.service.PaymentReason1099BoxService;
 
 /**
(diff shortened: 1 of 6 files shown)


cah292:
Note: UTEST is currently down, so we will need to wait for it to be available again before rerunning the unit tests on this PR.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRow1099Processor.java
```diff
@@ -925,7 +927,13 @@ private void checkForGeneralExclusionsAndFindTaxBox(ResultSet rs, Transaction109
         int idx;
         
         // Find tax box.
-        if (getPaymentReason1099BoxService().isPaymentReasonMappedTo1099Box(paymentReasonCodeP.value)) {
+        if (getDocumentType1099BoxService().isDocumentTypeMappedTo1099Box(docTypeP.value)) {
```
kyleoliveira: Not sure why there are so many instances of it, but shouldn't it be
```java
docTypeP.getValue()
```
rather than
```java
docTypeP.value
```
?

cah292: This has to do with how I had implemented the TAX module at the time, where the tax processor classes were intended to manipulate those fields directly and the superclass was intended to use the appropriate getValue() methods instead. I suppose getValue() could work in this case due to it being a string value, but doing the field access would be consistent with the rest of the code.

Perhaps the implementation of the TAX module can be revisited at a later date to try and make this cleaner, but that would be beyond the scope of this PR.

kyleoliveira: Alright, we'll leave as-is in that regard then.

kyleoliveira:
I think this will be good to go once the tests pass.

kyleoliveira:
retest this please

kyleoliveira:
retest this please

kyleoliveira:
Tests passed, merging.

PR #118, opened 2016-06-01

cah292:
This is nearly the same as the code from PR #106, except that the processing for PayeeACHAccountExtractServiceImpl now explicitly starts a new transaction. This is to prevent certain errors from showing up that complain about no active persistence broker transaction being active, which could occur when programmatically routing the PAAT documents.


cah292:
Note: UTEST may still be down, so the unit tests for this can't run until that database is back up.

cah292:
retest this please

nkimble:
Two line change introduced by this PR looks ok.

PR #119, opened 2016-06-01
[KFSPTS-4484] Restrict Size of DV check stub text to match limits in PDP - users can type info that will never make it to the check stub

kyleoliveira:


src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -1,67 +1,242 @@
...
+      throw new IllegalArgumentException("Invalid Max Notes Lines parameter, value: "+maxNoteLinesParam+" cannot be converted to an integer");
+    }
+
+    // The WordUtils should be sufficient for the majority of cases.  This method will
+    // word wrap the whole string based on the MAX_NOTE_LINE_SIZE, separating each wrapped
+    // word by a newline character.  The 'wrap' method adds line feeds to the end causing
+    // the character length to exceed the max length by 1, hence the need for the replace
+    // method before splitting.
+    String   wrappedText = WordUtils.wrap(text, CuDisbursementVoucherConstants.DV_EXTRACT_MAX_NOTE_LINE_SIZE);
+    String[] noteLines   = wrappedText.replaceAll("[\r]", "").split("\\n");
```
cah292: Has the NullPointerException problem mentioned in Cathy's user story comments been resolved by these changes? I think this particular line was the one causing issues before, and it looks like it could still cause problems if the "text" parameter is null (since WordUtils.wrap() will return null if the input is null).

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -1,67 +1,242 @@
...
-
-        // if payment method is CHECK and wire tab contains data, ask user to clear tab
-        if ((StringUtils.equals(KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_CHECK, dvDocument.getDisbVchrPaymentMethodCode()) || StringUtils.equals(KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_DRAFT, dvDocument.getDisbVchrPaymentMethodCode())) && hasWireTransferValues(dvWireTransfer)) {
-            String questionText = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(CUKFSKeyConstants.QUESTION_CLEAR_UNNEEDED_WIRE_TAB);
-
-            boolean clearTab = super.askOrAnalyzeYesNoQuestion(KFSConstants.DisbursementVoucherDocumentConstants.CLEAR_WIRE_TRANSFER_TAB_QUESTION_ID, questionText);
-            if (clearTab) {
-                // NOTE: Can't replace with new instance because Foreign Draft uses same object
-                clearWireTransferValues(dvWireTransfer);
+  /**
```
cah292: This file appears to be using 2 spaces per level of indentation instead of 4. I would suggest changing it to 4 spaces per level so that it's consistent with our other code files.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -1,67 +1,242 @@
...
+   * allowed in the FastTrack file format and optionally allows the user to go back and fix the issue.
+   * If the user chooses to fix the issue, it will return false, otherwise true.
+   *
+   * @param dvDocument submitted disbursement voucher document
+   * @return Returns true if the user wishes to correct the Check Stub Text, false otherwise.
+   */
+  @SuppressWarnings("deprecation")
+  protected boolean checkCheckStubTextOverflow(DisbursementVoucherDocument dvDocument) {
+    boolean checkStubTextOK = true;
+    int maxCheckStubLines = 4; // This is the number allowed in the FastTrack file format
```
cah292: I would suggest moving this number to at least a constant.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -1,67 +1,242 @@
...
+   * we're really only interested in how the existing Check Stub Text lines will be broken up).
+   */
+  private List<PaymentNoteText> buildNotesForCheckStubText(String text) {
+    PaymentNoteText pnt = null;
+    List<PaymentNoteText> pnts = new ArrayList<PaymentNoteText>();
+    String maxNoteLinesParam = SpringContext.getBean(ParameterService.class)
+                                            .getParameterValueAsString(KfsParameterConstants.PRE_DISBURSEMENT_ALL.class, 
+                                                                       PdpParameterConstants.MAX_NOTE_LINES);
+    int previousLineCount = 0;
+    int maxNoteLines = 3;
```
cah292: I would recommend moving the default value of "3" to a constant.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Ok, I'll fix that.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -1,67 +1,242 @@
...
+   * @param document submitted document
+   * @return true if pre-rules execute successfully
+   * @see org.kuali.rice.kns.rules.PromptBeforeValidationBase#doRules(org.kuali.rice.kns.document.MaintenanceDocument)
+   */
+  @Override
+  public boolean doPrompts(Document document) {
+    boolean preRulesOK = super.doPrompts(document);
+
+    DisbursementVoucherDocument dvDocument = (DisbursementVoucherDocument) document;
+    checkSpecialHandlingIndicator(dvDocument);
```
cah292: I think super.doPrompts() already calls this method; does it need to be called again here?

cah292:
Finished review; I did have some code suggestions plus a question about an issue recently reported on the user story, but other than that it looks good. Will wait for feedback before reviewing again.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Ok.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Ok.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Good catch, I'll add an exception for when the parameter is null.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: It does seem to call it. Perhaps there used to be a custom one that has been removed since the last time I worked on this? Regardless, I'll change this then.

kyleoliveira:
Ready for another look.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -64,4 +96,151 @@ protected boolean checkWireTransferTabState(DisbursementVoucherDocument dvDocume
         return tabStatesOK;
     }
 
+    /**
+     * This method notifies the user if the Check Stub Text will break into more lines than those
+     * allowed in the FastTrack file format and optionally allows the user to go back and fix the issue.
+     * If the user chooses to fix the issue, it will return false, otherwise true.
+     *
+     * @param dvDocument submitted disbursement voucher document
+     * @return Returns true if the user wishes to correct the Check Stub Text, false otherwise.
```
cah292: After further examination, I noticed that this line's comment is the opposite of what the paragraph above says about the return values. I would strongly recommend fixing these comments for consistency about when the method returns true and when it returns false.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -64,4 +96,151 @@ protected boolean checkWireTransferTabState(DisbursementVoucherDocument dvDocume
...
+     */
+    @SuppressWarnings("deprecation")
+    protected boolean checkCheckStubTextOverflow(DisbursementVoucherDocument dvDocument) {
+        boolean checkStubTextOK = true;
+        int maxCheckStubLines = CuDisbursementVoucherDocumentPreRules.FASTTRACK_MAX_LINE_COUNT;
+
+        List<PaymentNoteText> pntList = buildNotesForCheckStubText(dvDocument.getDisbVchrCheckStubText());
+
+        if(pntList.size() > maxCheckStubLines) {
```
cah292: If there's no need to change the value of maxCheckStubLines, then you could try using your new constant above directly.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -64,4 +96,151 @@ protected boolean checkWireTransferTabState(DisbursementVoucherDocument dvDocume
...
+    }
+
+    /* This is borrowed mostly from
+     * edu.cornell.kfs.document.service.impl.CuDisbursementVoucherExtractionHelperServiceImpl#buildNoteForCheckStubText
+     * with the modification of removing the setting for the previous line count (as we don't have previous lines, and
+     * we're really only interested in how the existing Check Stub Text lines will be broken up).
+     */
+    private List<PaymentNoteText> buildNotesForCheckStubText(String text) {
+        if(text == null) {
+            throw new IllegalArgumentException("Cannot build notes for the Check Stub text from a null string.");
```
cah292: I would advise checking if this is indeed the correct way to handle the null text argument, given that the caller is retrieving it from the DV document. If other checks already exist elsewhere to prevent submission of a DV with null check stub text, then I would advise doing some other handling here (like returning an empty list) if throwing the IllegalArgumentException ends up showing an incident report screen to the user.

cah292:
Reviewed again and found a couple more items that should be looked into, but otherwise the changes look good. Will wait for the next set of feedback.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Will do.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Will fix.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
kyleoliveira: Will do.

kyleoliveira:
Updated.

cah292:
Code looks good and the unit tests have passed. Merging.

PR #121, opened 2016-06-03
KFSPTS-5501: Added a Rice 2.5.x fix for immediately saving BO note adds/deletes, and tweaked it to only apply to Account maintenance docs.

cah292:
This adds an altered version of the Rice 2.5.x BO note add/delete fix, which has been specifically modified to only apply to Account maintenance documents for the present time. It requires some new Struts-related files, which also require a property change to KFS to load them in. (KFS should still be able to boot up without the property change, though.)

We may be able to remove this customization for our next KFS upgrade if users do not need it limited to Account documents afterwards.

### DIFF
--- a/src/test/java/edu/cornell/kfs/sys/document/web/struts/CuFinancialMaintenanceDocumentActionTest.java
+++ b/src/test/java/edu/cornell/kfs/sys/document/web/struts/CuFinancialMaintenanceDocumentActionTest.java
@@ -1,0 +1,190 @@
+package edu.cornell.kfs.sys.document.web.struts;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.easymock.EasyMock;
+import org.easymock.IMockBuilder;
+import org.junit.Before;
+import org.junit.Test;
+import org.kuali.kfs.coa.businessobject.Account;
+import org.kuali.kfs.coa.businessobject.Organization;
+import org.kuali.kfs.sys.businessobject.FinancialSystemDocumentHeader;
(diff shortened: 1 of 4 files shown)


src/test/java/edu/cornell/kfs/sys/document/web/struts/CuFinancialMaintenanceDocumentActionTest.java
```diff
@@ -0,0 +1,199 @@
...
+    protected org.kuali.rice.krad.bo.NoteType createMockNoteTypeBo(NoteType noteTypeEnum) {
+        org.kuali.rice.krad.bo.NoteType noteTypeBo = EasyMock.createMock(org.kuali.rice.krad.bo.NoteType.class);
+        EasyMock.expect(noteTypeBo.getNoteTypeCode()).andStubReturn(noteTypeEnum.getCode());
+        EasyMock.replay(noteTypeBo);
+        return noteTypeBo;
+    }
+
+
+
+    private static class TestCuFinancialMaintenanceDocumentAction extends CuFinancialMaintenanceDocumentAction {
```
bhutchinson: It looks like this class isn't doing much. Just overriding one method but calling the super implementation in it. Am I missing something, or could this just be eliminated?

bhutchinson:
Just one question, otherwise this looks good.

src/test/java/edu/cornell/kfs/sys/document/web/struts/CuFinancialMaintenanceDocumentActionTest.java
cah292: This override just expands the method visibility from protected to public. I suppose it's not strictly needed because the test is in the same package as the superclass (and hence can call the protected method); let me know if you think it's worth removing.

bhutchinson: Since it's not needed, I would remove it, thanks!

cah292:
The new unit test class has been updated as per developer feedback. This is ready for review again.

bhutchinson:
Thanks. Looks good, will merge once tests complete successfully.

PR #122, opened 2016-06-06
KFSPTS-5643: Require positive approval for PREQs whose POs' amounts equal or exceed a given limit.

cah292:
This is a newer version of the forced-positive-approval feature from PR #108, which now applies to both non-quantity and quantity-based POs on the given PREQs, and which now relies on a new parameter instead of reusing an existing one.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
@@ -3,6 +3,7 @@ package edu.cornell.kfs.module.purap.document.service.impl;
 import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -415,4 +418,39 @@ public String getPaymentRequestNoteTargetObjectId(String documentNumber) {
...
+    @Override
+    public boolean purchaseOrderForPaymentRequestIsWithinAutoApproveAmountLimit(PaymentRequestDocument document) {
+        PurchaseOrderDocument po = document.getPurchaseOrderDocument();
+        String amountParameterValue = parameterService.getParameterValueAsString(
+                PaymentRequestDocument.class, CUPurapParameterConstants.DEFAULT_PURCHASE_ORDER_POS_APRVL_LMT_FOR_PREQ);
+        KualiDecimal amountLimit = new KualiDecimal(amountParameterValue);
+        boolean withinLimit = po.getFinancialSystemDocumentHeader().getFinancialDocumentTotalAmount().isLessThan(amountLimit);
+        LOG.info("PayReq " + document.getDocumentNumber()
+                + (withinLimit
+                        ? " has the potential for auto-approval because the amount on the associated PO is below the limit of "
```
kyleoliveira: This is quite a long concatenation. The strings in this ternary statement should probably be turned into variables.

kyleoliveira:
Just one comment, but mostly ready to go I think.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
cah292: Ok, I will work on converting this string setup into something more readable.

cah292:
I've updated the one logging section in the code to make things a little clearer. Please review again when you get a chance.

kyleoliveira:
Looks good, just waiting for the unit tests to complete.

PR #124, opened 2016-06-07
KFSPTS-3599 Add new fields to Account Global

db483:


db483:
closing pull request, need to update code.

PR #125, opened 2016-06-07
KFSPTS-3599 Add new fields to Account Global doc

db483:


src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -8,146 +8,326 @@
...
 import org.kuali.rice.krad.bo.PersistableBusinessObject;
 import org.kuali.rice.krad.service.BusinessObjectService;
+import org.kuali.rice.krad.util.ObjectUtils;
+import org.kuali.rice.location.framework.campus.CampusEbo;
 
-public class CuAccountGlobal extends AccountGlobal {
+import edu.cornell.kfs.module.cg.businessobject.InvoiceFrequency;
+import edu.cornell.kfs.module.cg.businessobject.InvoiceType;
+
+public class CuAccountGlobal extends AccountGlobal implements GlobalBusinessObject{
```
cah292: AccountGlobal already implements GlobalBusinessObject, so there shouldn't be a need to have CuAccountGlobal specifically reference it here.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -8,146 +8,326 @@
...
 
     protected String majorReportingCategoryCode;
-    protected MajorReportingCategory majorReportingCategory;
+    protected String accountPhysicalCampusCode;
+    protected Date accountEffectiveDate;
+    protected boolean accountOffCampusIndicator;
+    protected boolean closed;
+    protected String accountTypeCode;
+    protected String appropriationAccountNumber;
+    protected boolean accountsFringesBnftIndicator;
```
cah292: For the applicable boolean fields that map to a matching field on the Account BO, I would suggest declaring them as Boolean object fields and requiring them to have non-null values in order for them to be included in the global updates. As an example, see the existing "pendingAcctSufficientFundsIndicator" account field, which is declared as a Boolean on the global version and which overrides the control type in the data dictionary.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -8,146 +8,326 @@
...
+	private void updateAccountValuesAndAddToPersistablesList(List<PersistableBusinessObject> persistables,AccountGlobalDetail accountGlobalDetail) {
+		Account account = updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccountGlobalDoc(accountGlobalDetail);
+		
+		if (ObjectUtils.isNotNull(account)){
+			persistables.add(account);   
+		}
+	}
     
-    public String getMajorReportingCategoryCode() {
+    private Account updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccountGlobalDoc(GlobalBusinessObjectDetailBase globalDetail) {
```
cah292: Please update the horizontal formatting of this method; its current arrangement makes it harder to spot the start of the subsequent method.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -8,146 +8,326 @@
...
+		
+		if (StringUtils.isNotBlank(accountPhysicalCampusCode)) {
+		    account.setAccountPhysicalCampusCode(accountPhysicalCampusCode);
+		}
+
+		if (ObjectUtils.isNotNull(accountEffectiveDate)) {
+		    account.setAccountEffectiveDate(accountEffectiveDate);
+		}
+		
+		if(accountOffCampusIndicator){
```
cah292: As per my previous comment related to Booleans, you may be able to update many of these similar comparisons to check for non-null values instead, allowing for setting the properties to either true or false on a global update. Unless there's a specific business requirement to only allow them to be set to true on a global update.

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
+import org.kuali.rice.krad.bo.GlobalBusinessObjectDetailBase;
+import org.kuali.rice.krad.bo.PersistableBusinessObjectBase;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+public abstract class GlobalObjectWithIndirectCostRecoveryAccounts extends PersistableBusinessObjectBase {
```
cah292: If this is only intended for use by global BOs, I would suggest making this class implement GlobalBusinessObject to avoid confusion. Since this is an abstract class, you don't have to specifically define the global BO methods here.

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
```diff
@@ -0,0 +1,109 @@
...
+		List<IndirectCostRecoveryAccount> addList = new ArrayList<IndirectCostRecoveryAccount>();
+
+		if (indirectCostRecoveryAccounts.size() > 0) {
+			for (IndirectCostRecoveryAccountChange newICR : indirectCostRecoveryAccounts) {
+				boolean found = false;
+
+				int temp = -1;
+				int i = 0;
+
+				while (i < icrAccounts.size() && (!found || (found && temp!=-1))) {
```
cah292: I realize that this is copied from a different section and that I think I had suggested an iterative solution for this in the past. However, based on our recent "Clean Code" meetings, this section could be changed around to improve clarity and without needing to adhere to an iterative solution. Sorry for the inconvenience.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,193 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
+		List<IndirectCostRecoveryAccount> currentActiveIndirectCostRecoveryAccountList = buildCurrentActiveICRAccountsAfterICRUpdates(icrUpdates, existingICRs, globalDetail, globalObjectWithIndirectCostRecoveryAccounts);
+
+		if ((ObjectUtils.isNull(currentActiveIndirectCostRecoveryAccountList) || (currentActiveIndirectCostRecoveryAccountList.size() == 0))) {
+			return true;
+		} else {
+
+			if (totalDistributionOnUpdatedActiveICRsIs100Percent(currentActiveIndirectCostRecoveryAccountList)) {
+				return true;
+			} else {
+				return false;
```
cah292: I would recommend returning the value of the "if" check directly instead of doing the "return true/false" branching.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,193 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
-			for (IndirectCostRecoveryAccount icrAccount : existingIcrAccounts) {
-				existingICRsWorkList.add(icrAccount);
+	protected boolean totalDistributionOnUpdatedActiveICRsIs100Percent(List<IndirectCostRecoveryAccount> currentActiveIndirectCostRecoveryAccountList) {
+		BigDecimal totalDistribution =  computeTotalDistribution(currentActiveIndirectCostRecoveryAccountList);
+		if (totalDistribution.compareTo(ONE_HUNDRED_PERCENT) == 0) {
+			return true;
+		}
+		else{
+			return false;
+		}
```
cah292: Another area where the "if" check result could be returned directly instead of using the "return true/false" branching.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,193 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
+				boolean found = false;
+				existingICRsToBeUpdated.clear();
+				for (IndirectCostRecoveryAccount icrAccount : icrAccountsNotUpdated) {
+					existingICRsToBeUpdated.add(icrAccount);
+				}
+				
+				IndirectCostRecoveryAccount tempIcrAccount = null;
+				int i = 0;
+				
+				while(i < existingICRsToBeUpdated.size() && (!found || (found && ObjectUtils.isNotNull(tempIcrAccount))) ){
```
cah292: Another area where the iterative solution could potentially be replaced with a clearer short-circuiting one. Sorry about suggesting the less-clear iterative strategy in the older review.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,193 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
-		icrAccount.setChartOfAccountsCode(chart);
-		icrAccount.setIndirectCostRecoveryAccountNumber(newICR.getIndirectCostRecoveryAccountNumber());
-		icrAccount.setIndirectCostRecoveryFinCoaCode(newICR.getIndirectCostRecoveryFinCoaCode());
-		icrAccount.setActive(newICR.isActive());
-		icrAccount.setAccountLinePercent(newICR.getAccountLinePercent());
+		return  currentActiveIndirectCostRecoveryAccountList;
+    }
+
+	private String buildMessageFromPrimaryKey(GlobalBusinessObjectDetailBase detail){
+		StringBuffer message = new StringBuffer();
```
cah292: Unless you have a specific need for using StringBuffer's thread-safety or regex-related features, I would suggest using StringBuilder instead for improved efficiency.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,193 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
+		if(detail instanceof AccountGlobalDetail){
+			AccountGlobalDetail accountGlobalDetail = (AccountGlobalDetail)detail;
+			message.append(accountGlobalDetail.getChartOfAccountsCode());
+			message.append("-");
+			message.append(accountGlobalDetail.getAccountNumber());
+		}
 		
-		return icrAccount;
-	}
+		if(detail instanceof SubAccountGlobalDetail){
```
cah292: If we don't expect a detail to be both an AccountGlobalDetail and a SubAccountGlobalDetail, then you could make this an "else if" instead.

src/main/java/org/kuali/kfs/coa/businessobject/AccountGlobal.java
```diff
@@ -0,0 +1,1105 @@
...
+import org.kuali.rice.location.framework.state.StateEbo;
+
+import edu.cornell.kfs.coa.businessobject.GlobalObjectWithIndirectCostRecoveryAccounts;
+import edu.cornell.kfs.coa.businessobject.IndirectCostRecoveryAccountChange;
+import edu.cornell.kfs.coa.businessobject.SubAccountGlobalDetail;
+
+/**
+ *
+ */
+public class AccountGlobal extends GlobalObjectWithIndirectCostRecoveryAccounts implements GlobalBusinessObject {
```
cah292: Is this class only being overwritten to extend from GlobalObjectWithIndirectCostRecoveryAccounts? If so, perhaps there's a better way to set this up on CuAccountGlobal so that we don't need to overwrite the base class directly. Unfortunately, that would mean converting GlobalObjectWithIndirectCostRecoveryAccounts into an interface, meaning its logic would have to be moved to a service class or similar location. Not sure if that would be more of a hassle than just overwriting this class.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+            Collection<SubFundProgram> retVals = bos.findMatching(SubFundProgram.class, fieldValues);
+            
+            if (retVals.isEmpty()) {
+                success = false;
+                putFieldError(CUKFSPropertyConstants.PROGRAM_CODE, CUKFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_PROGRAM_CODE_NOT_GROUP_CODE, new String[] {subFundProg, subFundGroupCode});
+            } else {
+            	for (SubFundProgram sfp : retVals) {
+            		if (!sfp.isActive()) {
+                        putFieldError(CUKFSPropertyConstants.PROGRAM_CODE, KFSKeyConstants.ERROR_INACTIVE, getFieldLabel(Account.class, CUKFSPropertyConstants.PROGRAM_CODE));
+                        success = false;
```
cah292: Given that the error message is not dependent on the SubFundProgram field values, it might be good to short-circuit this loop after finding an inactive one so as to avoid potentially adding duplicate messages for multiple inactive ones.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+            fieldValues.put(CUKFSPropertyConstants.APPROPRIATION_ACCT_NUMBER, appropriationAccountNumber);
+            
+            Collection<AppropriationAccount> retVals = bos.findMatching(AppropriationAccount.class, fieldValues);
+            
+            if (retVals.isEmpty()) {
+                success = false;
+                putFieldError(CUKFSPropertyConstants.APPROPRIATION_ACCT_NUMBER, CUKFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_APPROP_ACCT_NOT_GROUP_CODE, 
+                        new String[] {appropriationAccountNumber, subFundGroupCode});
+            } else {
+                for (AppropriationAccount sfp : retVals) {
```
cah292: You should rename the "sfp" variable to something that better reflects the AppropriationAccount name.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+                    if (!sfp.isActive()) {
+                        putFieldError(CUKFSPropertyConstants.APPROPRIATION_ACCT_NUMBER, KFSKeyConstants.ERROR_INACTIVE, 
+                                getFieldLabel(AccountGlobal.class, CUKFSPropertyConstants.APPROPRIATION_ACCT_NUMBER));
+                        success = false;
```
cah292: Another place where the loop could be short-circuited upon finding an inactive object, due to error message not using the object's field values.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
     }
 
+    protected boolean checkSubFundProgram(MaintenanceDocument document) {
+        boolean success = true;
+
+        String subFundGroupCode = newAccountGlobal.getSubFundGroupCode();
+        String subFundProg = newAccountGlobal.getProgramCode();
+        BusinessObjectService bos = SpringContext.getBean(BusinessObjectService.class);
+
+        if (!StringUtils.isBlank(subFundProg)) {
```
cah292: I would suggest using "StringUtils.isNotBlank()" instead of "!StringUtils.isBlank()" to improve clarity.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+            Map fieldValues = new HashMap();
```
cah292: You should use generics here, unless it's the only discernible way to call a method that expects an unchecked one.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+            	for (SubFundProgram sfp : retVals) {
+            		if (!sfp.isActive()) {
+                        putFieldError(CUKFSPropertyConstants.PROGRAM_CODE, KFSKeyConstants.ERROR_INACTIVE, getFieldLabel(Account.class, CUKFSPropertyConstants.PROGRAM_CODE));
+                        success = false;
+            		}
+            	}
+            }
+            
+        } else {
+            Map fieldValues = new HashMap();
```
cah292: Another place where generics should be used if possible.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+        fieldValues.put(propertyName, propertyValue);
+        
+        return getBoService().findByPrimaryKey(clazz, fieldValues);
+
+    }
+    
+	protected boolean checkOpenEncumbrances() {
+		boolean success = true;
+		for (AccountGlobalDetail detail : newAccountGlobal.getAccountGlobalDetails()) {
+			success &= checkOpenEncumbrances(detail);
```
cah292: Based on the "checkOpenEncumbrances" method below, this could be another short-circuiting candidate due to the error message contents not being derived from the object field values.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+        boolean result = true;
+
+        // Certain C&G fields are required if the Account belongs to the CG Fund Group
+        if (ObjectUtils.isNotNull(newAccount.getSubFundGroup())) {
+        	if (getSubFundGroupService().isForContractsAndGrants(newAccount.getSubFundGroup())) {
+        		result &= checkEmptyBOField(KFSPropertyConstants.ACCT_INDIRECT_COST_RCVY_TYPE_CD, newAccount.getAcctIndirectCostRcvyTypeCd(), replaceTokens(KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_ICR_TYPE_CODE_CANNOT_BE_EMPTY));
+	    		 result &= checkEmptyBOField(KFSPropertyConstants.FINANCIAL_ICR_SERIES_IDENTIFIER, newAccount.getFinancialIcrSeriesIdentifier(), replaceTokens(KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_ICR_SERIES_IDENTIFIER_CANNOT_BE_EMPTY));
+	
+	             // Validation for financialIcrSeriesIdentifier
+	             if (checkEmptyBOField(KFSPropertyConstants.FINANCIAL_ICR_SERIES_IDENTIFIER, newAccount.getFinancialIcrSeriesIdentifier(), replaceTokens(KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_ICR_SERIES_IDENTIFIER_CANNOT_BE_EMPTY))) {
```
cah292: The "checkEmptyBOField" check for financialIcrSeriesIdentifier is being duplicated here. I would recommend refactoring things to avoid the duplicate call, in case it ends up adding the same error message twice.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+         								break;
+         							}
+         						}
+         					}
+         					if(!foundInactivated){
+         						activeICRList.add(icrAccount);
+         					}
+         				}
+         				success &= !activeICRList.isEmpty();
+         	            if (!success){
```
cah292: A couple of these validation methods have the "if (!success)" blocks in an odd horizontal position. Please update the horizontal positioning of these blocks accordingly for readability.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+                    }
+            	}
+
+        	if(!success){
+        		if(ObjectUtils.isNotNull(newAccount.getAccountGlobalDetails()) && !newAccount.getAccountGlobalDetails().isEmpty()){
+        			for(AccountGlobalDetail accountGlobalDetail : newAccount.getAccountGlobalDetails()){
+        				accountGlobalDetail.refreshReferenceObject(KFSPropertyConstants.ACCOUNT);
+        				success &= !accountGlobalDetail.getAccount().getActiveIndirectCostRecoveryAccounts().isEmpty();
+        				if(success){
+        					continue;
```
cah292: This "continue" statement should be removed since it's at the end of the loop anyway.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+          			for(AccountGlobalDetail accountGlobalDetail : newAccount.getAccountGlobalDetails()){
+          				accountGlobalDetail.refreshReferenceObject(KFSPropertyConstants.ACCOUNT);
+          				boolean empty = accountGlobalDetail.getAccount().getActiveIndirectCostRecoveryAccounts().isEmpty();
+          				if(!empty){
+          					 putFieldError(KFSPropertyConstants.ACCOUNT_CHANGE_DETAILS, CUKFSKeyConstants.ERROR_DOCUMENT_ACCT_GLB_MAINT_ACCOUNT_CG_FIELDS_FILLED_FOR_NON_CG_ACCOUNT, new String[] { args, accountGlobalDetail.getAccountNumber() });
+          				   
+          				}
+          				success &= accountGlobalDetail.getAccount().getActiveIndirectCostRecoveryAccounts().isEmpty();
+          				if(success){
+          					continue;
```
cah292: Another "continue" statement that should be removed due to it being at the end of the loop.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+        	}
+        }
+        
+        if(!expectFilled){
+        	 success = newAccount.getActiveIndirectCostRecoveryAccounts().isEmpty();
+        	 
+        	 if (!success){
+                 putFieldError(KFSPropertyConstants.INDIRECT_COST_RECOVERY_ACCOUNTS, errorMessage, args);
+             }
+        	  if(success){
```
cah292: Since this immediately follows an "if (!success)" check, this could be changed to an "else" block.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+		boolean success = true;
+		if (expectFilled) {
+			success = !newAccount.getActiveIndirectCostRecoveryAccounts().isEmpty();
+			if (success) {
+				for (IndirectCostRecoveryAccountChange account : newAccount.getActiveIndirectCostRecoveryAccounts()) {
+					success &= StringUtils.isNotBlank(account.getIndirectCostRecoveryAccountNumber()) && StringUtils.isNotBlank(account.getIndirectCostRecoveryFinCoaCode());
+				}
+			}
+
+			if (!success) {
```
cah292: Another area where the "if" check could be changed to an "else".

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+				acctGlobalDetail.refreshReferenceObject(KFSPropertyConstants.ACCOUNT);
+
+				success &= !acctGlobalDetail.getAccount().getActiveIndirectCostRecoveryAccounts().isEmpty();
+			}
```
cah292: This block is confusing; "success" must be false to enter into it, but it will remain false after the boolean check and the only side effect is that the account object has been refreshed. Is this intentional, or is it meant to potentially succeed if the boolean check passes?

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+		}
+
+		if (!expectFilled) {
```
cah292: Unless "expectFilled" is allowed to change, I would suggest turning this into an "else" block that executed if the "if (expectFilled)" check above does not pass.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+     *
+     * @param errorConstant
+     * @return error string that has had tokens "{0}" and "{1}" replaced
+     */
+    protected String replaceTokens(String errorConstant) {
+        String cngLabel = getSubFundGroupService().getContractsAndGrantsDenotingAttributeLabel();
+        String cngValue = getSubFundGroupService().getContractsAndGrantsDenotingValueForMessage();
+        String result = getConfigService().getPropertyValueAsString(errorConstant);
+        result = StringUtils.replace(result, "{0}", cngLabel);
+        result = StringUtils.replace(result, "{1}", cngValue);
```
cah292: It would be cleaner to use "MessageFormat.format()" here instead of direct substitution.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+
+        if (ObjectUtils.isNull(newAccount.getContractControlFinCoaCode())) {
+            return result;
+        }
+        if (ObjectUtils.isNull(newAccount.getContractControlAccountNumber())) {
+            return result;
+        }
+        
+        //if no account global details exist then don't validate
+        if(ObjectUtils.isNull(newAccount.getAccountGlobalDetails()) || (ObjectUtils.isNotNull(newAccount.getAccountGlobalDetails()) && newAccount.getAccountGlobalDetails().size() == 0)){
```
cah292: This logic is effectively "x || (!x && y)"; I would recommend rewriting as "x || y" instead.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+        	return true;
+        }
+        	
+        if(ObjectUtils.isNotNull(newAccount.getAccountGlobalDetails()) && newAccount.getAccountGlobalDetails().size() == 1){
```
cah292: As per the previous "if" block, newAccount.getAccountGlobalDetails() should be non-null and non-empty by this point, so you only need the size check here.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+                    if (StringUtils.isBlank(buildingCode)) {
+                        putFieldError(errorPath, KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_CAMS_SUBFUNDGROUP_WITH_MISSING_BUILDING_CD, subFundGroupCode);
+                        success &= false;
+                    }
+
+                }
+                else {
+
+                    // if sub_fund_grp_cd is NOT 'PFCMR', campus code should NOT be entered
+                    if (!StringUtils.isBlank(campusCode)) {
```
cah292: This should be written as "StringUtils.isNotBlank()" for clarity.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+                        putFieldError(KFSPropertyConstants.ACCOUNT_DESCRIPTION + "." + KFSPropertyConstants.CAMPUS_CODE, KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_NONCAMS_SUBFUNDGROUP_WITH_CAMPUS_CD_FOR_BLDG, subFundGroupCode);
+                        success &= false;
+                    }
+
+                    // if sub_fund_grp_cd is NOT 'PFCMR' then bldg_cd should NOT be entered
+                    if (!StringUtils.isBlank(buildingCode)) {
```
cah292: This should be written as "StringUtils.isNotBlank()" for clarity.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+    
+    /**
+     * This method checks to see if any SubFund Group rules were violated Specifically: if SubFundGroup is empty or not "PFCMR" we
+     * cannot have a campus code or building code if SubFundGroup is "PFCMR" then campus code and building code "must" be entered
+     * and be valid codes
+     *
+     * @param maintenanceDocument
+     * @return false on rules violation
+     */
+    protected boolean checkSubFundGroup( AccountGlobalDetail detail) {
```
cah292: Lots of "!StringUtils.isBlank()" checks are in this method and should ideally be replaced with "StringUtils.isNotBlank()"; only a few are noted below.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +962,1065 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+     * This method checks to see if the contracts and grants fields are filled in or not
+     *
+     * @param account
+     * @param propertyName - property to attach error to
+     * @return false if the contracts and grants fields are blank
+     */
+    protected boolean checkCGFieldNotFilledIn(CuAccountGlobal account, String propertyName) {
+        boolean success = true;
+        Object value = ObjectUtils.getPropertyValue(account, propertyName);
+        if ((value instanceof String && !StringUtils.isBlank(value.toString())) || (value != null)) {
```
cah292: Another place where "StringUtils.isNotBlank()" should be used instead.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AccountGlobal.xml
```diff
@@ -10,14 +10,238 @@
...
+	<bean id="AccountGlobal-accountTypeCode-parentBean" abstract="true"
+		parent="Account-accountTypeCode">
+	</bean>
+	
+	<bean id="AccountGlobal-appropriationAccountNumber" parent="AccountGlobal-appropriationAccountNumber-parentBean" />
+	<bean id="AccountGlobal-appropriationAccountNumber-parentBean" abstract="true" parent="AppropriationAccount-appropriationAccountNumber">
+		<property name="required" value="false"/>
+	</bean>
+		
+	<bean id="AccountGlobal-accountsFringesBnftIndicator" parent="AccountGlobal-accountsFringesBnftIndicator-parentBean" />
```
cah292: As per my earlier comment, I would suggest updating these boolean fields (except possibly the "remove..." fields) to use the same control override as "AccountGlobal-pendingAcctSufficientFundsIndicator"

src/test/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccountsTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import edu.cornell.kfs.coa.fixture.AccountGlobalDetailFixture;
+import edu.cornell.kfs.coa.fixture.AccountGlobalFixture;
+import edu.cornell.kfs.coa.fixture.IndirectCostRecoveryAccountFixture;
+import edu.cornell.kfs.coa.fixture.SubAccountGlobalDetailFixture;
+import edu.cornell.kfs.coa.fixture.SubAccountGlobalFixture;
+import junit.framework.TestCase;
+
+
+public class GlobalObjectWithIndirectCostRecoveryAccountsTest extends TestCase {
```
cah292: It may be helpful to add some messages to the assertion checks to clarify what the problem is if an assertion fails.

src/test/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRuleTest.java
```diff
@@ -0,0 +1,103 @@
...
+	
+	public void testAccountGlobalCheckICRAccuntTotalDistributionOnDetailWillBe100PercentAfterUpdate_Pass() {
+		AccountGlobal accountGlobal = (AccountGlobal)AccountGlobalFixture.ACCT_GLOBAL_3333333_100_ACTIVE_1111111_2222222_98_2_INACTIVE.getAccountGlobal();
+		AccountGlobalDetail accountGlobalDetail = AccountGlobalDetailFixture.ACCOUNT_GLOBAL_DETAIL_1111111_2222222_98_2.getAccountGlobalDetail();
+		boolean result = globalIndirectCostRecoveryAccountsRule.checkICRAccuntTotalDistributionOnDetailWillBe100PercentAfterUpdate(
+				accountGlobal.getIndirectCostRecoveryAccounts(),
+				accountGlobalDetail.getAccount().getIndirectCostRecoveryAccounts(), 
+				accountGlobalDetail, accountGlobal);
+		LOG.info("Check Account Global updated ICR accounts distribution is 100%: " + result);
+		assertTrue("Updated Account Global ICR account distribution is 100%", result);
```
cah292: I would recommend updating the assertion messages in this class to replace "is" with "should have been" or equivalent, to avoid confusion if a failure occurs. Also, I don't think you need the LOG.info() calls since the assertions should already validate and clarify the values.

cah292:
Performed initial review and added comments accordingly. Will await feedback before examining further.

db483:
Thank you very much for reviewing this Chad. I am looking into your recommendations and will update the code.

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
db483: This code is in fact the same code you have previously reviewed but refactored into a common class that both AccountGlobal and SubAccountGlobal use.
As I have mentioned to you yesterday on Hipchat, my understanding during our Clean Code meeting was that 'while' is preferred to iteration with 'break'.  The condition for the while loop however I believe can be improved. I have extracted it into a function to improve clarity. The function name should be more descriptive.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
db483: Good point, I will also update the function name from totalDistributionOnUpdatedActiveICRsIs100Percent to isTotalDistributionOnUpdatedActiveICRs100Percent. Thank you!

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
db483: As before I am extracting the condition for the while loop into a function to improve clarity.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: Thank you for pointing this out. I will add a new error message that it includes the account that has open encumbrances. This way the user will know which accounts cannot be closed from the ones selected on the doc and can make a decision on how to proceed based on that.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: success might be true as it checks for active icr accounts on the account in the account global detail (on the account being editted). Initially it checked if any icr accounts were entered on the Account global doc.

cah292: My point is that in the block where it checks for ICR accounts on the global detail's account, it only gets to that code if "success" is already false, so the check basically does nothing because it won't change "success" back to true. Should the condition for that block be "if (success)" instead of "if (!success)"?

db483: good point, the code needs to be fixed and refactored to be more clear. If !=success check is fine but the &= inside does not do much good.

cah292: In that case, it would be good to clarify how the various checks in the "if (expectFilled)" block affect the outcome. Is your intention to have the final result be true if the global account has no active ICR accounts but the global detail's account does? If so, then should the final result always be false if the global account has active ICR accounts but at least one does not have the required chart and account information filled in (as suggested by the "for" loop)?

Perhaps an "if/else" setup could be used, which checks whether the global account has any active ICR accounts, and then performs the appropriate "for" loop check or global detail check accordingly.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: Turns out this method is not used in the code so I removed the code. It was replaced by checkICRCollectionDoesNotExistForNonCGSubFund and checkICRCollectionExistsForCGSubFund

bhutchinson:
retest this please

db483:
Hi Chad, I added the changes based on your recommendations and this is ready for review.

db483:
Hi Bryan, Chad will be out this week and I would like to reassign this to you for review. Thank you very much!

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
```diff
@@ -0,0 +1,40 @@
+package edu.cornell.kfs.coa.businessobject;
+
+import java.util.List;
+import java.util.Map;
+
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
+import org.kuali.rice.krad.bo.GlobalBusinessObjectDetailBase;
+
+public interface GlobalObjectWithIndirectCostRecoveryAccounts {
```
bhutchinson: Related to a question Chad had previously asked about this (when it was an abstract class), would it make sense for this to extend GlobalBusinessObject? SubAccountGlobal would only need to implement one interface if that was done.

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
```diff
@@ -0,0 +1,40 @@
...
+
+	public List<IndirectCostRecoveryAccountChange> getIndirectCostRecoveryAccounts();
+
+	public void setIndirectCostRecoveryAccounts(
+			List<IndirectCostRecoveryAccountChange> indirectCostRecoveryAccounts);
+
+	public List<IndirectCostRecoveryAccountChange> getActiveIndirectCostRecoveryAccounts();
+
+	public abstract List<? extends GlobalBusinessObjectDetailBase> getGlobalObjectDetails();
```
bhutchinson: Not a big deal, but since this is an interface, the public and abstract modifiers are redundant and can be removed.

src/main/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+				int positionForMatchWithSameActiveIndicator = -1;
+				int currentPosition = 0;
+				int maxPosition = icrAccounts.size();
+
+				while (noMatchFoundOrSameActiveIndicatorMatchFoundAndStillHaveICRAccountToCheck(currentPosition, maxPosition, foundMatch, positionForMatchWithSameActiveIndicator)) {
+					if (!alreadyUpdatedIndexes.containsKey(currentPosition)) {
+						IndirectCostRecoveryAccount existingICR = icrAccounts.get(currentPosition);
+						// check if we have a match on chart, account and percentage
+						if (newICR.matchesICRAccount(existingICR)) {
+							// set this to true if we have found a match
```
bhutchinson: I'm not sure these comments, this one especially, are adding much value, and I would be inclined to remove them.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +963,959 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+     * This method checks to see if the contracts and grants fields are filled in or not
+     *
+     * @param account
+     * @param propertyName - property to attach error to
+     * @return false if the contracts and grants fields are blank
+     */
+    protected boolean checkCGFieldNotFilledIn(CuAccountGlobal account, Account accountGlobalDetail, String propertyName) {
+        boolean success = true;
+        Object value = ObjectUtils.getPropertyValue(accountGlobalDetail, propertyName);
+        if ((value instanceof String && !StringUtils.isBlank(value.toString())) || (value != null)) {
```
bhutchinson: Another place where !StringUtils.isBlank could be replaced with StringUtils.isNotBlank.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -222,44 +281,47 @@ protected boolean checkGeneralRules(MaintenanceDocument maintenanceDocument) {
 
         if (!StringUtils.isBlank(newAccountGlobal.getAccountFiscalOfficerSystemIdentifier()) && (ObjectUtils.isNull(fiscalOfficer) || StringUtils.isEmpty(fiscalOfficer.getPrincipalId()) || !getDocumentHelperService().getDocumentAuthorizer(maintenanceDocument).isAuthorized(maintenanceDocument, KFSConstants.PermissionNames.SERVE_AS_FISCAL_OFFICER.namespace, KFSConstants.PermissionNames.SERVE_AS_FISCAL_OFFICER.name, fiscalOfficer.getPrincipalId()))) {
```
bhutchinson: Since you did replace the constants here, you could also replace the usages of !StringUtils.isBlank with StringUtils.isNotBlank.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +963,959 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
 
+    protected boolean checkSubFundProgram(MaintenanceDocument document) {
+        boolean success = true;
+
+        String subFundGroupCode = newAccountGlobal.getSubFundGroupCode();
+        String subFundProg = newAccountGlobal.getProgramCode();
+        BusinessObjectService bos = SpringContext.getBean(BusinessObjectService.class);
+
+        if (StringUtils.isNotBlank(subFundProg)) {
+            Map<String, String> fieldValues = new HashMap<String, String>();
```
bhutchinson: Not a big deal, but technically, the types are not needed on the right side of the assignment; could be:
Map<String, String> fieldValues = new HashMap<>();

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
db483: Good point. I've lost this one when refactoring from abstract class to interface, will add the change. Thank you.

src/main/java/edu/cornell/kfs/coa/businessobject/GlobalObjectWithIndirectCostRecoveryAccounts.java
db483: Thank you, I will clean this up.

src/main/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java
db483: I will clean this up, thank you.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: Will replace it.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: Will replace this one too.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +963,959 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+
+		return success;
+	}
+    
+    protected boolean checkICRCollectionDoesNotExistForNonCGSubFund(CuAccountGlobal newAccount, String subFundGroupCode){
+    	boolean success = true;
+    	success = newAccount.getActiveIndirectCostRecoveryAccounts().isEmpty();
+
+    	// if any active ICR accts entered then put error message; no need to check acct details
+    	if (!success){
```
bhutchinson: I believe this could be an else block after the if(success) check below.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -863,5 +963,959 @@ private boolean isIncomeStreamAccountRequired(String fundGroupCode, String subFu
...
+    	
+    	return success;	
+    }
+    
+    protected boolean checkICRCollectionDoesNotExistForNonCGSubFund(CuAccountGlobal newAccount, AccountGlobalDetail accountGlobalDetail, String subFundGroupCode){
+    	boolean success = true;
+    	success = newAccount.getActiveIndirectCostRecoveryAccounts().isEmpty();
+
+    	// if any active ICR accts entered then put error message; no need to check acct details
+    	if (!success){
```
bhutchinson: I believe this could be an else block after the if(success) check below.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: This is an interesting one, after changing it I get this warning:
'Map is a raw type. References to generic type Map<K,V> should be parameterized'

db483: Changed it to:
Map<String, String> fieldValues = new HashMap<>();

I think this is what you meant?

bhutchinson:
Daniela, I went through and it looks like the issues Chad raised have been addressed for the most part. I did have a couple questions related to his comments that I raised and I added a couple comments on things I noticed. Thanks.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
bhutchinson: Sorry, I meant this:
Map<String, String> fieldValues = new HashMap<>();

You need the types on the left side of the assignment, but not the right.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
db483: Thank you for pointing this out. I will refactor this code because it is not very clear and I think the current implementation for matching new with existing ICR accounts is incorrect as it does not take into account the percentage. I will update this and retest this functionality.

db483:
Hi Bryan, I have updated the PR and it is now ready for your review. Thank you.

bhutchinson:
Ok, I'll review it again, hopefully this afternoon. Thanks.

bhutchinson:
retest this please

bhutchinson:
Code changes look good; still waiting on unit tests - latest UTEST refresh was bad, will try again tomorrow after the next refresh.

bhutchinson:
retest this please

bhutchinson:
There were still four unit test failures, but these are not related to the changes in this PR so I'm going to merge it.

PR #126, opened 2016-06-08
KFSPTS-5301: Make CuYearEndSalaryExpenseTransferDocument and YearEndB…

nkimble:
…enefitsTransferDocument error processing consistent with each other. Consolidate duplicate year end salary expense transfer spring definition files into a single definition.


nkimble:
This code change is similar to PR#105.  The difference is that the user desired the removal of the yes/no prompt page for the YEST so that document presents the expired account error messages directly on the YEST edoc like the YEBT.

jhulslander:
The XML seems good, and particular reason naming this CuYearEndSalaryExpenseTransferDocument instead of just YearEndSalaryExpenseTransferDocument

nkimble:
That naming was already in place when the corresponding class file of the same name was created as part of a CU mod to deal with labor ledger fringe rates not being calculated correctly.  I adjusted what was present to get the functionality the users desired on the YEST document and incorporated the newer routing change into that pre-existing file.  I agree that this area should be refactored to remove the "Cu' but my thoughts are that it should be done for that entire mod and not for one file piecemeal. And, I am thinking that type of effort is outside the scope of this JIRA.

src/main/resources/edu/cornell/kfs/module/ld/document/datadictionary/CuYearEndSalaryExpenseTransferDocument.xml
```diff
@@ -42,4 +57,14 @@
     </property>
   </bean>
 
+  <bean id="YearEndSalaryExpenseTransferDocument-workflowAttributes" parent="YearEndSalaryExpenseTransferDocument-workflowAttributes-parentBean">
```
jhulslander: For consistency, lets rename this to CuYearEndSalaryExpenseTransferDocument and we will need to add it workflow attributes in CuYearEndSalaryExpenseTransferDocument.  Double check we don't need to worry about XML ingestion, I don't think we do.

nkimble: Making changes as requested.  Like you, I believe routing is valid as it currently stands as it is associated with parent document LaborDistributionYearEndTransactionalDocument.

jhulslander:
Sweet, after the tests complete, I will merge

PR #127, opened 2016-06-08
KFSPTS-5501: Added a Rice 2.5.x fix for immediately saving BO note adds/deletes, and tweaked it to only apply to Account maintenance docs.

cah292:
This is an updated version of the code from PR #121, which will now do note-service-only calls if the Account is in a state to accept BO notes, and will only do the forcible doc-saving actions for new/copied Accounts that have not been finalized. This does mean that much of the new code does not affect note deletion on pre-existing Accounts; however, the deletion operation is still calling out to the note service, and the deletion still appeared to work fine in my local testing.


bhutchinson:
Changes look good, I'll merge once tests pass.

PR #135, opened 2016-06-22
KFSPTS-5732 Add changes so that foreign vendors with payment reason i…

db483:
…n new sys param PAYMENT_REASONS_THAT_DO_NOT_REQUIRE_TAX_REVIEW_FOR_FOREIGN_VENDOR don't route for Tax Review and Income Class on Nonresident Alien Tax tab is set to N

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
+++ b/src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
@@ -60,6 +60,7 @@ import edu.cornell.kfs.fp.businessobject.CuDisbursementVoucherPayeeDetailExtensi
 import edu.cornell.kfs.fp.businessobject.DisbursementVoucherWireTransferExtendedAttribute;
 import edu.cornell.kfs.fp.document.interfaces.CULegacyTravelIntegrationInterface;
 import edu.cornell.kfs.fp.document.service.CULegacyTravelService;
+import edu.cornell.kfs.fp.document.service.CuDisbursementVoucherTaxService;
 import edu.cornell.kfs.fp.document.service.impl.CULegacyTravelServiceImpl;
 import edu.cornell.kfs.fp.service.CUPaymentMethodGeneralLedgerPendingEntryService;
 import edu.cornell.kfs.sys.CUKFSKeyConstants;
@@ -922,6 +923,56 @@ public class CuDisbursementVoucherDocument extends DisbursementVoucherDocument i
         }
 
         return false;
+    }
+    
+    @Override
+    protected boolean isTaxReviewRequired() {
+        if (isPayeePurchaseOrderVendorHasWithholding()) {
+            return true;
+        }
+
+        String payeeTypeCode = this.getDvPayeeDetail().getDisbursementVoucherPayeeTypeCode();
+        if (payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.EMPLOYEE)) {
+            return false;
+        } else if (payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.VENDOR)) {
+            if(getVendorService().isVendorInstitutionEmployee(this.getDvPayeeDetail().getDisbVchrVendorHeaderIdNumberAsInteger())){
+                return true;
+            }
+        }
+        
+        String paymentReasonCode = this.getDvPayeeDetail().getDisbVchrPaymentReasonCode();
+        Integer vendorHeaderId = getDvPayeeDetail().getDisbVchrVendorHeaderIdNumberAsInteger();
+        if (getCuDisbursementVoucherTaxService().isForeignVendorAndTaxReviewRequired(payeeTypeCode, paymentReasonCode, vendorHeaderId)) {
+        	return true;
+        }
+
+        String taxControlCode = this.getDisbVchrPayeeTaxControlCode();
+        if (StringUtils.equals(taxControlCode, DisbursementVoucherDocument.TAX_CONTROL_BACKUP_HOLDING) || StringUtils.equals(taxControlCode,DisbursementVoucherDocument.TAX_CONTROL_HOLD_PAYMENTS)) {
+            return true;
+        }
+
+       
+        if (this.getDvPymentReasonService().isDecedentCompensationPaymentReason(paymentReasonCode)) {
+            return true;
+        }
+
+        if (this.getDvPymentReasonService().isMovingPaymentReason(paymentReasonCode) && taxedCampusForMovingReimbursements()) {
+            return true;
+        }
+
+        if (getParameterEvaluatorService().getParameterEvaluator(this.getClass(), DisbursementVoucherDocument.PAYMENT_REASONS_REQUIRING_TAX_REVIEW_PARAMETER_NAME, paymentReasonCode).evaluationSucceeds()) {
+            return true;
+        }
+
+        return false;
+    }
+    
+    protected CuDisbursementVoucherTaxService getCuDisbursementVoucherTaxService() {
+        return SpringContext.getBean(CuDisbursementVoucherTaxService.class);
+    }
+    
+    protected ParameterEvaluatorService getParameterEvaluatorService(){
+    	return SpringContext.getBean(ParameterEvaluatorService.class);
     }
 
 
--- a/src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
+++ b/src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
@@ -1,0 +1,11 @@
+package edu.cornell.kfs.fp.document.service;
+
+public interface CuDisbursementVoucherTaxService {
+	
+	boolean isForeignVendorAndTaxReviewRequired(String payeeTypeCode,
+			String paymentReasonCode, Integer vendorHeaderId);
+	
+	boolean isForeignVendorAndTaxReviewNotRequired(String payeeTypeCode,
+			String paymentReasonCode, Integer vendorHeaderId);
+
+}
--- a/src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
@@ -3,16 +3,22 @@ package edu.cornell.kfs.fp.document.service.impl;
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.fp.document.DisbursementVoucherDocument;
 import org.kuali.kfs.fp.document.service.impl.DisbursementVoucherTaxServiceImpl;
-import org.kuali.kfs.fp.document.validation.impl.DisbursementVoucherNonResidentAlienInformationValidation;
 import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.KFSKeyConstants;
+import org.kuali.kfs.vnd.document.service.VendorService;
+import org.kuali.rice.core.api.parameter.ParameterEvaluatorService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 import org.kuali.rice.krad.util.GlobalVariables;
 import org.kuali.rice.krad.util.MessageMap;
 
+import edu.cornell.kfs.fp.document.CuDisbursementVoucherConstants;
+import edu.cornell.kfs.fp.document.service.CuDisbursementVoucherTaxService;
 import edu.cornell.kfs.fp.document.validation.impl.CuDisbursementVoucherNonResidentAlienInformationValidation;
 
-public class CuDisbursementVoucherTaxServiceImpl extends DisbursementVoucherTaxServiceImpl{
+public class CuDisbursementVoucherTaxServiceImpl extends DisbursementVoucherTaxServiceImpl implements CuDisbursementVoucherTaxService{
+	
+	protected VendorService vendorService;
+	protected ParameterEvaluatorService parameterEvaluatorService;
 
     public void processNonResidentAlienTax(DisbursementVoucherDocument document) {
         if (validateNRATaxInformation(document)) {
@@ -80,4 +86,43 @@ public class CuDisbursementVoucherTaxServiceImpl extends DisbursementVoucherTaxS
         }
         return true;
     }
+    
+    @Override
+	public boolean isForeignVendorAndTaxReviewRequired(String payeeTypeCode,
+			String paymentReasonCode, Integer vendorHeaderId) {
+
+		return isForeignVendor(payeeTypeCode, vendorHeaderId) && !paymentReasonDoesNotRequireTaxReviewForForeignVendor(paymentReasonCode);
+	}
+	
+    @Override
+	public boolean isForeignVendorAndTaxReviewNotRequired(String payeeTypeCode,
+			String paymentReasonCode, Integer vendorHeaderId) {
+
+		return isForeignVendor(payeeTypeCode, vendorHeaderId) && paymentReasonDoesNotRequireTaxReviewForForeignVendor(paymentReasonCode);
+	}
+	
+	protected boolean isForeignVendor(String payeeTypeCode, Integer vendorHeaderId){
+		return payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.VENDOR) && vendorService.isVendorForeign(vendorHeaderId);
+	}
+	
+	protected boolean paymentReasonDoesNotRequireTaxReviewForForeignVendor(String paymentReasonCode){
+		return parameterEvaluatorService.getParameterEvaluator(DisbursementVoucherDocument.class, CuDisbursementVoucherConstants.PAYMENT_REASONS_THAT_DO_NOT_REQUIRE_TAX_REVIEW_FOR_FOREIGN_VENDOR, paymentReasonCode).evaluationSucceeds();
+	}
+
+	public VendorService getVendorService() {
+		return vendorService;
+	}
+
+	public void setVendorService(VendorService vendorService) {
+		this.vendorService = vendorService;
+	}
+
+	public ParameterEvaluatorService getParameterEvaluatorService() {
+		return parameterEvaluatorService;
+	}
+
+	public void setParameterEvaluatorService(
+			ParameterEvaluatorService parameterEvaluatorService) {
+		this.parameterEvaluatorService = parameterEvaluatorService;
+	}
 }
--- a/src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
+++ b/src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
@@ -23,6 +23,8 @@ import java.util.List;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.WordUtils;
+import org.kuali.kfs.fp.businessobject.DisbursementVoucherPayeeDetail;
+import org.kuali.kfs.fp.document.DisbursementVoucherConstants;
 import org.kuali.kfs.fp.document.DisbursementVoucherDocument;
 import org.kuali.kfs.fp.document.validation.impl.DisbursementVoucherDocumentPreRules;
 import org.kuali.kfs.pdp.PdpParameterConstants;
@@ -33,13 +35,13 @@ import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.kfs.sys.service.impl.KfsParameterConstants;
 import org.kuali.rice.core.api.config.property.ConfigurationService;
 import org.kuali.rice.core.api.util.type.KualiInteger;
+import org.kuali.rice.coreservice.framework.parameter.ParameterService;
+import org.kuali.rice.krad.document.Document;
 
 import edu.cornell.kfs.fp.CuFPConstants;
 import edu.cornell.kfs.fp.document.CuDisbursementVoucherConstants;
+import edu.cornell.kfs.fp.document.service.CuDisbursementVoucherTaxService;
 import edu.cornell.kfs.sys.CUKFSKeyConstants;
-
-import org.kuali.rice.coreservice.framework.parameter.ParameterService;
-import org.kuali.rice.krad.document.Document;
 
 /**
  * Checks warnings and prompt conditions for dv document.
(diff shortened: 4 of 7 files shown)


src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
```diff
@@ -0,0 +1,11 @@
+package edu.cornell.kfs.fp.document.service;
+
+public interface CuDisbursementVoucherTaxService {
+	
+	public boolean isForeignVendorWithTaxReviewRequired(String payeeTypeCode,
+			String paymentReasonCode, Integer vendorHeaderId);
+	
+	public boolean isForeignVendoAndNOTaxReviewRequired(String payeeTypeCode,
```
cah292: Vendor is misspelled in method name; please correct that. Also, it might be clearer to rename the methods to something like "isForeignVendorAndTaxReviewIsRequired" for the first and "isForeignVendorAndTaxReviewIsNotRequired" for the second.

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
```diff
@@ -80,4 +86,48 @@ protected boolean validateNRATaxInformation(DisbursementVoucherDocument document
         }
         return true;
     }
+    
+	public boolean isForeignVendorWithTaxReviewRequired(String payeeTypeCode,
```
cah292: I would recommend adding the "@Override" annotation to any overridden or implemented methods, to help minimize errors when future code changes are made.

src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
db483: thank you, I will add the changes

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
db483: Will do

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
```diff
@@ -80,4 +86,48 @@ protected boolean validateNRATaxInformation(DisbursementVoucherDocument document
...
+
+		return isForeignVendor(payeeTypeCode, vendorHeaderId) && paymentReasonDoesNotRequireTaxReviewForForeignVendor(paymentReasonCode);
+	}
+	
+	protected boolean isForeignVendor(String payeeTypeCode, Integer vendorHeaderId){
+		if (payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.VENDOR) && vendorService.isVendorForeign(vendorHeaderId)){ 
+			return true;
+		}
+		
+		return false;
```
cah292: In this method and the one below it, I would suggest returning the result of the condition so that you don't need the extra "return true/false" statements.

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
```diff
@@ -80,4 +86,48 @@ protected boolean validateNRATaxInformation(DisbursementVoucherDocument document
...
+	}
+	
+	protected boolean paymentReasonDoesNotRequireTaxReviewForForeignVendor(String paymentReasonCode){
+		if (parameterEvaluatorService.getParameterEvaluator(DisbursementVoucherDocument.class, CuDisbursementVoucherConstants.PAYMENT_REASONS_THAT_DO_NOT_REQUIRE_TAX_REVIEW_FOR_FOREIGN_VENDOR, paymentReasonCode).evaluationSucceeds()) {
```
cah292: Parameter evaluators are capable of treating a parameter's values as a whitelist or blacklist depending on the parameter's Allow/Deny flag. There are other places in KFS that rely on this flag for treating the values as a blacklist. Your code should work as-is, but if you want to make things consistent with similar setup in KFS, you could change this around so that evaluationSucceeds() means tax review IS required, and then set the parameter's flag to Deny so that its elements will be treated as a blacklist. (This may also require a parameter name change for consistency.) Let me know if this makes sense in the event that you do intend to go forward with the change.

db483: This is a good point and I have considered this option. However I find it that this more clearly implements the requirements and would like to keep it as is unless there is a major concern about it.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -0,0 +1,291 @@
...
+	
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		disbursementVoucherTaxService = new CuDisbursementVoucherTaxServiceImpl();
+		disbursementVoucherTaxService.setVendorService(new MockVendorService());
+		disbursementVoucherTaxService.setParameterEvaluatorService(new MockParameterEvaluatorService());	
+	}
+	
+	public void testIsForeignVendorWithTaxReviewRequired_true()  {
```
cah292: You may need to rename these various test methods if you end up changing the CuDisbursementVoucherTaxService method names.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -0,0 +1,291 @@
...
+import org.kuali.kfs.vnd.businessobject.VendorHeader;
+import org.kuali.kfs.vnd.businessobject.VendorRoutingComparable;
+import org.kuali.kfs.vnd.document.service.VendorService;
+import org.kuali.rice.core.api.parameter.ParameterEvaluator;
+import org.kuali.rice.core.api.parameter.ParameterEvaluatorService;
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+import org.kuali.rice.krad.bo.Note;
+import org.kuali.rice.krad.document.Document;
+
+public class CuDisbursementVoucherTaxServiceImplTest extends TestCase {
```
cah292: I would suggest adding a few test cases where the DV payee is not a vendor (such as an employee payee).

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -0,0 +1,291 @@
...
+			return false;
+		}
+
+		@Override
+		public boolean isVendorForeign(Integer vendorHeaderGeneratedIdentifier) {
+			if(FOREIGN_VENDOR_HEADER_ID.equals(vendorHeaderGeneratedIdentifier)){
+				return true;
+			}
+			else
+				return false;
```
cah292: Another place where the return statement can be consolidated.

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImpl.java
cah292: Ok, that makes sense. I would recommend tweaking this method, though, so that it returns the result of the condition. That way, you don't need the extra "return true/false" statements.

cah292:
Changes look good in general, but I did have some suggestions for improvements. Will wait for feedback before reviewing again.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -923,6 +924,48 @@ protected boolean isCampusReviewRequired() {
 
         return false;
     }
+    
+    @Override
+    protected boolean isTaxReviewRequired() {
```
bhutchinson: Since the bulk of this method is the same as the one it overrides, it would be nice to eliminate the duplication and call super.isTaxReviewRequired(), however that looked like it might be difficult based on my quick review. Given that, I'm not asking for this to be changed, but I wanted to bring it up so others can think along the same lines (avoid duplicating code from overridden methods if possible) and just in case you can see a better way to do that here than I can.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -923,6 +924,48 @@ protected boolean isCampusReviewRequired() {
...
+        if (isPayeePurchaseOrderVendorHasWithholding()) {
+            return true;
+        }
+
+        String payeeTypeCode = this.getDvPayeeDetail().getDisbursementVoucherPayeeTypeCode();
+        if (payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.EMPLOYEE)) {
+            return false;
+        } else if (payeeTypeCode.equals(KFSConstants.PaymentPayeeTypes.VENDOR)) {
+            if(vendorService.isVendorInstitutionEmployee(this.getDvPayeeDetail().getDisbVchrVendorHeaderIdNumberAsInteger())){
```
bhutchinson: I know this was basically copied from the overridden method, but it makes me nervous that it's referencing the class variable rather than the getter. Since this service can't be injected, we're depending on the getter being called before this or we'll get an NPE. I would feel more comfortable with it calling the getter.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -923,6 +924,48 @@ protected boolean isCampusReviewRequired() {
...
+                return true;
+            }
+        }
+        
+        String paymentReasonCode = this.getDvPayeeDetail().getDisbVchrPaymentReasonCode();
+        Integer vendorHeaderId = getDvPayeeDetail().getDisbVchrVendorHeaderIdNumberAsInteger();
+        if (SpringContext.getBean(CuDisbursementVoucherTaxServiceImpl.class).isForeignVendorWithTaxReviewRequired(payeeTypeCode, paymentReasonCode, vendorHeaderId)) {
```
bhutchinson: Would be nice to extract this to a getter for better testability.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -923,6 +924,48 @@ protected boolean isCampusReviewRequired() {
...
+       
+        if (this.getDvPymentReasonService().isDecedentCompensationPaymentReason(paymentReasonCode)) {
+            return true;
+        }
+
+        if (this.getDvPymentReasonService().isMovingPaymentReason(paymentReasonCode) && taxedCampusForMovingReimbursements()) {
+            return true;
+        }
+
+        if (/*REFACTORME*/SpringContext.getBean(ParameterEvaluatorService.class).getParameterEvaluator(this.getClass(), DisbursementVoucherDocument.PAYMENT_REASONS_REQUIRING_TAX_REVIEW_PARAMETER_NAME, paymentReasonCode).evaluationSucceeds()) {
```
bhutchinson: Same as above, would be nice to extract this to a getter for better testability.

src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
```diff
@@ -0,0 +1,11 @@
+package edu.cornell.kfs.fp.document.service;
+
+public interface CuDisbursementVoucherTaxService {
+	
+	public boolean isForeignVendorWithTaxReviewRequired(String payeeTypeCode,
```
bhutchinson: since this is an interface, the public access modifier is redundant and can be removed

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -67,11 +70,26 @@ public boolean doPrompts(Document document) {
...
-    /**
+	private void setIncomeClassNonReportableForForeignVendorWithNoTaxReviewRequired(Document document) {
+		DisbursementVoucherDocument dvDoc = (DisbursementVoucherDocument) document;
+		DisbursementVoucherPayeeDetail dvPayeeDetail = dvDoc.getDvPayeeDetail();
+		
+		String payeeTypeCode = dvPayeeDetail.getDisbursementVoucherPayeeTypeCode();
+		String paymentReasonCode = dvPayeeDetail.getDisbVchrPaymentReasonCode();
+		Integer vendorHeaderId = dvPayeeDetail.getDisbVchrVendorHeaderIdNumberAsInteger();
+
+		if (SpringContext.getBean(CuDisbursementVoucherTaxServiceImpl.class).isForeignVendoAndNOTaxReviewRequired(payeeTypeCode,paymentReasonCode, vendorHeaderId)) {
```
bhutchinson: Another place where SpringContext.getBean() could be extracted into a getter for better testability.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
bhutchinson: FWIW, using JUnit 4, you don't need to extend TestCase, and you can use annotations - @Before for setup, @After for teardown and @Test for test methods. You can look at edu.cornell.kfs.pdp.batch.service.impl.PayeeACHAccountExtractServiceImplTest for an example.

src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
se12: Does the additional 'Is' in the proposed method name necessary?  It seems grammatically incorrect, does not flow right in my mind.

db483: Sandy, would it make more sense to rename them to: 
"isForeignVendorAndTaxReviewRequired" and "isForeignVendorAndTaxReviewNotRequired"? Thank you.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
db483: Thank you for pointing this out Bryan. Unfortunately I could not find a way to call super and eliminate duplication. I know this will make it difficult when we upgrade. If this method changes in foundation code it can be tricky to determine what was Cornell customization and what not, and the developer doing the upgrade would need to understand what changed in foundation code between different versions and what part was the intended Cornell customization. I thought about adding a comment surrounding the part of the code that was added for this change but comments can also be tricky. If any other developer can suggest a better way to implement this that would be greatly appreciated!

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
db483: Bryan, thank you very much for pointing out an example, I will update this.

src/main/java/edu/cornell/kfs/fp/document/service/CuDisbursementVoucherTaxService.java
se12: Yes, that sounds better, I think makes it more readable (Clean Code!).  Thanks

db483:
I have updated the code with the changes suggested during code review. Please let me know if you find any issues. Thank you!

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -67,11 +69,30 @@ public boolean doPrompts(Document document) {
...
+	private void setIncomeClassNonReportableForForeignVendorWithNoTaxReviewRequired(Document document) {
+		DisbursementVoucherDocument dvDoc = (DisbursementVoucherDocument) document;
+		DisbursementVoucherPayeeDetail dvPayeeDetail = dvDoc.getDvPayeeDetail();
+		
+		String payeeTypeCode = dvPayeeDetail.getDisbursementVoucherPayeeTypeCode();
+		String paymentReasonCode = dvPayeeDetail.getDisbVchrPaymentReasonCode();
+		Integer vendorHeaderId = dvPayeeDetail.getDisbVchrVendorHeaderIdNumberAsInteger();
+
+		if (getCuDisbursementVoucherTaxService().isForeignVendorAndTaxReviewNotRequired(payeeTypeCode,paymentReasonCode, vendorHeaderId)) {
+			dvDoc.getDvNonResidentAlienTax().setIncomeClassCode(DisbursementVoucherConstants.NRA_TAX_INCOME_CLASS_NON_REPORTABLE);
```
cah292: I'm wondering whether the PreRules is the correct place to be setting this value, since there doesn't appear to be any routing stage or user prompt restrictions to block the update. It would be good if someone else could chime in on whether they think such updates should be moved to a different part of the code if the restrictions mentioned earlier do not apply.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -0,0 +1,302 @@
...
+	
+	@Test
+	public void testIsForeignVendorAndTaxReviewRequired_true()  {
+		boolean isForeignVendorAndTaxReviewRequired = disbursementVoucherTaxService.isForeignVendorAndTaxReviewRequired(KFSConstants.PaymentPayeeTypes.VENDOR, PAYMENT_REASON_THAT_REQUIRES_TAX_REVIEW_FOR_FOREIGN_VENDOR, FOREIGN_VENDOR_HEADER_ID);
+        assertTrue("isForeignVendorAndTaxReviewRequired method should have returned true when input is foreign vendor and payment reason that requires tax review", isForeignVendorAndTaxReviewRequired);
+    }
+	
+	@Test
+	public void testIsForeignVendorAndTaxReviewRequired_false_PayeeTypeNotVendor()  {
+		boolean isForeignVendorAndTaxReviewRequired = disbursementVoucherTaxService.isForeignVendorAndTaxReviewRequired(KFSConstants.PaymentPayeeTypes.EMPLOYEE, PAYMENT_REASON_THAT_REQUIRES_TAX_REVIEW_FOR_FOREIGN_VENDOR, FOREIGN_VENDOR_HEADER_ID);
```
cah292: In the actual application, I think this method will get invoked with a null vendor header ID if the payee is not a vendor. Not sure if that needs to be reflected in this test.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -0,0 +1,302 @@
...
+	
+	@Test
+	public void testIsForeignVendorAndTaxReviewNotRequired_true(){
+		boolean isForeignVendorAndTaxReviewNotRequired = disbursementVoucherTaxService.isForeignVendorAndTaxReviewNotRequired(KFSConstants.PaymentPayeeTypes.VENDOR, PAYMENT_REASON_THAT_DOES_NOT_REQUIRE_TAX_REVIEW_FOR_FOREIGN_VENDOR, FOREIGN_VENDOR_HEADER_ID);
+        assertTrue("isForeignVendorAndTaxReviewNotRequired method should have returned true when input is foreign vendor and payment reason that does not require tax review", isForeignVendorAndTaxReviewNotRequired);
+	}
+	
+	@Test
+	public void testIsForeignVendorAndTaxReviewNotRequired_false_PayeeTypeNotVendor(){
+		boolean isForeignVendorAndTaxReviewNotRequired = disbursementVoucherTaxService.isForeignVendorAndTaxReviewNotRequired(KFSConstants.PaymentPayeeTypes.EMPLOYEE, PAYMENT_REASON_THAT_DOES_NOT_REQUIRE_TAX_REVIEW_FOR_FOREIGN_VENDOR, FOREIGN_VENDOR_HEADER_ID);
```
cah292: Another test where the service is being invoked for a non-vendor payee but with a non-null vendor header ID.

cah292:
I did find a couple minor items, but other than that the changes look good. Will wait for feedback before merging.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
db483: That is a good question. The income class field is set by the Tax reviewer when the document routes for tax review. In this case the doc will not route for tax review so the requirement was that a default value of 'N' is set in the income class field. The pre-rules are used for setting default values before business rules are applied so it seemed like the right place to do this. Bryan, do you have any suggestions? Thank you.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
db483: Good point, I will look into this and update it.

db483: Good point, I will update the test

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
bhutchinson: I am ok with this code being in the PreRules. It does seem consistent with what pre-rules do in other cases such as DisbursementVoucherDocumentPreRules.checkSpecialHandlingIndicator().

bhutchinson:
retest this please

bhutchinson:
retest this please

db483:
Hi Chad, I have update CuDisbursementVoucherTaxServiceImplTest.java to use null for vendorId when payee type is employee. Please review and let me know if you find any issues. Thank you.

cah292:
The changes look good and the unit tests have passed. Merging.

PR #136, opened 2016-06-22
KFSPTS-5519: eliminate overlay file

bhutchinson:


src/main/java/edu/cornell/kfs/module/cab/service/impl/CuCapitalAssetBuilderModuleServiceImpl.java
```diff
@@ -220,5 +221,37 @@ protected boolean validateAssetTagLocationLines(CapitalAssetInformation capitalA
...
+     * Build the appropriate note text being set to the purchase order document.
+     * Our version includes all assetNumbers, not just the first and last.
+     *
+     * @param documentType
+     * @param assetNumbers
+     * @return
+     */
+    @Override
+    protected String buildNoteTextForPurApDoc(String documentType, List<Long> assetNumbers) {
+        StringBuffer noteText = new StringBuffer();
```
jhulslander: it doesn't matter for this context, but stringbuffers perform better when they are pre-set to a reasonable size so Java doesn't have to increase the size of the buffer on the fly

jhulslander:
Looks fine to me, I will merge

PR #143, opened 2016-06-24
KFSPTS-3599 Add new fields to Account Global doc

db483:
I have added a fix for an issue with Account Global. The change was in GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java line 89, to deal with the case when there are no icr accounts entered on the Account Global document. I have also made a couple of small cleanup changes. I have changed one of the unit tests to test the service rather then a method on an object replaced GlobalObjectWithIndirectCostRecoveryAccountsTest with GlobalObjectWithIndirectCostRecoveryAccountsServiceImplTest.java. I have updated the unit tests so they don't extend TestCase and use the JUnit 4 notation @Before, @Test. Removed the public modifier from methods in GlobalObjectWithIndirectCostRecoveryAccountsService interface.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
+++ b/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
@@ -2,164 +2,824 @@ package edu.cornell.kfs.coa.businessobject;
 
 import java.sql.Date;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.businessobject.AccountGlobal;
 import org.kuali.kfs.coa.businessobject.AccountGlobalDetail;
+import org.kuali.kfs.coa.businessobject.AccountType;
+import org.kuali.kfs.coa.businessobject.BudgetRecordingLevel;
+import org.kuali.kfs.coa.businessobject.Chart;
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryType;
+import org.kuali.kfs.coa.businessobject.RestrictedStatus;
+import org.kuali.kfs.sys.KFSPropertyConstants;
 import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.core.api.datetime.DateTimeService;
+import org.kuali.rice.krad.bo.GlobalBusinessObjectDetailBase;
 import org.kuali.rice.krad.bo.PersistableBusinessObject;
 import org.kuali.rice.krad.service.BusinessObjectService;
+import org.kuali.rice.krad.util.ObjectUtils;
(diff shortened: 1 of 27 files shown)


bhutchinson:
retest this please

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -2,164 +2,824 @@
...
+	private Account updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccountGlobalDoc(
+			GlobalBusinessObjectDetailBase globalDetail) {
+		AccountGlobalDetail accountGlobalDetail = (AccountGlobalDetail) globalDetail;
+		Account account = SpringContext.getBean(BusinessObjectService.class).findByPrimaryKey(Account.class,accountGlobalDetail.getPrimaryKeys());
+
+		if (ObjectUtils.isNotNull(account)) {
+			updateAccountBasicFields(account);
+			updateAccountExtendedAttribute(account);
+			
+			if(ObjectUtils.isNotNull(this.getIndirectCostRecoveryAccounts()) && this.getIndirectCostRecoveryAccounts().size() > 0){
```
bhutchinson: since account.getIndirectCostRecoveryAccounts() is being passed to the updateIcrAccounts method, should it be what is checked in this if statement?

db483: The updateIcrAccounts method will update the ICR accounts collection on the account being edited by the global document with values entered on the account global doc. The ICR collection on the account global doc is this.getIndirectCostRecoveryAccounts(). If this is null or empty it means no ICR accounts have been entered on the Account Global document so there is no need to attempt to update ICR accounts on the account being edited.

bhutchinson: Ok, that makes sense, thanks.

bhutchinson:
I only had one question which has been answered, however, there are two tests failing in when the PR unit test suite runs. I was able to run them successfully when running GlobalIndirectCostRecoveryAccountsRuleTest by itself so maybe something is not getting cleaned up before the test runs when the entire suite runs. I'm looking into it, but if you could take a look as well @db483 that would be good. Thanks.

db483:
I am looking into it, thank you Bryan.

bhutchinson:
retest this please

bhutchinson:
retest this please

bhutchinson:
retest this please

bhutchinson:
retest this please

bhutchinson:
@db483 I tried multiple things to try to determine the source of the unit test failures without success. One thing I did notice is that the Jenkins job is doing a merge of your branch into develop and running the tests against that, not just your PR branch. Maybe we could try the same thing locally. I did do a rebase from develop which I though would pick up the changes, but the tests still passed locally for me after that. One other thing I'm wondering is if some logging in the GlobalIndirectCostRecoveryAccountsRule methods that are called by the tests and doing the work would be helpful in order to get more visibility into why the checkICRAccuntTotalDistributionOnDetailWillBe100PercentAfterUpdate method is returning false.

db483:
I am closing this PR as I have created a new one for this jira.

PR #144, opened 2016-06-24
KFSPTS-4490: Added checkbox on REQS notes to indicate whether they

bhutchinson:
should be copied to the PO. (KFSPTS-5924: put back what once was lost)

### DIFF
--- a/src/main/webapp/WEB-INF/tags/module/purap/camsDetail.tag
+++ b/src/main/webapp/WEB-INF/tags/module/purap/camsDetail.tag
@@ -234,8 +234,7 @@
                 <purap:camsLocation camsLocationAttributes="${camsLocationAttributes}" ctr="${ctr}"
                                     ctr2="new"
                                     camsAssetLocationProperty="${locationPrefix}newPurchasingCapitalAssetLocationLine"
-                                    availability="${availability}" poItemInactive="${poItemInactive}"
-                                    fullEntryMode="${fullEntryMode}"/>
+                                    availability="${availability}" poItemInactive="${poItemInactive}"/>
     	    </c:if>
 			
             <table class="standard">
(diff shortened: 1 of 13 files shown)


bhutchinson:
@cah292 NOTE after consulting with the Kuali Financials product team, I had to move the NoteExtendedAttribute from krad to sys to make this work with the latest version. We think that's a bug and James will be reporting one issue and I'll be reporting another.

src/main/webapp/WEB-INF/tags/module/purap/camsDetail.tag
```diff
@@ -234,8 +234,7 @@
                 <purap:camsLocation camsLocationAttributes="${camsLocationAttributes}" ctr="${ctr}"
                                     ctr2="new"
                                     camsAssetLocationProperty="${locationPrefix}newPurchasingCapitalAssetLocationLine"
-                                    availability="${availability}" poItemInactive="${poItemInactive}"
-                                    fullEntryMode="${fullEntryMode}"/>
+                                    availability="${availability}" poItemInactive="${poItemInactive}"/>
```
cah292: I don't recall this change being part of the original PR for this feature on the KFS 5.3.1 line. Was this supposed to be for a different PR?

cah292:
Changes look good overall, but I did have a question about a file change that I didn't think was originally present in the KFS 5.3.1 version of this feature. Will wait for feedback on that.

src/main/webapp/WEB-INF/tags/module/purap/camsDetail.tag
bhutchinson: It was something I had missed when making changes earlier and didn't notice until I tried testing this one by opening a REQS via the webapp. I could separate it out, but it seemed easier to just include the fix here.

cah292: Ok, sounds good given that this is for the upgrade branch.

cah292:
Changes are good and the extra file change is okay to include as well. Merging.

PR #148, opened 2016-06-30
KFSTPTS-5983: remove overlay files we no longer need

bhutchinson:


cah292:
Removals look good; confirmed with Bryan that the other files from the contribution have already been removed by other user stories. Merging.

PR #150, opened 2016-06-30
KFSPTS-5680 started using parameter instead of local static variable

jhulslander:


src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowBuilder.java
```diff
@@ -334,7 +337,7 @@ Document getWorkflowDocumentForTaxRow(String documentId, T summary) {
                     // If still greater than cached ID and more unfetched docs exist, perform next bulk retrieval.
                     if (idCompareResult > 0 && documentIdsForBulkQuery.hasNext()) {
                         StringBuilder docIdCriteria = new StringBuilder(DOC_ID_CRITERIA_SIZE);
-                        for (int i = 0; documentIdsForBulkQuery.hasNext() && i < DOC_SEARCH_BATCH_SIZE; i++) {
+                        for (int i = 0; documentIdsForBulkQuery.hasNext() && i < getMaxSearchSize(); i++) {
```
cah292: You may want to store the result of getMaxSearchSize() in a local variable, otherwise the method will get called on every iteration of the loop. Unless you think the parameter retrieval caching in that method can mitigate efficiency concerns (though I'm not certain if it can cache null values for nonexistent parameters).

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowBuilder.java
```diff
@@ -447,5 +450,22 @@ String getReplacementVendorTaxNumber(String payeeId, T summary) {
...
         return stats;
     }
+    
+    protected int getMaxSearchSize() {
+    	String searchLimit = getParameterService().getParameterValueAsString(KewApiConstants.KEW_NAMESPACE, 
+        		KRADConstants.DetailTypes.DOCUMENT_SEARCH_DETAIL_TYPE, KewApiConstants.DOC_SEARCH_RESULT_CAP);
+    	try {
+    		return Integer.parseInt(searchLimit);
+    	} catch (Exception e) {
+    		LOG.debug("Paramater look up of '" + KewApiConstants.DOC_SEARCH_RESULT_CAP + "' found a value of '" + searchLimit +
```
cah292: Minor typo in log label.

cah292:
The changes look good overall, though I did notice a couple minor items that may need tweaking. Will wait for a response.

jhulslander:
That was a good idea to use local variable, it allowed me to do logging like I wanted originally, and it did speed things up a bit.  Apparently parameter service doesn't cache NULL values (which makes sense I think) and we avoid having to convert a String to an integer every time.
this is ready for review again.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowBuilder.java
```diff
@@ -447,5 +452,26 @@ String getReplacementVendorTaxNumber(String payeeId, T summary) {
...
         return stats;
     }
+    
+    protected int getMaxSearchSize() {
+    	if (maxSearchResultSize == null) {
+	    	String searchLimit = getParameterService().getParameterValueAsString(KewApiConstants.KEW_NAMESPACE, 
+	        		KRADConstants.DetailTypes.DOCUMENT_SEARCH_DETAIL_TYPE, KewApiConstants.DOC_SEARCH_RESULT_CAP);
+	    	try {
+	    		LOG.debug("Found a value for KewApiConstants.DOC_SEARCH_RESULT_CAP, and it is '" + searchLimit + "'");
+	    		maxSearchResultSize = Integer.parseInt(searchLimit);
```
cah292: Implicit conversion of int to Integer, since Integer.parseInt() returns an int. Integer.valueOf() can be used to parse the string and return an Integer object instead. This code should still work fine as-is, but I would suggest updating this method to be consistent about implicit and explicit conversions, since an explicit one is occurring on the return statement.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowBuilder.java
```diff
@@ -447,5 +452,26 @@ String getReplacementVendorTaxNumber(String payeeId, T summary) {
...
+	    	} catch (Exception e) {
+	    		LOG.error("Unable to convert '" + searchLimit + 
+	    				"' to an integer.  Returning value of KewApiConstants.DOCUMENT_LOOKUP_DEFAULT_RESULT_CAP which is " + 
+	    				KewApiConstants.DOCUMENT_LOOKUP_DEFAULT_RESULT_CAP);
+	    		maxSearchResultSize = KewApiConstants.DOCUMENT_LOOKUP_DEFAULT_RESULT_CAP;
```
cah292: Another implicit int-to-Integer conversion. It should work fine as-is, but I would suggest updating things for consistency due to the explicit Integer-to-int conversion on the return statement.

cah292:
The code looks good in general, but I did have some suggestions to make things more consistent about implicit and explicit conversions between int and Integer. Will wait for feedback.

jhulslander:
OK, I cleaned up the implicit conversions

jhulslander:
oh, and ready for review again

cah292:

jhulslander:
thanks

PR #156, opened 2016-07-06
KFSPTS-5761 implement recurring DV document

jhulslander:
I fixed the issues cathy found, the unit test problems (I haven't done the re-engineer the unit tests per our discussions)


src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherForm.java
```diff
@@ -0,0 +1,37 @@
...
+	public String getTripUrl() {
+		return StringUtils.EMPTY;
+	}
+	
+	@Override
+	public String getTripID() {
+		return StringUtils.EMPTY;
+	}
+	
+	public boolean getDefaultOpenDVDetails() {
```
bhutchinson: since this returns a boolean, should it be isDefaultOpenDVDetails() or perhaps isDvDetailsDefaultOpen()?

bhutchinson:
I've reviewed the updates and just had one question about a method name.

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherForm.java
jhulslander: It's called by a tag so I think it needs to start with Get, correct.

jhulslander:
I sit corrected, GET isn't required .... I'll make the change to a more appropriate function name

jhulslander:

bhutchinson:
Looks good; I'll merge once tests pass.

PR #158, opened 2016-07-08
KFSPTS-5905: fix config issue related to dataobjects servlet

bhutchinson:


bhutchinson:
@jhulslander same three unit test failures for the same reason

jhulslander:
looks OK

PR #163, opened 2016-07-12
KFSPTS-5386: migrate dynamodb poc code to the cu-kfs overlay

bhutchinson:


src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -1430,9 +1430,9 @@ protected StringBuffer buildLoadSummary(ElectronicInvoiceLoad eInvoiceLoad) {
         finalText.append("======================================\n");
         finalText.append("               TOTALS\n");
         finalText.append("======================================\n\n");
-        finalText.append(ACCEPT).append("           : ").append(loadCounts.get(ACCEPT)!=null?loadCounts.get(ACCEPT):0).append("\n");
-        finalText.append(REJECT).append("           : ").append(loadCounts.get(REJECT)!=null?loadCounts.get(REJECT):0).append("\n");
-        finalText.append(EXTRACT_FAILURES).append(" : ").append(loadCounts.get(EXTRACT_FAILURES)!=null?loadCounts.get(EXTRACT_FAILURES):0).append("\n");
+        finalText.append(ACCEPT).append("           : ").append((Integer)(loadCounts.get(ACCEPT)!=null?loadCounts.get(ACCEPT):0)).append("\n");
+        finalText.append(REJECT).append("           : ").append((Integer)(loadCounts.get(REJECT)!=null?loadCounts.get(REJECT):0)).append("\n");
+        finalText.append(EXTRACT_FAILURES).append(" : ").append((Integer)(loadCounts.get(EXTRACT_FAILURES)!=null?loadCounts.get(EXTRACT_FAILURES):0)).append("\n");
```
cah292: Why is this explicit Integer cast needed? Were there some primitive-vs-object conflicts that were messing up the output?

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DynamoDBClient.java
```diff
@@ -0,0 +1,21 @@
...
+import edu.cornell.kfs.sys.dataaccess.DocumentstoreClient;
+
+public class DynamoDBClient implements DocumentstoreClient {
+
+    private AmazonDynamoDBClient dynamoDBClient;
+
+    public DynamoDBClient(String host, int port, String dbName, String username, String password) {
+        AWSCredentials credentials = new BasicAWSCredentials(username, password);
+        dynamoDBClient = new AmazonDynamoDBClient(credentials);
+        dynamoDBClient.setEndpoint("http://" + host + ":" + port);
```
cah292: Are we always expecting to use "http" endpoints here, or do we need to be able to reach "https" endpoints as well?

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/MongoDBClient.java
```diff
@@ -0,0 +1,32 @@
...
+public class MongoDBClient implements DocumentstoreClient {
+
+    private MongoTemplate mongoTemplate;
+
+    public MongoDBClient(String host, int port, String dbName, String username, String password) {
+        Mongo mongo = null;
+        try {
+            mongo = new Mongo(host, port);
+        } catch (UnknownHostException e) {
+            e.printStackTrace();
```
cah292: It might be better to use log4j instead of printStackTrace() here.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
bhutchinson: Good catch, that's a Java 8 vs. Java 7 thing. The POC code I pulled in and modified for DynamoDB requires Java 8, and without this cast, this code won't compile, I think it had to do with multiple possible return types being valid - Object vs. Integer.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DynamoDBClient.java
bhutchinson: Good question. We may need https once we're using a non-local version of DynamoDB. Let me look at improving this.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/MongoDBClient.java
bhutchinson: Yes it would be, I'll fix it.

src/main/java/edu/cornell/kfs/sys/dataaccess/DocumentstoreClient.java
```diff
@@ -0,0 +1,5 @@
+package edu.cornell.kfs.sys.dataaccess;
+
+public interface DocumentstoreClient {
```
cah292: It might be better to rename this "DocumentStoreClient" (capitalize "Store"), for consistency with some other base KFS classes with "DocumentStore" in their names.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/PreferencesDaoMongo.java
```diff
@@ -0,0 +1,204 @@
...
+import org.springframework.data.mongodb.core.index.IndexInfo;
+import org.springframework.data.mongodb.core.query.BasicQuery;
+
+import java.io.IOException;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class PreferencesDaoMongo implements PreferencesDao {
```
cah292: Is the abstraction of mongoTemplate behind the documentstoreClient the only difference between this and the default KFS PreferencesDaoMongo? If so, and if you expect getClient() to always return the same object, then perhaps this could be just a subclass of the existing DAO, and relies on a setter or other configuration to populate the mongoTemplate reference from the documentstoreClient. That is, unless this abstraction needs to be maintained in case getClient() can return various instances.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
cah292: Ok, makes sense.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -255,5 +255,19 @@
...
     </bean>
-	
+
+	<bean id="documentstoreClient" parent="documentstoreClient-parentBean"/>
+		<bean id="documentstoreClient-parentBean" abstract="true" class="${kfs.documentstore.client.class}">
+			<constructor-arg value="${kfs.documentstore.host}" type="java.lang.String" />
+		    <constructor-arg value="${kfs.documentstore.port}" type="int" />
+		    <constructor-arg value="${kfs.documentstore.dbname}" type="java.lang.String" />
+		    <constructor-arg value="${kfs.documentstore.username}" type="java.lang.String" />
+		    <constructor-arg value="${kfs.documentstore.password}" type="java.lang.String" />
```
cah292: Perhaps the constructor call could be replaced with setter calls, and then use an afterPropertiesSet() method to perform the initialization. Unless the implementations are not expected to share anything except the constructor argument lists, or if only constructor initialization needs to be enforced here.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/PreferencesDaoMongo.java
bhutchinson: That's a good suggestion, let me try that.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DynamoDBClient.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    public AmazonDynamoDBClient getClient() {
```
cah292: Both implementations have getClient() methods even though the interface does not include it. Perhaps Java generics and parameterized-type features could be used to add getClient() to the interface, unless that would make inheritance or implementation classes too ugly.

cah292:
Finished initial review, and had some suggestions for improvement. Will wait for feedback.

I also take it that DynamoDB and MongoDB may not be easy to set up in a micro-testable way, since I did not see any unit tests for this yet?

In addition, with the transition of cu-kfs to Java 8, it would be good to verify if there are any other compilation problems as a result. For example, in my local environment the post6-upgrade branch is complaining about certain Comparator implementations not implementing the reversed() method. Or maybe that's a consequence of the existing code specifying Java 7 compatibility even though I recently set up my IDE to use Java 8 by default?

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
bhutchinson: I don't think these implementations are expected to share anything except the constructor argument lists and we probably only need to enforce constructor initialization. What would be the benefit of replacing the constructor call with setter calls?

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DynamoDBClient.java
bhutchinson: Sure, I can do that.

bhutchinson:
I've responded to all of your comments and am working on implementing and testing your suggested improvements. 

I'm not sure what we would test with a micro-test since the guts of the implementations need Dynamo and Mongo. There are some preferences related micro-tests in the base code, but they abstract out the mongo implementation. I'm open to suggestions if you have any.

As far as compilation problems with Java 8, I haven't seen any others and the unit test suite has also compiled and passed on Jenkins using Java 8. It may be your compatibility settings are out of sync as you suggest.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
cah292: I was just wondering whether things would be cleaner with setter usage for the properties instead of constructor args. If enforcing constructor initialization is preferable in this case, then I'm fine with that.

bhutchinson: That's fair. I think it's ok in this case. Thanks.

bhutchinson:
PR has been updated per suggestions and I even added some micro tests.

src/test/java/edu/cornell/kfs/sys/dataaccess/impl/PreferencesDaoDynamoDBTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        protected Table retrieveTable(String tableName) {
+            return tables.get(tableName);
+        }
+
+    }
+
+    private class MockTable extends Table {
+
+        private Map items = new HashMap();
```
cah292: Is there a need for this MockTable to not use generics on this Map?

cah292:
Finished second review, just had one question about a raw type in the new micro-test class, but other than that the changes look good. Will wait for feedback before merging.

bhutchinson:
Fixed that, thanks.

cah292:

PR #166, opened 2016-07-15
KFSPTS-5928: update links config for custom paths

bhutchinson:


bhutchinson:
@cah292 Since the links of the new dashboard are stored in a documentstore (MongoDB ootb and DynamoDB for us, they are loaded via JSON. I have created a modified JSON file in our overlay project where we can tweak links as needed, and that is what this is.

cah292:
Does KFS and/or the documentstore require this particular config to be one massive JSON file, as opposed to a series of smaller JSON files? If the file needs to be updated frequently then it could increase the risk of merge conflicts, like on our KFS 5.x CU-ApplicationResources.properties file.

bhutchinson:
At this point yes, we are essentially storing one value for a key, so it hast to be one large JSON file. I think the product team creates a new one for each JIRA that changes it, and also has a process (mongo specific) to make and track more incremental changes. We might be able to modify that process for DynamoDB, but that is not in scope for this story. I'm also not sure how frequently it will need to be updated - once we get the dashboard set, I would anticipate infrequent changes, and there is also a mechanism to make dashboard changes via the UI which we might also use in the future. Do you think I should rename this file to be the JIRA # or include the JIRA # and create a new one for each change (I have multiple changes coming up shortly) or can we leave it for now and deal with any issues that come up later.

cah292:
I'm fine with leaving it as-is for now and dealing with other issues as they arise, though for future changes it would be helpful to have some kind of process for tracking updates. Merging.

PR #167, opened 2016-07-18
Kfspts 3599 new 3

nkimble:
Requesting code review.
Two commits present so Daniela can see the changes between her original code and my additions.
Please do NOT merge until she has had a change to review and comment.

A squash merge will most likely need to be performed to get a single commit before the PR can be performed.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
+++ b/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
@@ -124,9 +124,38 @@ public class CuAccountGlobal extends AccountGlobal implements GlobalObjectWithIn
 			if(ObjectUtils.isNotNull(this.getIndirectCostRecoveryAccounts()) && this.getIndirectCostRecoveryAccounts().size() > 0){
 				updateIcrAccounts(globalDetail, account.getIndirectCostRecoveryAccounts());
 			}
+
+			if (ObjectUtils.isNotNull(this.accountRestrictedStatusCode)) {
+				updateAccountRestrictedStatusCodeForAccountBeingEdited(this.accountRestrictedStatusCode, account);
+			}
 		}
 
 		return account;
+	}
+
+	/**
+	 * The business rules for the accountRestrictedStatusCode is as follows:
+	 * If the sub-fund of the account listed in the edit list of accounts on
+	 * the global maintenance edoc is one that has a default accountRestrictedStatusCode
+	 * listed, then any change entered in the global account maintenance edoc is ignored.
+	 *
+	 * If the sub-fund of the account listed in the edit list of accounts on
+	 * the global maintenance edoc is one that does NOT have a default accountRestrictedStatusCode
+	 * listed, then any change entered in the global account maintenance edoc overwrites
+	 * the accountRestrictedStatusCode value on that particular account from the edit list.
+	 *
+	 * @param globalAccountMaintenanceDocRestrictedStatusCode
+	 * @param accountBeingEdited
+	 */
+	private void updateAccountRestrictedStatusCodeForAccountBeingEdited(String globalAccountMaintenanceDocRestrictedStatusCode, Account accountBeingEdited)
+	{
+		if (ObjectUtils.isNotNull(accountBeingEdited.getSubFundGroup()) && StringUtils.isNotBlank(accountBeingEdited.getSubFundGroup().getAccountRestrictedStatusCode())) {
+			//ignore the global account maintenance edoc user entered value because the sub-fund for the account already has a default value listed that should be used
+		}
+		else {
+			//sub-fund associated to the account does NOT have a default accountRestrictedStatusCode identified, utilize the user entered value from the  global account maintenance edoc
+			accountBeingEdited.setAccountRestrictedStatusCode(globalAccountMaintenanceDocRestrictedStatusCode);
+		}
 	}
 
 	private void updateAccountBasicFields(Account account) {
@@ -247,9 +276,7 @@ public class CuAccountGlobal extends AccountGlobal implements GlobalObjectWithIn
 		    account.setReportsToAccountNumber(reportsToAccountNumber);
 		}
 
-		if (StringUtils.isNotBlank(accountRestrictedStatusCode)) {
-		    account.setAccountRestrictedStatusCode(accountRestrictedStatusCode);
-		}
+		// Method determineAccountRestrictedStatusCodeForAccountBeingEdited will determine the appropriate value to use.
 
 		if (ObjectUtils.isNotNull(accountRestrictedStatusDate)) {
 		    account.setAccountRestrictedStatusDate(accountRestrictedStatusDate);
--- a/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
+++ b/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
@@ -2,164 +2,824 @@ package edu.cornell.kfs.coa.businessobject;
 
 import java.sql.Date;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.businessobject.AccountGlobal;
 import org.kuali.kfs.coa.businessobject.AccountGlobalDetail;
+import org.kuali.kfs.coa.businessobject.AccountType;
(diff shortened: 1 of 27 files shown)


src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -124,11 +124,40 @@ private Account updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccount
...
+	 * the global maintenance edoc is one that does NOT have a default accountRestrictedStatusCode
+	 * listed, then any change entered in the global account maintenance edoc overwrites
+	 * the accountRestrictedStatusCode value on that particular account from the edit list.
+	 *
+	 * @param globalAccountMaintenanceDocRestrictedStatusCode
+	 * @param accountBeingEdited
+	 */
+	private void updateAccountRestrictedStatusCodeForAccountBeingEdited(String globalAccountMaintenanceDocRestrictedStatusCode, Account accountBeingEdited)
+	{
+		if (ObjectUtils.isNotNull(accountBeingEdited.getSubFundGroup()) && StringUtils.isNotBlank(accountBeingEdited.getSubFundGroup().getAccountRestrictedStatusCode())) {
```
bhutchinson: I don't really like having an empty if block just to have a comment; I think negating this condition and just having the logic in the else block (also without the comment) would be better. You have explained the logic in the method comment and the method name. If more explanation is needed, maybe better naming, or extracting some of the condition/logic into appropriate named methods would help.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -247,9 +276,7 @@ private void updateAccountBasicFields(Account account) {
 		    account.setReportsToAccountNumber(reportsToAccountNumber);
 		}
 
-		if (StringUtils.isNotBlank(accountRestrictedStatusCode)) {
-		    account.setAccountRestrictedStatusCode(accountRestrictedStatusCode);
-		}
+		// Method determineAccountRestrictedStatusCodeForAccountBeingEdited will determine the appropriate value to use.
```
bhutchinson: Please remove this comment.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -1607,49 +1584,21 @@ protected boolean checkCloseAccount(AccountGlobalDetail detail) {
...
-        Account existingAccount = accountGlobalDetail.getAccount();
 
         // get the expiration date, if any
-        Date expirationDate = existingAccount.getAccountExpirationDate();
+        Date expirationDate = newAccount.getAccountExpirationDate();
         if (ObjectUtils.isNull(expirationDate)) {
-        	expirationDate = newAccountGlobal.getAccountExpirationDate();
-        	if (ObjectUtils.isNull(expirationDate)) {
-            putFieldError("accountExpirationDate", CUKFSKeyConstants.ERROR_DOCUMENT_GLB_MAINT_CANNOT_BE_CLOSED_EXP_DATE_INVALID, new String[]{existingAccount.getChartOfAccountsCode(), existingAccount.getAccountNumber()});
+        	putFieldError(KFSPropertyConstants.ACCOUNT_EXPIRATION_DATE, KFSKeyConstants.ERROR_DOCUMENT_ACCMAINT_ACCT_CANNOT_BE_CLOSED_EXP_DATE_INVALID);
```
bhutchinson: Does there always have to be an expiration date? This logic seems to indicate that there does (but maybe I'm not seeing the entire context).

bhutchinson:
I reviewed the new changes (based on my diff with the prior PR, it looks like all the previous changes have been captured so I focused on the new ones) and only had a couple comments. I didn't dig into the logic changes too much, I'm interested in Daniela's thoughts, but could do that if necessary.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
nkimble: I will look at extracting the condition logic into an appropriately named method (even though it may be a on-liner).  I think that for future maintenance it will make the code more readable than having double negatives that others later on will need to sift through.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
nkimble: I will remove the comment.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -124,11 +124,40 @@ private Account updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccount
 			if(ObjectUtils.isNotNull(this.getIndirectCostRecoveryAccounts()) && this.getIndirectCostRecoveryAccounts().size() > 0){
 				updateIcrAccounts(globalDetail, account.getIndirectCostRecoveryAccounts());
 			}
+
+			if (ObjectUtils.isNotNull(this.accountRestrictedStatusCode)) {
```
db483: I would move this code inside the updateAccountBasicFields() method.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -1607,49 +1584,21 @@ protected boolean checkCloseAccount(AccountGlobalDetail detail) {
...
         Date todaysDate = new Date(getDateTimeService().getCurrentDate().getTime());
         todaysDate.setTime(DateUtils.truncate(todaysDate, Calendar.DAY_OF_MONTH).getTime());
-        // TODO: convert this to using Wes' Kuali KfsDateUtils once we're using Date's instead of Timestamp
-        
-        accountGlobalDetail.refreshReferenceObject(KFSPropertyConstants.ACCOUNT);
-        Account existingAccount = accountGlobalDetail.getAccount();
 
         // get the expiration date, if any
-        Date expirationDate = existingAccount.getAccountExpirationDate();
+        Date expirationDate = newAccount.getAccountExpirationDate();
```
db483: If the AccountGlobal expiration date field is empty and the close account field is yes then the user should be able to still close the account if the account being edited already has an expiration date on it. I am afraid after this change the user is always required to enter an expiration date on account global doc when closing an account.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -124,11 +124,40 @@ private Account updateAccountValuesOnAccountGlobalDetailWithNewValuesFromAccount
...
+	 * @param accountBeingEdited
+	 */
+	private void updateAccountRestrictedStatusCodeForAccountBeingEdited(String globalAccountMaintenanceDocRestrictedStatusCode, Account accountBeingEdited)
+	{
+		if (ObjectUtils.isNotNull(accountBeingEdited.getSubFundGroup()) && StringUtils.isNotBlank(accountBeingEdited.getSubFundGroup().getAccountRestrictedStatusCode())) {
+			//ignore the global account maintenance edoc user entered value because the sub-fund for the account already has a default value listed that should be used
+		}
+		else {
+			//sub-fund associated to the account does NOT have a default accountRestrictedStatusCode identified, utilize the user entered value from the  global account maintenance edoc
+			accountBeingEdited.setAccountRestrictedStatusCode(globalAccountMaintenanceDocRestrictedStatusCode);
```
db483: This will not work correctly if the Account global doc is changing the sub-fund group code. The restricted status code will be set from the existing sub fund group rather than updated one.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
nkimble: Done.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
nkimble: Adjusted the implementation to deal with the possibility sub-fund is being updated at the same time as the account restricted status code on the global doc.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
nkimble: Looking into haw the code can be adjusted appropriately.

PR #171, opened 2016-07-21
KFSPTS-6301: customize Dashboard

bhutchinson:


src/main/resources/edu/cornell/kfs/coa/document/workflow/SubObjectCodeGlobalEditMaintenanceDocument.xml
```diff
@@ -19,7 +19,7 @@
     <documentType>
       <name>GSOI</name>
       <parent>COA</parent>
-      <label>Global Sub-Object Code Inactivation</label>
+      <label>Sub-Object Code Global Inactivation</label>
```
cah292: This appears to be a change to a doc type instead of a change to the dashboard JSON. Does it still influence the appearance on the KFS dashboard? Or is it an extra fix that's still within the scope of this change?

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
```diff
@@ -683,6 +757,17 @@
         "documentTypeCode": "BRL",
         "linkType": "kfs"
       }, {
+        "label": "CG Reporting Code",
+        "link": "kr/lookup.do?methodToCall=start&businessObjectClassName=edu.cornell.kfs.coa.businessobject.ContractGrantReportingCode&docFormKey=88888888&hideReturnLink=true",
+        "linkType": "kfs",
+        "permission": {
+          "templateNamespace": "KR-SYS",
+          "templateName": "Initiate Document",
```
cah292: This appears to be a link to a lookup screen, but the link is restricted based on a document-initiation permission instead of a lookup or use-screen one. Is this intentional?

cah292:
Changes look good overall, but I did find a few little things related to the scope of the change and the permission setup.

src/main/resources/edu/cornell/kfs/coa/document/workflow/SubObjectCodeGlobalEditMaintenanceDocument.xml
bhutchinson: The code is able to automatically generate the link label from the label of the doc type. In this case the label was inconsistent, so the easiest way to get the correct link label was to change the doc type xml and let the code do it's thing (otherwise I would have had to add several custom lines to the json). We had previously made a similar change to another global doc for consistency as well.

cah292: Ok, sounds good.

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
bhutchinson: Well, I may have done this intentionally, but it's wrong, I'll fix it. Thanks for catching it.

bhutchinson:
I fixed the permission, so this should be ready for review again.

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
```diff
@@ -134,8 +137,16 @@
...
         }
       }, {
-        "businessObjectClass": "org.kuali.kfs.gl.businessobject.CurrentAccountBalance",
-        "linkType": "kfs"
+        "label": "Current Fund Balance",
+        "link": "kr/lookup.do?methodToCall=start&businessObjectClassName=edu.cornell.kfs.gl.businessobject.CurrentAccountBalance&docFormKey=88888888&hideReturnLink=true",
+        "linkType": "kfs",
+        "permission": {
+          "templateNamespace": "KR-NS",
+          "templateName": "Use Screen",
```
cah292: Given that this is also a lookup, should this be a "Look Up Records" permission as well, or should it be a "Use Screen" one like it is currently? Sorry for not pointing it out the first time, didn't think it was a problem at the time.

cah292:
Finished second round of review, had a second permission question based on the resolution of the first permission question. Will wait for feedback.

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
bhutchinson: I think I was trying to be consistent with the other balance inquiries, but they are not straight lookups like this, so you're probably right that a "Look Up Records" permission would probably be better. (FWIW, this customization will be probably be removed soon as long as we can change the label since this lookup is in the base code now.)

bhutchinson:
Second permission fixed, so this should be ready for review again.

cah292:
Changes look good, will wait for unit tests to pass before merging.

cah292:

PR #175, opened 2016-07-25
KFSPTS-6030: reintroduce cynergy dependency

bhutchinson:


cah292:
Changes appear to be good and unit tests have passed, but I had a question about the Cynergy dependency exclusions. It looks like KFS is intending to exclude all of the Rice JARs pulled into Cynergy. Is this still desired? If so, then it may be necessary to add "rice-krad-data" and "rice-krad-service-impl" to the list, since those are now present in the newer Rice versions.

bhutchinson:
Good catch, thanks; I'll add those to the exclusions list. I think the idea was to delegate to base kfs for the rice dependencies. We can certainly revisit that at some point if we want/need to.

bhutchinson:
Ok, I've updated the pom.

cah292:

cah292:

PR #176, opened 2016-07-28
KFSPTS-3599-transient

nkimble:
The only change between this branch (KFSPTS-3599-transient)  and the last merged branch for the Account Global additional attributes branch (KFSPTS-3599-new-4) is the addition of "transient" to the local variable GlobalObjectWithIndirectCostRecoveryAccountsService in class CuAccountGlobal.

If you do a comparison between these two stated branches, you will see a change that has been added to develop for CuDisbursementVoucherDocument, which the KFSPTS-3599-transient changes are based on so it will not be lost.

This single addition of "transient" deals with all of the stack traces that the users have reported thus far.


nkimble:
retest this please

PR #181, opened 2016-08-01
KFSPTS-6347: re-skin Create Disencumbrance

bhutchinson:


cah292:
Changes look good. The unit tests were running into some odd issues, but after further discussion it turns out this was the result of some in-progress DevOps work in Jenkins. Also, these are just changes to a JSP and should not affect the unit tests that we have. Merging.

PR #185, opened 2016-08-02
KFSPTS-6354: re-skin PreEncumbrance GLPE tab

bhutchinson:


src/main/webapp/WEB-INF/tags/fp/preEncumbranceGeneralLedgerPendingEntries.tag
```diff
@@ -69,65 +88,64 @@
 			<kul:htmlAttributeHeaderCell attributeEntry="${entryAttributes.financialDocumentReversalDate}" hideRequiredAsterisk="true" scope="col"/>
 		</tr>
 		<logic:iterate id="generalLedgerPendingEntry" name="KualiForm" property="${generalLedgerPendingEntriesProperty}" indexId="ctr">
-			<tr>
-				<th class="datacell center"><html:hidden property="${generalLedgerPendingEntryProperty}[${ctr}].transactionLedgerEntrySequenceNumber" write="true" value="${generalLedgerPendingEntry.transactionLedgerEntrySequenceNumber}" /></th>
-				<td class="datacell center">
+            <tr class="${ctr % 2 == 0 ? "highlight" : ""}">
```
jhulslander: I don't think it's a big deal, but I don't really like JSTL in line if statements like this.  They aren't as easy to read as breaking it out to a <c:choose> block.  I'll let you decide if it is worth chaning.

jhulslander:
Changes look good, I had one small comment, I shall wait for feedback.

src/main/webapp/WEB-INF/tags/fp/preEncumbranceGeneralLedgerPendingEntries.tag
bhutchinson: That's copied directly from the base code, so I would prefer to keep it as is.

jhulslander: fair enough

PR #191, opened 2016-08-04
KFSPTS-5988 Implement AWS billing integration

jhulslander:
Please don't merge this yet, I need to verify some outputs with Nicole, and see it run against the WS after it has been restored.  I would like some feedback on the code in the mean time.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
+++ b/src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
@@ -135,5 +135,26 @@ public class CuFPConstants {
         public static final String RECURRING_DV_EXPLANATION_TO_DV_NOTE_STARTER = "DV created by recurring DV: ";
         public static final String RECURRING_DETAILS_TAB_NAME = "RecurringDetails";
     }
+    
+    public static class AmazonWebServiceBillingConstants {
+        public static final String AWS_PROCESSING_DATE_PROPERTY_NAME = "AWS_PROCESSING_DATE";
+        public static final String AWS_FROM_ACCOUNT_PROPERTY_NAME = "AWS_FROM_ACCOUNT_NUMBER";
+        public static final String AWS_OBJECT_CODE_PROPERTY_NAME = "AWS_OBJECT_CODE";
+        public static final String AWS_DEFAULT_DOCUMENT_DESCRIPTION_PROPERTY_NAME = "AWS_DEFAULT_DOCUMENT_DESCRIPTION";
+        public static final String AWS_NOTE_HELP_TEXT_PROERTY_NAME = "AWS_NOTE_HELP_TEXT";
+        public static final String AWS_COMPENT_NAME = "LoadAWSBillsStep";
+        
+        public static final int PROCESSING_DAY_OF_MONTH = 15;
+        
+        public static final String DEFAULT_BILLING_PERIOD_PARAMETER = "CURRENT";
+        public static final String INTERNAL_KFS_ACCOUNT_DESCRIPTION = "internal";
+        public static final String AUTHORIZATION_HEADER_NAME = "Authorization";
+        public static final String AUTHORIZATION_TOKEN_VALUE_STARTER = "Token ";
+        public static final String URL_PARAMETER_YEAR  = "year=";
+        public static final String URL_PARAMETER_MONTH = "&month=";
+        public static final String HTTP_METHOD_GET_NAME = "GET";
+        public static final String DI_ROUTE_ANNOTATION = "DI automatically created by AWS Billing Batch Job";
+        public static final String TRANSACTION_DESCRIPTION_STARTER = "AWS CHARGES ";
+    }
 
 }
--- a/src/main/java/edu/cornell/kfs/fp/service/AmazonWebServicesBillingService.java
+++ b/src/main/java/edu/cornell/kfs/fp/service/AmazonWebServicesBillingService.java
@@ -1,0 +1,10 @@
+package edu.cornell.kfs.fp.service;
+
+public interface AmazonWebServicesBillingService {
+    
+    /**
+     * This function will call the AWS Billing web service and generate Distribution Of Income Documents.
+     */
+    void generateDistributionOfIncomeDocumentsFromAWSService();
+
+}
--- a/src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
@@ -1,0 +1,647 @@
+package edu.cornell.kfs.fp.service.impl;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.text.DateFormatSymbols;
+import java.util.Calendar;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.ws.rs.core.MediaType;
+
+import org.apache.commons.lang.StringUtils;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.kuali.kfs.coa.businessobject.Account;
+import org.kuali.kfs.coa.businessobject.Chart;
+import org.kuali.kfs.coa.businessobject.ObjectCode;
+import org.kuali.kfs.coa.businessobject.ProjectCode;
+import org.kuali.kfs.coa.businessobject.SubAccount;
+import org.kuali.kfs.coa.businessobject.SubObjectCode;
+import org.kuali.kfs.coa.service.AccountService;
+import org.kuali.kfs.coa.service.ChartService;
+import org.kuali.kfs.coa.service.ObjectCodeService;
+import org.kuali.kfs.coa.service.ProjectCodeService;
+import org.kuali.kfs.coa.service.SubAccountService;
+import org.kuali.kfs.coa.service.SubObjectCodeService;
+import org.kuali.kfs.fp.document.DistributionOfIncomeAndExpenseDocument;
+import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.businessobject.SourceAccountingLine;
+import org.kuali.kfs.sys.businessobject.TargetAccountingLine;
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+import org.kuali.rice.coreservice.framework.parameter.ParameterService;
+import org.kuali.rice.kew.api.exception.WorkflowException;
+import org.kuali.rice.krad.bo.DocumentHeader;
+import org.kuali.rice.krad.bo.Note;
+import org.kuali.rice.krad.service.BusinessObjectService;
+import org.kuali.rice.krad.service.DocumentService;
+import org.kuali.rice.krad.util.GlobalVariables;
+import org.kuali.rice.krad.util.KRADConstants;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+import com.sun.jersey.api.client.Client;
+import com.sun.jersey.api.client.ClientRequest;
+import com.sun.jersey.api.client.ClientResponse;
+import com.sun.jersey.api.client.config.ClientConfig;
+import com.sun.jersey.api.client.config.DefaultClientConfig;
+
+import edu.cornell.kfs.fp.CuFPConstants;
+import edu.cornell.kfs.fp.businessobject.AmazonBillingCostCenterDTO;
+import edu.cornell.kfs.fp.businessobject.AmazonBillingDistributionOfIncomeTransactionDTO;
+import edu.cornell.kfs.fp.document.CuDistributionOfIncomeAndExpenseDocument;
+import edu.cornell.kfs.fp.document.service.impl.CULegacyTravelServiceImpl;
+import edu.cornell.kfs.fp.service.AmazonWebServicesBillingService;
+import edu.cornell.kfs.fp.xmlObjects.AmazonAccountDetail;
+import edu.cornell.kfs.fp.xmlObjects.AmazonAccountDetailContainer;
+import edu.cornell.kfs.sys.CUKFSConstants;
+
+public class AmazonWebServicesBillingServiceImpl implements AmazonWebServicesBillingService, Serializable {
+
+    private static final long serialVersionUID = 7430710204404759511L;
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(AmazonWebServicesBillingServiceImpl.class);
+    
+    private String awsURL;
+    private String awsToken;
+    private String billingPeriodParameterValue;
+    private String defaultDocumentDescription;
+    private String transactionObjectCode;
+    private String transactionFromAccountNumber;
+    private String helpNoteText;
+
+    protected ParameterService parameterService;
+    protected DocumentService documentService;
+    protected ChartService chartService;
+    protected AccountService accountService;
+    protected SubAccountService subAccountService;
+    protected BusinessObjectService businessObjectService;
+    protected ObjectCodeService objectCodeService;
+    protected SubObjectCodeService subObjectCodeService;
+    protected ProjectCodeService projectCodeService;
+
+    @Override
+    public void generateDistributionOfIncomeDocumentsFromAWSService() {
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService(): AWS URL: " + getAwsURL());
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService(): Processing Year: " + findProcessYear() + "  Month: " + findProcessMonthNumber());
+        
+        String jsonResults = buildJsonOutput();
+        LOG.debug("JSON results: " + jsonResults);
+
+        List<AmazonAccountDetail> amazonAccountDetails = buildAmazonAcountListFromJson(jsonResults);
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService() Number of amazon account details received: " + amazonAccountDetails.size());
+        
+        boolean didProcessingCompleteWithoutErrors = true;
+        
+        if (amazonAccountDetails.size() == 0) {
+            LOG.error("generateDistributionOfIncomeDocumentsFromAWSService() No amazon account detail objects parsed from the webservice.");
+            didProcessingCompleteWithoutErrors =  false;
+        }
+        
+        didProcessingCompleteWithoutErrors = didProcessingCompleteWithoutErrors && processAmazonAccountDetails(amazonAccountDetails);
+        
+        clearMemberVariables();
+        if (!didProcessingCompleteWithoutErrors) {
+            throw new RuntimeException("There was error processing AWS Billing.");
+        }
+    }
+    
+    private String buildJsonOutput() {
+        ClientConfig clientConfig = new DefaultClientConfig();
+        Client client = Client.create(clientConfig);
+        
+        ClientResponse response = client.handle(buildClientRequest());
+        
+        response.bufferEntity();
+        String results = response.getEntity(String.class);
+        return results;
+    }
+    
+    protected ClientRequest buildClientRequest() {
+        ClientRequest.Builder builder = new ClientRequest.Builder();
+        builder.accept(MediaType.APPLICATION_JSON);
+        builder.header(CuFPConstants.AmazonWebServiceBillingConstants.AUTHORIZATION_HEADER_NAME, 
+                CuFPConstants.AmazonWebServiceBillingConstants.AUTHORIZATION_TOKEN_VALUE_STARTER + getAwsToken());
+        URI uri;
+        try {
+            uri = new URI(getAwsURL() + CuFPConstants.AmazonWebServiceBillingConstants.URL_PARAMETER_YEAR + findProcessYear() + 
+                    CuFPConstants.AmazonWebServiceBillingConstants.URL_PARAMETER_MONTH + findProcessMonthNumber());
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+        ClientRequest request = builder.build(uri, CuFPConstants.AmazonWebServiceBillingConstants.HTTP_METHOD_GET_NAME);
+        return request;
+    }
+    
+    protected List<AmazonAccountDetail> buildAmazonAcountListFromJson(String jsonResults) {
+        ObjectMapper mapper = new ObjectMapper();
+        AmazonAccountDetailContainer detailContainer = new AmazonAccountDetailContainer();
+        try {
+            detailContainer = mapper.readValue(jsonResults, AmazonAccountDetailContainer.class);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return detailContainer.getAccountDetail();
+    }
+
+    protected boolean processAmazonAccountDetails(List<AmazonAccountDetail> amazonAccountDetails) {
+        boolean didProcessingCompleteWithoutErrors = true;
+        int count = 0;
+        for (AmazonAccountDetail amazonAccountDetail : amazonAccountDetails) {
+            if (StringUtils.equalsIgnoreCase(CuFPConstants.AmazonWebServiceBillingConstants.INTERNAL_KFS_ACCOUNT_DESCRIPTION, amazonAccountDetail.getKfsAccount())) {
+                LOG.info("processAmazonAccountDetails() Found an internal kfs account, ignoring this transaction for AWS account " 
+                    + amazonAccountDetail.getAwsAccount());
+            } else {
+                didProcessingCompleteWithoutErrors = processExternalBill(amazonAccountDetail) && didProcessingCompleteWithoutErrors;
+            }
+            count ++;
+        }
+        LOG.info("Processed " + count + " AWS records");
+        return didProcessingCompleteWithoutErrors;
+    }
+
+    private boolean processExternalBill(AmazonAccountDetail amazonAccountDetail) {
+        boolean valid = true;
+        try {
+            AmazonBillingDistributionOfIncomeTransactionDTO transactionDTO = buildTransactionDTO(amazonAccountDetail);
+            if (transactionDTO.isTransactionInputError()) {
+                valid = false;
+                LOG.error("processAmazonAccountDetails() there errors that prevent a creation of a DI " + amazonAccountDetail.toString());
+            } else if (shouldBuildDistributionOfIncomeDocument(amazonAccountDetail, transactionDTO.getAmount())) {
+                buidAndRouteDistributionOfIncomeDocument(amazonAccountDetail, transactionDTO);
+            }
+        } catch (Throwable e) {
+            valid = false;
+            LOG.error("processAmazonAccountDetails() there was a problem building a DI for " + amazonAccountDetail.toString(), e);
+            GlobalVariables.getMessageMap().getErrorMessages().clear();
+        }
+        return valid;
+    }
+    
+    protected AmazonBillingDistributionOfIncomeTransactionDTO buildTransactionDTO(AmazonAccountDetail amazonAccountDetail) {
+        
+        AmazonBillingDistributionOfIncomeTransactionDTO transactionDTO = new AmazonBillingDistributionOfIncomeTransactionDTO();
+        transactionDTO.setLineDescription(buildAccountingLineDescription());
+        
+        updateTransactionDTOAmount(amazonAccountDetail, transactionDTO);
+        
+        try {
+            AmazonBillingCostCenterDTO costCenterDTO = convertCostCenterToAmazonBillingCostCenterDTO(amazonAccountDetail.getCostCenter());
+            updateTransactionDTOChartAccountSubAccount(transactionDTO, costCenterDTO, amazonAccountDetail.getKfsAccount());
+            updateTransactionDTOObjectCodes(transactionDTO, costCenterDTO);
+            updateTransactionDTOProjectCodeNumber(transactionDTO, costCenterDTO);
+            transactionDTO.setOrganizationReferenceId(costCenterDTO.getOrgReferenceId());
+        } catch (Exception e) {
+            transactionDTO.setTransactionInputError(true);
+            LOG.error("buildAmazonDistributionOfIncomeTransactionDTO: There was a problem parsing cost center.", e);
+        }
+        return transactionDTO;
+    }
+
+    private void updateTransactionDTOAmount(AmazonAccountDetail amazonAccountDetail, AmazonBillingDistributionOfIncomeTransactionDTO transactionDTO) {
+        transactionDTO.setAmount(convertCostStringToKualiDecimal(amazonAccountDetail.getCost()));
+        if (transactionDTO.getAmount().isLessEqual(KualiDecimal.ZERO)) {
+            transactionDTO.setTransactionInputError(true);
+        }
+    }
+    
+    protected KualiDecimal convertCostStringToKualiDecimal(String costString) {
+        KualiDecimal cost = new KualiDecimal(costString);
+        final double oneCent = .01;
+        if (cost.doubleValue() < oneCent) {
+            return KualiDecimal.ZERO;
+        }
+        return cost;
+    }
+    
+    protected AmazonBillingCostCenterDTO convertCostCenterToAmazonBillingCostCenterDTO(String costCenter) throws Exception {
+        AmazonBillingCostCenterDTO costCenterDTO = new AmazonBillingCostCenterDTO();
+        
+        if (StringUtils.contains(costCenter, KFSConstants.DASH)) {
+            String[] costCenterArray = StringUtils.splitPreserveAllTokens(costCenter, KFSConstants.DASH);
+            if (costCenterArray.length == 2) {
+                costCenterDTO.setAccountNumber(costCenterArray[0]);
+                costCenterDTO.setSubAccountNumber(costCenterArray[1]);
+            } else if (costCenterArray.length == 7) {
+                costCenterDTO.setChartCode(costCenterArray[0]);
+                costCenterDTO.setAccountNumber(costCenterArray[1]);
+                costCenterDTO.setSubAccountNumber(costCenterArray[2]);
+                costCenterDTO.setObjectCode(costCenterArray[3]);
+                costCenterDTO.setSubObjectCode(costCenterArray[4]);
+                costCenterDTO.setProjectCode(costCenterArray[5]);
+                costCenterDTO.setOrgReferenceId(costCenterArray[6]);
+            } else {
+                throw new Exception("There was an incorrect number of elements in the Cost Center");
+            }
+        } else {
+            costCenterDTO.setAccountNumber(costCenter);
+        }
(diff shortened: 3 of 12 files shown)


jhulslander:
The checks are failing because I also need to update the security.properties file with new information, on my list to do yet.

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/fp/service/AmazonWebServicesBillingService.java
```diff
@@ -0,0 +1,10 @@
+package edu.cornell.kfs.fp.service;
+
+public interface AmazonWebServicesBillingService {
+    
+    /**
+     * This function will call the AWS Billing web service and generates Distribution Of Income Documents.
```
bhutchinson: Nitpicky, but I would say generate (vs. generates) here.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+import edu.cornell.kfs.fp.service.AmazonWebServicesBillingService;
+import edu.cornell.kfs.fp.xmlObjects.AmazonAccountDetail;
+import edu.cornell.kfs.fp.xmlObjects.AmazonAccountDetailContainer;
+
+public class AmazonWebServicesBillingServiceImpl implements AmazonWebServicesBillingService, Serializable {
+
+    private static final long serialVersionUID = 7430710204404759511L;
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(AmazonWebServicesBillingServiceImpl.class);
+    
+    private static final String DEFAULT_BILLING_PERIOD_PARAMETER = "CURRENT";
```
bhutchinson: Is there a reason to have these constants in this class vs. in CuFPConstants with the other AmazonWebServiceBillingConstants?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        
+        valid = valid && processAmazonAccountDetails(amazonAccountDetails);
+        
+        clearMemberVariables();
+        if (!valid) {
+            throw new RuntimeException("There was error processing AWS Billing.");
+        }
+    }
+    
+    private String buildJSONoutput() {
```
bhutchinson: I would probably capitalize O in output. Either buildJSONOutput or buildJsonOutput (I'm seeing more of this format lately I think).

bhutchinson: JSON is a formatting implementation detail, and there isn't necessarily anything specific to JSON in this method - should the name be more generic (i.e. buildOutput or something like that)?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        
+        response.bufferEntity();
+        String results = response.getEntity(String.class);
+        return results;
+    }
+    
+    private ClientRequest buildClientRequest() {
+        ClientRequest.Builder builder = new ClientRequest.Builder();
+        builder.accept(MediaType.APPLICATION_JSON);
+        final String securityTokenValueStarter = "Token ";
```
bhutchinson: Another possible constant.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+        ClientRequest request = builder.build(uri, HTTP_METHOD_GET_NAME);
+        return request;
+    }
+    
+    public String findBillingYear() {
+        if (StringUtils.equalsIgnoreCase(getBillingPeriodParameterValue(), DEFAULT_BILLING_PERIOD_PARAMETER)) {
+            SimpleDateFormat dateFormater = new SimpleDateFormat("YYYY");
```
bhutchinson: Another possible constant.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+    private boolean processAmazonAccountDetails(List<AmazonAccountDetail> amazonAccountDetails) {
+        boolean valid = true;
+        int count = 0;
+        for (AmazonAccountDetail amazonAccountDetail : amazonAccountDetails) {
+            if (StringUtils.equalsIgnoreCase(INTERNAL_KFS_ACCOUNT_DESCRIPTION, amazonAccountDetail.getKfsAccount())) {
+                LOG.info("processAmazonAccountDetails() Found an internal kfs account, ignoring this transaction for AWS account " 
+                    + amazonAccountDetail.getAwsAccount());
+            } else {
+                Account account = determineWhichAccountToUse(amazonAccountDetail.getCostCenter(), amazonAccountDetail.getKfsAccount());
+                if (account != null) {
```
bhutchinson: If the account is null, do we need to log that we skipped it or do anything else?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+                    KualiDecimal cost = new KualiDecimal(Double.valueOf(amazonAccountDetail.getCost()));
+                    if (shouldBuildDistributionOfIncomeDocument(amazonAccountDetail, cost)) {
+                        try {
+                            buidAndRouteDistributionOfIncomeDocument(account, cost, amazonAccountDetail);
+                        } catch (Throwable e) {
+                            valid = false;
+                            LOG.error("processAmazonAccountDetails() there waw a problem building a DI for " +
```
bhutchinson: mispelling - was => was I think

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+    
+    private boolean validAccount(Account account) {
+        return account != null && account.isActive() && !account.isExpired();
+    }
+    
+    private boolean shouldBuildDistributionOfIncomeDocument(AmazonAccountDetail amazonAccountDetail, KualiDecimal cost) {
+        for (DocumentHeader dh : findDocumentHeadersForAmazonDetail(amazonAccountDetail)) {
+            try {
+                DistributionOfIncomeAndExpenseDocument di = (DistributionOfIncomeAndExpenseDocument) getDocumentService().getByDocumentHeaderId(dh.getDocumentNumber());
+                if (di.getTotalDollarAmount().equals(cost)) {
```
bhutchinson: This is only checking the cost, do you need to check the period as well? What if the cost is the same two months in a row?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+                LOG.error("shouldBuildDistributionOfIncomeDocument() There was a problems creating a DI document", e); 
+                throw new RuntimeException(e);
+            }
+        }
+        return true;
+    }
+    
+    private Collection<DocumentHeader> findDocumentHeadersForAmazonDetail(AmazonAccountDetail amazonAccountDetail) {
+        Map fieldValues = new HashMap();
+        fieldValues.put("EXPLANATION", buildDocumentExplanation(amazonAccountDetail.getAwsAccount()));
```
bhutchinson: Do you need to have period in the criteria to only find documents created for the same period you're trying to create the DI for now?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,446 @@
...
+        line.setAmount(amount);
+        line.setFinancialDocumentLineDescription(buildAccountingLineDescription());
+        line.setSequenceNumber(new Integer(1));
+        return line;
+    }
+    
+    private String buildAccountingLineDescription() {
+        return "AWS CHARGES " + findBillingMonthName() + " " + findBillingYear();
+    }
+
```
bhutchinson: a bit of extra whitespace here

bhutchinson:
In general looks good, just had a few comments/questions.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
jhulslander: I changed the name to the case you mentioned, but I thought including json in the function name seemed appropriate as later one we run buildAmazonAcountListFromJson and that uses json field mappings to do the conversion.  I am not married to including json in the function name, but seems more helpful to me.

bhutchinson: Ok, you've convinced me, thanks.

jhulslander:
I made the suggested changes and fixed a few bugs. This is read for review again.  Nicole is in the process of validating some files I sent her, so this is NOT ready to be merged yet.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,464 @@
...
+    }
+    
+    private ClientRequest buildClientRequest() {
+        ClientRequest.Builder builder = new ClientRequest.Builder();
+        builder.accept(MediaType.APPLICATION_JSON);
+        builder.header(CuFPConstants.AmazonWebServiceBillingConstants.AUTHORIZATION_HEADER_NAME, 
+                CuFPConstants.AmazonWebServiceBillingConstants.AUTHORIZATION_TOKEN_VALUE_STARTER + getAwsToken());
+        URI uri;
+        try {
+            final String yearUrlParam = "year=";
```
bhutchinson: possible constants I missed last time

bhutchinson:
Changes look good. Commented on a couple more possible comments I missed last time. Also, would it be possible to write a test for AmazonWebServicesBillingService? I know that it would be tricky to make that a microservice, but probably not impossible.

jhulslander:
retest this please

jhulslander:
This is ready for review again, I have completed local verification, so this is ready to be merged after a review, many changes this time.

jhulslander:
Shawn had some additional thoughts I need to review with Nicole, please review but hold off on merging

src/main/java/edu/cornell/kfs/fp/businessobject/TransactionDTO.java
```diff
@@ -0,0 +1,106 @@
+package edu.cornell.kfs.fp.businessobject;
+
+import java.io.Serializable;
+
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+public class TransactionDTO implements Serializable {
```
bhutchinson: This is kind of a generic name - maybe you can come up with a more specific one that indicates exactly what this is?

src/main/java/edu/cornell/kfs/fp/businessobject/TransactionDTO.java
```diff
@@ -0,0 +1,106 @@
...
+    private String objectCodeNumber;
+    private String subObjectCodeNumber;
+    private String projectCodeNumber;
+    private String organizationReferenceId;
+    private KualiDecimal amount;
+    private String lineDescription;
+    private boolean failingError;
+    
+    public TransactionDTO() {
+        failingError = false;
```
bhutchinson: This will default to false, I don't think you need the constructor.

src/main/java/edu/cornell/kfs/fp/businessobject/TransactionDTO.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public String getLineDescription() {
+        return lineDescription;
+    }
+
+    public void setLineDescription(String lineDescription) {
+        this.lineDescription = lineDescription;
+    }
+
+    public boolean isFailingError() {
```
bhutchinson: It's not clear to me what this is based on it's name. Maybe this is another place to try improving the naming?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+
+    protected boolean processAmazonAccountDetails(List<AmazonAccountDetail> amazonAccountDetails) {
+        boolean valid = true;
+        int count = 0;
+        for (AmazonAccountDetail amazonAccountDetail : amazonAccountDetails) {
+            if (StringUtils.equalsIgnoreCase(CuFPConstants.AmazonWebServiceBillingConstants.INTERNAL_KFS_ACCOUNT_DESCRIPTION, amazonAccountDetail.getKfsAccount())) {
+                LOG.info("processAmazonAccountDetails() Found an internal kfs account, ignoring this transaction for AWS account " 
+                    + amazonAccountDetail.getAwsAccount());
+            } else {
+                valid = processExternalBill(amazonAccountDetail) && valid;
```
bhutchinson: Maybe this is nitpicky, but I think the normal idiom is valid = valid && newCondition.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+        boolean valid = true;
+        try {
+            TransactionDTO transactionDTO = buildTransactionDTO(amazonAccountDetail);
+            if (transactionDTO.isFailingError()) {
+                valid = false;
+                LOG.error("processAmazonAccountDetails() there errors that prevent a creation of a DI " + amazonAccountDetail.toString());
+            } else if (shouldBuildDistributionOfIncomeDocument(amazonAccountDetail, transactionDTO.getAmount())) {
+                buidAndRouteDistributionOfIncomeDocument(amazonAccountDetail, transactionDTO);
+            }
+        } catch (Throwable e) {
```
bhutchinson: We typically don't want to catch Throwable, but a type of Exception instead; this should probably be changed to catch what is actually being thrown.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+        TransactionDTO transactionDTO = new TransactionDTO();
+        transactionDTO.setLineDescription(buildAccountingLineDescription());
+        
+        transactionDTO.setAmount(convertCostStringToKualiDecimal(amazonAccountDetail.getCost()));
+        if (transactionDTO.getAmount().isLessEqual(KualiDecimal.ZERO)) {
+            transactionDTO.setFailingError(true);
+        }
+        
+        if (StringUtils.contains(amazonAccountDetail.getCostCenter(), KFSConstants.DASH)) {
+            //could be account & subaccount or whole transaction string
```
bhutchinson: Not sure this comment is helpful. I'd lean towards removing it.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+            String[] costCenterArray = StringUtils.split(amazonAccountDetail.getCostCenter(), KFSConstants.DASH);
+            if (costCenterArray.length == 2) {
+                updateTransactionDTOAccountSubAccount(amazonAccountDetail, transactionDTO);
```
bhutchinson: I wonder if there is a way to make this behavior polymorphic and eliminate the if/else if. Maybe there could be sub-classes of the TransactionDTO that implement their own version of updateTransactionDTOAccountString or something?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+            }
+        } else {
+            updateTransactionDTOAccountNumber(amazonAccountDetail, transactionDTO);
+        }
+        
+        return transactionDTO;
+    }
+    
+    protected KualiDecimal convertCostStringToKualiDecimal(String costString) {
+        KualiDecimal cost = KualiDecimal.ZERO;
```
bhutchinson: There's no code that already does this elsewhere in the system? Seems pretty generic.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+        try {
+            double costDouble = Double.valueOf(costString);
+            final double oneCent = .01;
+            if (costDouble >= oneCent) {
+                cost = new KualiDecimal(costDouble);
+            }
+        } catch (Throwable e) {
```
bhutchinson: Another place that should be catching a specific exception instead of Throwable.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+            sb.append(" costCenterSubObjectCode ").append(KRADConstants.SINGLE_QUOTE).append(costCenterSubObjectCode).append(KRADConstants.SINGLE_QUOTE);
+            sb.append(" costCenterProjectCode ").append(KRADConstants.SINGLE_QUOTE).append(costCenterProjectCode).append(KRADConstants.SINGLE_QUOTE);
+            sb.append(" costCenterObjectRefId ").append(KRADConstants.SINGLE_QUOTE).append(costCenterObjectRefId).append(KRADConstants.SINGLE_QUOTE);
+            LOG.debug(sb.toString());
+        }
+    }
+
+    private void updateTransactionDTOAccountNumber(AmazonAccountDetail amazonAccountDetail, TransactionDTO transactionDTO) {
+        Account account = determineWhichAccountToUse(amazonAccountDetail.getCostCenter(), amazonAccountDetail.getKfsAccount());
+        if (account != null) {
```
bhutchinson: I would probably use ObjectUtils.isNotNull here.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+            updateTransactionDTOChartAccountObjectCodes(transactionDTO, account);
+        } else {
+            transactionDTO.setFailingError(true);
+        }
+    }
+    
+    private void updateTransactionDTOAccountNumbers(AmazonAccountDetail amazonAccountDetail, TransactionDTO transactionDTO, String costCenterAccountNumber,
+            String costCenterSubAccountNumber) {
+        Account account = determineWhichAccountToUse(costCenterAccountNumber, amazonAccountDetail.getKfsAccount());
+        if (account != null) {
```
bhutchinson: A couple more places I would probably use ObjectUtils.isNotNull.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+    private Account determineWhichAccountToUse(String costCenter, String kfsAccountString) {
+        Account costAccount = buildAccount(costCenter);
+        if (validAccount(costAccount)) {
+            return costAccount;
+        }
+        Account kfsAccount = buildAccount(kfsAccountString);
+        if (validAccount(kfsAccount)) {
+            return kfsAccount;
+        }
+        return null;
```
bhutchinson: I wonder if you could avoid returning null by just returning a new Account, or something like that.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+            }
+        }
+    }
+    
+    private Account buildAccount(String accountNumber) {
+        if (StringUtils.isNotBlank(accountNumber)) {
+            try {
+                Account account = getAccountService().getUniqueAccountForAccountNumber(accountNumber);
+                return account;
+            } catch (Exception e) {
```
bhutchinson: Does the method above throw something more specific than Exception? If so, the catch should have that specific exception. If not, bummer.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+                LOG.error("shouldBuildDistributionOfIncomeDocument() There was a problems creating a DI document from a document header: " + dh.getDocumentNumber(), e); 
+                throw new RuntimeException(e);
+            }
+        }
+        return true;
+    }
+    
+    private Collection<DocumentHeader> findDocumentHeadersForAmazonDetail(AmazonAccountDetail amazonAccountDetail) {
+        Map fieldValues = new HashMap();
+        fieldValues.put("EXPLANATION", buildDocumentExplanation(amazonAccountDetail.getAwsAccount()));
```
bhutchinson: A couple Constant candidates here.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,649 @@
...
+        line.setChartOfAccountsCode(fromAccount.getChartOfAccountsCode());
+        line.setFinancialObjectCode(getTransactionObjectCode());
+        line.setAmount(amount);
+        line.setFinancialDocumentLineDescription(buildAccountingLineDescription());
+        line.setSequenceNumber(new Integer(1));
+        return line;
+    }
+    
+    protected String buildAccountingLineDescription() {
+        return "AWS CHARGES " + findMonthName() + " " + findProcessYear();
```
bhutchinson: another possible constant

bhutchinson:
I've reviewed the latest and added a few more comments. You'll also need to squash your commits before this PR can be merged.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
jhulslander: I always want to run processExternalBill() even if vaid is already false, so I think the reverse makes sense here

jhulslander:
I believe I have adjusted the code per the last code review and is ready for review again.  @db483 I assigned to you to keep this going

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
jhulslander: I wanted to make sure that any error happening in be transaction doesn't prevent subsequent ones

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService() Number of amazon account details received: " + amazonAccountDetails.size());
+        
+        boolean valid = true;
+        
+        if (amazonAccountDetails.size() == 0) {
+            LOG.error("generateDistributionOfIncomeDocumentsFromAWSService() No amazon account detail objects parsed from the webservice.");
+            valid =  false;
+        }
+        
+        valid = valid && processAmazonAccountDetails(amazonAccountDetails);
```
db483: do you still want to do the processing here if valid is false?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService(): AWS URL: " + getAwsURL());
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService(): Processing Year: " + findProcessYear() + "  Month: " + findProcessMonthNumber());
+        
+        String jsonResults = buildJsonOutput();
+        LOG.debug("JSON results: " + jsonResults);
+
+        List<AmazonAccountDetail> amazonAccountDetails = buildAmazonAcountListFromJson(jsonResults);
+        LOG.info("generateDistributionOfIncomeDocumentsFromAWSService() Number of amazon account details received: " + amazonAccountDetails.size());
+        
+        boolean valid = true;
```
db483: I think a more descriptive name for this boolean would be better

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        try {
+            detailContainer = mapper.readValue(jsonResults, AmazonAccountDetailContainer.class);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return detailContainer.getAccountDetail();
+    }
+
+    protected boolean processAmazonAccountDetails(List<AmazonAccountDetail> amazonAccountDetails) {
+        boolean valid = true;
```
db483: Since this method is not doing validation could there be a better name for this variable?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
db483: Just checking, have you looked into Bryan's question?

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
db483: Seems that this concern hasn't been addressed.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+            transactionDTO.setAccountNumber(account.getAccountNumber());
+            updateTansactionDTOChart(transactionDTO, costCenterDTO, account);
+            updateTransactionDTOSubAccount(transactionDTO, costCenterDTO, account);
+        } else {
+            LOG.info("updateTransactionDTOChartAccountSubAccount() Invalid account.");
+            transactionDTO.setTransactionInputError(true);
+        }
+    }
+    
+    private Account buildAccount(String accountNumber) {
```
db483: This method name is rather misleading since it is only retrieving an account rather than building one based on the input data

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        } else {
+            costCenterDTO.setAccountNumber(costCenter);
+        }
+        
+        return costCenterDTO;
+    }
+    
+    private void updateTransactionDTOChartAccountSubAccount(AmazonBillingDistributionOfIncomeTransactionDTO transactionDTO, AmazonBillingCostCenterDTO costCenterDTO,
+            String amazonDetailKFSAccount) {
+        Account account;
```
db483: seems like the first part of this method could be extracted into a retrieve account method

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        if (StringUtils.isNotBlank(costCenterDTO.getAccountNumber())) {
+            account = buildAccount(costCenterDTO.getAccountNumber());
+        } else {
+            account = buildAccount(amazonDetailKFSAccount);
+        }
+        if (validAccount(account)) {
+            transactionDTO.setAccountNumber(account.getAccountNumber());
```
db483: probably the second part could be extracted into an update method

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+            if (ObjectUtils.isNotNull(subAccount)) {
+                transactionDTO.setSubAccountNumber(subAccount.getSubAccountNumber());
+            } else {
+                LOG.info("updateTransactionDTOSubAccount() Invalid Sub Account.");
+                transactionDTO.setTransactionInputError(true);
+            }
+        }
+    }
+    
+    private SubAccount buildSubAccount(Account account, String subAccountNumber) {
```
db483: this method again looks like it is only retrieving the sub account rather then building it from input data, probably would be good to rename the method

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+    private Collection<DocumentHeader> findDocumentHeadersForAmazonDetail(AmazonAccountDetail amazonAccountDetail) {
+        Map fieldValues = new HashMap();
+        fieldValues.put(CUKFSConstants.DOCUMENT_HEADER_EXPLANATION_FIELD, buildDocumentExplanation(amazonAccountDetail.getAwsAccount()));
+        fieldValues.put(CUKFSConstants.DOCUMENT_HEADER_DESCRIPTION_FIELD, buildDocumentDescription(amazonAccountDetail.getBusinessPurpose()));
+        Collection<DocumentHeader> documentHeaders = getBusinessObjectService().findMatching(DocumentHeader.class, fieldValues);
+        return documentHeaders;
+    }
+    
+    private void buidAndRouteDistributionOfIncomeDocument(AmazonAccountDetail amazonAccountDetail, AmazonBillingDistributionOfIncomeTransactionDTO transactionDTO) throws WorkflowException {
+        CuDistributionOfIncomeAndExpenseDocument diDocument = (CuDistributionOfIncomeAndExpenseDocument) getDocumentService().getNewDocument(
```
db483: could  extract the first part of this method into a method that creates and populates the DI doc

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
```diff
@@ -0,0 +1,636 @@
...
+                CuDistributionOfIncomeAndExpenseDocument.class);
+        
+        diDocument.getDocumentHeader().setDocumentDescription(buildDocumentDescription(amazonAccountDetail.getBusinessPurpose()));
+        diDocument.getDocumentHeader().setExplanation(buildDocumentExplanation(amazonAccountDetail.getAwsAccount()));
+        diDocument.getNotes().add(buildDINote());
+        diDocument.getTargetAccountingLines().add((buildToAccountingLine(diDocument.getDocumentNumber(), transactionDTO)));
+        diDocument.getSourceAccountingLines().add(buildFromAccountLine(diDocument.getDocumentNumber(), transactionDTO.getAmount()));
+        diDocument.setTripAssociationStatusCode(CULegacyTravelServiceImpl.TRIP_ASSOCIATIONS.IS_NOT_TRIP_DOC);;
+        
+        getDocumentService().saveDocument(diDocument);
```
db483: is there a need to save the DI doc first? I am not familiar with this doc

db483:
I have done a first round of review and added my comments.

src/main/java/edu/cornell/kfs/fp/service/impl/AmazonWebServicesBillingServiceImpl.java
jhulslander: Valid is only false when there aren't any records in amazonAccountDetails, so there are no records to process anyway

jhulslander:
Thanks, I adjusted the code per code review, and is ready for review again.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -135,5 +135,26 @@ public static ScheduleTypes fromName(String name) {
...
+    
+    public static class AmazonWebServiceBillingConstants {
+        public static final String AWS_PROCESSING_DATE_PROPERTY_NAME = "AWS_PROCESSING_DATE";
+        public static final String AWS_FROM_ACCOUNT_PROPERTY_NAME = "AWS_FROM_ACCOUNT_NUMBER";
+        public static final String AWS_OBJECT_CODE_PROPERTY_NAME = "AWS_OBJECT_CODE";
+        public static final String AWS_DEFAULT_DOCUMENT_DESCRIPTION_PROPERTY_NAME = "AWS_DEFAULT_DOCUMENT_DESCRIPTION";
+        public static final String AWS_NOTE_HELP_TEXT_PROERTY_NAME = "AWS_NOTE_HELP_TEXT";
+        public static final String AWS_COMPENT_NAME = "LoadAWSBillsStep";
+        
+        public static final int PROCESSING_DAY_OF_MONTH = 15;
```
db483: nothing wrong with the code here but just wondering if this would have been better off in a system parameter

db483:
Code looks good. I had a question about whether a constant would be better off as a system parameter. I will merge the code to develop so that testing can begin for this change and if it turns out you need a system param you can update the code later.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
jhulslander: This value is used to create a calendar object, and the day of month actually isn't used.  Since this will be run starting on the 15th of a month, 15 seemed like a reasonable number to go with.

db483: thanks for explaining this Jay

PR #192, opened 2016-08-04
KFSPTS-5907: fix NPE with Account Global Lookup

bhutchinson:


nkimble:
Code looks ok. Merging.

PR #194, opened 2016-08-10
KFSPTS-3599 add missing error messages

db483:
Hi Bryan,

I compared the develop and master files and found 3 error messages missing. They were all added on this commit.


src/main/resources/CU-ApplicationResources.properties
```diff
@@ -3,6 +3,8 @@ message.pdp.paymentLoad.invalidSubObject=Sub Object code {0}-{1}-{2}-{3} is inva
 warning.dv.payee.mustBeActive = Only active employees may be paid on a Disbursement Voucher.
 
 error.document.accountMaintenance.programCodeNotAssociatedWithGroupCode=Sub-Fund Program Code {0} is not associated with Sub-Fund Group Code {1}.
+error.document.accountMaintenance.acct.programCodeNotAssociatedWithGroupCode=Sub-Fund Program Code {0} is not associated with Sub-Fund Group Code {1} for Account {2}.
```
db483: These keys should have pointed to account global maintenance. I think we had others like this that were cleaned during code review but these were missed as they were not part of the commit.

bhutchinson: Yeah, maybe we can clean them up at some point in the future.

PR #196, opened 2016-08-10
KFSPTS-4095: Added Payee ACH Account Extract job to read in ACH data from .csv files delivered from Workday.

nkimble:
I conferred with functional user. Bank routing number and bank account number for ACH should be validated as being numeric.

Changes have been rebased to original modification and are ready for final code review and merge.


nkimble:
retest this please

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,702 @@
...
+        // For now, return true even if files or rows did not load successfully. Functionals will address the failed rows/files accordingly.
+        return true;
+    }
+
+    /**
+     * Create a collection of the files to process with the mapped value of the BatchInputFileType
+     */
+    // This is a slightly-tweaked copy of a method from CustomerLoadServiceImpl.
+    protected Map<String,BatchInputFileType> getListOfFilesToProcess() {
+        Map<String,BatchInputFileType> inputFileTypeMap = new LinkedHashMap<String, BatchInputFileType>();
```
cah292: Not sure if this is an issue, but the calling code is iterating over this Map's keys and also calling its get() method within the loop (and LinkedHashMap is modified structurally when get() is called). There are numerous other areas in KFS where this iteration is used, including the CustomerLoadServiceImpl processing that I had originally based portions of this service from, so I'm fine with it going out as-is.

If problems do arise from potential structural modification while iterating, we can address that in a separate user story so that we can update all of the affected services at once.

cah292:
Changes appear to be good, and unit tests have passed. There's one minor area of concern about a particular Map implementation usage, but that's part of a larger KFS issue and I'm fine with deferring it to another user story.

Merging.

PR #199, opened 2016-08-17
KFSPTS-3599 Add new fields to Account Global document.

db483:


db483:
Hi Chad,

The main changes on this new commit are to:
- add transient to  GlobalObjectWithIndirectCostRecoveryAccountsService on SubAccountGlobal 
- properly add some error messages that before were not part of the commit related on this jira. This error messages existed in develop due to some erroneous reverts. However they did not exist in master causing an issue when code was moved to test.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -2,164 +2,903 @@
...
     private static final long serialVersionUID = 1L;
 
-    protected String majorReportingCategoryCode;
-    protected MajorReportingCategory majorReportingCategory;
-    
-    @Override
-    public List<PersistableBusinessObject> generateGlobalChangesToPersist() {
+    protected DateTimeService dateTimeService;
+    protected BusinessObjectService businessObjectService;
+    protected transient GlobalObjectWithIndirectCostRecoveryAccountsService globalObjectWithIndirectCostRecoveryAccountsService;
```
cah292: Is there a reason why the third service is transient but the others are not? Should they all be transient?

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -2,164 +2,903 @@
...
+    protected boolean removeContinuationChartAndAccount;
+    protected String financialIcrSeriesIdentifier;
+    protected Boolean everify;
+    protected boolean removeIncomeStreamChartAndAccount;  
     
-                // LABOR BENEFIT RATE CATEGORY CODE
-                if (StringUtils.isNotBlank(getLaborBenefitRateCategoryCode())) {
-                    account.setLaborBenefitRateCategoryCode(getLaborBenefitRateCategoryCode());
-                }
+	protected MajorReportingCategory majorReportingCategory;
```
cah292: Small indentation issue on this line; not worth fixing unless other changes are needed.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -2,164 +2,903 @@
...
+	 */
+	private void updateAccountRestrictedStatusCodeForAccountBeingEdited(SubFundGroup globalAccountMaintenanceDocSubFundGroup, String globalAccountMaintenanceDocRestrictedStatusCode, Account accountBeingEdited)
+	{
+	    if (ObjectUtils.isNull(globalAccountMaintenanceDocSubFundGroup)) {
+	        if ( !isDefaultAccountRestrictedStatusCodeSetOnSubFundGroup(accountBeingEdited.getSubFundGroup()) ) {
+	            accountBeingEdited.setAccountRestrictedStatusCode(globalAccountMaintenanceDocRestrictedStatusCode);
+	        }
+	    }
+	    else {
+	        if ( !isDefaultAccountRestrictedStatusCodeSetOnSubFundGroup(globalAccountMaintenanceDocSubFundGroup) ) {
```
cah292: In this case, you could rewrite it as an "else if" instead of an "else" with a nested "if".

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -2,164 +2,903 @@
...
+	/**
+	 *
+	 * @param subFundGroup
+	 * @return true when a default account restricted status code set on the sub-fund; false otherwise
+	 */
+	private boolean isDefaultAccountRestrictedStatusCodeSetOnSubFundGroup(SubFundGroup subFundGroup) {
+		if (ObjectUtils.isNotNull(subFundGroup) && StringUtils.isNotBlank(subFundGroup.getAccountRestrictedStatusCode())) {
+			return true;
+		}
+		return false;
```
cah292: It would probably be simpler and cleaner to just return the result of the condition instead.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
db483: good question, I was wondering the same. The third one is the only one that caused an exception, see https://jira.cornell.edu/browse/KFSPTS-5794. On save/submit we got an exception: Caused by: java.io.NotSerializableException: edu.cornell.kfs.coa.service.impl.GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl

No exception however for the other 2

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,194 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
-		icrAccount.setActive(newICR.isActive());
-		icrAccount.setAccountLinePercent(newICR.getAccountLinePercent());
 		
-		return icrAccount;
+		return  currentActiveIndirectCostRecoveryAccountList;
+    }
+
+	private boolean noMatchFoundOrSameActiveIndicatorMatchFoundAndStillHaveICRAccountsToCheck(int currentPosition, int maxPosition,
+			boolean foundMatch, IndirectCostRecoveryAccount matchingICRAccountWithSameActiveIndicator) {
+		return currentPosition < maxPosition && (!foundMatch || (foundMatch && ObjectUtils.isNotNull(matchingICRAccountWithSameActiveIndicator)));
```
cah292: This condition is currently written as "x && (!y || (y && z))". I would suggest rewriting it as "x && (!y || z)", which is simpler and should still yield the same result.

src/main/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+    
+	public void updateIcrAccounts(
+			GlobalObjectWithIndirectCostRecoveryAccounts globalObjectWithIndirectCostRecoveryAccounts,
+			GlobalBusinessObjectDetailBase globalDetail,
+			List<IndirectCostRecoveryAccount> icrAccounts) {
+    	List<IndirectCostRecoveryAccount> updatedIcrAccounts = buildUpdatedIcrAccounts(globalObjectWithIndirectCostRecoveryAccounts, globalDetail, icrAccounts);
+    	globalObjectWithIndirectCostRecoveryAccounts.updateGlobalDetailICRAccountCollection(globalDetail, updatedIcrAccounts);
+    }
+    
+	public List<IndirectCostRecoveryAccount> buildUpdatedIcrAccounts(
```
cah292: The search logic in this method is similar to that from the "buildCurrentActiveICRAccountsAfterICRUpdates" method in this PR's GlobalIndirectCostRecoveryAccountsRule class. Is there some way that the search logic could be consolidated into one common spot, or are there enough differences between the two to make that impractical?

src/main/java/edu/cornell/kfs/sys/CUKFSKeyConstants.java
```diff
@@ -175,4 +176,36 @@
...
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_ICR_NOT_EMPTY_FOR_NON_CG_ACCOUNT = "error.document.accountGlobal.acct.cgICRNotEmptyForNonCGAccount";
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_EXP_DT_AND_REMOVE_CONT_ACCT = "error.document.accountGlobal.acct.expirationDate.removeContAcct";
+
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_REMOVE_INC_STR_CHART_AND_ACCT_CHECKED_WHEN_INC_STR_REQ = "error.document.accountGlobalMaintenance.removeIncStrChartAndAcctChecked.incStrReq";
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_REMOVE_INC_STR_CHART_AND_ACCT_CHECKED_WHEN_INC_STR_REQ_FOR_ACCT = "error.document.accountGlobalMaintenance.removeIncStrChartAndAcctChecked.incStrReqForAcct";
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_CLOSED_CHECKED_WHEN_ACCOUNT_HAS_OPEN_ENCUMBRENCES = "error.document.accountGlobalMaintenance.accountCannotCloseOpenEncumbrance";
+    public static final String ERROR_DOCUMENT_ACCT_GLB_MAINT_CANNOT_BE_CLOSED_EXP_DATE_NOT_ENTERED_EXISTING_DATE_INVALID = "error.document.accountGlobalMaintenance.accountCannotBeClosedExpDateNotExnteredExistingDateInvalid";
```
cah292: Small typo in variable value, should be "...NotEntered..." instead of "...NotExntered..."

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -479,3 +480,31 @@ message.dv.payee.invalid.payment.type=The Payee didn't have a valid Payee Type f
...
+error.document.accountGlobal.acct.cgICRNotEmptyForNonCGAccount=The Sub-Fund Group {0} for account {1} is not associated with Contracts and Grants; the Indirect Cost Recovery Accounts collection should be empty.
+error.document.accountGlobal.acct.expirationDate.removeContAcct=Remove Continuation Account cannot be checked. Account {0} {1} has an expiration date.
+error.document.accountGlobalMaintenance.removeIncStrChartAndAcctChecked.incStrReq=Remove Income Stream Chart and Account cannot be checked when Income stream Chart and Account is required.
+error.document.accountGlobalMaintenance.removeIncStrChartAndAcctChecked.incStrReqForAcct=Remove Income Stream Chart and Account cannot be checked when Income stream Chart and Account is required for Account {0} {1}.
+error.document.accountGlobalMaintenance.accountCannotCloseOpenEncumbrance = Account {0}-{1} cannot be closed because it has an open Encumbrance.
+error.document.accountGlobalMaintenance.accountCannotBeClosedExpDateNotExnteredExistingDateInvalid=Account {0}-{1} cannot be closed. It has an existing Expiration Date in the future and you have not entered an Expiration Date.
```
cah292: Small typo in key, should be "...NotEntered..." instead of "...NotExntered..."

src/test/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRuleTest.java
```diff
@@ -0,0 +1,281 @@
...
+	private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger
+			.getLogger(GlobalIndirectCostRecoveryAccountsRuleTest.class);
+
+	@Before
+	public void setUp() {
+		globalIndirectCostRecoveryAccountsRule = new GlobalIndirectCostRecoveryAccountsRule();
+	}
+
+	@Test
+	public void testAccountGlobalCheckICRAccuntTotalDistributionOnDetailWillBe100PercentAfterUpdate_Pass() {
```
cah292: I would suggest updating the test method names to use "...ICRAccount..." instead of "...ICRAccunt..."

src/test/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRuleTest.java
```diff
@@ -0,0 +1,281 @@
...
+		LOG.debug("updates");
+		List<IndirectCostRecoveryAccountChange> updates = accountGlobal
+				.getIndirectCostRecoveryAccounts();
+		for (IndirectCostRecoveryAccountChange icrChange : updates) {
+			LOG.debug("Chart " + icrChange.getIndirectCostRecoveryFinCoaCode()
+					+ " account "
+					+ icrChange.getIndirectCostRecoveryAccountNumber()
+					+ " percent " + icrChange.getAccountLinePercent()
+					+ " active " + icrChange.isActive());
+		}
```
cah292: There seems to be a lot of duplication in these test methods, especially for these debugging loops. I would recommend creating some new methods to consolidate the repeating parts where it makes sense.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/GlobalIndirectCostRecoveryAccountsRule.java
```diff
@@ -183,208 +169,194 @@ protected boolean checkIndirectCostRecoveryAccount(IndirectCostRecoveryAccountCh
...
-				IndirectCostRecoveryAccount icrAcct = new IndirectCostRecoveryAccount();
-				icrAcct.setChartOfAccountsCode(a21ICRAccount.getChartOfAccountsCode());
-				icrAcct.setAccountNumber(a21ICRAccount.getAccountNumber());
-				icrAcct.setIndirectCostRecoveryAccountNumber(a21ICRAccount.getIndirectCostRecoveryAccountNumber());
-				icrAcct.setIndirectCostRecoveryFinCoaCode(a21ICRAccount.getIndirectCostRecoveryFinCoaCode());
-				icrAcct.setAccountLinePercent(a21ICRAccount.getAccountLinePercent());
-				icrAcct.setActive(a21ICRAccount.isActive());
-				existingIcrAccounts.add(icrAcct);
-			}
+	protected boolean checkICRAccuntTotalDistributionOnDetailWillBe100PercentAfterUpdate(List<IndirectCostRecoveryAccountChange> icrUpdates, List<IndirectCostRecoveryAccount> existingICRs, GlobalBusinessObjectDetailBase globalDetail, GlobalObjectWithIndirectCostRecoveryAccounts globalObjectWithIndirectCostRecoveryAccounts) {
```
cah292: I recommend changing the method name to use "...ICRAccount..." instead of "...ICRAccunt..."

src/test/java/edu/cornell/kfs/coa/fixture/A21SubAccountFixture.java
```diff
@@ -0,0 +1,65 @@
...
+			A21IndirectCostRecoveryAccountFixture.A21_ICR_2222222_2_PERCENT_INACTIVE
+					.getA21IndirectCostRecoveryAccountChange());
+
+	public A21IndirectCostRecoveryAccount icrAccount1;
+	public A21IndirectCostRecoveryAccount icrAccount2;
+	public A21IndirectCostRecoveryAccount icrAccount3;
+
+	private A21SubAccountFixture(A21IndirectCostRecoveryAccount icrAccount1,
+			A21IndirectCostRecoveryAccount icrAccount2,
+			A21IndirectCostRecoveryAccount icrAccount3) {
```
cah292: I would suggest changing this fixture to instead use an array of ICR accounts, and to update the constructor to use a var-args parameter instead of three individual ones. This will greatly simplify the setup process.

src/test/java/edu/cornell/kfs/coa/fixture/AccountFixture.java
```diff
@@ -0,0 +1,64 @@
...
+			IndirectCostRecoveryAccountFixture.ICR_2222222_2_PERCENT_INACTIVE
+					.getIndirectCostRecoveryAccountChange());
+
+	public IndirectCostRecoveryAccount icrAccount1;
+	public IndirectCostRecoveryAccount icrAccount2;
+	public IndirectCostRecoveryAccount icrAccount3;
+
+	private AccountFixture(IndirectCostRecoveryAccount icrAccount1,
+			IndirectCostRecoveryAccount icrAccount2,
+			IndirectCostRecoveryAccount icrAccount3) {
```
cah292: Another fixture that could use an ICR Account array and a var-args constructor parameter.

src/test/java/edu/cornell/kfs/coa/fixture/A21IndirectCostRecoveryAccountFixture.java
```diff
@@ -0,0 +1,40 @@
...
+	A21_ICR_1111111_98_PERCENT_INACTIVE("IT", "1111111",  new BigDecimal(98), false),
+	A21_ICR_2222222_2_PERCENT_ACTIVE("IT", "2222222",  new BigDecimal(2), true),
+	A21_ICR_2222222_2_PERCENT_INACTIVE("IT", "2222222",  new BigDecimal(2), false),
+	A21_ICR_3333333_100_PERCENT_ACTIVE("IT", "3333333",  new BigDecimal(100), true),
+	A21_ICR_3333333_100_PERCENT_INACTIVE("IT", "3333333",  new BigDecimal(100), false);
+	
+    public String indirectCostRecoveryFinCoaCode;
+    public String indirectCostRecoveryAccountNumber;
+    public BigDecimal accountLinePercent;
+    public boolean active;
```
cah292: In this fixture and all the other new fixtures being added, I would strongly recommend declaring these fields as "final", to prevent testing code from accidentally updating them.

src/test/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+import org.kuali.kfs.coa.businessobject.AccountGlobalDetail;
+import org.kuali.kfs.coa.businessobject.IndirectCostRecoveryAccount;
+
+import edu.cornell.kfs.coa.businessobject.CuAccountGlobal;
+import edu.cornell.kfs.coa.businessobject.GlobalObjectWithIndirectCostRecoveryAccounts;
+import edu.cornell.kfs.coa.fixture.AccountGlobalDetailFixture;
+import edu.cornell.kfs.coa.fixture.AccountGlobalFixture;
+import edu.cornell.kfs.coa.fixture.IndirectCostRecoveryAccountFixture;
+
+public class GlobalObjectWithIndirectCostRecoveryAccountsServiceImplTest {
```
cah292: There are some odd line breaks and indentation in this test class. I would suggest cleaning up the formatting to make it easier to read.

cah292:
I've finished my initial review. I did have some various suggestions for clean-up and improved unit test setup, but the changes seem good overall. Will wait for feedback before reviewing again.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
cah292: It must have been because the third service contained some non-serializable objects.

Since the getters for these services also appear to be auto-initializing the variables if needed, I would suggest declaring all 3 as transient so that we're not passing them around when serializing CuAccountGlobal.

src/main/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java
db483: The methods are similar but rather difficult to combine in common code.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
db483: I made the change, thank you

src/main/java/edu/cornell/kfs/coa/service/impl/GlobalObjectWithIndirectCostRecoveryAccountsServiceImpl.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		return updatedIcrAccounts;
+    }
+
+	private boolean noMatchFoundOrSameActiveIndicatorMatchFoundAndStillHaveICRAccountToCheck(int currentPosition, 
+			int maxPosition, 
+			boolean foundMatch, 
+			int positionForMatchWithSameActiveIndicator) {
+		return currentPosition < maxPosition && (!foundMatch || (foundMatch && positionForMatchWithSameActiveIndicator !=-1));
```
cah292: If you're not planning to consolidate those two ICR search pieces, then I would suggest updating this condition use the "x && (!y || z)" logic that I had mentioned for the other search.

db483: will do

db483:
I have updated the PR with the suggested changes.

src/test/java/edu/cornell/kfs/coa/fixture/IndirectCostRecoveryAccountChangeFixture.java
```diff
@@ -0,0 +1,57 @@
...
+	ICR_CHANGE_3333333_100_PERCENT_ACTIVE("IT", "3333333", new BigDecimal(100),
+			true),
+
+	ICR_CHANGE_3333333_100_PERCENT_INACTIVE("IT", "3333333",
+			new BigDecimal(100), false);
+
+	public String indirectCostRecoveryFinCoaCode;
+	public String indirectCostRecoveryAccountNumber;
+	public BigDecimal accountLinePercent;
+	public boolean active;
```
cah292: It looks like only some of the fixtures were updated to have final variables. There are a few (including this one) that should be declaring these as "final" to prevent accidental updates after construction.

src/test/java/edu/cornell/kfs/coa/fixture/A21SubAccountFixture.java
```diff
@@ -0,0 +1,56 @@
...
+
+	A21_SUB_ACCOUNT_3333333_100_INACTIVATE(
+			A21IndirectCostRecoveryAccountFixture.A21_ICR_3333333_100_PERCENT_ACTIVE
+					.getA21IndirectCostRecoveryAccountChange(),
+			A21IndirectCostRecoveryAccountFixture.A21_ICR_1111111_98_PERCENT_INACTIVE
+					.getA21IndirectCostRecoveryAccountChange(),
+			A21IndirectCostRecoveryAccountFixture.A21_ICR_2222222_2_PERCENT_INACTIVE
+					.getA21IndirectCostRecoveryAccountChange());
+
+	public final A21IndirectCostRecoveryAccount[] icrAccounts;
```
cah292: Just wanted to confirm if you saw my edits to my older comment regarding these. (Sorry if you didn't see those edits.) With regards to my edited older comment, declaring the array as final does not prevent the array contents from being updated. If you want to make this truly immutable, you should use an unmodifiable list or similar data structure, and I would suggest keeping the ICR accounts in their fixture form until the getA21SubAccount() method gets called. (You can still use a var-args parameter in the constructor and then have it wrap the values in a list.)

The same goes for the other fixtures that are set up similar to this one.

cah292:
Finished second round of review. The changes look good in general, I just had a few suggestions for further improvement of the unit tests. Will wait for more feedback.

Also, I'm noticing some areas in the new code that are using tabs instead of spaces. As per some earlier team discussions, I would suggest using spaces instead of tabs where possible (especially in brand new files) if it's not too much trouble to change.

db483:
Hi Chad,

I have updated the PR with the fixture related fixes. As discussed on Hipchat, will skip the tab/space issue for now.

Thank you.

cah292:
Changes look good, and unit tests have passed.

One other quick note, though. It appears that AccountGlobalRule had so many changes that it couldn't be displayed in this PR, so I had accidentally skipped over it in my earlier reviews. I examined the changes locally in Eclipse, and I did see some unneeded comments and an odd use of a static variable, but I didn't see any show-stoppers and I'm fine with it going forward as-is. However, I would suggest that the functionals thoroughly test out these business rules to make sure things are working as expected, due to the sheer number of changes to that one file.

Merging.

PR #200, opened 2016-08-17
KFSPTS-4095: Added Payee ACH Account Extract batch job to read in ACH data from .csv files delivered from Workday.

nkimble:
The main differences between this version of the code and the previous commit are:
a)  a single bad data line will no longer fail the entire data file
b) expanded on report / log output to indicate number of failure rows per file
c) indicate in report / log file data rows detected as "update"  for which functionality is not yet implemented.


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+
+import edu.cornell.kfs.pdp.CUPdpConstants;
+import edu.cornell.kfs.pdp.CUPdpParameterConstants;
+import edu.cornell.kfs.pdp.batch.PayeeACHAccountExtractStep;
+import edu.cornell.kfs.pdp.batch.service.PayeeACHAccountExtractService;
+import edu.cornell.kfs.pdp.businessobject.PayeeACHAccountExtractDetail;
+
+public class PayeeACHAccountExtractServiceImpl implements PayeeACHAccountExtractService {
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(PayeeACHAccountExtractServiceImpl.class);
+    private static final String NEW_LINE = "\n";
```
db483: A constant for new line already exists in KFSConstants.java file.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+    @Override
+    public boolean processACHBatchDetails() {
+        LOG.info("Beginning processing of ACH input files");
+        
+        int numSuccess = 0;
+        int numPartial = 0;
+        int numFail = 0;
+        
+
+        // create a list of the files to process.
```
db483: this comment does not add much value

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+        Map<String,BatchInputFileType> fileNamesToLoad = getListOfFilesToProcess();
+        LOG.info("Found " + fileNamesToLoad.size() + " file(s) to process.");
+        
+        // process each file in turn
```
db483: this comment does not add much value

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+        PayeeACHAccount employeeAccount = achService.getAchInformation(
+                PayeeIdTypeCodes.EMPLOYEE, payee.getEmployeeId(), getDirectDepositTransactionType());
+        
+        String processingResults = KFSConstants.EMPTY_STRING;
+        // Add or update Entity ID account.
+        if (ObjectUtils.isNull(entityAccount)) {
+            processingResults = addACHAccount(payee, achDetail, PayeeIdTypeCodes.ENTITY);
+            LOG.info("Created new ACH Account of Entity type for payee " + payee.getPrincipalName());
+        } else {
+            processingResults = updateACHAccountIfNecessary(payee, achDetail, entityAccount);
```
db483: I think it would be cleaner to log the info about the processing in the add and update ACHAccount methods instead of returning a String to process and log here. 
I would change the add and update methods to either return a boolean or thow and exception that you catch here and log accordingly. Probably returning a boolean and logging the info about the processed data in the add/update methods would be better.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+                             .append(" of type '")
+                             .append(achAccount.getPayeeIdentifierTypeCode())
+                             .append("' matches what is already in KFS; no updates will be made for this entry.");
+        }
+        processingResults.append(" Update was not performed.");
+        return processingResults.toString();
+    }
+
+    private String getFailRequestMessage(Exception e) {
+        if (e instanceof ValidationException) {
```
db483: are you getting these validation exceptions when trying to route the document?
I would log the errors here instead of returning them.

db483:
Nancy I have added my comments. You might want to wait for Chad's input before updating the code.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
db483: I wonder if it might be better that you validate the doc first and only route if validation passes

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: I agree, that sounds like a good idea.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
nkimble: Changing to use KFSConstants and removing local one.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
nkimble: removing comment

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: The issue with a validate-first approach is that the validation could end up being executed twice (once manually and once for the routing), unless we tried to use a non-validating route-document method (which may not be desirable). Not sure how much that would slow things down.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
nkimble: removing comment.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,765 @@
...
+            PayeeACHAccount achAccount = (PayeeACHAccount) maintainable.getDataObject();
+            
+            // Setup payee ID and type.
+            if (PayeeIdTypeCodes.ENTITY.equals(payeeType)) {
+                achAccount.setPayeeIdNumber(payee.getEntityId());
+            } else if (PayeeIdTypeCodes.EMPLOYEE.equals(payeeType)) {
+                achAccount.setPayeeIdNumber(payee.getEmployeeId());
+            } else {
+                throw new WorkflowException("Unexpected payee ID type for automated Payee ACH Account creation: " + payeeType);
+            }
```
nkimble: I believe that this should also be changed to returning a boolean and logging the info about the problem instead of throwing an exception.  Daniela and ChadH : your thoughts?

db483: yes, I agree

cah292: Yes, sounds good.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
nkimble: Making that suggested change to both methods.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
nkimble: making this change as well for consistency.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
db483: good point Chad

nkimble: Not altering the code at this time based on Chad's comment.

nkimble:
retest this please

nkimble:
PR has been updated with the suggested changes. Please review those changes.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -0,0 +1,771 @@
...
+        } else {
+            processingResults.append(" Input file's account information for payee ")
+                             .append(achDetail.getNetID())
+                             .append(" of type '")
+                             .append(achAccount.getPayeeIdentifierTypeCode())
+                             .append("' matches what is already in KFS; no updates will be made for this entry.");
+        }
+        processingResults.append(" Update was NOT performed.");
+        LOG.info(processingResults.toString());
+        return true;
```
db483: is this method always returning true? it seem it is not worth returning a boolean in that case unless I am missing the return false path.

nkimble: The "update" functionality is to be addressed in a future user story. Since it is considered bad practice to change method signatures and knowing the functionality is coming, I structured this method signature to match the "add" so that the calling routine should not need to be changed. The future user story should only need to adjust this method.

db483: thank you Nancy, makes sense now

db483:
Code changes look good. merging.

PR #201, opened 2016-08-19
KFSPTS-6384 clear pdp statuses on RCDV copy

db483:


db483:
retest this please

db483:
retest this please

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherAction.java
```diff
@@ -54,6 +55,14 @@ public ActionForward route(ActionMapping mapping, ActionForm form, HttpServletRe
         return forward;
     }
     
+    @Override
+    public ActionForward copy(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+    	RecurringDisbursementVoucherForm recurringForm = (RecurringDisbursementVoucherForm) form;
+    	recurringForm.getPdpStatuses().clear();
+    	ActionForward forward = super.copy(mapping, form, request, response);
```
jhulslander: should the getting of the form and clearing the arrayList come after super()?

I think we also want to loop through recurring Details, and clear the DV document number from them.

jhulslander: recurringForm.getPdpStatuses() calls getRecurringDisbursementVoucherDocumentService().findPdpStatuses((RecurringDisbursementVoucherDocument)getDocument()

And that looks at reucrring Details for DVs, and then finds the PDP statuses, so yes, we want to clear the DV document numbers from the recurring details

db483: There is nothing in the super.copy() that will alter the pdpStatuses list but I can move that after just to be cautious.

The recurringDisbursementVoucherDetails are on the RCDV document and the list is reset on the copy method of the document so there does not seem to be a need to loop over this list and reset the DV doc number. Am I looking at the wrong list?

jhulslander:
I had a couple notes

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherAction.java
jhulslander: OK, I see what happens, the form has the PDP statuses cached, and the recurring DV document clears what needs cleared, so your solution makes sense I think.  the tests pass, I shall merge

PR #202, opened 2016-08-22
KFSPTS-6399 Log exceptions in route step and don't fail job

db483:


db483:
This change is to prevent the IncomingACHandWiresBatchJob batch job to fail in the routeAdvanceDepositDocumentsStep step. The exceptions that occur in this step are logged now instead of thrown so that the job can continue with the processing.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -186,16 +192,31 @@ public boolean routeAdvanceDepositDocuments() {
...
+                LOG.error("Error routing document # " + documentId + " due to exception: " + e.getMessage());
+                logException(e);
             }
         }
 
         return true;
+    }  
+
+	private void logException(Exception e) {
+        if (e instanceof ValidationException) {
```
nkimble: Formatting for this section of the file looks off.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
```diff
@@ -186,16 +192,31 @@ public boolean routeAdvanceDepositDocuments() {
                     LOG.info("Routing advance deposit document # " + documentId + ".");
                 }
                 documentService.routeDocument(advanceDocument, "document routed by achIncome batch job", null);
-            } catch (WorkflowException e) {
-                LOG.error("Error routing document # " + documentId + " " + e.getMessage());
-                throw new RuntimeException(e.getMessage(), e);
+            }
```
nkimble: There is a RuntimeException also being thrown on line 181 of this method.  Should that exception also be addressed in a more graceful manner?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
db483: not quite sure why it looks off, it looks fine in Eclipse. Maybe because it is using spaces rather then tabs...

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
db483: Good question. Seems though that if there is a problem retrieving the docs that is kind of a legitimate failure and throwing the exception seems fine.

db483:
I have updated the formatting, thank you Nancy! Does it look better now?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImpl.java
nkimble: Sounds appropriate for this data condition.

nkimble:
Formatting looks better.  Waiting for unit tests to complete before merging.

nkimble:
Unit tests passed. Merging.

PR #203, opened 2016-08-23
KFSPTS-6324 Implement cancel of spawned DVs from recurring DV

jhulslander:


jhulslander:
This is ready for a first review.  I've added both @cah292  and @db483 as reviews since I've worked with both of you on different details of this.

src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -134,6 +134,7 @@ public static ScheduleTypes fromName(String name) {
         public static final String RECURRING_DV_DOCUMENT_TYPE_NAME = "RCDV";
         public static final String RECURRING_DV_EXPLANATION_TO_DV_NOTE_STARTER = "DV created by recurring DV: ";
         public static final String RECURRING_DETAILS_TAB_NAME = "RecurringDetails";
+        public static final String RECURRING_DV_CANCEL_PARMENTS_PERMISSION_NAME = "CANCEL RECURRING DISBURSEMENT VOUCHER PAYMENTS";
```
cah292: Minor typo in variable name, should be "..._PAYMENTS_..." instead of "...PARMENTS..."

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherAction.java
```diff
@@ -61,6 +68,44 @@ public ActionForward copy(ActionMapping mapping, ActionForm form, HttpServletReq
...
     	ActionForward forward = super.copy(mapping, form, request, response);
         return forward;
+        
+    }
+    
+    public ActionForward confirmAndCancel(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+        RecurringDisbursementVoucherForm recurringForm = (RecurringDisbursementVoucherForm) form;
+        RecurringDisbursementVoucherDocument recurringDV = (RecurringDisbursementVoucherDocument) recurringForm.getDocument();
+        
+        if (StringUtils.isEmpty(recurringDV.getPaymentCancelReason())) {
```
cah292: You may want to use StringUtils.isBlank() instead of StringUtils.isEmpty() here, unless you have a need for allowing blank input.

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherAction.java
```diff
@@ -61,6 +68,44 @@ public ActionForward copy(ActionMapping mapping, ActionForm form, HttpServletReq
...
+       if (isSetPopulated(canceledDVs)) {
+           KNSGlobalVariables.getMessageList().add(CUKFSKeyConstants.INFO_RCDV_LIST_OF_DVS_CANCELED, StringUtils.join(canceledDVs, ", "));
+       }
+       if (!isSetPopulated(canceledPaymentGroups) && !isSetPopulated(canceledDVs)) {
+           KNSGlobalVariables.getMessageList().add(CUKFSKeyConstants.INFO_RCDV_NO_PAYMENTS_OR_DVS_CANCELED);
+       }
+    }
+    
+    private boolean isSetPopulated(Set<String> setToTest) {
+        return setToTest != null && !setToTest.isEmpty();
```
cah292: If the methods on RecurringDisbursementVoucherDocumentService are not expected to return a null Set, then you can probably remove this method and replace its usage with isEmpty() calls on the Sets instead. (Generally, I think we want our Kuali services to avoid returning null collections or maps.)

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherForm.java
```diff
@@ -38,6 +49,26 @@ public String getTripID() {
 		return StringUtils.EMPTY;
 	}
 	
+	public boolean isCancelPDPPaymentsActionAvailable() {
```
cah292: There's some odd indentation in this new method. I would suggest changing this method's tabs into spaces to clean up the formatting.

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherForm.java
```diff
@@ -38,6 +49,26 @@ public String getTripID() {
...
+	    if (getRecurringDisbursementVoucherPaymentMaintenanceService().hasCancelPermission(GlobalVariables.getUserSession().getPerson())) {
+    	    Collection<PaymentDetail> details = getRecurringDisbursementVoucherDocumentService().findPaymentDetailsFromRecurringDisbursementVoucher((RecurringDisbursementVoucherDocument)getDocument());
+    	    for (PaymentDetail detail : details) {
+    	        String paymentDetailStatus = detail.getPaymentGroup().getPaymentStatusCode();
+    	        Person person = GlobalVariables.getUserSession().getPerson();
+    	        if (paymentDetailStatus != null && 
+    	                getRecurringDisbursementVoucherDocumentService().isPaymentCancelable(person, paymentDetailStatus)) {
```
cah292: For consistency with our other code, I would suggest moving the "&&" to the second line of the "if" statement.

src/main/java/edu/cornell/kfs/fp/service/RecurringDisbursementVoucherDocumentService.java
```diff
@@ -35,11 +37,33 @@
 	List<RecurringDisbursementVoucherPDPStatus> findPdpStatuses(RecurringDisbursementVoucherDocument recurringDisbursementVoucherDocument);
 	
 	/**
+	 * Cancels all DVs spawned by the Recurring Disbursement Voucher that are in saved status.
+	 * @param recurringDisbursementVoucherDocument
+	 * @param cancelMessage
+	 * @return A list of DV document numbers that have been canceled.
```
cah292: This method and the one below it are mentioning lists in their Javadocs even though they actually return Sets. Please updated accordingly to avoid confusion.

src/main/java/edu/cornell/kfs/fp/service/RecurringDisbursementVoucherDocumentService.java
```diff
@@ -35,11 +37,33 @@
...
 	 * @return
 	 */
 	Collection<PaymentDetail> findPaymentDetailsFromRecurringDisbursementVoucher(RecurringDisbursementVoucherDocument recurringDisbursementVoucherDocument);
 	
-	
+	/**
+	 * Determins if an individual payment can be canceled from a recurring DV.
+	 * @param user
+	 * @param paymentDetailStatus
+	 * @return
```
cah292: Please clean up the Javadocs for this method, especially the misspelling in the description.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -311,7 +339,72 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+        Note noteBase = new Note();
+        noteBase.setAuthorUniversalIdentifier(GlobalVariables.getUserSession().getPerson().getPrincipalId());
+        noteBase.setNotePostedTimestampToCurrent();
+        return noteBase;
+    }
+    
+    @Override
+    public boolean isPaymentCancelable(Person user, String paymentDetailStatus) {
+        return paymentDetailStatus.equalsIgnoreCase(PdpConstants.PaymentStatusCodes.OPEN) 
+                && getRecurringDisbursementVoucherPaymentMaintenanceService().hasCancelPermission(user);
```
cah292: If you rewrite the first part as "PdpConstants.PaymentStatusCode.OPEN.equalsIgnoreCase(paymentDetailStatus)", then you can remove the extra null check in the cancelPDPPayments() method above.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -311,7 +339,72 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
         }
         return paymentDetails;
     }
- 
+    
+    private void noteChangeOnRecurringDV(RecurringDisbursementVoucherDocument recurringDV, String noteText, Set<String> setOfStrings) {
+        if (setOfStrings != null && !setOfStrings.isEmpty()) {
```
cah292: You don't need the null check if you're certain that this private method is only being invoked with a non-null Set. Alternatively, you could use CollectionUtils.isNotEmpty() to keep things simpler.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -311,7 +339,72 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+                }
+            });
+        } catch (Exception e) {
+            throw new RuntimeException("cancelDVAsSystemUser() Unable to cancel DV: " + dv.getDocumentNumber(), e);
+        }
+    }
+    
+    private boolean isDVCancelable(CuDisbursementVoucherDocument dv) {
+        return getRecurringDisbursementVoucherPaymentMaintenanceService().hasCancelPermission(GlobalVariables.getUserSession().getPerson()) && 
+                StringUtils.equalsIgnoreCase(dv.getDocumentHeader().getWorkflowDocument().getStatus().getLabel(), DocumentStatus.SAVED.getLabel());
```
cah292: Another spot where the "&&" could be moved to the second line.
Also, if you're comparing KEW document statuses and not application document statuses, then I would suggest comparing the status codes instead of the status labels.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -311,7 +339,72 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+                cancelDVAsSystemUser(cancelMessage, dv);
+                canceledDVs.add(dv.getDocumentNumber());
+            }
+        }
+        noteChangeOnRecurringDV(recurringDisbursementVoucherDocument, "The following Disbursement Vouchers were canceled: ", canceledDVs);
+        return canceledDVs;
+    }
+
+    private void cancelDVAsSystemUser(final String cancelMessage, CuDisbursementVoucherDocument dv) {
+        final String docuemntNumber = dv.getDocumentNumber();
```
cah292: Minor typo in variable name.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -311,7 +339,72 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+    public boolean isPaymentCancelable(Person user, String paymentDetailStatus) {
+        return paymentDetailStatus.equalsIgnoreCase(PdpConstants.PaymentStatusCodes.OPEN) 
+                && getRecurringDisbursementVoucherPaymentMaintenanceService().hasCancelPermission(user);
+    }
+    
+    @Override
+    public Set<String> cancelDisbursementVouchers(RecurringDisbursementVoucherDocument recurringDisbursementVoucherDocument, final String cancelMessage) {
+        Set<String> canceledDVs = new HashSet<String>();
+        for (RecurringDisbursementVoucherDetail detail : recurringDisbursementVoucherDocument.getRecurringDisbursementVoucherDetails()) {
+            CuDisbursementVoucherDocument dv = retrieveCUDisbursementVoucherDocument(detail);
```
cah292: This code has the potential of retrieving the same document twice: Once to check if it's cancelable and again to perform the actual cancellation. Because the isDVCancelable() check only appears to require a permission check and a route status check, you could instead use a method such as RouteHeaderService.getDocumentStatus() to get the status for the can-cancel check, and then wait until the actual cancellation point to retrieve the full document.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherPaymentMaintenanceServiceImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import org.kuali.rice.kim.api.permission.PermissionService;
+import org.kuali.rice.krad.util.GlobalVariables;
+import org.kuali.rice.krad.util.ObjectUtils;
+
+import edu.cornell.kfs.fp.CuFPConstants;
+import edu.cornell.kfs.fp.service.RecurringDisbursementVoucherPaymentMaintenanceService;
+
+public class RecurringDisbursementVoucherPaymentMaintenanceServiceImpl extends PaymentMaintenanceServiceImpl implements RecurringDisbursementVoucherPaymentMaintenanceService, Serializable {
+
+    private static final long serialVersionUID = -1061956872976755617L;
```
cah292: Why does this service need to be Serializable? Can the variables that use it be declared as transient instead and the getters updated accordingly?

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherPaymentMaintenanceServiceImpl.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+    private void changePaymentStatusToCancelAndAddUpdatePaymentDetail(Integer paymentDetailId, String noteText, Person user, PaymentGroup paymentGroup, String paymentStatus) {
+        changeStatus(paymentGroup, PdpConstants.PaymentStatusCodes.CANCEL_PAYMENT, PdpConstants.PaymentChangeCodes.CANCEL_PAYMENT_CHNG_CD, noteText, user);
+        updatePaymentDetail(paymentDetailId, noteText);
+        LOG.debug("changePaymentStatusToCancelAndAddUpdatePaymentDetail() Pending payment cancelled; exit method.");
+    }
+
+    private void updatePaymentDetail(Integer paymentDetailId, String noteText) {
+        Map primaryKeys = new HashMap();
```
cah292: Unless the BO service is unable to accept wildcard-related Maps, I would suggest using generics here.

cah292:
Changes seem good overall, though I did have a few suggestions for improvement. Will wait for feedback and the review from @db483.

Also, I'm wondering what the feasibility is for adding micro-tests to some of the new services or service methods. I guess that would depend on how feasible it is to set up the appropriate mock objects, given that these changes use the doc service and the permission service and many others.

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -297,9 +303,31 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+            if (doCancel) {
+                LOG.debug("cancelPDPPayments: About to Cancel " + detail.getId() + " for a reason of " + cancelMessage);
+                canceledPaymentGroups.add(detail.getPaymentGroupId().toString());
+                getRecurringDisbursementVoucherPaymentMaintenanceService().cancelPendingPayment(detail.getPaymentGroupId().intValue(), 
+                        detail.getId().intValue(), cancelMessage, GlobalVariables.getUserSession().getPerson());
+            } else {
+                LOG.debug("cancelPDPPayments: Not going to cancel "+ detail.getId());
+            }
+        }
+        noteChangeOnRecurringDV(recurringDisbursementVoucherDocument, "The following payments were canceled: ", canceledPaymentGroups);
```
db483: Should you add this note only if there were some payments cancelled? I could be confusing to the user if a note is added with an empty list.

jhulslander: noteChangeOnRecurringDV will only add a note if the list has entries

src/main/java/edu/cornell/kfs/fp/document/web/struts/RecurringDisbursementVoucherAction.java
```diff
@@ -61,6 +68,44 @@ public ActionForward copy(ActionMapping mapping, ActionForm form, HttpServletReq
...
+        if (StringUtils.isEmpty(recurringDV.getPaymentCancelReason())) {
+            String paymentCancelFieldName = AccountingDocumentRuleBaseConstants.ERROR_PATH.DOCUMENT_ERROR_PREFIX + 
+                    CUKFSPropertyConstants.RECURRING_DV_PAYMENT_CANCEL_REASON_FIELD_NAME;
+            GlobalVariables.getMessageMap().putError(paymentCancelFieldName, CUKFSKeyConstants.ERROR_RCDV_PAYMENT_CANCEL_REASON_REQUIRED);
+        } else {
+            Set<String> canceledPaymentGroups = getRecurringDisbursementVoucherDocumentService().cancelPDPPayments(recurringDV, recurringDV.getPaymentCancelReason());
+            Set<String> canceledDVs = getRecurringDisbursementVoucherDocumentService().cancelDisbursementVouchers(recurringDV, recurringDV.getPaymentCancelReason());
+            reportCanceledItemsToBrowser(canceledPaymentGroups, canceledDVs);
+        }
+        forceRefreshOfPDPStatuses(recurringForm);
```
db483: Is it possible to get here even though some error occurred during cancelation? then the pdp statuses would be incorrectly updated to null

src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -297,9 +303,31 @@ private RecurringDisbursementVoucherPDPStatus buildRecurringDisbursementVoucherP
...
+    public Set<String> cancelPDPPayments(RecurringDisbursementVoucherDocument recurringDisbursementVoucherDocument, String cancelMessage) {
+        Collection<PaymentDetail> paymentDetails = findPaymentDetailsFromRecurringDisbursementVoucher(recurringDisbursementVoucherDocument);
+        Set<String> canceledPaymentGroups = new HashSet<String>();
+        for (PaymentDetail detail : paymentDetails) {
+            String paymentDetailStatus = detail.getPaymentGroup().getPaymentStatusCode();
+            boolean doCancel = paymentDetailStatus != null && 
+                    isPaymentCancelable(GlobalVariables.getUserSession().getPerson(), paymentDetailStatus);
+            if (doCancel) {
+                LOG.debug("cancelPDPPayments: About to Cancel " + detail.getId() + " for a reason of " + cancelMessage);
+                canceledPaymentGroups.add(detail.getPaymentGroupId().toString());
```
db483: cancelPendingPayment returns false if there was a problem canceling the payment. Should we add the paymentGroupId to cancelled payment groups only if it was successfully cancelled?

db483: I wonder if payment group id should be added to cancelled list only if all details were successfully canceled

db483:
Jay, I have also reviewed the code and had a couple of questions. I will review again once PR is updated. Thank you.

jhulslander:
Thanks @cah292 and @db483 I have made the recommended changes, this is ready for review again

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/fp/service/RecurringDisbursementVoucherDocumentService.java
```diff
@@ -35,11 +37,33 @@
...
 	 * @return
 	 */
 	Collection<PaymentDetail> findPaymentDetailsFromRecurringDisbursementVoucher(RecurringDisbursementVoucherDocument recurringDisbursementVoucherDocument);
 	
-	
+	/**
+	 * Determines if the payment can be canceled from the Recurring DV but the user.
+	 * @param user
+	 * @param paymentDetailStatus
+	 * @return True if the payment can be canceled.  False if ii can not be.
```
cah292: There's a couple small typos above, in the description and the @return line.

cah292:
Finished second round of review. Changes look good to me; I did notice a few comments that could be tweaked, but I'm not certain how important it is to fix them at this time.

Will wait for @db483 to finish her review.

jhulslander:
@cah292 I fixed the typo you found in the last review.  
I spoke with @db483 on hip chat and her concern was verifying canceling payments are transactional.  I was able to verify that if a second payment fails to cancel, the first payment cancel gets rolled back.  So I believe the functions are transactional as we expect.
I think if you feel the code is fine, we should be clear to merge this in.

jhulslander:
retest this please

db483:
Thank you Jay! Code looks good to me. Chad please go ahead and merge the code if you are also good with the changes

cah292:
Changes seem good to me. I did notice one tiny comment typo, but I don't think it's worth having it hold up this PR.

Will wait for the unit tests to pass before merging.

cah292:

jhulslander:
Thanks all

PR #208, opened 2016-08-31
KFSPTS-5876 Add new everify field on Award

db483:


src/main/java/edu/cornell/kfs/module/cg/businessobject/AwardExtendedAttribute.java
```diff
@@ -88,5 +89,13 @@ public KualiDecimal getBudgetTotalAmount() {
 	public void setBudgetTotalAmount(KualiDecimal budgetTotalAmount) {
 		this.budgetTotalAmount = budgetTotalAmount;
 	}
+	
+	public boolean isEerify() {
```
jhulslander: this is missing a V isn't it?

jhulslander:
Code looks good aside from one missing V in a function.  I suspect CI being down is causing the unit test faiures

db483:
Thank you for catching that, I accidentally deleted that while messing with formatting code.

bhutchinson:
Since the unit test failures are because the CI env is down, and it's not coming back up quickly, I think it's safe to merge this PR once you are happy with the code changes.

jhulslander:
Code looks good now, I shall merge

PR #212, opened 2016-09-12
KFSPTS-6386: add table prefix for dynamodb table names

bhutchinson:


bhutchinson:
retest this please

bhutchinson:
Apparently this isn't ready for primetime. I'll fix the failing tests before re-assigning it for review.

cah292:
It looks like the unit tests are failing because the new property does not exist on the Mongo implementation of the DAO. The property will likely need to be added there as well.

bhutchinson:
retest this please

bhutchinson:
I changed the unit test config to be dynamodb instead of mongo so it is consistent with our other environments even though that config isn't actually leveraged in our unit tests. I also added a getter which looked like it might actually be necessary when running some unit tests locally.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -269,6 +269,7 @@
 	<bean id="preferencesDao" parent="preferencesDao-parentBean"/>
 	<bean id="preferencesDao-parentBean" abstract="true" class="${kfs.documentstore.class}">
 		<property name="documentStoreClient" ref="documentStoreClient" />
+		<property name="tableNamePrefix" value="${kfs.documentstore.tablename.prefix}" />
```
cah292: This bean setup assumes that the DAO implementation has a tableNamePrefix setter. However, this changeset is only adding the setter to the Dynamo implementation, so this configuration will not work for our custom Mongo implementation bean. Perhaps the setter should be added to our Mongo implementation as well; or, perhaps a sub-interface of PreferencesDao could be used to require the setter or to at least provide a Java 8 "default" implementation of it.

cah292:
Finished my review. I have a concern about the bean configuration being incompatible with our Mongo DAO implementation bean, but otherwise the changes look good to me. Will wait for feedback.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/PreferencesDaoDynamoDB.java
```diff
@@ -193,11 +195,21 @@ public void saveUserPreferences(String principalName, Map<String, Object> prefer
 
     protected Table retrieveTable(String tableName) {
         DynamoDB dynamoDB = new DynamoDB(documentStoreClient.getClient());
+        if (StringUtils.isNotBlank(tableNamePrefix)) {
+            tableName = tableNamePrefix + tableName;
```
cah292: If tableName is null, will this end up being prefix + "null"? Will dynamoDB.getTable() be able to handle the name accordingly?

bhutchinson: It never gets called with a null tableName - constants are used for those. I think if somehow it did end up with an invalid table name because it had prefix + null, I think later on down the line you'd just get an error that the table didn't exist.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
bhutchinson: I don't think we would ever actual user our custom Mongo implementation - if we decide to go back to Mongo at some point in the future, I think we would eliminate the entire customization. That said, it probably makes sense to make a change for consistency. I will add default get/set implementations in a new interface as that seems like the most straightforward thing to do at this point and it also lets me try out Java8 default interface methods.

bhutchinson: The default interface methods didn't work well with Spring and BeanUtils. Instead of trying to figure that out (if even possible), I just added the new field, setter and getter to CuPreferencesDaoMongo.

bhutchinson:

cah292:
Changes look good, and unit tests have passed. Merging.

PR #213, opened 2016-09-23
KFSPTS-6507: update to 2016-09-16 release

bhutchinson:
I updated to the 2016-09-16 version of base Financials code and Rice 2.5.17. In doing so for this PR, I:
* removed our custom code related to one contribution from sprint 10
* applied changes in base code to files we overlay where appropriate (most of these were license and formatting changes - sorry for all the noise)
* updated our documentstore json
* made sure everything compiles ok and the app started (locally)
* made sure unit tests pass (when run locally)


src/main/java/org/kuali/kfs/fp/document/web/struts/DisbursementVoucherAction.java
```diff
@@ -202,7 +203,7 @@ public ActionForward printDisbursementVoucherCoverSheet(ActionMapping mapping, A
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         DisbursementVoucherCoverSheetService coverSheetService = SpringContext.getBean(DisbursementVoucherCoverSheetService.class);
 
-        coverSheetService.generateDisbursementVoucherCoverSheet(directory, DisbursementVoucherConstants.DV_COVER_SHEET_TEMPLATE_NM, document, baos);
+        coverSheetService.generateDisbursementVoucherCoverSheet(document, baos);
```
db483: It looks like the local directory variable is not needed any longer, should that be removed?

bhutchinson:
I believe I have fixed the failing unit test. There were code changes in a unit test class that we shadow with our overlay that I hadn't applied to our test class; I have applied them now.

src/main/java/org/kuali/kfs/fp/document/web/struts/DisbursementVoucherAction.java
bhutchinson: It probably should (and I'll mention it to the product team) but it's still there in base code and since this is a file we overlay, I think we should keep it as close to base as possible. Thanks.

db483: Makes sense, thank you.

PR #216, opened 2016-09-30
KFSPTS-6555: remove doc type that was removed in base

bhutchinson:


jhulslander:
Year End Depreciation Assets exists in prod, but we don't seem to have used the doc type, and the unit tests pass, so will merge this in.

bhutchinson:
Yeah, it was removed from the dashboard in the latest base code and we don't have any references to it in our project so I figured it was safe to remove. Maybe we want to consider removing the doc type too, but it's probably not causing harm if no one can do anything with it.

jhulslander:
I'd lean toward deactivating it versus remove it just to keep our system as clean as possible.  However, like you mentioned, it's not hurting anything.

bhutchinson:
Good point, thanks.

PR #219, opened 2016-10-10
KFSPTS-6312 updating permission check from institution preferences

jhulslander:
This updates the JSON to use the correct permission to look up to determine if links should be shown or not.  I also added headers and footers so this JSON can be directly loaded into dynamo DB.  I am not sure if we should put that here or not, but seems like a descent idea to make it easier to apply JSON changes

### DIFF
--- a/src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
+++ b/src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
@@ -602,7 +602,7 @@
               "templateNamespace": "KR-NS",
               "templateName": "Look Up Records",
               "details": {
-                "namespaceCode": "KFS*"
+                "namespaceCode": "KFS-BC"
               }
             }
           },
@@ -626,7 +626,7 @@
               "templateNamespace": "KR-NS",
               "templateName": "Look Up Records",
               "details": {
-                "namespaceCode": "KFS*"
+                "namespaceCode": "KFS-BC"
               }
             }
           },
@@ -2565,7 +2565,7 @@
       }
     }
   ],
-  "menu": [
+   "menu": [
     {
       "label": "Feedback",
       "link": "kr/kualiFeedbackReport.do"


src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
```diff
@@ -2245,7 +2250,14 @@
         "activities": [
           {
             "documentTypeCode": "CCAP",
-            "linkType": "kfs"
+            "linkType": "kfs",
```
bhutchinson: Wouldn't this be using an Initiate Document permission?

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
```diff
@@ -1,4 +1,9 @@
-{
+var params = {
```
bhutchinson: I would prefer to not have the DynamoDB headers and footers included, for multiple reasons.

- We haven't decided how we are going to load the data into DynamoDB for dev, test and prod yet.
- I think this will work locally, but not for a remote DynamoDB instance.
- It's easy enough to add these to a local script to run.
- We may not always be using DynamoDB.
- It feels cleaner.

src/main/resources/edu/cornell/kfs/core/documentstore/InstitutionPreferences.json
```diff
@@ -602,7 +607,7 @@
               "templateNamespace": "KR-NS",
               "templateName": "Look Up Records",
               "details": {
-                "namespaceCode": "KFS*"
+                "componentName": "BudgetConstructionPosition"
```
bhutchinson: Based on possible changes from the other related SQL review, maybe this becomes KFS-BC?

jhulslander:
retest this please

jhulslander:
Please note, I removed my changes to add permission checks to CCAP.  No changes to that seem to be needed there to hide the links.

jhulslander:

PR #221, opened 2016-10-11
KFSPTS-6344: add cynergy attachment customizations to kfs version

bhutchinson:


bhutchinson:
retest this please

db483:
Hi Bryan, 

The new unit tests have failed, do you need to update the new test file before I review the code? thank you

bhutchinson:
Daniela,
I'm touching base with ChadB on those failures - they only happen on the Jenkins server, not locally. I may need to update the unit test class, so you could review the other files and that one later, or just wait until later to review everything. It's up to you. I don't expect any changes to the non-test artifacts and only slight, if any changes to the unit test class. Thanks.

db483:
thank you Bryan

bhutchinson:
retest this please

bhutchinson:
@db483 With help from ChadB, I've fixed the unit tests (only a one line change in the test class - line 101), so this should be good for review now. Thanks.

src/main/java/org/kuali/kfs/krad/config/KRADConfigurer.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    @Override
+    public void addAdditonalToConfig() {
+        configureDataSource();
+    }
+
+    @Override
+    public List<String> getAdditionalSpringFiles() {
+        final String files = ConfigContext.getCurrentContextConfig().getProperty("kfs." + getModuleName() + ".additionalSpringFiles");
```
db483: This seems like something other institutions would also have to do, can it be included in the kuali co version of the code in the future?

src/main/java/org/kuali/kfs/krad/service/impl/AttachmentServiceImpl.java
```diff
@@ -0,0 +1,290 @@
...
+        String fullPathUniqueFileName = getDocumentDirectory(attachment.getNote().getRemoteObjectIdentifier()) + File.separator + attachment.getAttachmentIdentifier();
+        File attachmentFile = new File(fullPathUniqueFileName);
+        attachmentFile.delete();
+    }
+
+    private String getPendingDirectory() {
+        return this.getDocumentDirectory("");
+    }
+
+    protected String getDocumentDirectory(String objectId) {
```
db483: The only change in this class is the visibility of thus method, can we create a contribution for that we can remove this in the future?

db483:
Thank you Bryan. Code changes look good and unit tests have passed, I am merging this.

PR #228, opened 2016-10-25
KFSPTS-6604: Added one day to the processDate to be used in the ACH Payment file header record as the headerDate. Cleaned up the code. Removed implementation class that duplicated this functionality and was not being referenced.

nkimble:
The cleanup performed removed the dependency on Spring, removed dead code, and removed extraneous comments. These changes cannot be verified locally and will need to be verified on the server based on the type of data that is placed in the generated output file.  These changes do compile cleanly and allow the server to start.


bhutchinson:
Why can't these changes be verified locally? Would it be possible to write a unit test that verified your changes?

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -61,33 +62,31 @@
     */
     @Override
     public void extractAchPayments() {
-        LOG.debug("AchBundlerExtractPaymentServiceImpl MOD - extractAchPayments() started");
+        LOG.debug("CuExtractPaymentServiceImpl MOD - extractAchPayments() started");
```
jhulslander: The logger will include the pull Class name in the log entry, so I don't think we need to include the class name in the log entry, just the function should do

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -61,33 +62,31 @@
...
 
-        Date processDate = SpringContext.getBean(DateTimeService.class).getCurrentDate();
+        Date processDate = dateTimeService.getCurrentDate();
         SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
-        PaymentStatus extractedStatus = (PaymentStatus) SpringContext.getBean(BusinessObjectService.class)
-                .findBySinglePrimaryKey(PaymentStatus.class, PdpConstants.PaymentStatusCodes.EXTRACTED);
+        PaymentStatus extractedStatus = (PaymentStatus) businessObjectService.findBySinglePrimaryKey(PaymentStatus.class, PdpConstants.PaymentStatusCodes.EXTRACTED);
     
-        String achFilePrefix = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(PdpKeyConstants.ExtractPayment.ACH_FILENAME);
+        String achFilePrefix = this.kualiConfigurationService.getPropertyValueAsString(PdpKeyConstants.ExtractPayment.ACH_FILENAME);
```
jhulslander: I think we can remove the "this"

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -61,33 +62,31 @@
...
         achFilePrefix = MessageFormat.format(achFilePrefix, new Object[] { null });
     
         String filename = getOutputFile(achFilePrefix, processDate);
-        LOG.debug("AchBundlerExtractPayment MOD: extractAchPayments() filename = " + filename);
+        LOG.debug("CuExtractPaymentServiceImpl MOD: extractAchPayments() filename = " + filename);
```
jhulslander: No need to include the class in the leg entry here as well

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -96,26 +95,22 @@ public void extractAchPayments() {
...
-    * A custom method that goes through and extracts all pending ACH payments and bundles them by payee/disbursement nbr.
     * 
     * @param extractedStatus
     * @param filename
     * @param processDate
     * @param sdf
     */
    protected void writeExtractBundledAchFile(PaymentStatus extractedStatus, String filename, Date processDate, SimpleDateFormat sdf) {
-       LOG.info("AchBundledExtractPaymentServiceImpl.writeExtractBundledAchFile started.");
+       LOG.info("CuExtractPaymentServiceImpl.writeExtractBundledAchFile started.");
```
jhulslander: no need for the class in here I think

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -170,11 +163,10 @@ protected void writeExtractBundledAchFile(PaymentStatus extractedStatus, String
            }
            writeCloseTag(os, 0, "achPayments");
            
-           // send summary email
            getPaymentFileEmailService().sendAchSummaryEmail(unitCounts, unitTotals, SpringContext.getBean(DateTimeService.class).getCurrentDate());
        }
        catch (IOException ie) {
-           LOG.error("AchBunderlExtract MOD: extractAchFile() Problem reading file:  " + filename, ie);
+           LOG.error("CuExtractPaymentServiceImpl MOD: extractAchFile() Problem reading file:  " + filename, ie);
```
jhulslander: another log entry not needing class name

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -2183,15 +2108,11 @@ protected String repeatThis(String s, int n) {
         return  String.format(String.format("%%0%dd", n), 0).replace("0",s);
     }
     
-    //KFSPTS-1460
     public AchBundlerHelperService getAchBundlerHelperService() {
         return achBundlerHelperService;
```
jhulslander: It looks like cu-spring-pdp.xml is already injecting this service.  I am not sure why spring context was used above.  Is there a way to test this locally? I am wondering if there was a spring mapping problem that was worked around by using spring context above.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -170,11 +163,10 @@ protected void writeExtractBundledAchFile(PaymentStatus extractedStatus, String
            }
            writeCloseTag(os, 0, "achPayments");
            
-           // send summary email
            getPaymentFileEmailService().sendAchSummaryEmail(unitCounts, unitTotals, SpringContext.getBean(DateTimeService.class).getCurrentDate());
```
bhutchinson: Looks like this should be referencing dateTimeService (or getDateTimeService() if you prefer) to avoid the SpringContext.getBean call.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Removing class name.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Removing

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Removing

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Removing

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Removed SpringContext call

nkimble:
reset this please

nkimble:
Unit tests to verify that header record date is being created correctly and used in the output file correctly were created.  I believe that I have updated the other concerns as well.  Please let me know if I missed anything.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,187 @@
...
+        cuExtractPaymentServiceImpl.setAchBundlerHelperService(new TestAchBundlerHelperService());
+        
+        PaymentStatus extractedStatus = new PaymentStatus();
+        Date processDate = new Date(Calendar.getInstance().getTimeInMillis());
+        List<String> notificationEmailAddresses = new ArrayList<String>(Arrays.asList(NOTIFICATION_EMAIL_ADDRESSES));
+        
+        cuExtractPaymentServiceImpl.writeExtractAchFileMellonBankFastTrack(extractedStatus, ACH_EXTRACT_TEST_PATH_AND_FILE, processDate, yyyyMMddHHmmss_DATE_FORMATTER, notificationEmailAddresses);
+        
+        Date expectedHeaderDate = cuExtractPaymentServiceImpl.calculateHeaderDate(processDate);
+        assertEquals("ACH Extract Payments File Header Date is not what was expected..", true, isDateInOutputFileWhereExpected(HEADER_LINE_IN_OUTPUT_FILE, yyyyMMddHHmmss_DATE_FORMATTER.format(expectedHeaderDate)));
```
jhulslander: This is a cool unit test, I like it.  I am confused about this line though  Do we expect the "expectedHeaderDate" to be in the header row of the output file?

Not worth changing (IMHO) unless other code review things come up, but I would use assertTrue instead of assertEquals here.

jhulslander:
I think this looks pretty good, I had one question, and one small note that isn't worth changing unless other code review items come up.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: Yes, the header date should be in the first line of the output file.  The users and the bank reference that line in the file under multiple names. Meaning it can be called the header record or the header row.  

Your suggestion of using assertTrue is a valid change.  I will be pushing that change shortly.

nkimble:
retest this please

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Arrays;
+import java.util.Date;
+import java.util.Calendar;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+
+public class CuExtractPaymentServiceImplTest {
+    
+    private static final String ACH_BATCH_TEST_FILE_PATH = "src/test/java/edu/cornell/kfs/pdp/batch/service";
```
bhutchinson: I would set this to simply "test" so that you're not creating and removing directories under a path that is under version control.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+    public void setUp() throws Exception {
+        cuExtractPaymentServiceImpl = new TestCuExtractPaymentServiceImpl();
+        File testOutputDirectory = new File(ACH_EXTRACT_TEST_FILE_PATH);
+        testOutputDirectory.mkdir();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        cuExtractPaymentServiceImpl = null;
+        FileUtils.forceDelete(new File(ACH_EXTRACT_TEST_FILE_PATH).getAbsoluteFile());
```
bhutchinson: If you do change ACH_BATCH_TEST_FILE_PATH as suggested above, you might want to forceDelete the absolute path for that here, otherwise, your project directory will be dirty after the tests run.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+        boolean expectedWasFound = false;
+        try {
+            String lineWeWant = null;
+            inputReader = new BufferedReader(new FileReader(new String (ACH_EXTRACT_TEST_PATH_AND_FILE_FOR_VALIDATION)));
+            for (int i=0; i<numberOfLinesToRead; i++) {
+                lineWeWant = inputReader.readLine();
+            }
+            expectedWasFound = lineWeWant.contains(formattedDateExpected);
+        }
+        catch (IOException ie) {
```
bhutchinson: Not necessarily something I would make you change, but something I wanted to point out: The simplest thing for a unit test would be to just throw the IOException from this method and not bother with the try/catch. If there is an IOException, the test will still fail, and you let the testing framework do the work for you. The only time I catch Exceptions in unit tests is when I need to do processing after the point in which the exception occurs. Otherwise I just throw them, or use annotations to tell JUnit to expect them.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+    private class TestAchBundlerHelperService implements AchBundlerHelperService {
+
+        private static final String BANKCODE_1 = "bankCode1";
+        private final Integer DISB_ID_1 = new Integer(1);
+
+        @Override
+        public Iterator<PaymentDetail> getPendingAchPaymentDetailsByDisbursementNumberAndBank(Integer disbursementNumber, String bankCode) {
+            ArrayList<PaymentDetail> paymentDetails = new ArrayList<PaymentDetail>();
+
+            PaymentDetail paymentDetail_1 = new PaymentDetail();
```
bhutchinson: It looks like you only have one PaymentDetail, PaymentGroup and Bank - is there a need for the "_1" suffixes? Not a huge deal, they're just a bit visually distracting. Same question for the constants above.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+            paymentGroup_1.setDisbursementNbr(new KualiInteger(123456789));
+
+            paymentDetail_1.setPaymentGroup(paymentGroup_1);
+            paymentDetails.add(paymentDetail_1);
+            return paymentDetails.iterator();
+        }
+
+        @Override
+        public Iterator<PaymentDetail> getPendingAchPaymentDetailsByBank(String bankCode) {
+            // TODO Auto-generated method stub
```
bhutchinson: unnecessary comment

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
```diff
@@ -0,0 +1,188 @@
...
+    private static final String ACH_EXTRACT_TEST_FILE_NAME = "testAchExtractPaymentsOuput.txt";
+    private static final String ACH_EXTRACT_TEST_PATH_AND_FILE = ACH_EXTRACT_TEST_FILE_PATH + File.separator + ACH_EXTRACT_TEST_FILE_NAME;
+    private static final String ACH_EXTRACT_TEST_PATH_AND_FILE_FOR_VALIDATION = ACH_EXTRACT_TEST_PATH_AND_FILE + ".READY";
+    private static final SimpleDateFormat yyyyMMddHHmmss_DATE_FORMATTER = new SimpleDateFormat("yyyyMMddHHmmss");
+    private static final SimpleDateFormat yyyyMMdd_DATE_FORMATTER = new SimpleDateFormat("yyyyMMdd");
+    private static final String[] NOTIFICATION_EMAIL_ADDRESSES = {"abc@xyz.edu", "def@xyz.edu", "ghi@xyz.edu"};
+    private static final int HEADER_LINE_IN_OUTPUT_FILE = 1;
+    private static final int FIRST_PAYEE_LINE_IN_OUTPUT_FILE = HEADER_LINE_IN_OUTPUT_FILE + NOTIFICATION_EMAIL_ADDRESSES.length + 1;
+    private static final  Logger LOG = org.apache.log4j.Logger.getLogger(CuExtractPaymentServiceImplTest.class);
+
```
bhutchinson: Not something I'd require in and of itself, but since there are other requested changes, I'd remove a couple of these blank lines since they seem extraneous.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: Changed.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: Done.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: Changed the forceDelete.  Also had to change the unit test setUp method to perform a forceMkdir with the directory path change.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: Removed.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImplTest.java
nkimble: The suffix values were used with the intent that future expansion on this unit test would most likely require multiple processing lines in the output file and using from the start would decrease maintenance.

bhutchinson: I would leave off the suffixes and refactor these names if/when future expansion of this unit test necessitates it. In general, I prefer to avoid coding for future expansion when first introducing code because often the future expansion never occurs or when it does the needs are different and the original code ends up adding unintentional technical debt.

nkimble:
retest this please

jhulslander:
+1 from me

PR #229, opened 2016-10-27
KFSPTS-6382: update to Rice 2.1.16

bhutchinson:


src/main/resources/ESAPI.properties
```diff
@@ -0,0 +1,431 @@
...
+# http://www.opensource.org/licenses/ecl2.php
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# CU Customization: Added the Rice 2.3.13 version of this file.
```
cah292: Looks like this was copied from the cynergy-standalone version of the file, where I had added this comment to clarify that it was an addition or overlay from existing Rice code. Should this be updated or removed accordingly on the KFS side?

src/main/webapp/WEB-INF/tld/esapi.tld
```diff
@@ -0,0 +1,372 @@
...
+    http://www.opensource.org/licenses/ecl2.php
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<!-- CU Customization: Added this tag library from the Rice 2.3.13 release. -->
```
cah292: Another comment copied from the cynergy-standalone version of the file.

bhutchinson: Good catch, I'll remove, thanks.

cah292:
As per the recent unit test failure, it looks like cu-kfs was having trouble finding the new WebUtils.sanitizeBackLocation() method. I had added this in cornell_customizations, perhaps the unit tests are somehow using an outdated version of the cornell_customizations JAR?

bhutchinson:
The unit test job was looking at the current version of the jar, I've modified it to use the develop version by default and the current version when running against the post6-upgrade branch.

bhutchinson:
This should be ready for review again.

PR #230, opened 2016-11-01
KFSPTS-6618: update 1042S report format

bhutchinson:


src/main/resources/edu/cornell/kfs/tax/batch/Default1042STaxOutputDefinition.xml
```diff
@@ -29,83 +29,87 @@
...
+        <field name="B74" length="1" type="BLANK" /><!-- B74 :: LPR steps -->
+        <field name="B75" length="1" type="BLANK" /><!-- B75 :: Tax Home Connection -->
+        <field name="B76" length="1" type="BLANK" /><!-- B76 :: Claim Treaty benefits -->
+        <field name="B77" length="4" type="BLANK" /><!-- B77 :: Primary Withholding Agent's ID -->
+        <field name="B78" length="40" type="BLANK" /><!-- B78 :: Primary Withholding Agent's Name -->
         <!--
-            ==== NOTE: The current spec does not have any fields with names B75 through B96. ====
+            ==== NOTE: The current spec does not have any fields with names B79 through B96. ====
          -->
         <field name="B97" length="25" type="BLANK" /><!-- B97 :: Filler -->
```
cah292: Does this field still need to contain the exact number of spaces for the tab-delimited format, or is it fine as-is? The same goes for B99 below.

src/main/resources/edu/cornell/kfs/tax/batch/Default1042STaxOutputDefinition.xml
```diff
@@ -128,28 +132,30 @@
...
+        <field name="D29" length="40" type="BLANK" /><!-- D29 :: Intermediary/FLW-THR Name Line 1 -->
+        <field name="D30" length="4" type="BLANK" /><!-- D30 :: Other Payer ID -->
+        <field name="D31" length="40" type="BLANK" /><!-- D31 :: Other Payer Name -->
+        <field name="D32" length="4" type="BLANK" /><!-- D32 :: Primary Withholding Agent's ID -->
+        <field name="D33" length="40" type="BLANK" /><!-- D33 :: Primary Withholding Agent's Name -->
         <!--
-            ==== NOTE: The current spec does not have any fields with names D32 through D96. ====
+            ==== NOTE: The current spec does not have any fields with names D34 through D96. ====
          -->
         <field name="D97" length="25" type="BLANK" /><!-- D97 :: Filler -->
```
cah292: Another spot where it's unclear whether manually including spaces is still needed; same goes for D99 below.

src/main/resources/edu/cornell/kfs/tax/batch/Default1042STaxOutputDefinition.xml
bhutchinson: It may not be needed, but it doesn't seem to hurt anything, and if we leave as is, we could easily convert back to fixed width instead of tab delimited (although whether we will have that need is another question). Do you think there's a downside to including it and we should remove it?

cah292: The current cu-kfs tax job will only output filler spaces for "BLANK"-type fields if the output type is fixed-width. Lori might know if the tab-delimited file still needs pre-existing spaces in those fields.

bhutchinson: Lori said the output files are fine, so I think we're ok then.

PR #235, opened 2016-11-09
KFSPTS-6749: update to 2016-11-03.1 version of Finacials base code

bhutchinson:
This is one of three related PRs, one in config, one in nonprod-sql and this one.


src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
```diff
@@ -641,6 +641,11 @@ public PersistableBusinessObject retrieve(PersistableBusinessObject object) {
         }
 
         @Override
+        public <T extends BusinessObject> Collection<T> findMatching(Class<T> aClass, Map<String, ?> map, int i, int i1, String[] strings) {
```
jhulslander: this seems odd, do we really want to return null on this look up?

jhulslander:
retest this please

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
bhutchinson: I think it's fine. It's in a MockBusinessObjectService just used for a test class. This method was recently added, so I needed to implement it to fix the compiler error, but most of the methods in that Mock, included another findMatching version, return null.

jhulslander: oh, I see now, my bad

PR #245, opened 2016-12-06
KFSPTS-6707: remediate KFS maintenance XML comversion feature

bhutchinson:


src/main/java/edu/cornell/kfs/krad/service/impl/CuDocumentServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Maintenance documents ordinarily convert the old/new maintenance XML
+ * sections separately if they need it, and are not capable of converting
+ * legacy BO notes XML. To support better conversion of such content,
+ * the default MaintainableXMLConversionService has been switched over
+ * to a no-op implementation, and instead this class will perform
+ * the XML conversion on the entire maintenance XML payload if necessary.
+ * This allows for adding the features without overlaying Rice code.
+ *
+ * There are three get-by-ID service methods that do not save documents
```
cah292: The paragraph beginning at this line does not accurately reflect the KFS-side DocumentService customization, which appears to be overlaying the DocumentService to make a private method protected (instead of doing full overrides of the 3 get-by-ID methods that Cynergy does). It might be a good idea to update this Javadoc for accuracy.

src/main/resources/edu/cornell/kfs/krad/config/CUKRADSpringBeans.xml
```diff
@@ -30,4 +30,46 @@
...
+    <bean id="cf.maintainableXMLConversionService" class="edu.cornell.kfs.krad.service.impl.NoOpMaintainableXMLConversionServiceImpl" />
+
+    <bean id="cuMaintainableXMLConversionService" parent="cuMaintainableXMLConversionService-parentBean"/>
+    <bean id="cuMaintainableXMLConversionService-parentBean" abstract="true" class="edu.cornell.kfs.krad.service.impl.CuMaintainableXMLConversionServiceImpl">
+        <property name="conversionRuleFiles">
+            <list>
+                <value>classpath:edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml</value>
+            </list>
+        </property>
+        <property name="conversionRuleFile" value="classpath:edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml"/>
```
cah292: This bean is setting both a "conversionRuleFiles" and a "conversionRuleFile" property, is it necessary for it to set both?

src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
```diff
@@ -0,0 +1,724 @@
...
+            </pattern>
+            <pattern>
+                <match>indirectCostRecoveryAcctNbr</match>
+                <replacement></replacement>
+            </pattern>
+        </pattern>
+        <!-- KDO-873 end -->
+        <!-- KDO-934 -->
+        <!--
+            NOTE: We will need to revisit these VendorHeader conversion rules with the next
```
cah292: We still have a vendorLocale property that's exclusive to our vendor header extension objects, but it appears to be getting removed by the rules below (due to current limitations of our XML conversion service customization). Is that presenting an issue with older docs? Or is there a user story to analyze this particular case further?

src/main/java/org/kuali/kfs/krad/service/impl/DocumentServiceImpl.java
```diff
@@ -0,0 +1,1164 @@
...
+    }
+
+    /**
+     * Performs required post-processing for every document from the documentDao
+     *
+     * @param documentHeaderId
+     * @param workflowDocument
+     * @param document
+     */
+    protected Document postProcessDocument(String documentHeaderId, WorkflowDocument workflowDocument, Document document) {
```
cah292: Just to clarify, is this method's increased visibility the only change that was made by this overlay?

src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
bhutchinson: I was able to open an old vendor doc without error. What type of issue would you anticipate? I do expect us to do more functional validation during QA and I will be creating a story for validating this feature, but I can create an additional story if we need to.

src/main/java/org/kuali/kfs/krad/service/impl/DocumentServiceImpl.java
bhutchinson: Yes, that is correct. And I'm hopeful we could contribute that back to base code and eliminate this overlay, possibly as soon as next sprint.

src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
cah292: If the vendorLocale is being fully removed, then any values for it on older vendor docs will not show up. It may not be much of an issue for vendor docs that have finished routing, but it could pose problems for enroute ones. Deferring further validation of this issue to another story or an existing QA story is fine with me.

bhutchinson: I'll include this in the QA validation story. It doesn't look like we have very many vendor extension records with a locale (only 129 out of 65253) so I'm not sure how big of a problem this might be. I did find another issue while I was trying to test this though so I'll need to fix that and update this PR with that fix (as well as the other changes above) before it is merged.

bhutchinson:
I'm closing this for now. Once I get it cleaned up and fix the new bug I found, I'll open a new one.

PR #249, opened 2016-12-14
KFSPTS-4977: introduce institutional-config.properties and cleanup ov…

bhutchinson:
…erride properties

This is one of two PRs that should be reviewed together, the other one is in the config repo. However, the one in the config repo should be merged first and then the unit test PR job should be re-run on this PR (so that both sets of changes are taken into account) before it is merged.

### DIFF
--- a/src/main/resources/institutional-config.properties
+++ b/src/main/resources/institutional-config.properties
@@ -1,0 +1,132 @@
+app.context.name=kfs
+context.names.rice=cynergy
+production.environment.code=kfs-prod
+logs.directory=/infra/logs
+work.directory=/infra/work
+data.xml.root.location=${staging.directory}/workflow/
+useQuartzDatabase=false
+threadPool.size=20
+message.persistence=false
+RouteQueue.maxRetryAttempts=3
+RouteQueue.timeIncrement=1000
+bus.refresh.rate=120
+rice.kew.enableKENNotification=false
+initialDelaySecs=60
+actionlist.outbox=false
+dailyEmail.active=false
+weeklyEmail.active=false
+dailyEmail.cronExpression=0 0 1 * * ?
+weeklyEmail.cronExpression=0 0 2 ? * 2
+kim.identityArchiveServiceImpl.executionIntervalSeconds=600
+session.document.cache.size=1000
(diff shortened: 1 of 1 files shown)


src/main/resources/institutional-config.properties
```diff
@@ -0,0 +1,144 @@
...
+work.directory=/infra/work
+reports.directory=/infra/work/reports
+staging.directory=/infra/work/staging
+temp.directory=/infra/work/temp
+batch.file.lookup.root.directories=/infra/work/staging;/infra/work/reports
+templates.directory=/infra/work/templates
+attachment.dir.location=/infra/work/attachments
+staging.directory.sys.batchContainer=/infra/work/staging/sys/batchContainer
+data.xml.root.location=/infra/work/staging/workflow
+data.xml.pending.location=/infra/work/staging/workflow/pending
```
cah292: There seems to be several properties that would benefit from "staging.directory" and/or "data.xml.root.location" placeholders. Would it make sense to use such placeholders for some of these directory-related properties, or are these all being defined without placeholders for a specific reason?

bhutchinson: They were all listed out without placeholders in the template file from puppet I was using as reference, and I also wasn't sure about the interaction between the default properties files, the new institition-config.properties and the command-line property overrides file. That said, you're correct that they would probably benefit from placeholders so let me try it out and see what I can do.

src/main/resources/institutional-config.properties
```diff
@@ -0,0 +1,144 @@
...
+logs.directory=/infra/logs
+attachments.directory=/infra/work/attachments
+attachments.pending.directory=/infra/work/attachments/pending
+work.directory=/infra/work
+reports.directory=/infra/work/reports
+staging.directory=/infra/work/staging
+temp.directory=/infra/work/temp
+batch.file.lookup.root.directories=/infra/work/staging;/infra/work/reports
+templates.directory=/infra/work/templates
+attachment.dir.location=/infra/work/attachments
```
bhutchinson: probably, see below comment

bhutchinson:

cah292:
retest this please

PR #251, opened 2016-12-16
KFSPTS-6979 add test connection end point and ba…

db483:
…sic authentication filter

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
+++ b/src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
@@ -1,0 +1,31 @@
+package edu.cornell.kfs.concur.rest.application;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+import edu.cornell.kfs.concur.rest.resource.ConcurTestConnectionResource;
+
+@ApplicationPath("/systems/v1.0")
+public class ConcurTestConnectionApplication extends Application {
+
+    protected Set<Object> singletons = new HashSet<>();
+    protected Set<Class<?>> clazzes = new HashSet<>();
+
+    public ConcurTestConnectionApplication() {
+        singletons.add(new ConcurTestConnectionResource());
+    }
+
+    @Override
+    public Set<Class<?>> getClasses() {
+        return clazzes;
+    }
+
+    @Override
+    public Set<Object> getSingletons() {
+        return singletons;
+    }
+
+}
--- a/src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
+++ b/src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
@@ -1,0 +1,53 @@
+package edu.cornell.kfs.concur.rest.authentication;
+
+import java.io.IOException;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.sys.context.SpringContext;
+
+import edu.cornell.kfs.concur.ConcurConstants;
+import edu.cornell.kfs.concur.service.ConcurAuthenticationService;
+
+public class ConcurAuthenticationFilter implements Filter {
+
+    @Override
+    public void destroy() {
+
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
+        HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+        String authorizationHeader = httpServletRequest.getHeader(ConcurConstants.AUTHORIZATION_PROPERTY);
+
+        if (StringUtils.isBlank(authorizationHeader)) {          
+            httpServletResponse.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        } else {
+            final String encodedUserNameAndPasswordAsToken = authorizationHeader.replaceFirst(ConcurConstants.BASIC_AUTHENTICATION_SCHEME + " ", "");
+
+            if (!getConcurAuthenticationService().isConcurTokenValid(encodedUserNameAndPasswordAsToken)) {
+                httpServletResponse.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            }
+        }
+    }
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+
+    }
+
+    private ConcurAuthenticationService getConcurAuthenticationService() {
+        return SpringContext.getBean(ConcurAuthenticationService.class);
+    }
+
+}
--- a/src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
+++ b/src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2007-2008 The Kuali Foundation Licensed under the Educational 
+	Community License, Version 2.0 (the "License"); you may not use this file 
+	except in compliance with the License. You may obtain a copy of the License 
+	at http://www.opensource.org/licenses/ecl2.php Unless required by applicable 
+	law or agreed to in writing, software distributed under the License is distributed 
+	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either 
+	express or implied. See the License for the specific language governing permissions 
+	and limitations under the License. -->
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://www.springframework.org/schema/p"
+	xsi:schemaLocation="http://www.springframework.org/schema/beans
+        http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
+
+	<bean id="concurModuleConfiguration" parent="concurModuleConfiguration-parentBean" />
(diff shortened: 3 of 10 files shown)


bhutchinson:
retest this please

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurApplication.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+import edu.cornell.kfs.concur.service.impl.ConcurWebServiceImpl;
+
+@ApplicationPath("/concur/v1.0")
```
bhutchinson: I think changing this to just v1 would be more consistent with the base financials APIs. (i.e. "/concur/v1")

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.ws.rs.ApplicationPath;
+import javax.ws.rs.core.Application;
+
+import edu.cornell.kfs.concur.service.impl.ConcurTestConnectionWebServiceImpl;
+
+@ApplicationPath("/systems/v1.0")
```
bhutchinson: another place I would remove the .0 from the version

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+    public void destroy() {
+
+    }
+
+    @Override
+    public void doFilter(ServletRequest servletRequest,
+            ServletResponse response, FilterChain arg2) throws IOException,
+            ServletException {
+
+        if (servletRequest instanceof HttpServletRequest) {
```
bhutchinson: Would we get anything other than an HttpServletRequest here? I'm wondering if this is necessary. (For reference, I don't see this check in the base code AuthenticationTokenFilter).

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+            String authorizationHeader = httpServletRequest
+                    .getHeader(ConcurConstants.AUTHORIZATION_PROPERTY);
+
+            if (StringUtils.isBlank(authorizationHeader)) {
+                if (response instanceof HttpServletResponse) {
+                    HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+                    httpServletResponse
+                            .setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+                }
+                return;
```
bhutchinson: This return and the other doesn't seem quite right to me; I'm wondering if you could rework the code a bit to eliminate them? Perhaps the code after this if could be in an else?

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+            ServletException {
+
+        if (servletRequest instanceof HttpServletRequest) {
+            HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
+
+            String authorizationHeader = httpServletRequest
+                    .getHeader(ConcurConstants.AUTHORIZATION_PROPERTY);
+
+            if (StringUtils.isBlank(authorizationHeader)) {
+                if (response instanceof HttpServletResponse) {
```
bhutchinson: another check where I wonder if it is really necessary

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurWebServiceImpl.java
```diff
@@ -0,0 +1,15 @@
+package edu.cornell.kfs.concur.service.impl;
+
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+
+@Path("/validate")
+public class ConcurWebServiceImpl {
```
bhutchinson: if this PR is for the test connection should this class be included?

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurApplication.java
```diff
@@ -0,0 +1,31 @@
...
+public class ConcurApplication extends Application {
```
bhutchinson: if this PR is for the test connection should this class be included?

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
```diff
@@ -0,0 +1,31 @@
...
+public class ConcurTestConnectionApplication extends Application {
+
+    protected Set<Object> singletons = new HashSet<>();
+    protected Set<Class<?>> clazzes = new HashSet<>();
+
+    public ConcurTestConnectionApplication() {
+        singletons.add(new ConcurTestConnectionWebServiceImpl());
```
bhutchinson: The pattern in the base Financials code is that Resource classes are registered in the Application and the Resources calls services. I know this service isn't complex, but it would probably be best to follow that pattern here as well.

src/main/resources/edu/cornell/kfs/concur/cu-ojb-concur.xml
```diff
@@ -0,0 +1,20 @@
+<!--
```
bhutchinson: since this file doesn't actually include any OJB descriptors is it necessary to include at this time?

src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
```diff
@@ -0,0 +1,48 @@
...
+		abstract="true">
+		<property name="namespaceCode" value="KFS-CNC" />
+		<property name="initializeDataDictionary" value="true" />
+		<property name="packagePrefixes">
+			<list>
+				<value>edu.cornell.kfs.concur</value>
+			</list>
+		</property>
+
+		<property name="databaseRepositoryFilePaths">
```
bhutchinson: I think this could probably be eliminated for now.

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
db483: Concur expects this endpoint to be: {servername}/system/v1.0/testconnection

I am not sure this that can be customized. I will double check with Concur. thank you

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
db483: I will update the code

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurWebServiceImpl.java
db483: I will remove code unrelated to test connection

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurTestConnectionApplication.java
bhutchinson: Ok, thanks @db483. If this is a Concur requirement, so be it.

src/main/java/edu/cornell/kfs/concur/rest/application/ConcurApplication.java
db483: I removed this code for now as it is not related to testconnection. When I add it back with the next user story I will change it to use v1.

db483:
I have made the requested changes and this is now ready for review. Thank you.

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest httpServletRequest = (HttpServletRequest) servletRequest;
+        HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+        String authorizationHeader = httpServletRequest.getHeader(ConcurConstants.AUTHORIZATION_PROPERTY);
+
+        if (StringUtils.isBlank(authorizationHeader)) {          
+            httpServletResponse.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        } else {
+            final String encodedUserNameAndPasswordAsToken = authorizationHeader.replaceFirst(ConcurConstants.BASIC_AUTHENTICATION_SCHEME + " ", "");
```
bhutchinson: since it doesn't look like we are using this variable for anything other than passing it to the next method call, I would consider eliminating it and inlining the authorizationHeader.replaceFirst call

src/test/java/edu/cornell/kfs/concur/rest/services/ConcurAuthenticationServiceImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+    private ConcurAuthenticationServiceImpl concurAuthenticationService;
+
+    @Before
+    public void setUp() {
+        concurAuthenticationService = new ConcurAuthenticationServiceImpl();
+        concurAuthenticationService.setConcurToken(CONCUR_TOKEN);
+    }
+
+    @Test
+    public void testIsConcurTokenValid_pass() {
```
bhutchinson: My preference would be to eliminate underscores from method names and just stick with camel casing throughout. Also, since the test is annotated, you could remote "test" from the start of the method name (that seems to be a newer test naming paradigm). So isConcurTokenValidPass or even isConcurTokenValidGoodToken / isConcurTokenValidBadToken / isConcurTokenValidEmptyToken for example.

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
db483: then the if would become rather long:
 if (!getConcurAuthenticationService().isConcurTokenValid(authorizationHeader.replaceFirst(ConcurConstants.BASIC_AUTHENTICATION_SCHEME + " ", ""))

it seems cleaner the way it is now.

bhutchinson: that's fair, thanks

db483:
I have updated the unit test and the code is ready for review again. Thank you!

PR #254, opened 2016-12-23
KFSPTS-7056 account validation web service initial implementation

db483:


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+public class ConcurEventNotificationProcessingServiceImpl implements ConcurEventNotificationProcessingService {
+    
+    protected ConcurEventNotificationService concurEventNotificationService;
+    protected ConcurAccountValidationService concurAccountValidationService;
+
+    @Override
+    public void processConcurEventNotifications() {
+        Collection<ConcurEventNotification> concurEventNotifications = concurEventNotificationService.retrieveConcurEventNotificationsForProcessing();
+
+        if (concurEventNotifications != null && CollectionUtils.isNotEmpty(concurEventNotifications)) {
```
bhutchinson: Since the body of the for loop won't be executed if the collection is empty, you could eliminate the CollectionUtils.isNotEmpty check. Also, you could use ObjectUtils.isNotNull for the null check.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+                ExpenseDetailedReportDTO expenseDetailedReport = retrieveExpenseDetailedReportFromConcur(concurEventNotification);
+                ConcurAccountInfo concurAccountInfo = extractAccountInfo(expenseDetailedReport);
+                concurAccountValidationService.validateConcurAccountInfo(concurAccountInfo);
+                updateExpenseReportStatusInConcur();
+            }
+        }
+    }
+
+    @Override
+    public ExpenseDetailedReportDTO retrieveExpenseDetailedReportFromConcur(
```
bhutchinson: It looks a little odd to me to have the method signatures of these methods split across two lines. They don't seem too long to be on one line, but maybe that's just personal preference.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+            ConcurEventNotification concurEventNotification) {
+        return null;
```
bhutchinson: Is the plan to implement these three methods later?

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotification.java
```diff
@@ -0,0 +1,101 @@
+package edu.cornell.kfs.concur.businessobjects;
+
+import java.sql.Date;
+
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+
+public class ConcurEventNotification extends PersistableBusinessObjectBase{
+    private static final long serialVersionUID = 1L;
```
bhutchinson: Should this be a real serial version ID value instead of 1?

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
```diff
@@ -0,0 +1,52 @@
...
+    public void addMessage(String message) {
+        if (messages == null) {
+            messages = new ArrayList<String>();
+        }
+        messages.add(message);
+    }
+
+    @Override
+    public String toString() {
+        StringBuffer result = new StringBuffer();
```
bhutchinson: do we care about including valid in the toString results?

src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseDetailedReportDTO.java
```diff
@@ -0,0 +1,5 @@
+package edu.cornell.kfs.concur.rest.xmlObjects;
+
+public class ExpenseDetailedReportDTO {
```
bhutchinson: Is the plan to implement this class later? It seems odd to include an empty class here.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -0,0 +1,17 @@
...
+
+import edu.cornell.kfs.concur.businessobjects.ConcurAccountInfo;
+import edu.cornell.kfs.concur.businessobjects.ValidationResult;
+import edu.cornell.kfs.concur.service.ConcurAccountValidationService;
+
+public class ConcurAccountValidationServiceImpl implements ConcurAccountValidationService {
+
+    @Override
+    public ValidationResult validateConcurAccountInfo(ConcurAccountInfo concurAccountInfo) {
+        ValidationResult validationResult = new ValidationResult(false, new ArrayList<String>());
```
bhutchinson: I'm thinking the plan is to implement more robust validation logic later?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    protected DateTimeService dateTimeService;
+
+    @Override
+    public ConcurEventNotification convertConcurEventNotification(ConcurEventNotificationDTO concurEventNotificationDTO) {
+        ConcurEventNotification concurEventNotification = new ConcurEventNotification();
+        concurEventNotification.setContext(concurEventNotificationDTO.getContext());
+        try {
+            concurEventNotification.setEventDateTime(dateTimeService.convertToSqlDate(concurEventNotificationDTO.getEventDateTime()));
+        } catch (ParseException e) {
+            concurEventNotification.setEventDateTime(dateTimeService.getCurrentSqlDate());
```
bhutchinson: Should this maybe log an error or warning that the event date time was in an invalid format? And we're ok with defaulting to today?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
db483: I will update it to use ObjectUtils.isNotNull. Thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
db483: Good point. Formatter issue. Are you using a template in your environment?

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotification.java
db483: Good point, I will update it.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
db483: That is a great question! I will double check if there is any message they expect when account is valid.

src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseDetailedReportDTO.java
db483: I was still working on figuring out what the fields will be. This will be implemented in the next Sprint. It seemed better to have retrieveExpenseDetailedReportFromConcur signature return a ExpenseDetailedReportDTO even if it is just empty for now.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: yes, current account validation cannot be used. A new service will be implemented to do account validation. I will be working on that as a separate story.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
db483: This event date time will be used to fail requests that could not be processed on our side in 48 hours (still need to create a story for this, maybe we an talk about this after the break). That is the time the time the event notification was generated. It seemed having it defaulted to current time rather than failing the whole thing would make more sense. But maybe I should send back a failure response in this case?

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
db483: I have checked the specs and there is no message specified for the case when the account info is valid. When account info is valid the next step is to approve the Expense Report. I have sent the question to Sherry to see if a message would be needed. I can update this when the account validation with error messages is implemented in Sprint 16 if there is a need for it.

db483:
I have updated the code and this is ready for review again. Thank you.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
db483: Sherry has replied that no message is needed when account info is valid, the report just moves along the workflow to completed.

db483:
Hi Bryan,

Do you have any further suggestions for improvement on this PR? Thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
bhutchinson: I think I do have a template, but I'm using Intellij IDEA so I'm not sure it's sharable if you're using Eclipse. For reference, here are the Coding Standards - IDE Settings from the product team: https://kualico.atlassian.net/wiki/x/DgDrAg

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+                ConcurAccountInfo concurAccountInfo = extractAccountInfo(expenseDetailedReport);
+                concurAccountValidationService.validateConcurAccountInfo(concurAccountInfo);
+                updateExpenseReportStatusInConcur();
+            }
+        }
+    }
+
+    @Override
+    public ExpenseDetailedReportDTO retrieveExpenseDetailedReportFromConcur(ConcurEventNotification concurEventNotification) {
+        //TODO add implementation
```
bhutchinson: FWIW, I don't like TODO comments. I would much prefer having a JIRA for the future work and leaving the comment out of the code.

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotification.java
```diff
@@ -0,0 +1,99 @@
+package edu.cornell.kfs.concur.businessobjects;
+
+import java.sql.Date;
+
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+
+public class ConcurEventNotification extends PersistableBusinessObjectBase{   
+    protected int concurEventNotificationId;
```
bhutchinson: It looks like the serialVersionUID was removed; was that intentional?

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
bhutchinson: Ok, maybe I'm just confused. Usually toString, would output the string representation of the object for things like logging, in which case I would expect the value of valid to be included somehow. I understand if no message is specified when the account info is valid, but how does that relate to the implementation of toString? Is toString being used somewhere in the logic to display/send the ValidationResults messages?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
db483: thank you. I have already updated this code but I will check the kuali co page for future code.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public ConcurEventNotification convertConcurEventNotification(ConcurEventNotificationDTO concurEventNotificationDTO) {
+        ConcurEventNotification concurEventNotification = new ConcurEventNotification();
+        concurEventNotification.setContext(concurEventNotificationDTO.getContext());
+        try {
+            concurEventNotification.setEventDateTime(dateTimeService.convertToSqlDate(concurEventNotificationDTO.getEventDateTime()));
+        } catch (ParseException e) {
+            LOG.warn("Incoming EventDateTime was invalid and it was reset to the current date");
+            concurEventNotification.setEventDateTime(dateTimeService.getCurrentSqlDate());
```
bhutchinson: I think this is fine for now, but we should talk about this logic a little more and make sure it is really what we want. If not, we can change it when working on a future story.

bhutchinson:
Sorry for the delay. I've reviewed the latest and added my questions/comments.

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotification.java
db483: I don't think the serialVersionUID is in fact needed in this case, it was intentionally removed.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
db483: I think I better understand your concern now. Yes, my plan was to use the toString method to get all the error messages in a formatter String that is sent back to Concur. I think it would be best to have a new method implemented for this and not use toString. I will update the code.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
db483: so maybe we should not accept the request if the event date time is not correctly formatted and throw an exception here and return a 406 response? I will check with the Concur developer as well to see how this is usually handled. thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
db483: Fair enough. I thought adding TODOs will eliminate the confusion on what may be erroneous code and what are just stubs waiting to be implemented in future jiras. Jira's already exist for this work.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
db483: I changed the code to throw the ParseException. when the end point calls this conversion method, if a ParseException occurs I throw a javax.ws.rs.BadRequestException. This will return a n HTTP 400. Since a valid date is expected in the incoming xml I think a 400 error would be appropriate in this case.

db483:
Hi Bryan, I have updated the code and it is now ready for review again. Thank you!

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
bhutchinson: ok, that makes sense, thanks

bhutchinson:
Changes look good and unit tests have passed I will merge this (and the corresponding nonprod-sql PR).

PR #255, opened 2017-01-04
KFSPTS-7178 moving utility classes to service architecture

jhulslander:


jhulslander:
This is ready for review, it just moves a few remaining utility classes into services.  They will be refactored when we turn on the end points they are used in.

src/main/java/edu/cornell/kfs/paymentworks/batch/PaymentWorksUploadSuppliersStep.java
```diff
@@ -33,9 +33,10 @@
 import edu.cornell.kfs.paymentworks.batch.report.SupplierUploadSummaryLine;
 import edu.cornell.kfs.paymentworks.businessobject.PaymentWorksVendor;
 import edu.cornell.kfs.paymentworks.service.PaymentWorksKfsService;
+import edu.cornell.kfs.paymentworks.service.PaymentWorksSupplierConversionService;
 import edu.cornell.kfs.paymentworks.service.PaymentWorksVendorService;
 import edu.cornell.kfs.paymentworks.service.PaymentWorksWebService;
-import edu.cornell.kfs.paymentworks.util.PaymentWorksSupplierConversionUtil;
+import edu.cornell.kfs.paymentworks.service.impl.PaymentWorksSupplierConversionServiceImpl;
```
nkimble: edu.cornell.kfs.paymentworks.service.PaymentWorksSupplierConversionService is imported on line 36. 
Why is the implementation for this service also being imported?

src/main/java/edu/cornell/kfs/paymentworks/service/impl/PaymentWorksWebServiceImpl.java
```diff
@@ -57,9 +57,7 @@
 
 @Transactional
 public class PaymentWorksWebServiceImpl implements PaymentWorksWebService {
-
-	private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger
-			.getLogger(PaymentWorksWebServiceImpl.class);
+	private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(PaymentWorksWebServiceImpl.class);
 
 	private static final String NEW_VENDOR_REQUEST_PENDING_VENDORS = "new-vendor-requests/";
 	private static final String NEW_VENDOR_REQUEST_UPDATE_STATUS = "new-vendor-requests/bulk/";
```
nkimble: I realize this code was not modified with these changes but should directory structures be hard coded this way?  Wouldn't it be better to have these two items in a properties file that allows for easier modification later on?

src/main/java/edu/cornell/kfs/paymentworks/batch/PaymentWorksUploadSuppliersStep.java
jhulslander: Good catch, I will remove this import

src/main/java/edu/cornell/kfs/paymentworks/service/impl/PaymentWorksWebServiceImpl.java
jhulslander: these are URL rest call paths. Perhaps they would be better be handled in another way.  I will put a story in the backlog to review that, and do it consistently with what we are doing amazon and concur.

nkimble:
Overall the changes look good. Once a few questions I had are answered, I will re-review.

jhulslander:
retest this please

jhulslander:
This is ready for review again .... I removed the un-needed import and answered teh other two questions.

nkimble:
Unneeded import was removed. 
Questions were answered as being issues that would be addressed during later refactoring efforts of those services when the developer turns on the end points where/when the service are used. Merging code based on that.

PR #258, opened 2017-01-10
KFSPTS-7251: create custom login filter

bhutchinson:


db483:
Hi Bryan, changes look good, I had one small question about some formatting.

I do however wonder if this was also supposed to be configured in the web.xml file? I don't understand how this filter is configured. Thank you.

src/main/java/edu/cornell/kfs/web/filter/CornellLoginFilter.java
```diff
@@ -0,0 +1,50 @@
...
+
+        String remoteUser = getRemoteUser(httpRequest);
+        HttpServletRequestWrapper request =  new HttpServletRequestWrapper(httpRequest) {
+            @Override
+            public String getRemoteUser() {
+                return remoteUser;
+            }
+        };
+
+        filterChain.doFilter(request,httpResponse);
```
bhutchinson: Ideally, but since this is the only comment on this review, I'm willing to live with it if you are.

bhutchinson:
Thanks Daniela. The filter is configured through a combination of the login.filter.className (set to the new CornellLoginFilter class) and the Delegating Authentication Filter (which was created after 5.3.1). More information is in the Kuali Documentation Confluence space: https://kualico.atlassian.net/wiki/x/oYAZ. Please let me know if you have any other questions.

src/main/java/edu/cornell/kfs/web/filter/CornellLoginFilter.java
db483: fair enough

db483:
thank you for explaining this. Code looks good, will merge.

bhutchinson:
Thanks Daniela

PR #262, opened 2017-01-12
KFSPTS-7184 add refresh concur access token funcionality

db483:


bhutchinson:
Looks like you may need to add a concur.refresh.access.token.url property to the upgrade unit test pr job security properties file to help the unit tests pass.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -3,7 +3,18 @@
 public class ConcurConstants {
     public static final String AUTHORIZATION_PROPERTY = "Authorization";
     public static final String BASIC_AUTHENTICATION_SCHEME = "Basic";
-    
+    public static final String OAUTH_AUTHENTICATION_SCHEME = "OAuth ";
```
bhutchinson: mostly out of curiosity - should there be a space in this constant?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -3,7 +3,18 @@
...
+
     public static final int VALIDATION_RESULT_MESSAGE_MAX_LENGTH = 2000;
 
+    public static final String REFRESH_TOKEN_URL_PARAM = "refresh_token=";
+    public static final String CLIENT_ID_URL_PARAM = "&client_id=";
+    public static final String CLIENT_SECRET_URL_PARAM = "&client_secret=";
```
bhutchinson: this is probably ok, but usually, we would set constants like this to just the string (ex: "refresh_token") and then concatenate them with more global constants for equals (=) and ampersand (&)

src/main/java/edu/cornell/kfs/concur/service/ConcurAccessTokenService.java
```diff
@@ -0,0 +1,16 @@
+package edu.cornell.kfs.concur.service;
+
+
+public interface ConcurAccessTokenService {
+    
+    String getConcurAccessToken();
```
bhutchinson: Since the Interface has Concur in it, I wonder if Concur is also necessary in these method names.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import edu.cornell.kfs.concur.ConcurPropertyConstants;
+import edu.cornell.kfs.concur.rest.xmlObjects.AccessTokenDTO;
+import edu.cornell.kfs.concur.service.ConcurAccessTokenService;
+import edu.cornell.kfs.sys.businessobject.WebServiceCredential;
+import edu.cornell.kfs.sys.service.WebServiceCredentialService;
+
+public class ConcurAccessTokenServiceImpl implements ConcurAccessTokenService {
+    private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurAccessTokenServiceImpl.class);
+
+    protected String concurRefreshAccessTokenURL;
```
bhutchinson: Could this string variable be private even if the service is protected (which we normally do because of possible overriding)?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        try {
+            uri = new URI(getConcurRefreshAccessTokenURL()
+                    + ConcurConstants.REFRESH_TOKEN_URL_PARAM
+                    + getConcurRefreshToken()
+                    + ConcurConstants.CLIENT_ID_URL_PARAM
+                    + getConcurConsumerKey()
+                    + ConcurConstants.CLIENT_SECRET_URL_PARAM
+                    + getConcurSecretKey());
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
```
bhutchinson: It might be helpful to include a message along with the exception when creating this new RuntimeException to add context that could help with possible troubleshooting.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        ClientRequest request = builder.build(uri, HttpMethod.GET);
+        return request;
```
bhutchinson: If this variable request is never used, I would probably remove it and return build.build(uri, HttpMethod.GET) directly.

bhutchinson:
retest this please

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
db483: Probably I should remove the space from this constant and concatenate it when building the string.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
db483: This makes sense, I will update the code.

src/main/java/edu/cornell/kfs/concur/service/ConcurAccessTokenService.java
db483: Good point, I will update the names.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
db483: private makes sense, I will update.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
db483: Will add a message, thank you.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
db483: that would make sense, I will update the code.

db483:
I updated the code and is ready for review. thank you.

PR #267, opened 2017-01-17
KFSPTS-7078 add concur account validation service

db483:


src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -6,7 +6,23 @@
     public static final String OAUTH_AUTHENTICATION_SCHEME = "OAuth";
 
     public static final int VALIDATION_RESULT_MESSAGE_MAX_LENGTH = 2000;
-
+    
+    public static class AccountingStringValidationErrorMessages {
+        public static final String ERROR_CHART_OF_ACCTS_REQUIRED = "Chart is a required field.";
```
bhutchinson: Normally, we would put an error message in our properties file and have a corresponding key constant used to retrieve it from there and build it with token replacement if desired. Should we follow that pattern here? Also, some of these messages look pretty similar to existing messages in the system, can we reuse them?

If not, and we decide to keep these as-is, I would probably spell out the words in the constant name to help improve readability.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,218 @@
...
+    protected AccountService accountService;
+    protected ObjectCodeService objectCodeService;
+    protected SubAccountService subAccountService;
+    protected SubObjectCodeService subObjectCodeService;
+    protected ProjectCodeService projectCodeService;
 
     @Override
     public ValidationResult validateConcurAccountInfo(ConcurAccountInfo concurAccountInfo) {
-        ValidationResult validationResult = new ValidationResult(false, new ArrayList<String>());
+        ValidationResult validationResult = checkAccountingString(
```
bhutchinson: I would consider eliminating this intermediate variable, validationResult, and just returning the result of the checkAccountingString() call directly.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,218 @@
...
+        ValidationResult requiredFieldsValidationResult = checkRequiredAccountInfo(chartOfAccountsCode, accountNumber, objectCode);
+
+        if (requiredFieldsValidationResult.isNotValid()) {
+            return requiredFieldsValidationResult;
+        } else {
+            ValidationResult accountValidationResult = checkAccount(chartOfAccountsCode, accountNumber);
+            if (accountValidationResult.isNotValid()) {
+                return accountValidationResult;
+            } else {
+                ValidationResult objectCodeValidationResult = checkObjectCode(chartOfAccountsCode, objectCode);
```
bhutchinson: This is a lot of code that could probably be extracted into a separate method.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,218 @@
...
+        }
+    }
+
+    public ValidationResult checkRequiredAccountInfo(String chartOfAccountsCode, String accountNumber, String objectCode) {
+        ValidationResult validationResult = new ValidationResult(true, new ArrayList<String>());
+        if (chartOfAccountsCode == null || chartOfAccountsCode.isEmpty()) {
+            validationResult.setValid(false);
+            validationResult.addMessage(ConcurConstants.AccountingStringValidationErrorMessages.ERROR_CHART_OF_ACCTS_REQUIRED);
+            return validationResult;
+        } else if (accountNumber == null || accountNumber.isEmpty()) {
```
bhutchinson: I'm wondering if these else if's are needed or if they should just be separate if blocks. As is, if the accountNumber is missing, that message won't be returned if the chart is missing too. Wouldn't it be helpful to have all the messages? Or are we limited to one message to return?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,218 @@
...
+        }
+        return validationResult;
+    }
+
+    public ValidationResult checkObjectCode(String chartOfAccountsCode, String objectCodeParm) {
+        ValidationResult validationResult = new ValidationResult(true, new ArrayList<String>());
+        
+        ObjectCode objectCode = objectCodeService.getByPrimaryIdForCurrentYear(chartOfAccountsCode, objectCodeParm);
+    
+        if (objectCode == null || objectCode.toString().isEmpty()) {
```
bhutchinson: I don't know if this would make the code cleaner or not, but this block is pretty similar between most of the accounting string components. I think you could extract it into a method with three parameters (Inactivatable object, String missingMessage, String inactiveMessage) with common logic for all of them.

src/test/java/edu/cornell/kfs/concur/services/ConcurAccountValidationServiceTest.java
```diff
@@ -0,0 +1,444 @@
...
+    public void isAccountValid(){
+        ValidationResult validationResult = concurAccountValidationService.checkAccount(ConcurAccountValidationTestConstants.VALID_CHART, ConcurAccountValidationTestConstants.VALID_ACCT_NBR);
+        assertEquals("Validation was expected to pass", true, validationResult.isValid());
+        assertEquals("No error messages expected", 0, validationResult.getMessages().size());    
+    }
+    
+    @Test
+    public void isAccountNull(){
+        ValidationResult validationResult = concurAccountValidationService.checkAccount(ConcurAccountValidationTestConstants.VALID_CHART, null);
+        assertEquals("Validation was expected to fail but returned true", false, validationResult.isValid());
```
bhutchinson: This is nitpicky, but you could use assertFalse instead of assertEquals for these.

src/test/java/edu/cornell/kfs/concur/services/ConcurAccountValidationServiceTest.java
```diff
@@ -0,0 +1,444 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.kuali.kfs.sys.KFSConstants;
+
+import edu.cornell.kfs.concur.ConcurConstants;
+import edu.cornell.kfs.concur.businessobjects.ConcurAccountInfo;
+import edu.cornell.kfs.concur.businessobjects.ValidationResult;
+import edu.cornell.kfs.concur.service.impl.ConcurAccountValidationServiceImpl;
+import static org.junit.Assert.assertEquals;
```
bhutchinson: Per Kuali standards, we would like to avoid static imports. We can just import the Assert class instead and change the method calls to Assert.assertEquals(...).

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
db483: Thank you for pointing this out. I was wondering if it would make more sense to use the KFS error messages in the properties files or use constants. I used constants because initially I thought that if we move this into a micro-service that calls KFS endpoints to check on whether accounts, object codes, etc exist it would be better not to depend on the KFS error messages in the properties files. After further discussions about possible micro-services I think we talked that perhaps the validation service could be an end point in KFS in which case using the KFS error messages in the properties files would make sense. What do you think?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: good point, I will update this.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: I will do that.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: Good point, I don't think we need the else if. I mostly used the code from the existing account verification web service and did not think that this would be better without the else if. I will update it.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: I will work on extracting this into a separate method, thank you for the suggestion.

src/test/java/edu/cornell/kfs/concur/services/ConcurAccountValidationServiceTest.java
db483: will update, thank you.

src/test/java/edu/cornell/kfs/concur/services/ConcurAccountValidationServiceTest.java
db483: will update, thank you.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
bhutchinson: I agree that the validation service could be an end point in KFS and using the error messages from the properties file would make sense in that case, thanks.

db483: thank you, I will update it.

db483:
Hi Bryan, I have made the requested changes and this is ready for review. Thank you.

src/main/java/edu/cornell/kfs/sys/service/impl/ErrorMessageUtilsServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.rice.core.api.config.property.ConfigurationService;
+
+import edu.cornell.kfs.sys.service.ErrorMessageUtilsService;
+
+public class ErrorMessageUtilsServiceImpl implements ErrorMessageUtilsService{
+    protected ConfigurationService configurationService;
+    
+    public String createErrorString(String errorKey, String... params) {
```
bhutchinson: I'm trying to understand the purpose of this. Do you anticipate sending in errorKey values that aren't actually keys but error messages themselves?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,173 @@
...
+        ValidationResult requiredFieldsValidationResult = checkRequiredAccountInfo(chartOfAccountsCode, accountNumber, objectCode);
+
+        if (requiredFieldsValidationResult.isNotValid()) {
+            return requiredFieldsValidationResult;
+        } else {
+            ValidationResult accountValidationResult = checkAccount(chartOfAccountsCode, accountNumber);
+            if (accountValidationResult.isNotValid()) {
+                return accountValidationResult;
+            } else {           
+                ValidationResult validationResult = new ValidationResult(true, new ArrayList<String>());
```
bhutchinson: I would say this section could still be extracted into another method and then there would just be the one line call in this else. If there aren't other changes to make, I probably wouldn't insist on just changing this though.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -2,17 +2,173 @@
...
+        }
+        if (objectCode == null || objectCode.isEmpty()) {
+            validationResult.setValid(false);
+            validationResult.addMessage(errorMessageUtilsService.createErrorString(KFSKeyConstants.ERROR_REQUIRED, ConcurConstants.AccountingStringFieldNames.OBJECT_CODE));
+        }
 
         return validationResult;
     }
     
+    private ValidationResult checkMissingOrInactive(Inactivatable inactivatableObject, String missingMessage, String inactiveMessage){
```
bhutchinson: Not a big deal, but from a clean code perspective, I would probably move this method below where it is first called.

src/main/java/edu/cornell/kfs/sys/service/impl/ErrorMessageUtilsServiceImpl.java
db483: I am not sure I understand the question. Is there a easier way to do this?

I have basically copied this from AttributeValidatingTypeServiceBase and extracted it in this utility service that can be used outside of KFS documents too.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: that makes sense, I will extract this in a separate method. thank you.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
db483: will move it. thank you.

db483:
Hi Bryan,

I have updated the code based on your latest comments and it is now ready for your review again. Thank you!

PR #268, opened 2017-01-17
KFSPTS-7244: pull in Sprint 16 changes (batch file lookup performance), cleanup pom and fix errors

bhutchinson:


bhutchinson:
@jhulslander, I think this one should probably be a merge commit instead of a squash merge since we probably want to preserve the individual commits I cherry-picked from develop to be consistent with what's on master.

src/main/java/edu/cornell/kfs/sys/service/BatchFileDirectoryService.java
```diff
@@ -0,0 +1,15 @@
...
+import org.kuali.kfs.sys.businessobject.SystemOptions;
+import org.kuali.rice.core.api.util.KeyValue;
+import org.springframework.cache.annotation.Cacheable;
+
+import java.io.File;
+import java.util.List;
+
+public interface BatchFileDirectoryService {
+
+    @Cacheable(value= SystemOptions.CACHE_NAME, key="'buildDirectoryKeyValuesList' + #p0")
```
jhulslander: Just out of curiosity, what does the #p0 do for the cache key?

src/test/resources/edu/cornell/kfs/sys/businessobject/lookup/fixture/sub1/testlookupfile1.txt
```diff
@@ -0,0 +1 @@
+just a placeholder file for testing
```
jhulslander: Probably not worth changing since it is already in prod, but I would suggest having the unit create these test files, and then delete them after the test have completed.  It feels better than keeping files like this in our code repo.  What are your thoughts on this?

src/main/java/edu/cornell/kfs/sys/service/impl/BatchFileDirectoryServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BatchFileDirectoryServiceImpl implements BatchFileDirectoryService {
+
+    @Override
+    public List<KeyValue> buildDirectoryKeyValuesList(List<File> rootDirectories) {
```
jhulslander: Would it make more sense to define the "Cacheable" on the implementation rather than on the interface?

src/main/java/edu/cornell/kfs/sys/service/BatchFileDirectoryService.java
bhutchinson: That matches the parameter to this method (0 for first; in this case there is only one). So the cache key includes the parameter passed to the method so different parameters will have different cache entries.

src/test/resources/edu/cornell/kfs/sys/businessobject/lookup/fixture/sub1/testlookupfile1.txt
bhutchinson: They are essentially test fixtures - I don't have a problem keeping them in our code repo. Off the top of my head I feel like adding code to create and delete them overcomplicates things.

src/main/java/edu/cornell/kfs/sys/service/impl/BatchFileDirectoryServiceImpl.java
bhutchinson: Why? Are you concerned that some implementations may be cacheable and some may not be? I saw other places (mostly Rice I think), where the annotation was on the interface, so I followed that pattern here.

jhulslander: The reason for my question was essentially I would consider caching an implementation decision, so by that way of thinking my first thought would be to put in the impl.  However, consistency with Rice seems most appropriate, so we should go with that.

bhutchinson: Yeah, we'll see what the product team says when we contribute these changes.

bhutchinson:
Thanks Jay

PR #269, opened 2017-01-18
KFSPTS-7262: Fix DI doc spinning on selection due to non-existent travel instance.

nkimble:


bhutchinson:
Good catch, @jhulslander. Just to be sure, we should probably check with Cathy to confirm how a copied DI (with an associated trip) should behave.

jhulslander:
Double checking with Cathy sounds good to me.
If a DI associated with a trip does get copied and the trip association values are copied as well, I think there may be some issues we need to consider.  We can dive deeper if we need to.

Oh, and I forgot to include the tripID value should be nulled in the toCopy function as well.

bhutchinson:
Also, if it can be done quickly, maybe that's ok, but technically this was only to handle initiated DIs, so copied DIs would be a bit of scope creep.

nkimble:
I talked with Cathy and reviewed the code pertaining to the DI toCopy functionality.  
Per Cathy: The DI should function the way it currently does when a copy is performed. Meaning the link that shows in the DI from which the copy is made disappears in the copy.  The DI toCopy method was already clearing the trip id. I just needed to set the tripAssociationStatusCode to the value that indicates it is not a trip.  Change is in place and ready for re-review.

jhulslander:
Looks good, thanks, will merge

PR #272, opened 2017-01-19
KFSPTS-6979 add missing payload element, log additional info, change access type to NONE, update date field

db483:


src/main/java/edu/cornell/kfs/concur/rest/resource/ConcurAccountValidationResource.java
```diff
@@ -28,6 +29,7 @@ public Response validate(ConcurEventNotificationDTO concurEventNotificationDTO)
         try {
             concurEventNotification = getConcurEventNotificationConversionService().convertConcurEventNotification(concurEventNotificationDTO);
         } catch (ParseException e) {
+            LOG.error(e.getMessage());
```
jhulslander: This should probably include the function name in the error message.  Also, should the error be logged as well as the message?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
```diff
@@ -15,10 +15,13 @@
     public ConcurEventNotification convertConcurEventNotification(ConcurEventNotificationDTO concurEventNotificationDTO) throws ParseException {
         ConcurEventNotification concurEventNotification = new ConcurEventNotification();
         concurEventNotification.setContext(concurEventNotificationDTO.getContext());
-        concurEventNotification.setEventDateTime(dateTimeService.convertToSqlDate(concurEventNotificationDTO.getEventDateTime()));
+        if(concurEventNotificationDTO.getEventDateTime() != null){
```
jhulslander: should this use objectUtils.isNotNull?

src/main/java/edu/cornell/kfs/concur/rest/resource/ConcurAccountValidationResource.java
db483: Good point, I will update this.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationConversionServiceImpl.java
db483: This is a field in the incoming payload xml so checking for null is enough, this is not a check on a proxied object that may or may not be materialized yet so ObjectUtils would not work here.

jhulslander: OK, thanks

db483:
Hi Jay,

I have updated the code and it is now ready for review again. Thank you!

db483:
Hi Jay,

I have updated the code and it is now ready for review again. Thank you!

PR #273, opened 2017-01-24
KFSPTS-7155: add lastUpdateTimestamp to BOs with maint docs

bhutchinson:


bhutchinson:
This is ready to review again. There are mixed tabs/spaces all throughout those files. That makes the formatting look off sometimes. Perhaps we could go through and reformat them as a future cleanup task, but not along with any other work so as not to make the review of the change diff too difficult.

PR #278, opened 2017-01-30
KFSPTS-7403 when validation errors happen we still save the payment works request information.  Also did some small clean up tasks

jhulslander:


jhulslander:
thank you

PR #280, opened 2017-01-30
KFSPTS-7338: fix bugs preventing Award from being opened

bhutchinson:


bhutchinson:
@cah292 We should be able to expect the unit tests to pass now, thanks.

PR #281, opened 2017-01-31
KFSPTS-7346: fix file location so e-shop cart can be returned to KFS

bhutchinson:


db483:
Change look good to me, will merge once unit tests finish running.

PR #284, opened 2017-01-31
KFSPTS-7351: fix DV copy so vendors aren't cleared incorrectly

bhutchinson:


bhutchinson:
retest this please

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -652,6 +653,26 @@ protected void clearFieldsThatShouldNotBeCopied() {
         setTripId(null);
     }
 
+    /**
+     * This overrides the code in the parent base class because of an issue where vendorDetail is null when it shouldn't be.
+     * Might be related to OJB and proxy objects or something like that.
+     * Hopefully we can contribute a fix to base code and eliminate this duplicate method in the future.
+     */
+    protected void clearInvalidPayee() {
```
jhulslander: Should we add an override annotation to this function?

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -639,7 +641,6 @@ public void toCopy() throws WorkflowException {
 
         super.toCopy();
 
-        KNSGlobalVariables.getMessageList().clear();
```
jhulslander: Just out of curiosity,  was clearing the message list causing part of the problem?

bhutchinson: It wasn't causing the part where stuff was being cleared that shouldn't, but was causing an issue where messages from the clearing method weren't being displayed like they should be.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
bhutchinson: probably; I can fix that

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
jhulslander: ah, thanks

bhutchinson:
In addition to the one small change, there is one failing test related to this change. It doesn't fail locally, so I'm still trying to figure out what the issue is. We should wait to merge until the tests are green.

src/test/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocumentIntegrationTest.java
```diff
@@ -91,6 +93,12 @@ public void testToCopy() throws WorkflowException {
         assertNull(cuDisbursementVoucherDocument.getCancelDate());
         assertEquals(KFSConstants.DocumentStatusCodes.INITIATED, cuDisbursementVoucherDocument.getFinancialSystemDocumentHeader().getFinancialDocumentStatusCode());
 
+        MessageList messageList = KNSGlobalVariables.getMessageList();
+        System.err.println("KNSGlobalVariables.getMessageList().size(): " + messageList.size());
```
jhulslander: I don't think we want a system.err message here do we?

src/test/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocumentIntegrationTest.java
```diff
@@ -91,6 +93,12 @@ public void testToCopy() throws WorkflowException {
...
+        for (ErrorMessage errorMessage: messageList) {
+            System.err.println("errorMessage.getErrorKey(): " + errorMessage.getErrorKey());
```
jhulslander: another system.err .... shouldn't these be LOG.error?

src/test/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocumentIntegrationTest.java
bhutchinson: It's temporary to help figure out why this test fails in Jenkins but not locally for me.

src/test/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocumentIntegrationTest.java
bhutchinson: again, temporary

bhutchinson:
This should be ready for final review now. The unit test failure is fixed and the temporary debugging is cleaned up.

PR #285, opened 2017-02-01
KFSPTS-7077 retrieve report details from Concur and validate account info

db483:


src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurAccountInfo.java
```diff
@@ -64,4 +76,27 @@ public void setProjectCode(String projectCode) {
     public void setOrgRefId(String orgRefId) {
         this.orgRefId = orgRefId;
     }
+    
+    @Override
+    public String toString() {
+        StringBuffer result = new StringBuffer();
```
jhulslander: You may want to use StringBuilder here instead of StringBuffer.  StringBuilders are thread safe, so if this is single threaded, StringBuilder will perform a bit better.  Chad H suggested I start using StringBuilder.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccountValidationServiceImpl.java
```diff
@@ -92,7 +95,8 @@ public ValidationResult checkRequiredAccountInfo(String chartOfAccountsCode, Str
 
     public ValidationResult checkAccount(String chartOfAccountsCode, String accountNumber) {
         Account account = accountService.getByPrimaryId(chartOfAccountsCode, accountNumber);
-        return checkMissingOrInactive(account, MessageFormat.format(configurationService.getPropertyValueAsString(KFSKeyConstants.ERROR_EXISTENCE), ConcurConstants.AccountingStringFieldNames.ACCOUNT_NUMBER), MessageFormat.format(configurationService.getPropertyValueAsString(KFSKeyConstants.ERROR_INACTIVE), ConcurConstants.AccountingStringFieldNames.ACCOUNT_NUMBER));
+        String  accountErrorMessageString = ConcurConstants.AccountingStringFieldNames.ACCOUNT_NUMBER + CUKFSConstants.COLON + chartOfAccountsCode + KFSConstants.COMMA + accountNumber;
```
jhulslander: there are a few cases where you are building an errorMessageString using a similar format: fieldName:value,value .  Would it be helpful to break that into a function?  I am asking about this.

db483:
Thank you very much for reviewing this Jay. I am working on updating the code based on your suggestions.

db483:
Hi Jay,

I have updated the code based on your suggestions and it is now ready for your review again. Thank you!

PR #287, opened 2017-02-01
KFSPTS-7336: Added fix to remove certain empty DynamoDB preferences.

cah292:
KFS's REST preferences endpoint is adding certain empty preferences by default, which don't appear to be used anywhere else yet but are causing problems with saving them to DynamoDB. This PR adds a fix adapted from one developed by Bryan, which will remove the empty default preferences for now. This may need revisiting in the future if a KFS patch ends up adding functionality for these new preferences.


bhutchinson:
My understanding from talking to the Financials Product Team is that these empty fields were added to setup a foundation for future analytics work. However, that work is on hold now and may take a different shape in the future. As stated above, we will want to keep an eye on this and if these fields are used in the future adjust our code as needed.

PR #292, opened 2017-02-06
KFSPTS-7538: Fixed transaction issue with doc-related KFS batch jobs.

cah292:
For some reason, KFS 7.x is not performing proper transaction interception for DocumentService operations, even though DocumentServiceImpl has the "TransactionalNoValidationExceptionRollback" annotation. KFS 5.x was using the Rice-delivered DocumentDaoProxy around the OJB DAO (where the proxy had the "Transactional" annotation), so perhaps that was why we didn't encounter issues in the past.

This PR introduces a workaround similar to the one in use by Rice and earlier KFS versions. Rather than creating an explicit DAO proxy class, though, this fix just creates a DAO subclass that has the "Transactional" annotation on it, and overrides methods as needed to allow the related Spring/Rice utilities to perform interception accordingly.

I went with this workaround since it's closer to what base Rice code uses, and to also avoid messing with annotations on the DocumentService until Kuali can examine the issue further. If you think this warrants further investigation on our end to come up with a different solution, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/krad/dao/impl/CuDocumentDaoOjb.java
+++ b/src/main/java/edu/cornell/kfs/krad/dao/impl/CuDocumentDaoOjb.java
@@ -1,0 +1,53 @@
+package edu.cornell.kfs.krad.dao.impl;
+
+import java.util.List;
+
+import org.kuali.kfs.krad.dao.BusinessObjectDao;
+import org.kuali.kfs.krad.dao.impl.DocumentDaoOjb;
+import org.kuali.kfs.krad.document.Document;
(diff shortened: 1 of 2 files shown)


cah292:
After further discussion with @bhutchinson, I will look into other alternatives for fixing this, possibly by annotating the batch-related services instead. I'm closing this for now while I investigate other solutions and while Bryan follows up with the product team.

cah292:
As per feedback from @bhutchinson and product team, a "Transactional" annotation will eventually be placed back on the document DAO. We will move forward with our version of the change, and then remove it once things are fixed in base code.

cah292:
retest this please

PR #293, opened 2017-02-08
KFSPTS-7079 update report status in Concur

db483:


src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -25,10 +25,20 @@
...
+
     public static final String DEFAULT_TRAVEL_REQUEST_OBJECT_CODE_PARAMETER_NAME = "DEFAULT_TRAVEL_REQUEST_OBJECT_CODE";
 
+    public static final String EXPENSE_AWAIRING_EXTERNAL_VALIDATION_STATUS_CODE = "A_EXTV";
+    public static final String REQUEST_AWAIRING_EXTERNAL_VALIDATION_STATUS_CODE = "Q_EXTV";
+
+    public static final String APPROVE_ACTION = "Approve";
+    public static final String SEND_BACK_TO_EMPLOYEE_ACTION = "Send Back To Employee";
+    public static final String APPROVE_COMMENT = "Approved via Concur Connect";
+    public static final String EXPENSE_WORKFLOW_UPDATE_NAMESPACE = "http://www.concursolutions.com/api/expense/expensereport/2011/03";
```
nkimble: Won't this value have the possibility of changing over time requiring rebuilds and redeploys?  Would this item be better served as a configuration parameter instead of a constant?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -25,10 +25,20 @@
...
+    public static final String REQUEST_WORKFLOW_UPDATE_NAMESPACE = "http://www.concursolutions.com/api/travelrequest/2012/06";
```
nkimble: Same comment as above. Shouldn't this be a configuration parameter?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
```diff
@@ -35,35 +36,49 @@ public void processConcurEventNotifications() {
...
+                    validationResult.addMessage(configurationService.getPropertyValueAsString(ConcurKeyConstants.INCORRECT_CONCUR_STATUS_CODE));
                 }
                 
-                LOG.info("Update report status in Concur");
-                concurReportsService.updateExpenseReportStatusInConcur(concurReport.getWorkflowURI(), validationResult);  
-                LOG.info("Update ConcurEventNotification flags and validationMessage in KFS database");
-                concurEventNotificationService.updateConcurEventNotificationFlagsAndValidationMessage(concurEventNotification, false, true, validationResult.isValid(), validationResult.getErrorMessagesAsOneFormattedString());
+                LOG.info("Update ConcurEventNotification flags and validationMessage in KFS database: " + validationResult.isValid() + "," + validationResult.getErrorMessagesAsOneFormattedString());
+                concurEventNotificationService.updateConcurEventNotificationFlagsAndValidationMessage(concurEventNotification, false, true, validationResult.isValid(), validationResult.getErrorMessagesAsOneFormattedString());
```
nkimble: Using local variables with descriptive names would give clarity as to what "false" and "true" are representing in this method call.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
```diff
@@ -59,8 +61,28 @@ public String getErrorMessagesAsOneFormattedString() {
...
         }
-        
-        return result.toString();
+
+        return addMessageHeaderAndTruncate(result.toString(), ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH);
+    }
+
+    private String addMessageHeaderAndTruncate(String message, int maxLength) {
+        String errorMessagesString = addMessageHeader(message);
+        errorMessagesString = truncateMessageLength(errorMessagesString, maxLength);
```
nkimble: Curious about this. I realize there are message limits that need to be adhered to and why you are doing this, but is there an understanding that there could be more errors than what is being returned due to the truncation being performed?  Meaning, will the users somehow understand that they could go through this validation cycle multiple times just because they were not given all the errors the first time?

src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseEntryDTO.java
```diff
@@ -18,6 +18,9 @@
 
     @XmlElement(name = "OrgUnit6")
     private String orgUnit6;
+      
+    @XmlElement(name = "Allocations")
+    List<AllocationsDTO> allocations;
```
nkimble: Should this be marked private?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -25,10 +25,20 @@
...
-    
+
     public static final String EXPENSE_REPORT_URI_INDICATOR = "/expense/expensereport/";
     public static final String TRAVEL_REQUEST_URI_INDICATOR = "/travelrequest/";
-    
+
     public static final String DEFAULT_TRAVEL_REQUEST_OBJECT_CODE_PARAMETER_NAME = "DEFAULT_TRAVEL_REQUEST_OBJECT_CODE";
 
+    public static final String EXPENSE_AWAIRING_EXTERNAL_VALIDATION_STATUS_CODE = "A_EXTV";
+    public static final String REQUEST_AWAIRING_EXTERNAL_VALIDATION_STATUS_CODE = "Q_EXTV";
```
db483: Good catch, will update

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
db483: Good point, I think that it would probably make sense to have this in a config param

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEventNotificationProcessingServiceImpl.java
db483: Good point, will add local variables.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
db483: Good question. The requirements specs specifically asked for the error message to be truncated to 2000 characters. I will check with Sherry Wilson, the BA, to make sure the users understand this. Thank you!

src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseEntryDTO.java
db483: yes, it should, will update it

PR #299, opened 2017-02-14
KFSPTS-7605: Fixed error when performing EIRT doc searches.

cah292:
I needed to add some currency-related config to the EIRT "Grand Total" field so that EIRT doc searches could handle the attribute value properly.


src/main/resources/edu/cornell/kfs/module/purap/document/datadictionary/CuElectronicInvoiceRejectDocument.xml
```diff
@@ -19,5 +19,14 @@
     <property name="documentClass" value="edu.cornell.kfs.module.purap.document.CuElectronicInvoiceRejectDocument"/>
    <!--   <property name="baseDocumentClass" value="edu.cornell.kfs.module.purap.document.ElectronicInvoiceRejectDocument"/> -->
```
nkimble: Could this dead code be removed since we are modifying this file anyway?

cah292: Seems reasonable, since this is a small file and not much is being modified anyway. I'll make that change.

cah292:
I've made some more minor changes based on initial feedback. This is now ready for review again.

cah292:
retest this please

PR #301, opened 2017-02-15
KFSPTS-7625: put back letter of credit fund group

bhutchinson:


bhutchinson:
Good question. I don't know of any plans to remove this db column. Since it and the related changes with LOC Group Fund are related to CGB and I think that is on hold until a partner implements it, and since I believe we would like to look at CGB here at Cornell, I would propose leaving it as is until we have more clarity around any future directions. We may decide we want to move it to an extended attribute, or we may decide we can move our data to another location at a future time.

PR #302, opened 2017-02-15
KFSPTS-7616: Fixed text field widths on PCDO accounting lines.

cah292:
Many of the PCDO accounting line beans are extending from beans that do not have the improved field width setup. This change fixes that by updating the related "parentBean" beans for those whose field widths are too small to accommodate the lookup icons. The overrides were done at this level because the issue is an inheritance-related one.


bhutchinson:
@cah292, when this issue was reported, the user said it was not happening in the KualiCo environment. Based on your comments and this fix it seems like the problem is with base code. Could you take a look and see if the problem is or isn't happening in the KualiCo environment so we can determine if this is a good fix, or if maybe something else is going on here. Thanks.

@nkimble, please hold off on merging this PR until @cah292 can do that analysis. Thanks.

nkimble:
Waiting for additional analysis as requested.

cah292:
I confirmed that the KualiCo environments are not having this problem.

After further analysis, it seems that the problem has to do with a different bean file altogether. Our cu-kfs project has its own copy of KfsBaseAttributeDefinitions.xml that overrides (but does not overlay) the one delivered with KFS. However, our version of the file has out-of-sync changes compared with the KFS one. I will need to investigate further to determine the extent of the changes needed.

I'm closing this PR since the fix will likely not use the new bean files.

bhutchinson:
Thanks Chad!

PR #303, opened 2017-02-15
KFSPTS-7616: Update base DD bean overrides with KFS 7.x changes.

cah292:
Our base DD attributes file was out-of-sync with the KFS 7.x version of it. This brings it up-to-date while still preserving our existing customizations to it.

If you think the file should be modified to only include the things that we have specifically overridden, please let me know.


bhutchinson:
The changes look fine, but yes, if it's not too much effort, I think it would be better if the file could be modified to only include the things that we have specifically overridden, thanks.

cah292:
That makes sense; I'll update the code accordingly. Thanks!

cah292:
I've updated the bean file accordingly based on feedback. This is now ready for review again.

cah292:
retest this please

PR #304, opened 2017-02-15
Kfspts 7617 e invoice

jhulslander:


jhulslander:
This is brings in a copy of electronicInvoice.xsd and overrides purapTypes element with the pre-existing purapTypes.xsd that we had already built.  For some reason, we had lost a mapping to our purapTypes.xsd for einvoice.

PR #318, opened 2017-03-01
KFSPTS-7810 fix autoCancelBatchJob

bhutchinson:


src/main/java/edu/cornell/kfs/sys/dataaccess/impl/AutoCancelBatchDaoJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(AutoCancelBatchDaoJdbc.class);
+
+    private BusinessObjectService businessObjectService;
+    private DocumentService documentService;
+    private ParameterService parameterService;
+    private SessionDocumentService sessionDocumentService;
+    private WorkflowDocumentService workflowDocumentService;
+
+    private Collection<String> cancelDocumentTypes = Collections.emptySet();
```
cah292: Is this field necessary? The code below is just assigning it the values of a parameter if currently empty, and the ParameterService already uses the ParameterRepositoryService's caching to minimize database calls anyway.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/AutoCancelBatchDaoJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+            dbmsOutput = new DbmsOutput(conn);
+            dbmsOutput.enable(1000000);
+            result = cs.execute();
+            dbmsOutput.show();
+        } catch (SQLException e) {
+            throw new RuntimeException("Issue encountered with cancelFYIsAndAcknowledgements", e);
+        } finally {
+            if (dbmsOutput != null) {
+                try {
+                    dbmsOutput.close();
```
cah292: It looks like the current implementation of the DbmsOutput.close() method has the potential to leave some of its internal statements open if one of them throws an exception on close. If fixing that is out of scope for this PR, then we should at least create a new user story to get this fixed. (DbmsOutput is a CU-specific class, so we don't need to worry about fixing it at the KualiCo level.)

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/AutoCancelBatchDaoJdbc.java
```diff
@@ -0,0 +1,206 @@
...
+            String docTypId = (String) result.get("DOC_TYP_ID");
+            LOG.info("Doc ID : " + docId + " Doc Typ ID : " + docTypId);
+            ids.put(docId, docTypId);
+        }
+
+        return ids;
+    }
+
+    private boolean canAutoCancelDocType(String docTypeId) {
+        Map<String, DocumentType> docTypes = new HashMap<String, DocumentType>();
```
cah292: This helper Map only exists locally in this method, but the code below is acting as though the Map is maintaining its state between method calls. If there is a need to temporarily cache these document types, then a new field should be added to this class accordingly.

bhutchinson:
This is ready for review again, thanks.

PR #319, opened 2017-03-01
KFSPTS-7377: New services created to parse Concur Request Extract file into POJO and perform high level header row file validation.

nkimble:
KFSPTS-7377: New services created to parse Concur Request Extract file into POJO and perform high level header row file validation.


nkimble:
Updated code for review with an additional commit that removed references to TransientBusinessObjectBase after talking with Jay to make the business objects really be POJOs.

nkimble:
retest this please

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    protected ConcurRequestExtractFileValidationService concurRequestExtractFileValidationService;
+
+    public List<String> getUnprocessedRequestExtractFiles() {
+        List<String> fileNamesToLoad = getBatchInputFileService().listInputFileNamesWithDoneFile(getBatchInputFileType());
+        return fileNamesToLoad;
+    }
+
+    public boolean processFile(String requestExtractFileName) {
+        boolean processingSuccess = true;
+        Object parsedFile;
```
db483: Since this is not used outside the try I think it would make sense to move this inside the try where the parsedFile is first used

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+public class ConcurRequestExtractFileServiceImpl implements ConcurRequestExtractFileService {
+    private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurRequestExtractFileServiceImpl.class);
+    protected String acceptedDirectoryName;
+    protected String rejectedDirectoryName;
+    protected BatchInputFileType batchInputFileType;
+    protected BatchInputFileService batchInputFileService;
+    protected ConcurRequestExtractFileValidationService concurRequestExtractFileValidationService;
+
+    public List<String> getUnprocessedRequestExtractFiles() {
+        List<String> fileNamesToLoad = getBatchInputFileService().listInputFileNamesWithDoneFile(getBatchInputFileType());
```
db483: Is there any reason why you are using the getter methods instead of the properties themselves? it makes the code more difficult to read and I am not sure I see the value in this case. These properties are injected using spring and they are protected so probably it would make sense that only the setter is defined here.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    public boolean requestExtractHeaderRowValidatesToFileContents(Object parsedFile) {
+        boolean headerValidationPassed = false;
+        List<ConcurRequestExtractFile> requestExtractFiles = (ArrayList<ConcurRequestExtractFile>) parsedFile;
+        if (requestExtractFiles.isEmpty() || requestExtractFiles.size() != 1) {
+            LOG.error("Header validation should have received single parsed file. More or less than one parsed file was detected.");
+        }
+        else {
+            ConcurRequestExtractFile requestExtractFile = requestExtractFiles.get(0);
+            LOG.info("batchDate=" + requestExtractFile.getBatchDate() + "=   recordCount=" + requestExtractFile.getRecordCount() + "=     totalApprovedAmount=" + requestExtractFile.getTotalApprovedAmount() + "=");
+            headerValidationPassed = (getConcurRequestExtractFileValidationService().fileRowCountMatchesHeaderRowCount(requestExtractFile) &&
```
db483: Again, I think it would make sense to use the concurRequestExtractFileValidationService directly.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    public void performRejectedRequestExtractFileTasks(String pathAndFileName) {
+        moveFile(pathAndFileName, getRejectedDirectoryName());
+    }
+
+    public void performAcceptedRequestExtractFileTasks(String pathAndFileName) {
+        moveFile(pathAndFileName, getAcceptedDirectoryName());
+    }
+
+    public void performDoneFileTasks(String requestExtractFileName) {
+        File doneFile = new File(StringUtils.substringBeforeLast(requestExtractFileName, ".") + ".done");
```
db483: I know each batch service we have defines their own method to remove these .done files but I am wondering if you could use some generic service to do this. I see a new FileStorageService was created in 7.x and I wonder if it would make more sense to use that?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        this.concurRequestExtractFileValidationService = concurRequestExtractFileValidationService;
+    }
+
+    public ConcurRequestExtractFileValidationService getConcurRequestExtractFileValidationService() {
+        return this.concurRequestExtractFileValidationService;
+    }
+
+    private Object loadFile(String fileName) {
+        byte[] fileByteContent = safelyLoadFileBytes(fileName);
+        Object parsedObject;
```
db483: these two lines could be combined in one. no need for a declaration on a separate line.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        parsedObject = getBatchInputFileService().parse(getBatchInputFileType(), fileByteContent);
+        return parsedObject;
+    }
+
+    private byte[] safelyLoadFileBytes(String fileName) {
```
db483: I see both you and Jay are using this method to load file bytes and I have seen similar code in other batch jobs. Let's extract this method into a utility service so that it can be re-used. Thank you!

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        } catch (IOException e1) {
+            LOG.error("IO Exception loading: [" + fileName + "]. " + e1.getMessage());
+            throw new RuntimeException("IO Exception loading: [" + fileName + "]. " + e1.getMessage());
+        } finally {
+            IOUtils.closeQuietly(fileContents);
+        }
+        return fileByteContent;
+    }
+
+    public void process(String fileName, Object parsedFileContents) {
```
db483: Since you are not implementing any of these methods (process, validate, getFileName) do you still need to FlatFileDataHandler?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+public class ConcurRequestExtractFileValidationServiceImpl implements ConcurRequestExtractFileValidationService {
+
+    private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurRequestExtractFileValidationServiceImpl.class);
+
+    public boolean fileRowCountMatchesHeaderRowCount(ConcurRequestExtractFile requestExtractFile) {
+        int countEntryDetailLines = 0;
+        List<ConcurRequestExtractRequestDetailFileLine> requestDetailLines = requestExtractFile.getRequestDetails();
```
db483: this code should be extracted in a separate method something like in getTotalEntryDetailLines(). Also, is it possible for this requestExtractFile.getRequestDetails() to be null?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        for (ConcurRequestExtractRequestDetailFileLine detailLine : requestDetailLines) {
+            countEntryDetailLines = countEntryDetailLines + detailLine.getRequestEntryDetails().size();
+        }
+        if ((requestDetailLines.size() + countEntryDetailLines) == requestExtractFile.getRecordCount()) {
```
db483: I would move this computation in a new getTotalDetailLines method and use the result in this comparison.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        else {
+            LOG.error("Header row count validation failed. Header record count was =" +requestExtractFile.getRecordCount()+"= while file line count was =" +(requestDetailLines.size() + countEntryDetailLines)+ "=");
+            return false;
+        }
+    }
+
+    public boolean fileApprovedAmountsMatchHeaderApprovedAmount(ConcurRequestExtractFile requestExtractFile) {
+        double detailLinesApprovedAmountSum = 0;
+        List<ConcurRequestExtractRequestDetailFileLine> requestDetailLines = requestExtractFile.getRequestDetails();
```
db483: extract this code in a separate method to calculate the total and use that in the check for matching totals.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,46 @@
...
+        for (ConcurRequestExtractRequestDetailFileLine detailLine : requestDetailLines) {
+            detailLinesApprovedAmountSum = detailLinesApprovedAmountSum + detailLine.getRequestAmount().doubleValue();
+        }
+        KualiDecimal detailLinesAmountSumKualiDecimal = new KualiDecimal(detailLinesApprovedAmountSum);
+        if (detailLinesAmountSumKualiDecimal.doubleValue() == requestExtractFile.getTotalApprovedAmount().doubleValue()) {
+            return true;
+        }
+        else {
+            LOG.error("Header amount validation failed. Header amount was =" +requestExtractFile.getTotalApprovedAmount().toString()+"= while file calculated amount was =" +detailLinesAmountSumKualiDecimal.toString()+ "=");
```
db483: you're missing a space after the + sign, might be safe to format the file

src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
```diff
@@ -163,20 +163,141 @@
...
+        </property>
+        <property name="directoryPath">
+            <value>${staging.directory}/concur/requestExtract</value>
+        </property>
+        <property name="fileExtension">
+            <value>txt</value>
+        </property>
+        <property name="fileNamePrefix" value="Standard_Travel_Request_Extract_"/>
+        <property name="fileTypeIdentifier" value="concurRequestExtractInputFileType"/>
+        <property name="titleKey" value="message.batchUpload.title.achIncome"/>
```
db483: the title key doesn't make sense: message.batchUpload.title.achIncome. You might want to define a new one for this batch

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+    
+    @After
+    public void tearDown() throws Exception {
+        concurRequestExtractFileValidationService = null;
+    }
+    
+    @Test
+    public void testHeaderAmountMatches() {
+        ConcurRequestExtractFile testFile = buildConcurRequestExtractFile();
```
db483: I think you need to separate tests here, one for a test case where the amounts match and one for a test case where the amounts don't match.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+        testFile.setRecordCount(new Integer(3));
+        if (concurRequestExtractFileValidationService.fileRowCountMatchesHeaderRowCount(testFile)) {
+            assertTrue("Invalid: Header row count matched file row count. They should NOT have matched.", false);
+        }
+        else {
+            assertTrue("Valid: Header row count did NOT match file row count.", true);
+        } 
+    }
+    
+    protected ConcurRequestExtractFile buildConcurRequestExtractFile() {
```
db483: You could use a fixture for the test data.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+        if (concurRequestExtractFileValidationService.fileApprovedAmountsMatchHeaderApprovedAmount(testFile)) {
+            assertTrue("Valid: Header amount matched sum of row amounts from file.", true);
+        }
+        else {
+            assertTrue("Invalid: Header amount did NOT match sum of row amounts from file.", false);
+        } 
+    }
+    
+    @Test
+    public void testHeaderAmountDoesNotMatch() {
```
db483: I think you need to separate tests here, one for a test case where the amounts match and one for a test case where the amounts don't match.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+            assertTrue("Invalid: Header amount matched row sum. They should NOT have matched.", false);
+        }
+        else {
+            assertTrue("Valid: Header amount did NOT match sum of row amounts from file.", true);
+        } 
+    }
+    
+    @Test
+    public void testHeaderRowCountMatches() {
+        ConcurRequestExtractFile testFile = buildConcurRequestExtractFile();
```
db483: I think you need to separate tests here, one for a test case where the amounts match and one for a test case where the amounts don't match.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+            assertTrue("Valid: Header row count file row count.", true);
+        }
+        else {
+            assertTrue("Invalid: Header should NOT have matched file row count", false);
+        } 
+    }
+    
+    @Test
+    public void testHeaderRowCountDoesNotMatch() {
+        ConcurRequestExtractFile testFile = buildConcurRequestExtractFile();
```
db483: I think you need to separate tests here, one for a test case where the amounts match and one for a test case where the amounts don't match.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: I normally utilize getters/setters for class level attributes. It makes it clearer that the data is not local  to the method and is at the class level. In my experience it also makes it easier to implement a maintenance change when a getter is used.  You implement it once in the getter method and do not have to sprinkle it throughout the code.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: Will adjust the code.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: There is a method in that service that operates on a list of files.  Will investigate its use on a singleton.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: Adjusting code.

src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
nkimble: Created new title key.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: This will be dealt with in the follow-on user story KFSPTS-7997

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: This test was modeled after ConcurStandardAccountingExtractServiceImplTest.java 
Why would I need a fixture and that test not?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: This comment pertains to this concern and the next three. These four tests are in pairs with match test first followed by mis-match test next.   I think the issue here is the assert processing and I will rework that to make it clearer what is happening.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: See comment above.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: See comment above

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: See comment above.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
db483: Good catch, I missed that in the ConcurStandardAccountingExtractServiceImplTest.java class. I will add a comment there too. thank you!

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
db483: Maybe I am missing something here, but I think what we want to do with a unit test is to test that if we expect fileApprovedAmountsMatchHeaderApprovedAmount to return true when given input A, we are in fact getting true as result. If fileApprovedAmountsMatchHeaderApprovedAmount should return false when given input B then we do in fact get false as a result. The test as is written now, will always pass, no matter what fileApprovedAmountsMatchHeaderApprovedAmount returns.

nkimble: We are saying the same thing.  I will be changing the tests to look like:
    @Test
    public void testHeaderAmountMatches() {
        ConcurRequestExtractFile testFile = buildConcurRequestExtractFile();
        assertTrue("Expected Result: Header amount SHOULD match sum of row amounts from file.", concurRequestExtractFileValidationService.fileApprovedAmountsMatchHeaderApprovedAmount(testFile));
    }
    
    @Test
    public void testHeaderAmountDoesNotMatch() {
        ConcurRequestExtractFile testFile = buildConcurRequestExtractFile();
        testFile.setTotalApprovedAmount(new KualiDecimal(9.87));
        assertFalse("Expected Result: Header amount should NOT match sum of row amounts from file.", concurRequestExtractFileValidationService.fileApprovedAmountsMatchHeaderApprovedAmount(testFile));
    }

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: Fixed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: Technically this attribute should not be null as it is initialized in the class constructor but that condition will be taken into account when requested new method is created.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: New method will be created as requested.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: The FlatFileDataHandler is the class that contains the "processor" to deal with parsing each file into the POJO data map that the rest of this processing utilizes. The KFS server will generate a 404 error on startup if the file service interface does not extend it.

nkimble:
Changes have been made. Ready for re-review.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
db483: The FlatFileDataHandler is indeed extended by the processor class but since you are not implementing any of the process, validate and getFileName I am not sure you need this processor class. I think that if you remove the  <property name="processor" ref="concurRequestExtractFileService"/> from the concurRequestExtractInputFileType bean definition then you will not get the 404 exception. Alternatively you can use the validate method of the processor and move the validation into that method. The validate method is called by the batchInputFileType validate method so you can call that one to validate the file contents. In your case the batchInputFileType implementation is FlatFileParserBase, you can take a look at that implementation.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        else {
+            LOG.error("Header row count validation failed. Header record count was =" + requestExtractFile.getRecordCount() + "= while file line count was =" + fileLineCount + "=");
+            return false;
+        }
+    }
+
+    public boolean fileApprovedAmountsMatchHeaderApprovedAmount(ConcurRequestExtractFile requestExtractFile) {
+        KualiDecimal detailLinesAmountSumKualiDecimal = getTotalRequestFileRequestAmount(requestExtractFile);
+        if (detailLinesAmountSumKualiDecimal.doubleValue() == requestExtractFile.getTotalApprovedAmount().doubleValue()) {
```
db483: I am not sure why the comparison between the doubleValue is needed here. You can use the KualiDecimal equals or compareTo methods for this test. (equals is calling the compareTo)

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        else {
+            return requestDetailLine.getRequestEntryDetails().size();
+        }
+    }
+
+    private KualiDecimal getTotalRequestFileRequestAmount(ConcurRequestExtractFile requestExtractFile) {
+        double detailLinesApprovedAmountSum = 0;
+        if ( (requestExtractFile.getRequestDetails() == null) || (requestExtractFile.getRequestDetails().isEmpty()) ) {
+            return new KualiDecimal(detailLinesApprovedAmountSum);
```
db483: you should use the KualiDecimal.ZERO

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        else {
+            List<ConcurRequestExtractRequestDetailFileLine> requestDetailLines = requestExtractFile.getRequestDetails();
+            for (ConcurRequestExtractRequestDetailFileLine detailLine : requestDetailLines) {
+                detailLinesApprovedAmountSum = detailLinesApprovedAmountSum + detailLine.getRequestAmount().doubleValue();
```
db483: no need to use the doubleValue, you should use the KualiDecimal.add method.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -0,0 +1,76 @@
...
+        if ( (requestDetailLine.getRequestEntryDetails() == null) || (requestDetailLine.getRequestEntryDetails().isEmpty()) ) {
+            return 0;
+        }
+        else {
+            return requestDetailLine.getRequestEntryDetails().size();
+        }
+    }
+
+    private KualiDecimal getTotalRequestFileRequestAmount(ConcurRequestExtractFile requestExtractFile) {
+        double detailLinesApprovedAmountSum = 0;
```
db483: this should be a KualiDecimal

src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
```diff
@@ -160,23 +158,143 @@
...
+        <property name="directoryPath">
+            <value>${staging.directory}/concur/requestExtract</value>
+        </property>
+        <property name="fileExtension">
+            <value>txt</value>
+        </property>
+        <property name="fileNamePrefix" value="Standard_Travel_Request_Extract_"/>
+        <property name="fileTypeIdentifier" value="concurRequestExtractInputFileType"/>
+        <property name="titleKey" value="message.concur.requestExtract.title.createPdpFeedFile"/>
+        <property name="processor" ref="concurRequestExtractFileService"/>
```
db483: If the processor is not used then you can remove this line and the ConcurRequestExtractFileService doesn't have to implement FileDataHandler.

nkimble:
retest this please

nkimble:
All unit tests have passed. Ready for re-review.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractCreatePdpFeedServiceImpl.java
```diff
@@ -1,45 +1,43 @@
...
-import edu.cornell.kfs.concur.service.ConcurRequestExtractFileService;
-import edu.cornell.kfs.concur.service.ConcurRequestExtractFileValidationService;
+import edu.cornell.kfs.concur.batch.service.ConcurRequestExtractFileService;
+import edu.cornell.kfs.concur.batch.service.ConcurRequestExtractFileValidationService;
 
 public class ConcurRequestExtractCreatePdpFeedServiceImpl implements ConcurRequestExtractCreatePdpFeedService {
     private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurRequestExtractCreatePdpFeedServiceImpl.class);
+    protected FileStorageService fileStorageService;
     protected ConcurRequestExtractFileService concurRequestExtractFileService;
     protected ConcurRequestExtractFileValidationService concurRequestExtractFileValidationService;
```
db483: looks like the validation service is not used in this class. it should be removed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+            Object parsedFile = loadFile(requestExtractFileName);
+            processingSuccess = requestExtractHeaderRowValidatesToFileContents(parsedFile);
+        } catch (RuntimeException re) {
+            LOG.error("Caught exception trying to load: " + requestExtractFileName, re);
+            processingSuccess = false;
+        }
+        return processingSuccess;
+    }
+
+    public boolean requestExtractHeaderRowValidatesToFileContents(Object parsedFile) {
```
db483: I think it would make more sense for this method to be in the validation service.

jhulslander:
Nancy, looks like there are conflicts merging this to develop now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        List<String> fileNamesToLoad = getBatchInputFileService().listInputFileNamesWithDoneFile(getBatchInputFileType());
+        return fileNamesToLoad;
+    }
+
+    public boolean processFile(String requestExtractFileName) {
+        boolean processingSuccess = true;
+        try {
+            Object parsedFile = loadFile(requestExtractFileName);
+            processingSuccess = requestExtractHeaderRowValidatesToFileContents(parsedFile);
+        } catch (RuntimeException re) {
```
db483: is there any reason why we want to catch a RuntimeException here? Is this to catch a parse exception while parsing the file? Might make more sense to add a catch to the try in the createPdpFeedsFromRequestExtracts method in the ConcurRequestExtractCreatePdpFeedServiceImpl class

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public void setConcurRequestExtractFileValidationService(ConcurRequestExtractFileValidationService concurRequestExtractFileValidationService) {
+        this.concurRequestExtractFileValidationService = concurRequestExtractFileValidationService;
+    }
+
+    public ConcurRequestExtractFileValidationService getConcurRequestExtractFileValidationService() {
+        return this.concurRequestExtractFileValidationService;
+    }
+
+    private Object loadFile(String fileName) {
```
db483: should this return List<ConcurRequestExtractFile> instead of Object?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractCreatePdpFeedServiceImpl.java
```diff
@@ -1,45 +1,43 @@
...
         } else {
+            LOG.info("Found " + filesToProcess.size() + " file(s) to process.");
+
             for (String requestExtractFileName : filesToProcess) {
-                createPdpFeedFromRequestExtract(requestExtractFileName);
-                LOG.info("Request Extract File " + requestExtractFileName + " was processed.");
+                try {
+                    LOG.info("Begin processing for filename: " + requestExtractFileName + ".");
+                    getConcurRequestExtractFileService().processFile(requestExtractFileName);
+                } finally {
```
db483: might make sense to add a catch here too

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractCreatePdpFeedServiceImpl.java
nkimble: Removing.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractCreatePdpFeedServiceImpl.java
nkimble: Spoke with Daniela off-line. Leaving as is for now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: Removing catch.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: I believe that this should be left as an object since method loadFile will be placed in a utility class with user story KFSPTS-7997 and will need to be used by multiple file types.

db483: got it. thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: Moved method to validation class. Changed visibility of existing validation methods as they should no longer be public with this method now being in the validation class. Had to update tests as well due to this change.

nkimble:
Requested changes have been made. Unit tests have passed. Ready for re-review.

db483:
changes look good and unit tests have passed. Merging.

PR #323, opened 2017-03-02
KFSPTS-7809: update to 2017-02-23 version of base code

bhutchinson:


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPdpEmailServiceImpl.java
```diff
@@ -204,9 +199,9 @@ private MailMessage createAdviceMessageAndPopulateHeader(PaymentGroup paymentGro
             message.setSubject(customer.getAdviceSubjectLine());
         }
         else {
-            message.addToAddress(mailService.getBatchMailingList());
-            message.addCcAddress(mailService.getBatchMailingList());
-            message.addBccAddress(mailService.getBatchMailingList());
+            message.addToAddress(emailService.getDefaultFromAddress());
+            message.addCcAddress(emailService.getDefaultFromAddress());
+            message.addBccAddress(emailService.getDefaultFromAddress());
```
cah292: Should these three properties be using the default "to" address instead of the default "from" one?

src/main/java/edu/cornell/kfs/sys/batch/CuJobListener.java
```diff
@@ -22,24 +22,23 @@
...
-            MailMessage mailMessage = new MailMessage();
-            mailMessage.setFromAddress(mailService.getBatchMailingList());
+            BodyMailMessage mailMessage = new BodyMailMessage();
+            mailMessage.setFromAddress(emailService.getDefaultFromAddress());
             if (jobExecutionContext.getMergedJobDataMap().containsKey(REQUESTOR_EMAIL_ADDRESS_KEY) && !StringUtils.isBlank(jobExecutionContext.getMergedJobDataMap().getString(REQUESTOR_EMAIL_ADDRESS_KEY))) {
                 mailMessage.addToAddress(jobExecutionContext.getMergedJobDataMap().getString(REQUESTOR_EMAIL_ADDRESS_KEY));
             }
             if (SchedulerService.FAILED_JOB_STATUS_CODE.equals(jobStatus) || SchedulerService.CANCELLED_JOB_STATUS_CODE.equals(jobStatus)) {
-                mailMessage.addToAddress(mailService.getBatchMailingList());
+                mailMessage.addToAddress(emailService.getDefaultFromAddress());
```
cah292: Another area where we should be using default "to" instead of default "from".

bhutchinson:
I have updated the code and this is ready for review again. As to the null default email addresses, I have mentioned that to Cathy and noted it in the DEV validation sub-task for this story. We will need to keep an eye on them, and may want to set the values to what is in the mailing.list.batch property (a little tricky to script in SQL since that is different between prod and non-prod).

PR #324, opened 2017-03-06
KFSPTS-8009: Added generic batch step for moving files.

cah292:
This adds a general-purpose batch step for moving batch files from one directory to another, and for automatically creating .done files for them. None of the non-unit-test code is actively using the new batch step, but I did add an abstract bean to simplify setting up other batch jobs to use this step.

This batch step assumes that all of the files matching a given regex should be moved to the destination directory. It also assumes the source directory does not have .done files, though because of the step's current setup, it should theoretically be able to move such files successfully. If you think there needs to be some .done-related restrictions for the files to be moved, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
@@ -16,6 +16,8 @@ public class CUKFSConstants {
     public static final String RULE_CODE_CA = "CA";
     public static final String DELIMITER = ".";
     public static final String DATE_FORMAT_YYYY = "YYYY";
+    public static final String FILE_EXTENSION_DELIMITER = ".";
+    public static final String DONE_FILE_EXTENSION = FILE_EXTENSION_DELIMITER + "done";
 
     public static class SysKimApiConstants {
         public static final String CONTRACTS_AND_GRANTS_PROCESSOR = "Contracts & Grants Processor";
--- a/src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
+++ b/src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
@@ -1,0 +1,128 @@
+package edu.cornell.kfs.sys.batch;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Date;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.sys.batch.AbstractStep;
+
+import edu.cornell.kfs.sys.CUKFSConstants;
+
+/**
+ * Generic batch step for moving files from one directory to another.
+ * 
+ * At a minimum, the sourcePath and targetPath must be set to valid pre-existing directories,
+ * and the fileNamePattern must be set to a valid Java regex String. Only files whose names
+ * match the regex will be moved from the source directory to the target directory.
+ * This step will also automatically create ".done" files in the target directory for each moved file.
+ * 
+ * The source/target directory paths are not required to have trailing slashes. This step
+ * will automatically add them as needed.
+ */
+public class MoveFileStep extends AbstractStep {
+
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(MoveFileStep.class);
+
+    protected String sourcePath;
+    protected String targetPath;
+    protected Pattern fileNamePattern;
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        if (StringUtils.isBlank(sourcePath)) {
+            throw new IllegalStateException("sourcePath cannot be blank");
+        } else if (StringUtils.isBlank(targetPath)) {
+            throw new IllegalStateException("targetPath cannot be blank");
+        } else if (fileNamePattern == null) {
+            throw new IllegalStateException("fileNamePattern cannot be null");
+        }
+        
+        super.afterPropertiesSet();
+    }
+
+    @Override
+    public boolean execute(String jobName, Date jobRunDate) throws InterruptedException {
+        return moveFiles();
+    }
+
+    protected boolean moveFiles() {
+        verifyDirectoriesExist(sourcePath, targetPath);
+        
+        try {
+            File sourceDirectory = new File(sourcePath);
+            Matcher fileNameMatcher = fileNamePattern.matcher(StringUtils.EMPTY);
+            int numFilesMoved = 0;
+            
+            LOG.info("Moving files from " + sourcePath + " to " + targetPath);
+            
+            for (File sourceFile : FileUtils.listFiles(sourceDirectory, null, false)) {
+                fileNameMatcher.reset(sourceFile.getName());
+                if (fileNameMatcher.matches()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Moving file: " + sourceFile.getName());
+                    }
+                    moveFileAndMarkAsReady(sourceFile);
+                    numFilesMoved++;
+                }
+            }
+            
+            LOG.info("File move completed.");
+            LOG.info(Integer.toString(numFilesMoved) + " file(s) have been moved to the target directory.");
+        } catch (Exception e) {
+            LOG.error("Encountered error while moving files", e);
+            return false;
+        }
+        
+        return true;
+    }
+
+    protected void verifyDirectoriesExist(String... directoryPaths) {
+        for (String directoryPath : directoryPaths) {
+            File directory = new File(directoryPath);
+            if (!directory.exists() || !directory.isDirectory()) {
+                throw new IllegalStateException("Directory does not exist or does not represent a directory path: " + directoryPath);
+            }
+        }
+    }
+
+    protected void moveFileAndMarkAsReady(File sourceFile) throws IOException {
+        String targetFilePath = buildFilePath(targetPath, sourceFile.getName());
+        File targetFile = new File(targetFilePath);
+        FileUtils.moveFile(sourceFile, targetFile);
+        
+        String doneFilePath = StringUtils.substringBeforeLast(targetFilePath, CUKFSConstants.FILE_EXTENSION_DELIMITER)
+                + CUKFSConstants.DONE_FILE_EXTENSION;
+        File doneFile = new File(doneFilePath);
+        FileUtils.touch(doneFile);
+    }
+
+    protected String buildFilePath(String prefix, String fileName) {
+        String actualPrefix = prefix;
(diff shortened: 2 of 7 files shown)


cah292:
Hi Daniela/Nancy/Jay,

Because of the impact that this PR will have on the various Concur batch jobs you have all been developing, I've assigned you all as reviewers on this so that you can weigh in on whether this batch step's setup makes sense. Please let me know if you have questions or any suggestions for improvement.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+ * The source/target directory paths are not required to have trailing slashes. This step
+ * will automatically add them as needed.
+ */
+public class MoveFileStep extends AbstractStep {
+
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(MoveFileStep.class);
+
+    protected String sourcePath;
+    protected String targetPath;
+    protected Pattern fileNamePattern;
```
db483: Just a note: the SAE to PDP and collector batch job will need to move both pdp files and collector files. In that case we will need two separate steps to accomplish that: one to move the pdp files and one to move the collector files.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Moving file: " + sourceFile.getName());
+                    }
+                    moveFileAndMarkAsReady(sourceFile);
+                    numFilesMoved++;
+                }
+            }
+            
+            LOG.info("File move completed.");
+            LOG.info(Integer.toString(numFilesMoved) + " file(s) have been moved to the target directory.");
```
db483: Is the Integer.toString needed? I think you can use just: numFilesMoved + " file(s) have been moved to the target directory." instead

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
```diff
@@ -16,6 +16,8 @@
     public static final String RULE_CODE_CA = "CA";
     public static final String DELIMITER = ".";
     public static final String DATE_FORMAT_YYYY = "YYYY";
+    public static final String FILE_EXTENSION_DELIMITER = ".";
```
nkimble: Why is another constant needed? Couldn't DELIMITER which is already present be used?

db483:
Hi Jay,

What are the pdp feed and collector feed file names? I thought the names would be unique with each run. I am not sure I understand how we would pass the exact file names in.

I agree that we need to make sure files are moved only if both pdp feed and collector file have been correctly generated. As this is the second step to the job that would only run if the first step succeeds I thought this would not even run unless both files were successfully generated.

jhulslander:
I am not sure how one step passes information to the other.  When we create the pdp feed file or the collector file, we have control over the file name we create with.  My thought was to use the file name used on the original SAE file from Concur.

db483:
Hi Jay,

Since you place the output files in a specific output folder for pdp and collector the move file step will just look into that folder for the generated files.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
cah292: My goal was to make this step generic enough so that multiple instances of this class could be used to move the files accordingly, given that the source and target directories would be different for each. (Also, this step may be needed for the request-extract-to-PDP file move.) That way, you could have multiple beans representing a specific configured version of this step.

If you need me to, I can try to implement multiple source-to-target behavior for a single step instance, but I think that would introduce too much complexity.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+        
+        super.afterPropertiesSet();
+    }
+
+    @Override
+    public boolean execute(String jobName, Date jobRunDate) throws InterruptedException {
+        return moveFiles();
+    }
+
+    protected boolean moveFiles() {
```
cah292: Ok, I'll see if I can separate this into its own service.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+        verifyDirectoriesExist(sourcePath, targetPath);
```
cah292: I tried using FileSystemFileStorageServiceImpl for implementing many of the different operations for this. However, the service assumes that either the directory path is an absolute one, or that the path should be prefixed with another absolute or relative path. I can look into using that service further if desired, but I would likely need to create a separately-configured service bean that has an appropriate path prefix configured.

jhulslander:
Correct, but what about the scenario of multiple files.  TWO PDP feel files exist in the requisition folder, but only one should be moved to be processed by PDP.  How do we handle that situation?  I think we want to process the files we can.  However, we may want to not process any, and review what went awry.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+        
+        try {
+            File sourceDirectory = new File(sourcePath);
+            Matcher fileNameMatcher = fileNamePattern.matcher(StringUtils.EMPTY);
```
cah292: Thanks, I'll work on refactoring this. I was trying to reuse the same Matcher instance for efficiency, but perhaps I can do so through some additional helper methods.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+            int numFilesMoved = 0;
+            
+            LOG.info("Moving files from " + sourcePath + " to " + targetPath);
+            
+            for (File sourceFile : FileUtils.listFiles(sourceDirectory, null, false)) {
```
cah292: Yes, we could add that if we need to. We should discuss that at the future meeting you had mentioned in the standup.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+                fileNameMatcher.reset(sourceFile.getName());
+                if (fileNameMatcher.matches()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Moving file: " + sourceFile.getName());
+                    }
+                    moveFileAndMarkAsReady(sourceFile);
```
cah292: Good point, I'll change the name accordingly.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
cah292: Yes, I think the implicit way is clearer. I'll change this accordingly.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    }
+
+    protected void moveFileAndMarkAsReady(File sourceFile) throws IOException {
+        String targetFilePath = buildFilePath(targetPath, sourceFile.getName());
+        File targetFile = new File(targetFilePath);
+        FileUtils.moveFile(sourceFile, targetFile);
+        
+        String doneFilePath = StringUtils.substringBeforeLast(targetFilePath, CUKFSConstants.FILE_EXTENSION_DELIMITER)
```
cah292: I originally went with FileSystemFileStorageServiceImpl for this, but later decided not to because of the prefixing issues mentioned in my earlier comment. I'll need to investigate further to determine if there's a way to get around the prefix problem.

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
cah292: I think the other DELIMITER constant was meant to be used in the context of property delimiters. If you think that the existing constant is also okay to use in the context of file delimiters, then I'll change this accordingly.

cah292: See my response to Jay's comment.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
```diff
@@ -0,0 +1,128 @@
...
+            Matcher fileNameMatcher = fileNamePattern.matcher(StringUtils.EMPTY);
+            int numFilesMoved = 0;
+            
+            LOG.info("Moving files from " + sourcePath + " to " + targetPath);
+            
+            for (File sourceFile : FileUtils.listFiles(sourceDirectory, null, false)) {
+                fileNameMatcher.reset(sourceFile.getName());
+                if (fileNameMatcher.matches()) {
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("Moving file: " + sourceFile.getName());
```
nkimble: All of the other logging is at the info level.  Shouldn't this one be too?

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
jhulslander: I think it's fine, but I'll let others way in.

cah292:
Jay,

I was under the impression that the batch steps creating the PDP/collector files would place them in unique directories specific to that portion of the processing, and then multiple instances/beans of this step would then be used to move the files into the appropriate processing folders. For example, the request-extract-to-PDP process, SAE-to-PDP process, and SAE-to-collector process would each have their own unique folders where they would place their output files. Then we can use three separate instances/beans of this step to move the files from those folders to the existing KFS PDP/collector file folders.

Given that we may be integrating the SAE-to-PDP and SAE-to-collector processing into a single batch step (but will still likely use two separate services for performing the work), we could configure that step to relocate or remove the generated PDP files accordingly if the collector file generation fails. This will prevent the PDP files from being moved to the destination.

It may be a good idea to have a meeting to discuss this batch step further, like what Daniela suggested at today's standup.

src/main/java/edu/cornell/kfs/sys/batch/MoveFileStep.java
cah292: I didn't know if we needed the file logging to be this fine-grained in a Production environment. I could certainly change it to INFO level if you think it would be more consistent with how things are currently set up in similar batch jobs.

cah292:
After further discussion, it was decided that the request-to-PDP and SAE-to-PDP/collector jobs will be responsible for creating the PDP/collector files in the destination folders, and for only creating .done files for those that are actually ready to be processed. Therefore, this extra file-moving batch step will not be needed.

I'm closing this PR as a result.

PR #325, opened 2017-03-07
KFSPTS-7979: asynchronously pre-load batch file lookup cache

bhutchinson:


src/main/java/edu/cornell/kfs/sys/service/BatchFileDirectoryService.java
```diff
@@ -4,12 +4,14 @@
...
 
 public interface BatchFileDirectoryService {
 
-    @Cacheable(value= SystemOptions.CACHE_NAME, key="'buildDirectoryKeyValuesList' + #p0")
-    List<KeyValue> buildDirectoryKeyValuesList(List<File> rootDirectories);
+    @Cacheable(value= SystemOptions.CACHE_NAME, key="'buildStagingDirectories'")
+    List<KeyValue> buildBatchFileStagingDirectories();
+
+    @Cacheable(value= SystemOptions.CACHE_NAME, key="'buildStagingAndReportsDirectories'")
+    List<KeyValue> buildBatchFileLookupDirectories();
```
cah292: The cache key is using a value other than the name of the method, unlike the method above it. Is this intentional?

src/main/java/edu/cornell/kfs/sys/service/impl/BatchFileDirectoryServiceImpl.java
```diff
@@ -11,9 +14,28 @@
...
+        return directoryKeyValueList;
+    }
+
     @Override
-    public List<KeyValue> buildDirectoryKeyValuesList(List<File> rootDirectories) {
-        List<KeyValue> keyValues = new ArrayList<KeyValue>();
+    public List<KeyValue> buildBatchFileLookupDirectories() {
+        LOG.debug("Starting BatchFileDirectoryServiceImpl.buildStagingAndReportsDirectories().");
+        List<KeyValue> directoryKeyValueList = buildDirectoryKeyValuesList(BatchFileUtils.retrieveBatchFileLookupRootDirectories());
+        LOG.debug("Finished BatchFileDirectoryServiceImpl.buildStagingAndReportsDirectories().");
```
cah292: Another area where the strings are using one name but the method is using another name.

bhutchinson:
Yes, those were leftover from renaming. I've fixed them now and this should be ready for review again, thanks.

PR #326, opened 2017-03-08
KFSPTS-8016 fix code that updates report status in Concur and add mor…

db483:
…e logging info

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseReportDetailsDTO.java
+++ b/src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseReportDetailsDTO.java
@@ -12,6 +12,9 @@ import javax.xml.bind.annotation.XmlRootElement;
 @XmlAccessorType(XmlAccessType.NONE)
 public class ExpenseReportDetailsDTO {
     
+    @XmlElement(name = "ReportId")
+    private String reportId;
+    
     @XmlElement(name = "ApsKey")
     private String concurStatusCode;
     
@@ -44,6 +47,14 @@ public class ExpenseReportDetailsDTO {
 
     public void setConcurStatusCode(String concurStatusCode) {
         this.concurStatusCode = concurStatusCode;
+    }
+
+    public String getReportId() {
+        return reportId;
+    }
+
+    public void setReportId(String reportId) {
+        this.reportId = reportId;
     }
 
 }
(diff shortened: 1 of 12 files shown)


src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseReportDetailsDTO.java
```diff
@@ -12,6 +12,9 @@
 @XmlAccessorType(XmlAccessType.NONE)
 public class ExpenseReportDetailsDTO {
     
+    @XmlElement(name = "ReportId")
+    private String reportId;
```
cah292: This area uses "ReportId" and "reportId", but the ConcurReport class uses "reportID" instead. (The other XML DTO is also using "RequestID" and "requestID".) Does this DTO need to capitalize the "D" as well, or does Concur or KFS expect it to be just "Id" and not "ID"?

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurReportsServiceImpl.java
```diff
@@ -197,9 +204,32 @@ private String buildWorkflowUpdateXML(String workflowURI, String action, String
...
+    private String addMessageHeader(String message) {
+        if (message.length() > 0) {
+            message = ConcurConstants.ERROR_MESSAGE_HEADER + message;
+        }
+        return message;
+    }
+
+    private String truncateMessageLength(String message, int maxLength) {
+        if (message.length() > maxLength) {
+            message = message.substring(0, maxLength + 1);
```
cah292: This line will actually return a substring of length maxLength + 1. If the substring indeed needs to be no longer than maxLength, then replace "maxLength + 1" with just "maxLength".

src/main/java/edu/cornell/kfs/concur/rest/xmlObjects/ExpenseReportDetailsDTO.java
db483: good point. Concur expects ReportId in this case.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurReportsServiceImpl.java
db483: good catch, the +1 needs to be removed.

db483:
I fixed the truncate method. I would like to leave the reportId property as is although there is some inconsistency. Concur uses different naming for different reports.

db483:

PR #327, opened 2017-03-08
KFSPTS-7909: Template SAE-to-collector handling: WIP: DO NOT MERGE YET

cah292:
PLEASE DO NOT MERGE THIS YET! I wanted to get your opinions on this code before I finalize it.

This changes the SAE-to-PDP batch step so that it includes template code for the SAE-to-collector part. It also has some template code for eventually implementing the SAE-to-KFS reporting requirements.

I've nearly finished the regular code, and I'm still in the process of implementing the unit tests, so it's not quite ready for merging.

Please let me know if this seems like the way to go for implementing the requirements, or if you have suggestions for improvements.


db483:
Hi Chad,

I think for this user story you should only rename the step to ConcurStandardAccountingExtractToPdpAndCollectorStep.java and update the bean definition in cu-spring-concur.xml file accordingly. This way it will be easier to track what changed and also it should be less disruptive for Jay to rebase.

There are separate stories for the reporting part and they are currently a lower priority so I would leave that out for now.

Also, the new BatchFileSystemFileStorageServiceImpl.java should be on a separate story if we decide that is the way to go.

Any BOs specific to the collector part of the story should be added on a separate story, probably KFSPTS-7911.

I will set up a zoom meeting to touch base on the stories for the SAE to collector and brain storm about how to best split this work. Thank you!

cah292:
After further discussion with Daniela, we've narrowed down the required scope for this user story, and I will create a new branch for this work.

PR #328, opened 2017-03-09
KFSPTS-7909: Add template SAE-to-collector processing.

cah292:
As per the recent discussion, the changes for this PR just rename the batch job/step and add a new service method for the SAE-to-collector processing.


db483:
Changes look good, will merge once unit tests pass.

PR #330, opened 2017-03-09
KFSPTS-7911: Added Report End Date to the SAE detail POJO.

cah292:
This adds another SAE file field to the detail POJO, which will be needed for the Collector processing.


db483:
Changes look good to me, merging.

PR #331, opened 2017-03-14
KFSPTS-7868 sae to PDP

jhulslander:
I need to do some more validation that Marcia provided, however I would like to start getting some feedback on this code.  Please do NOT merge this yet.


src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -49,8 +49,30 @@
 
     public static final String PDP_XML_NAMESPACE = "http://www.kuali.org/kfs/pdp/payment";
     
+    public static final String DATE_FORMAT = "MM/dd/YYYY";
```
jhulslander: eliminate the extra lines and move

src/main/java/edu/cornell/kfs/concur/batch/service/ConcurStandardAccountingExtractValidationService.java
```diff
@@ -14,6 +15,10 @@
     
     boolean validateDebitCreditField(String debitCredit);
     
+    boolean validateEmployeeGroupId(String employeeGroupId);
```
jhulslander: move this below validate date

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +129,222 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
     }
 
     @Override
-    public boolean extractPdpFeedFromStandardAccounitngExtract(ConcurStandardAccountingExtractFile concurStandardAccountingExtractFile) {
-        return true;
+    public String extractPdpFeedFromStandardAccounitngExtract(ConcurStandardAccountingExtractFile concurStandardAccountingExtractFile) {
```
jhulslander: fix the order of these new changes

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +129,222 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
...
+    }
+    
+    protected String buildPdpOutputFileName(String originalFileName) {
+        return StringUtils.replace(originalFileName, GeneralLedgerConstants.BatchFileSystem.TEXT_EXTENSION, ConcurConstants.XML_FILE_EXTENSION);
+    }
+
+    private PdpFeedFileBaseEntry buildPdpFeedFileBaseEntry(ConcurStandardAccountingExtractFile concurStandardAccountingExtractFile) {
+        KualiDecimal pdpTotal = KualiDecimal.ZERO;
+        PdpFeedFileBaseEntry pdpFeedFileBaseEntry = new PdpFeedFileBaseEntry();
+        pdpFeedFileBaseEntry.setVersion("1.0");
```
jhulslander: make 1.0 a constant

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +129,222 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
...
+        pdpFeedFileBaseEntry.setHeader(buildPdpFeedHeaderEntry(concurStandardAccountingExtractFile.getBatchDate()));
+        
+        for (ConcurStandardAccountingExtractDetailLine line : concurStandardAccountingExtractFile.getConcurStandardAccountingExtractDetailLines()) {
+            if (StringUtils.equalsIgnoreCase(line.getPaymentCode(), ConcurConstants.StandardAccountingExtractPdpConstants.PAYMENT_CODE_CASH)) {
+                if (getConcurStandardAccountingExtractValidationService().validateConcurStandardAccountingExtractDetailLine(line)) {
+                    PdpFeedGroupEntry currentGroup = getGroupForLine(pdpFeedFileBaseEntry, line);
+                    PdpFeedDetailEntry currentDetail = getDetailEntryForLine(currentGroup, line);
+                    PdpFeedAccountingEntry currentAccounting = getAccountingEntryForLine(currentDetail, line);
```
jhulslander: see about refactoring this method to shorten it up

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -4,8 +4,20 @@
...
+import org.kuali.kfs.coa.businessobject.SubAccount;
+import org.kuali.kfs.coa.businessobject.SubObjectCode;
+import org.kuali.kfs.coa.service.AccountService;
+import org.kuali.kfs.coa.service.ObjectCodeService;
+import org.kuali.kfs.coa.service.ProjectCodeService;
+import org.kuali.kfs.coa.service.SubAccountService;
+import org.kuali.kfs.coa.service.SubObjectCodeService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 
+import com.ctc.wstx.util.StringUtil;
```
jhulslander: This is probably wrong

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
```diff
@@ -105,15 +125,51 @@ else if ((paymentDetail.getNetPaymentAmount() == null) && (paymentDetail.isDetai
...
                 errorMap.putError(KFSConstants.GLOBAL_ERRORS, PdpKeyConstants.ERROR_PAYMENT_LOAD_MAX_NOTE_LINES, Integer.toString(groupCount), Integer.toString(noteLineCount), Integer.toString(getMaxNoteLines()));
             }
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("After processGroupValidation: " + printErrorMap(errorMap));
+            }
+        }
+    }
+
+    @Override
+    public void doHardEdits(PaymentFileLoad paymentFile, MessageMap errorMap) {
```
jhulslander: eliminate these functions that don't do anything

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
```diff
@@ -1107,21 +1107,22 @@ public String getStreet() {
...
-        fieldValues.put("vendorDetailAssignedIdentifier", headerDetails[1]);
-        
-        List<VendorAddress> addrs = (List<VendorAddress>)bos.findMatching(VendorAddress.class, fieldValues);
-        if (addrs.size() == 1) {
-            VendorAddress addr = (VendorAddress) addrs.get(0);
-            setVendorAddress(addr);
-        } else {
-            throw new RuntimeException("Invalid Address [ "+customerInstitutionNumber+" ] for payee [ "+payeeId + " ]");
-            // Need to handle bad data.
+        if (payeeId.split("-").length > 1) {
```
jhulslander: is there a better way to do this?

cah292: You could use a regex to see if the payee ID matches the "headerId-detailId" format.

jhulslander: Do you know of an example I can use as a reference to do this?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImplTest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Test
+    public void buildPayeeNameOverFortyCharsEndWithComma2() {
+        String results = concurStandardAccountExtractPdpEntryServiceImpl.buildPayeeName("zxcvbnmasdfghjklqwer", "qwertyuiopasdfghi", "l");
+        String expected = "zxcvbnmasdfghjklqwer, qwertyuiopasdfghi";
+        assertEquals("The names should be the same", expected, results);
+    }
+    
+    @Test
+    public void formatDate() {
+        Date testDate = new Date(2017-1900, 1, 7);
```
cah292: Is this year subtraction intentional for the test? Also, it would be clearer to use the Calendar.FEBRUARY constant instead of a literal 1 for the month.

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
```diff
@@ -1107,21 +1107,22 @@ public String getStreet() {
     }
     
     public void validateVendorIdAndCustomerInstitutionIdentifier() {
```
cah292: I think the custom vendor-related methods in this file are being used by CU-specific pieces of code elsewhere. It might make more sense to customize those areas instead of this file, depending on how difficult it is to get those other pieces of code to work for non-vendor payment groups.

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
cah292: I don't know of an example off the top of my head, but it's possible that this check could be relocated or removed. See my comment a couple lines above this one.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +118,134 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
...
+    }
+    
+    private PdpFeedFileBaseEntry buildPdpFeedFileBaseEntry(ConcurStandardAccountingExtractFile concurStandardAccountingExtractFile) {
+        KualiDecimal pdpTotal = KualiDecimal.ZERO;
+        PdpFeedFileBaseEntry pdpFeedFileBaseEntry = new PdpFeedFileBaseEntry();
+        pdpFeedFileBaseEntry.setVersion(ConcurConstants.StandardAccountingExtractPdpConstants.FEED_FILE_ENTRY_HEADER_VERSION);
+        pdpFeedFileBaseEntry.setHeader(
+                getConcurStandardAccountExtractPdpEntryService().buildPdpFeedHeaderEntry(concurStandardAccountingExtractFile.getBatchDate()));
+        for (ConcurStandardAccountingExtractDetailLine line : concurStandardAccountingExtractFile.getConcurStandardAccountingExtractDetailLines()) {
+            if (StringUtils.equalsIgnoreCase(line.getPaymentCode(), ConcurConstants.StandardAccountingExtractPdpConstants.PAYMENT_CODE_CASH)) {
```
db483: it might make sense to extract this code into a separate method that creates/updates and accounting entry from the current line in the file

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
jhulslander: this was a previous Cornell customization, it presumes a vendor number, we need to pass in an employee ID.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,24 @@
...
+        public static final String EMPLOYEE_STATUS_CODE = "EMPLOYEE";
+        public static final String EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "E";
+        public static final String NON_EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "Y"; 
+        public static final String DEFAULT_CHART_CODE = "IT";
+        public static final String DEFAULT_UNIT = "CRNL";
+        public static final String DEFAULT_SUB_UNIT = "CNCR";
+        public static final String FEED_FILE_ENTRY_HEADER_VERSION = "1.0";
+    }
+    
+    public static final String DATE_FORMAT = "MM/dd/YYYY";
```
jhulslander: ah, good catch, thanks

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        currentGroup.setBankCode(ConcurConstants.StandardAccountingExtractPdpConstants.BANK_CODE);
+        currentGroup.setCustomerInstitutionIdentifier("");
+        return currentGroup;
+    }
+    
+    protected String buildPayeeName(String lastName, String firstName, String middleInitial) {
+        String seperator = KFSConstants.COMMA + KFSConstants.BLANK_SPACE;
+        String fullName = lastName + seperator + firstName;
+        if (StringUtils.isNotBlank(middleInitial)) {
+            fullName = fullName + seperator + middleInitial + KFSConstants.DELIMITER;
```
jhulslander: This was what was in the spec, but I will confirm.  I created some unit tests so I thought I was getting the results I planed for

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        String sourceDocNumber = ConcurConstants.StandardAccountingExtractPdpConstants.DOCUMENT_TYPE + reportId;
+        return StringUtils.substring(sourceDocNumber, 0, ConcurConstants.SOURCE_DOCUMENT_NUMBER_FIELD_SIZE);
+    }
+
+    @Override
+    public PdpFeedAccountingEntry buildPdpFeedAccountingEntry(ConcurStandardAccountingExtractDetailLine line) {
+        PdpFeedAccountingEntry currentAccountingEntry =  new PdpFeedAccountingEntry();
+        currentAccountingEntry.setCoaCd(line.getChartOfAccountsCode());
+        currentAccountingEntry.setAccountNbr(line.getAccountNumber());
+        currentAccountingEntry.setObjectCd(line.getJournalAccountCode());
```
jhulslander: I will ping Sherry about that.  I hadn't seen anything about special object handling.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +118,134 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
...
+                if (getConcurStandardAccountingExtractValidationService().validateConcurStandardAccountingExtractDetailLine(line)) {
+                    PdpFeedGroupEntry currentGroup = getGroupEntryForLine(pdpFeedFileBaseEntry, line);
+                    PdpFeedDetailEntry currentDetail = getDetailEntryForLine(currentGroup, line);
+                    PdpFeedAccountingEntry currentAccounting = getAccountingEntryForLine(currentDetail, line);
+                    
+                    pdpTotal = pdpTotal.add(line.getJournalAmount());
+                    String newAmount = addAmounts(currentAccounting.getAmount(), line.getJournalAmount());
```
jhulslander: I will check with Sherry

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
jhulslander: I'll need to investigate

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    public PdpFeedGroupEntry buildPdpFeedGroupEntry(ConcurStandardAccountingExtractDetailLine line) {
+        PdpFeedGroupEntry currentGroup = new PdpFeedGroupEntry();
+        currentGroup.setPayeeName(buildPayeeName(line.getEmployeeLastName(), line.getEmployeeFirstName(), 
+                line.getEmployeeMiddleInitital()));
+        currentGroup.setPayeeId(buildPayeeIdEntry(line));
+        currentGroup.setPaymentDate(formatDate(line.getBatchDate()));
+        currentGroup.setCombineGroupInd(ConcurConstants.StandardAccountingExtractPdpConstants.COMBINDED_GROUP_INDICATOR);
```
jhulslander: I can parameterize this, however I suspect if there is a need to no longer combine groups, we will need to make other changes

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        currentGroup.setBankCode(ConcurConstants.StandardAccountingExtractPdpConstants.BANK_CODE);
```
jhulslander: DISB is from the spec document

db483: Looking at the spec, the Customer Profile defines the bank code for ACH and Check. What are those used for? Are they used in the PDP processing?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
db483: In that case I think you should leave it like this, thank you!

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: I don't know for certain, I was just plugging in values Sherry defined in the PDP out pul file requirements   I can track it down further

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,24 @@
         public static final String CREDIT = "CR";
         public static final String DEBIT = "DR";
```
cah292: I'm also going to need this debit/credit codes for the Collector processing. It might make sense to move them into the top-level class or to at least rename the nested class.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
db483: Checking back with Sherry sounds like a good idea. Thank you!

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        currentAccountingEntry.setAccountNbr(line.getAccountNumber());
+        currentAccountingEntry.setObjectCd(line.getJournalAccountCode());
+        currentAccountingEntry.setSubObjectCd(line.getSubObjectCode());
+        currentAccountingEntry.setOrgRefId(line.getOrgRefId());
+        currentAccountingEntry.setProjectCd(line.getProjectCode());
+        currentAccountingEntry.setAmount(KualiDecimal.ZERO.toString());
+        return currentAccountingEntry;
+    }
+    
+    protected String formatDate(Date date) {
```
jhulslander: yep, that works

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
```diff
@@ -1158,19 +1159,25 @@ public void setVendorAddress(VendorAddress addr) {
...
-        BusinessObjectService bos = SpringContext.getBean(BusinessObjectService.class);
-        String[] headerDetails = payeeId.split("-");      
-        Map<String, String> fieldValues = new HashMap<String, String>();
-        fieldValues.put("vendorHeaderGeneratedIdentifier", headerDetails[0]/*payeeId*/);
-        fieldValues.put("vendorDetailAssignedIdentifier", headerDetails[1]);
-        
-        List<VendorDetail> details = (List<VendorDetail>)bos.findMatching(VendorDetail.class, fieldValues);
-        if (details.size() == 1) {
-            payeeName=details.get(0).getVendorName();
+        if (string.split("-").length == 1) {
```
jhulslander: We had a previously customized PaymentGroup and that customization presumed it was just vendors, and not employee IDs.  I am not sure if this solution is what we need yet.  Marcia gave me some more details at testing PDP.

src/main/java/edu/cornell/kfs/concur/batch/ConcurStandardAccountingExtractToPdpAndCollectorStep.java
```diff
@@ -41,11 +43,23 @@ protected boolean processCurrentFileAndExtractPdpFeedFromSAEFile(String saeFileN
...
                         .extractCollectorFeedFromStandardAccountingExtract(concurStandardAccoutingExtractFile);
             }
+            if (success) {
+                try {
+                    getConcurStandardAccountingExtractService().createDoneFileForPdpFile(outputFileName);
+                } catch (IOException e) {
+                    LOG.error("processCurrentFileAndExtractPdpFeedFromSAEFile, unable to create .done file: ", e);
+                    success = false;
+                }
+            }
```
jhulslander: I think we want to leave the PDP xml in place.  There won't be a .done file so it won't process. The file may be helpful in analyzing a problem.  Also, if we need to re-process the same file, the existing pdp xml will get over-written.

src/main/java/edu/cornell/kfs/concur/batch/ConcurStandardAccountingExtractToPdpAndCollectorStep.java
```diff
@@ -22,7 +24,7 @@ public boolean execute(String jobName, Date jobRunDate) throws InterruptedExcept
         List<String> listOfSaeFileNames = getConcurStandardAccountingExtractService().buildListOfFileNamesToBeProcessed();
         boolean success = true;
         for (String saeFileName : listOfSaeFileNames) {
-            LOG.info("execute, processing: " + saeFileName);
+            LOG.debug("execute, processing: " + saeFileName);
```
jhulslander: I do log it with info elsewhere, so this turns out to be a duplicate log message in most cases.  I wanted to leave it as debug in case we need to trace things more indepthly.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
cah292: If this is being changed to use the DateTimeService instead like Daniela suggested, we'll need to make sure that the default date format remains as "MM/dd/yyyy" when performing future upgrades. (It should already default to that format in the current Rice and KFS code.)

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,24 @@
...
+        public static final String EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "E";
+        public static final String NON_EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "Y"; 
+        public static final String DEFAULT_CHART_CODE = "IT";
+        public static final String DEFAULT_UNIT = "CRNL";
+        public static final String DEFAULT_SUB_UNIT = "CNCR";
+        public static final String FEED_FILE_ENTRY_HEADER_VERSION = "1.0";
+    }
+    
+    public static final String DATE_FORMAT = "MM/dd/YYYY";
+    public static final String EMPLOYEE_GROUP_ID = "CORNELL";
```
jhulslander: My requirement was that the eighth element in the SAE must be Cornell, otherwise it is an error.  I don't have a strong opinion on what to call it.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: I changed the constant as you suggested, and updated this function per Daniela's note so it is 
getDateTimeService().toString(date, ConcurConstants.DATE_FORMAT);

I have a unit test to verify the results are what we want

cah292: Ah yes, I forgot that the DateTimeService has a public method that accepts a date pattern. That would work for this, so that we're not relying on the Rice default pattern.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,24 @@
         public static final String CREDIT = "CR";
         public static final String DEBIT = "DR";
     }
+    
+    public class StandardAccountingExtractPdpConstants {
+        public static final String PAYMENT_CODE_CASH = "CASH";
+        public static final String COMBINDED_GROUP_INDICATOR = "Y";
+        public static final String BANK_CODE = "DISB";
+        public static final String FS_ORIGIN_CODE = "Z6";
```
nkimble: I need this value for the Request Extract processing so that would make sense.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
nkimble: This is a candidate to be a common concur utility method. I need to perform the same transformation for the Request Extract to PDP processing.

jhulslander:
I have made a bunch of changes from the code review.  Please take a look.  I have asked Sherry some of the questions we discussed.  I also need to do some further testing, so please don't merge this code

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public PdpFeedPayeeIdEntry buildPayeeIdEntry(ConcurStandardAccountingExtractDetailLine line) {
+        PdpFeedPayeeIdEntry payeeIdEntry = new PdpFeedPayeeIdEntry();
+        payeeIdEntry.setContent(line.getEmployeeId());
+        if (StringUtils.equalsIgnoreCase(line.getEmployeeStatus(), ConcurConstants.StandardAccountingExtractPdpConstants.EMPLOYEE_STATUS_CODE)) {
+            payeeIdEntry.setIdType(ConcurConstants.StandardAccountingExtractPdpConstants.EMPLOYEE_PAYEE_STATUS_TYPE_CODE);
+        } else {
+            payeeIdEntry.setIdType(ConcurConstants.StandardAccountingExtractPdpConstants.NON_EMPLOYEE_PAYEE_STATUS_TYPE_CODE);
```
nkimble: You are assuming that all other values not equal to EMPLOYEE are to be set to Y. The spec for both SAE and request extract indicates that the data value of "NON-EMPLOYEE" should be set to Y. To me this implies there could be other values sent from Concur in that column besides EMPLOYEE and NON-EMPLOYEE. Thoughts?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public PdpFeedDetailEntry buildPdpFeedDetailEntry(ConcurStandardAccountingExtractDetailLine line) {
+        PdpFeedDetailEntry currentDetailEntry  = new PdpFeedDetailEntry();
+        currentDetailEntry.setSourceDocNbr(buildSourceDocumentNumber(line.getReportId()));
+        currentDetailEntry.setFsOriginCd(getConcurParamterValue(ConcurConstants.DEFAULT_ORIGINATION_CODE_PARAMETER_NAME));
+        currentDetailEntry.setFdocTypCd(getConcurParamterValue(ConcurConstants.DEFAULT_SAW_PDP_DOCUMENT_TYPE_PARAMETER_NAME));
+        currentDetailEntry.setInvoiceNbr("");
+        currentDetailEntry.setPoNbr("");
```
nkimble: Should StringUtils.EMPTY be used for these lines?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,25 @@
...
+    }
+    
+    public static final String DEFAULT_ORIGINATION_CODE_PARAMETER_NAME = "CONCUR_AP_PDP_ORIGINATION_CODE";
+    public static final String DEFAULT_SAW_PDP_DOCUMENT_TYPE_PARAMETER_NAME = "CONCUR_SAE_PDP_DOCUMENT_TYPE";
+    public static final String CUSTOMER_PROFILE_LOCATION_PARAMETER_NAME = "CONCUR_CUSTOMER_PROFILE_LOCATION";
+    public static final String CUSTOMER_PROFILE_UNIT_PARAMETER_NAME = "CONCUR_CUSTOMER_PROFILE_UNIT";
+    public static final String CUSTOMER_PROFILE_SUB_UNIT_PARAMETER_NAME = "CONCUR_CUSTOMER_PROFILE_SUB_UNIT";
+    
+    public static final String DATE_FORMAT = "MM/dd/yyyy";
+    public static final String EMPLOYEE_GROUP_ID = "CORNELL";
```
cah292: Should this value be parameter-derived as well?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -86,5 +102,55 @@ public boolean validateDate(Date date) {
...
+        } else {
+            LOG.error("validateReportId, no valid report ID");
+        }
+        return valid;
+    }
+    
+    /*
+    private boolean validateAccountingInformation(ConcurStandardAccountingExtractDetailLine line) {
+        return validateRequiredElementsExists(line) && validateAccouningLine(line);
+    }*/
```
cah292: Please remove this commented-out code.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -86,5 +102,55 @@ public boolean validateDate(Date date) {
...
+
+    private boolean validateAccouningLine(ConcurStandardAccountingExtractDetailLine line) {
```
cah292: Minor typo in method name.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -86,5 +102,55 @@ public boolean validateDate(Date date) {
...
+        boolean valid = StringUtils.isNotBlank(line.getAccountNumber()) && 
+                StringUtils.isNotBlank(line.getChartOfAccountsCode()) &&
+                StringUtils.isNotBlank(line.getJournalAccountCode());
+        if (valid) {
+            LOG.debug("validateRequiredElementsExists, found a chart, account, and object code");
+        } else {
+            LOG.error("validateRequiredElementsExists, chart, account, or object code was empty.");
+        }
+        return valid;
+    }*/
```
cah292: Please remove this commented-out code.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
```diff
@@ -82,20 +99,23 @@ else if ((paymentDetail.getNetPaymentAmount() == null) && (paymentDetail.isDetai
 
                 // compare net to accounting segments
                 if (paymentDetail.getAccountTotal().compareTo(paymentDetail.getNetPaymentAmount()) != 0) {
+                    LOG.debug("processGroupValidation, account total (" + paymentDetail.getAccountTotal()  + ") not equal to net amount tatal (" + paymentDetail.getNetPaymentAmount() + ")");
```
cah292: Minor typo in one of the "total" words.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImplTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @Test
+    public void buildPayeeNameOverFortyCharsEndWithComma2() {
+        String results = concurStandardAccountExtractPdpEntryServiceImpl.buildPayeeName("zxcvbnmasdfghjklqwer", "qwertyuiopasdfghi", "l");
+        String expected = "zxcvbnmasdfghjklqwer, qwertyuiopasdfghi";
+        assertEquals("The names should be the same", expected, results);
+    }
+    
+    @Test
+    public void formatDate() {
+        Date testDate = new Date(2017-1900, Calendar.FEBRUARY, 7);
```
cah292: Now I see why you're subtracting 1900, based on the Date Javadoc. However, I think this constructor is now deprecated. If you need another easy means of creating a Date, you could construct a Joda DateTime object and then convert it to a Date. (If converting from DateTime to a SQL Date, you may have to call DateTime.getMillis() and pass it to the Date constructor.)

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -86,5 +102,55 @@ public boolean validateDate(Date date) {
...
+    
+    /*
+    private boolean validateAccountingInformation(ConcurStandardAccountingExtractDetailLine line) {
+        return validateRequiredElementsExists(line) && validateAccouningLine(line);
+    }*/
+
+    private boolean validateAccouningLine(ConcurStandardAccountingExtractDetailLine line) {
+        ConcurAccountInfo concurAccountInfo = new ConcurAccountInfo(line.getChartOfAccountsCode(), line.getAccountNumber(), 
+                line.getSubAccountNumber(), line.getJournalAccountCode(), line.getSubObjectCode(), line.getProjectCode(), line.getOrgRefId());
+        ValidationResult validationResults = getConcurAccountValidationService().validateConcurAccountInfo(concurAccountInfo);
```
cah292: Does the ConcurAccountValidationService validate the object code? If so, the service may need to be updated to allow for the "Pending Client" object code, or the the replacement value for "Pending Client" will need to be passed to the validation service instead. (For the SAE-to-Collector processing, I still need to process rows which use that code, so I can't use this validation code for the Collector processing unless it can handle the "Pending Client" code.)

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -53,4 +53,25 @@
...
+        public static final String PAYMENT_CODE_CASH = "CASH";
+        public static final String COMBINED_GROUP_INDICATOR = "Y";
+        public static final String BANK_CODE = "DISB";
+        public static final String EMPLOYEE_STATUS_CODE = "EMPLOYEE";
+        public static final String EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "E";
+        public static final String NON_EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "Y"; 
+        public static final String FEED_FILE_ENTRY_HEADER_VERSION = "1.0";
+    }
+    
+    public static final String DEFAULT_ORIGINATION_CODE_PARAMETER_NAME = "CONCUR_AP_PDP_ORIGINATION_CODE";
```
db483: On the same note, I think usually the constant name and parameter name are the same.

db483:
Jay, I have reviewed the code again. There are still some suggestions from the first round of review that have not been addressed

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: yes, I think that is true

nkimble: Then, should this be an if-else if-else construct instead?

jhulslander: Yes

jhulslander:
retest this please

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
jhulslander: I was able to eliminate this code here by updating the calling code. However the setPayeeId function is called by XML configuration.

jhulslander:
I have given Marcia O some test files to verify and I have made adjustments to the code based on code review.  This is ready to reviewed again.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    
+    public Integer getPayeeNameFieldSize() {
+        if (payeeNameFieldSize == null) {
+            payeeNameFieldSize = getDataDictionaryService().getAttributeMaxLength(PaymentGroup.class, PdpConstants.PaymentDetail.PAYEE_NAME);
+        }
+        return payeeNameFieldSize;
+    }
+    
+    public String getConcurParameterValue(String parameterName) {
+        String paramterValue = getParameterService().getParameterValueAsString(CUKFSConstants.ParameterNamespaces.CONCUR,
```
cah292: Minor typo in variable name.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
```diff
@@ -98,21 +122,142 @@ protected void logDetailedInfoForConcurStandardAccountingExtractFile(ConcurStand
...
+        pdpFeedFileBaseEntry.setTrailer(getConcurStandardAccountExtractPdpEntryService().buildPdpFeedTrailerEntry(pdpFeedFileBaseEntry, pdpTotal));
+        return pdpFeedFileBaseEntry;
+    }
+    
+    private void replacePendingClientWithOverride(ConcurStandardAccountingExtractDetailLine line) {
+        if (StringUtils.equalsIgnoreCase(line.getJournalAccountCode(), ConcurConstants.PENDING_CLIENT)) {
+            String overrideObjectCode = getParameterService().getParameterValueAsString(CUKFSConstants.ParameterNamespaces.CONCUR, 
+                    CUKFSParameterKeyConstants.ALL_COMPONENTS, ConcurConstants.ParameterNames.CONCUR_PENDING_CLIENT_OBJECT_CODE_OVERRIDE);
+            LOG.error("replacePendingClientWithOverride, found a Pending Client object code, replacing it with " + overrideObjectCode);
+            line.setJournalAccountCode(overrideObjectCode);
```
cah292: The SAE-to-Collector processing also needs to know which lines have the "Pending Client" object code, so setting the replacement on the detail POJO directly will prevent the Collector side from reusing that POJO unless you're doing object-copying somewhere. Also, you should confirm with Sherry whether such lines need to stand apart from others for PDP (since on the Collector side I need such lines to remain standalone in the final file).

Depending on what changes are needed for this processing, the validation code may also need to be updated to handle "Pending Client" object codes.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -60,21 +70,38 @@ public boolean validateAmountsAndDebitCreditCode(ConcurStandardAccountingExtract
...
+        boolean valid = StringUtils.equalsIgnoreCase(employeeGroupId, expectedGroupId);
+        if (valid) {
+            LOG.debug("Found a valid employee group id.");
+        } else {
+            LOG.error("Found an invalid employee group id: " + employeeGroupId + ".  We expected the group ID to be " + expectedGroupId);
+        }
+        return valid;
+    }
+    
+    protected String findEmploueeGroupId() {
```
cah292: Minor typo in method name.

src/main/java/edu/cornell/kfs/sys/xmladapters/KualiDecimalXmlAdapter.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+public class KualiDecimalXmlAdapter extends XmlAdapter<String, KualiDecimal> {
+
+    @Override
+    public KualiDecimal unmarshal(String v) throws Exception {
+        BigDecimal doubleValue = new BigDecimal(v);
+        return new KualiDecimal(doubleValue);
```
cah292: KualiDecimal has a constructor that accepts a String as input, so you don't need to pass in an explicit BigDecimal unless you're running into problems with the other constructor.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImplTest.java
```diff
@@ -116,5 +141,12 @@ private void setBadDebitCredit() {
     private void setBadRecordCount() {
         file.setRecordCount(new Integer(5));
     }
+    
+    private class TestableConcurStandardAccountingExtractValidationServiceImpl extends ConcurStandardAccountingExtractValidationServiceImpl {
+        @Override
+        protected String findEmploueeGroupId() {
+            return "CORNELL";
```
cah292: It might make sense to create a testing-only constant for this value, given that at least one other test class relies on it (and the unit tests for my story may need it as well).

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
jhulslander: If we cam update the validation service to accept client pending as a good object code then I can so this check elsewhere and not update the same pojo. 

Something to consider we only update cash lines here.  Which I think are only pro lines.  Does collector use cash lines?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -49,8 +47,32 @@
...
+        public static final String COMBINED_GROUP_INDICATOR = "Y";
+        public static final String BANK_CODE = "DISB";
+        public static final String EMPLOYEE_STATUS_CODE = "EMPLOYEE";
+        public static final String NON_EMPLOYEE_STATUS_CODE = "NON-EMPLOYEE";
+        public static final String EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "E";
+        public static final String NON_EMPLOYEE_PAYEE_STATUS_TYPE_CODE = "Y"; 
+        public static final String FEED_FILE_ENTRY_HEADER_VERSION = "1.0";
+    }
+    
+    public static class ParameterNames {
```
db483: I think Chad had a comment about this too, let's move the parameters to a separate ConcurParameterConstants class

src/main/java/edu/cornell/kfs/concur/batch/ConcurStandardAccountingExtractToPdpAndCollectorStep.java
```diff
@@ -22,7 +24,7 @@ public boolean execute(String jobName, Date jobRunDate) throws InterruptedExcept
         List<String> listOfSaeFileNames = getConcurStandardAccountingExtractService().buildListOfFileNamesToBeProcessed();
         boolean success = true;
```
db483: Jay, I don't see a change or comment regarding this matter, I wonder if it got missed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractServiceImpl.java
cah292: Yes, the SAE-to-Collector process needs to handle lines with CASH payment codes, among others.

PR #336, opened 2017-03-23
KFSPTS-8043: avoid OLE when inactivating a security model

bhutchinson:


bhutchinson:
It is inactivating the role. It only gets called if modelRole.isActive returns false, and the call to Role.Builder.create(modelRole) sets active to be that value.

cah292:
I thought that a false return value from newSecurityModel.isActive triggered the method, because the model role should still be active prior to that point so that the membership updates can be performed in KIM. The method being overridden is supposed to perform the final step of inactivating the role in KIM.

bhutchinson:
let me look again more closely

bhutchinson:
You are correct of course. I was making assumptions and seeing what I wanted to see. Thanks for catching that. I've fixed that and it should be ready for review again now.

bhutchinson:
Thanks Chad!

PR #337, opened 2017-03-24
KFSPTS-8076 fix function that retrieves field value from concur so th…

db483:
…at the sub object code can be correctly extracted

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/ConcurUtils.java
+++ b/src/main/java/edu/cornell/kfs/concur/ConcurUtils.java
@@ -16,7 +16,7 @@ public class ConcurUtils {
     }
 
     public static String extractKFSInfoFromConcurString(String concurString) {
-        String kfsInfo = StringUtils.EMPTY;
+        String kfsInfo = concurString;
         if (doesStringContainOpenCloseParenthesis(concurString)) {
             kfsInfo = concurString.substring(concurString.lastIndexOf(CUKFSConstants.LEFT_PARENTHESIS) + 1, concurString.lastIndexOf(CUKFSConstants.RIGHT_PARENTHESIS));
         }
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/concur/ConcurUtils.java
```diff
@@ -16,7 +16,7 @@ public static boolean isTravelRequestURI(String URI) {
     }
 
     public static String extractKFSInfoFromConcurString(String concurString) {
-        String kfsInfo = StringUtils.EMPTY;
+        String kfsInfo = concurString;
         if (doesStringContainOpenCloseParenthesis(concurString)) {
             kfsInfo = concurString.substring(concurString.lastIndexOf(CUKFSConstants.LEFT_PARENTHESIS) + 1, concurString.lastIndexOf(CUKFSConstants.RIGHT_PARENTHESIS));
```
jhulslander: I am curious about this function.  As written, if "(some bad text (GOOD TEST) some more bad text)" is passed in, the function will return "GOOD TEXT) some more bad text".  I am not sure if you need to be worried about that case or not, but wanted to throw that out here.

I don't think the name of the function says what it is doing.  I am not sure what a concise alternative would be either, so I don't think it needs to be changed right now.

db483: That is a good point. I think this could be cleaned up to be more clear. Some fields in the concur xml report details contain the value in the format: "(code) description". In this case we only care about the code so we extract only that. In other cases it just contains a value and we pull the whole value in. I think it would be better to separate the two based on specific fields instead of trying to treat all values with the same utility function. I will update the code and try to make it more clear. Thank you.

db483:
Hi Jay,

I have updated the code and it is now ready for review. Thank you.

PR #340, opened 2017-03-27
KFSPTS-7869: Concur Request Extract file to PDP Feed File base code.

nkimble:


nkimble:
This is the initial base code to create the PDP feed file from a Concur request extract file. This code will be used to build the full functionality for the Request Extract to PDP feed file imports.

nkimble:
This PR depends on nonprod-sql  PR#71.  They must go in together.

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
```diff
@@ -0,0 +1,23 @@
+package edu.cornell.kfs.concur.batch;
+
+public class ConcurRequestExtractPdpConstants {
+
+    public static class ValidationConstants {
+        public static final String CASH_ADVANCE_INDICATOR = "CASH ADVANCE";
+        public static final String EMPLOYEE_INDICATOR = "EMPLOYEE";
+        public static final String NON_EMPLOYEE_INDICATOR = "NON-EMPLOYEE";
```
jhulslander: there are constants for employee and non-employee is ConcurConstants, can those be used?

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static class ValidationErrorMessages {
+        public static final String NOT_CASH_ADVANCE_DATA_LINE = "Request Extract Detail Line is not a Cash Advance. ";
+        public static final String CASH_ADVANCE_USED_IN_EXPENSE_REPORT = "Cash Advance has been used in expense report.";
+        public static final String REQUEST_ID_INVALID = "Request ID was detected as being NULL or blank.";
+        public static final String EMPLOYEE_ID_NULL_OR_BLANK = "Employee ID was detected as being NULL or blank.";
+        public static final String EMPLOYEE_ID_NOT_FOUND_IN_KFS = "Person for provided Employee ID could not be found in KFS.";
+        public static final String PAYEE_ID_TYPE_INVALID = "Payee ID type was not specified as EMPLOYEE or NON-EMPOYEE.";
+        public static final String REQUEST_AMOUNT_INVALID = "Requested cash advance amount was detected as being NULL or blank.";
+        public static final String DUPLICATE_CASH_ADVANCE_DETECTED = "Duplicate cash advance request detected.";
```
jhulslander: Should these be in applicationResources.properties?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+        PdpFeedFileBaseEntry pdpFeedFileDataObject = buildPdpFeedBaseEntry(requestExtractFile);
+        if (pdpFeedFileDataObject.getTrailer().getDetailCount().intValue() != 0) {
+            String fullyQualifiedPdpFileName = buildFullyQualifiedPdpOutputFileName(getPaymentImportDirectory(), requestExtractFile.getFileName());
+            pdpFileSuccessfullyCreated = createPdpFeedFile(pdpFeedFileDataObject, fullyQualifiedPdpFileName);
+            if (pdpFileSuccessfullyCreated) {
+                LOG.debug("fullyQualifiedPdpFileName [" + fullyQualifiedPdpFileName + "]  was created for requestExtractFile [" + requestExtractFile.getFileName() + "]");
+                requestExtractFile.setFullyQualifiedPdpFileName(fullyQualifiedPdpFileName);
+            }
+            else {
+                LOG.debug("FAILED TO CREATE: fullyQualifiedPdpFileName [" + fullyQualifiedPdpFileName + "] for requestExtractFile [" + requestExtractFile.getFileName() + "]");
```
jhulslander: should this be a LOG.error

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+            }
+        }
+        return pdpFileSuccessfullyCreated;
+    }
+
+    public void createDoneFileForPdpFile(String fullyQualifiedPdpFileName) throws IOException {
+        String fullFilePath = StringUtils.replace(fullyQualifiedPdpFileName, ConcurConstants.XML_FILE_EXTENSION, GeneralLedgerConstants.BatchFileSystem.DONE_FILE_EXTENSION);
```
jhulslander: There is a function is the batch file service to create a done file based on an inputted file.  Would that work here?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+    }
+
+    private PdpFeedFileBaseEntry buildPdpFeedBaseEntry(ConcurRequestExtractFile requestExtractFile) {
+        int totalPdpDetailRecordsCount = 0;
+        KualiDecimal totalPdpDetailRecordsAmount = KualiDecimal.ZERO;
+        PdpFeedFileBaseEntry pdpBaseEntry = new PdpFeedFileBaseEntry();
+        pdpBaseEntry.setHeader(buildPdpFeedHeaderEntry(requestExtractFile.getBatchDate()));
+        List<PdpFeedGroupEntry> groupEntries = new ArrayList<PdpFeedGroupEntry>();
+        for (ConcurRequestExtractRequestDetailFileLine detailFileLine : requestExtractFile.getRequestDetails()) {
+            if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isValidCashAdvanceLine()) {
```
jhulslander: would it make sense to refactor the code inside the for loop into it's own function

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+    private PdpFeedTrailerEntry buildPdpFeedTrailerEntry(int totalDetailRecordCount, KualiDecimal totalDetailRecordAmount) {
+        PdpFeedTrailerEntry trailer = new PdpFeedTrailerEntry();
+        trailer.setDetailCount(totalDetailRecordCount);
+        trailer.setDetailTotAmt(totalDetailRecordAmount);
+        return trailer;
     }
 
-    public void setOutgoingDirectoryName(String outgoingDirectoryName) {
-        this.outgoingDirectoryName = outgoingDirectoryName;
+    //common utility method
```
jhulslander: Just to confirm, these functions have been set to be refactored into a utility later on, yes?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -64,6 +91,24 @@ public ConcurRequestExtractFileValidationService getConcurRequestExtractFileVali
...
 
+    private String parseRequestExtractFileNameFrom(String requestExtractFullyQualifiedFileName) {
+        //plus 1 because we do not want the forward slash
+        return requestExtractFullyQualifiedFileName.substring((requestExtractFullyQualifiedFileName.lastIndexOf(ConcurConstants.FORWARD_SLASH)) + 1);
+    }
+
+    private ConcurRequestExtractFile loadFileIntoParsedDataObject(String requestExtractFullyQualifiedFileName) {
+        Object parsedFile = loadFile(requestExtractFullyQualifiedFileName);
+        List<ConcurRequestExtractFile> requestExtractFiles = (ArrayList<ConcurRequestExtractFile>) parsedFile;
+        if (requestExtractFiles.isEmpty() || requestExtractFiles.size() != 1) {
```
jhulslander: why can't this IF statement just be the size check not equal 1, en empty array will return zero for the size, right?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestedCashAdvanceServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+        Collection<ConcurRequestedCashAdvance> concurRequestedCashAdvances = null;
+        Map<String, String> fieldValues = new HashMap<String, String>();
+        
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.EMPLOYEE_ID, concurRequestedCashAdvance.getEmployeeId());
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.REQUEST_ID, concurRequestedCashAdvance.getRequestId());
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.PAYMENT_AMOUNT, concurRequestedCashAdvance.getPaymentAmount().toString());
+        
+        concurRequestedCashAdvances = businessObjectService.findMatching(ConcurRequestedCashAdvance.class, fieldValues);
+        
+        if ((concurRequestedCashAdvances == null) || (concurRequestedCashAdvances.isEmpty()) ){
```
jhulslander: why not just return the conditional statement?

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
nkimble: I'll use employee and non-employee from ConcurConstants.

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
nkimble: I'll move the constants part of this section into the ConcurKeyConstants to conform to other applicationResources.properties lookups and put the error string in the applicationResources.properties.

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
```diff
@@ -0,0 +1,23 @@
+package edu.cornell.kfs.concur.batch;
+
+public class ConcurRequestExtractPdpConstants {
```
nkimble: With bulk of the code in this file being placed in other files, I will also be moving the two remaining constants to the ConcurConstants file and totally deleting this constants class.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestedCashAdvanceServiceImpl.java
nkimble: The conditional is looking for when there is nothing returned to indicate there is no duplicate. 
I thought having an explicit conditional statement was clearer than negating the result of the conditional.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
db483: I agree this should be extracted in a separate method

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
db483: Given all these utility methods listed in this class I think it would be easier to extract them in a utility class now and use that utility class here. Then Jay and Chad can use that utility class and add any other methods as needed. When we get to the jira about the utility class we can do any remaining clean up if needed.

Only move your methods to the utility class for now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -64,6 +91,24 @@ public ConcurRequestExtractFileValidationService getConcurRequestExtractFileVali
         return this.concurRequestExtractFileValidationService;
     }
 
+    private String parseRequestExtractFileNameFrom(String requestExtractFullyQualifiedFileName) {
+        //plus 1 because we do not want the forward slash
+        return requestExtractFullyQualifiedFileName.substring((requestExtractFullyQualifiedFileName.lastIndexOf(ConcurConstants.FORWARD_SLASH)) + 1);
```
db483: I am not sure I understand why this is needed

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
```diff
@@ -64,6 +91,24 @@ public ConcurRequestExtractFileValidationService getConcurRequestExtractFileVali
...
+    }
+
+    private ConcurRequestExtractFile loadFileIntoParsedDataObject(String requestExtractFullyQualifiedFileName) {
+        Object parsedFile = loadFile(requestExtractFullyQualifiedFileName);
+        List<ConcurRequestExtractFile> requestExtractFiles = (ArrayList<ConcurRequestExtractFile>) parsedFile;
+        if (requestExtractFiles.isEmpty() || requestExtractFiles.size() != 1) {
+            LOG.error("Single physical file " + requestExtractFullyQualifiedFileName + " should have translated into single parsed file. More or less than one parsed file was detected.");
+            throw new RuntimeException("Single physical file " + requestExtractFullyQualifiedFileName + " should have translated into single parsed file. More or less than one parsed file was detected.");
```
db483: do you want to throw runtime exception here? if there are multiple files to be processed and one file fails will these cause the whole batch job to fail?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
-
     private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurRequestExtractFileValidationServiceImpl.class);
+    protected ConcurRequestedCashAdvanceService concurRequestedCashAdvanceService;
+    protected ConcurAccountValidationService concurAccountValidationService;
+    protected ParameterService parameterService;
+    protected PersonService personService;
 
-    public boolean requestExtractHeaderRowValidatesToFileContents(List<ConcurRequestExtractFile> requestExtractFiles) {
+    public boolean requestExtractHeaderRowValidatesToFileContents(ConcurRequestExtractFile requestExtractFile) {
         boolean headerValidationPassed = false;
```
db483: I don't think this initialization is needed here.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+        }
+    }
+
+    /**
+     * A request detail line is considered a cash advance when request entry detail type is set to the cash advance indicator.
+     *
+     * @param detailFileLine
+     * @return boolean
+     */
+    private boolean requestDetailFileLineIsCashAdvance(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
```
db483: this method name is misleading as this method does not just perform validation and returns a boolean, it is also changing the input object.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+    }
+
+    /**
+     * If a cash advance has been used in an expense report, the expense report id will have information in it.
+     * Bypass that transaction.
+     *
+     * @param detailFileLine
+     * @return boolean
+     */
+    private boolean requestedCashAdvanceHasNotBeenUsedInExpenseReport(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
```
db483: this method name is misleading as this method does not just perform validation and returns a boolean, it is also changing the input object.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+
+    /**
+     * If a cash advance is duplicated, it will have already been paid and there will be an entry in our tracking database table
+     * Bypass that transaction.
+     *
+     * @param detailFileLine
+     * @return boolean
+     */
+    private boolean requestedCashAdvanceIsNotBeingDuplicated(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        boolean haveDataForLookup = (requestIdIsValid(detailFileLine) && employeeIdIsValid(detailFileLine) && payeeIdTypeIsValid(detailFileLine) && requestAmountIsValid(detailFileLine));
```
db483: this boolean value is not used

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
+
+    private boolean requestIdIsValid(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if (StringUtils.isEmpty(detailFileLine.getRequestId())) {
+            detailFileLine.getValidationResult().addMessage(new String(ConcurRequestExtractPdpConstants.ValidationErrorMessages.REQUEST_ID_INVALID));
```
db483: here you are adding an error message but no validation flag is switched

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+            return false;
+        }
+        else {
+            return true;
+        }
+    }
+
+    private boolean employeeIdIsValid(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if (StringUtils.isEmpty(detailFileLine.getEmployeeId())) {
+            detailFileLine.getValidationResult().addMessage(new String(ConcurRequestExtractPdpConstants.ValidationErrorMessages.EMPLOYEE_ID_NULL_OR_BLANK));
```
db483: same here, you are adding an error message but no validation flag is switched

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+            return false;
+        }
+        else {
+            Person employee = getPersonService().getPersonByEmployeeId(detailFileLine.getEmployeeId());
+            if (ObjectUtils.isNotNull(employee)) {
+                return true;
+            }
+            else {
+                detailFileLine.getValidationResult().addMessage(new String(ConcurRequestExtractPdpConstants.ValidationErrorMessages.EMPLOYEE_ID_NOT_FOUND_IN_KFS));
```
db483: same here, you are adding an error message but no validation flag is switched

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+    }
+
+    private boolean payeeIdTypeIsValid(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if ( StringUtils.isNotEmpty(detailFileLine.getPayeeIdType()) &&
+             (detailFileLine.getPayeeIdType().equalsIgnoreCase(ConcurRequestExtractPdpConstants.ValidationConstants.EMPLOYEE_INDICATOR) ||
+              detailFileLine.getPayeeIdType().equalsIgnoreCase(ConcurRequestExtractPdpConstants.ValidationConstants.NON_EMPLOYEE_INDICATOR)) ){
+            return true;
+        }
+        else {
+            detailFileLine.getValidationResult().addMessage(new String(ConcurRequestExtractPdpConstants.ValidationErrorMessages.PAYEE_ID_TYPE_INVALID));
```
db483: same here, you are adding an error message but no validation flag is switched

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+            return false;
+        }
+    }
+
+    private boolean requestAmountIsValid(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if ((detailFileLine.getRequestAmount() == null) || (detailFileLine.getRequestAmount().isNegative()) || (detailFileLine.getRequestAmount().isZero())) {
+            detailFileLine.getValidationResult().addMessage(new String(ConcurRequestExtractPdpConstants.ValidationErrorMessages.REQUEST_AMOUNT_INVALID));
```
db483: same here, you are adding an error message but no validation flag is switched

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -73,18 +246,68 @@ private int computedTotalRequestEntryDetailLinesForRequestDetail(ConcurRequestEx
         }
     }
 
-    private KualiDecimal getTotalRequestFileRequestAmount(ConcurRequestExtractFile requestExtractFile) {
-        KualiDecimal detailLinesApprovedAmountSum = KualiDecimal.ZERO;
+    private KualiDecimal getTotalRequestFileTotalApprovedAmount(ConcurRequestExtractFile requestExtractFile) {
+        KualiDecimal detailLinesTotalApprovedAmountSum = KualiDecimal.ZERO;
         if ( (requestExtractFile.getRequestDetails() == null) || (requestExtractFile.getRequestDetails().isEmpty()) ) {
-            return detailLinesApprovedAmountSum;
+            return detailLinesTotalApprovedAmountSum;
```
db483: if might be better to just return KualiDecimal.ZERO here
Or, you could only do the addition when there are details and return detailLinesTotalApprovedAmountSum at the end

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -73,18 +246,68 @@ private int computedTotalRequestEntryDetailLinesForRequestDetail(ConcurRequestEx
...
+                                  detailFileLine.getSubObjectCode(), detailFileLine.getProjectCode(), detailFileLine.getOrgRefId());
+        ValidationResult validationResults = getConcurAccountValidationService().validateConcurAccountInfo(concurAccountInfo);
+        if (validationResults.isNotValid()) {
+            detailFileLine.getValidationResult().addMessages(validationResults.getMessages());
         }
+        return validationResults.isValid();
+    }
+
+    //common utility method
+    protected String getConcurParamterValue(String parameterName) {
```
db483: It might be better to extract this in the utility service now

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
```diff
@@ -0,0 +1,23 @@
...
+        public static final String CASH_ADVANCE_USED_IN_EXPENSE_REPORT = "Cash Advance has been used in expense report.";
+        public static final String REQUEST_ID_INVALID = "Request ID was detected as being NULL or blank.";
+        public static final String EMPLOYEE_ID_NULL_OR_BLANK = "Employee ID was detected as being NULL or blank.";
+        public static final String EMPLOYEE_ID_NOT_FOUND_IN_KFS = "Person for provided Employee ID could not be found in KFS.";
+        public static final String PAYEE_ID_TYPE_INVALID = "Payee ID type was not specified as EMPLOYEE or NON-EMPOYEE.";
+        public static final String REQUEST_AMOUNT_INVALID = "Requested cash advance amount was detected as being NULL or blank.";
+        public static final String DUPLICATE_CASH_ADVANCE_DETECTED = "Duplicate cash advance request detected.";
+    }
+    
+    public static final String WHITESPACE = "    ";
```
nkimble: This is four spaces used to format the information sent to the log files and in the future user stories, to the reports. Business object toString methods and service error message construction use it.  If I call it FOUR_SPACES_FOR_OUTPUT would that be descriptive enough?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Changing from LOG.debug to LOG.error and changing the "if" LOG.debug to LOG.info

src/main/java/edu/cornell/kfs/concur/batch/ConcurRequestExtractPdpConstants.java
db483: Thank you for explaining this Nancy. Perhaps SPACING_STRING_FOR_OUTPUT? this way if you have to change it to 5 spaces, or 3 spaces, or to use tabs, you don't need to update the constant name.

nkimble: sounds good

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Changing to use FileStorageServicesto create the done file.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+        }
+        return pdpFileSuccessfullyCreated;
+    }
+
+    public void createDoneFileForPdpFile(String fullyQualifiedPdpFileName) throws IOException {
+        String fullFilePath = StringUtils.replace(fullyQualifiedPdpFileName, ConcurConstants.XML_FILE_EXTENSION, GeneralLedgerConstants.BatchFileSystem.DONE_FILE_EXTENSION);
+        FileUtils.touch(new File(fullFilePath));
+    }
+
+    public void setPaymentImportDirectory(String paymentImportDirectory) {
```
nkimble: I will move the getter/setters to the end at your request.  The pattern I was following was to have all public methods first and private/protected methods after them.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: At this point, I am not going to making that assumption. Having both checks makes the code and logic clearer IMHO.  I only want to process the valid cash advance lines, or only output the data to the log for invalid cash advance lines, and skip any lines that are not cash advances.  With the data I have encountered, I'd prefer to leave both checks in.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+        int totalPdpDetailRecordsCount = 0;
+        KualiDecimal totalPdpDetailRecordsAmount = KualiDecimal.ZERO;
+        PdpFeedFileBaseEntry pdpBaseEntry = new PdpFeedFileBaseEntry();
+        pdpBaseEntry.setHeader(buildPdpFeedHeaderEntry(requestExtractFile.getBatchDate()));
+        List<PdpFeedGroupEntry> groupEntries = new ArrayList<PdpFeedGroupEntry>();
+        for (ConcurRequestExtractRequestDetailFileLine detailFileLine : requestExtractFile.getRequestDetails()) {
+            if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isValidCashAdvanceLine()) {
+                PdpFeedDetailEntry pdpDetailEntry = buildPdpFeedDetailEntry(detailFileLine, buildPdpFeedAccountingEntry(detailFileLine));
+                List<PdpFeedDetailEntry> pdpDetailEntries = new ArrayList<PdpFeedDetailEntry>();
+                pdpDetailEntries.add(pdpDetailEntry);
```
nkimble: No I cannot.  I need to reference attributes of the pdpDetailEntry on line 147 below.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -1,36 +1,299 @@
...
+
+                PdpFeedGroupEntry groupEntry = buildPdpFeedGroupEntry(detailFileLine, buildPdpFeedPayeeIdEntry(detailFileLine), pdpDetailEntries);
+                groupEntries.add(groupEntry);
```
nkimble: Changing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Breaking out into another method.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: If you are questioning the addition of 1, the reason is:
We want just the file name to add to the database table entry to track duplicates.
Method lastIndexOf returns the index of the slash.
Method substring starts at the index provided and goes to the end of the string sent to it.
Thus, the returned value from the substring call was including the forward slash which is not part of the filename.

src/main/resources/edu/cornell/kfs/concur/cu-spring-concur.xml
```diff
@@ -259,17 +258,32 @@
...
 	</bean>
 
 	<bean id="concurCashAdvancePdpFeedFileService" parent="concurCashAdvancePdpFeedFileService-parentBean" />
 	<bean id="concurCashAdvancePdpFeedFileService-parentBean" class="edu.cornell.kfs.concur.batch.service.impl.ConcurCashAdvancePdpFeedFileServiceImpl" abstract="true">
-	    <property name="outgoingDirectoryName" value="${staging.directory}/concur/cashAdvance"/>
+	    <property name="concurRequestedCashAdvanceService" ref="concurRequestedCashAdvanceService"/>
+	    <property name="cuMarshalService" ref="cuMarshalService"/>
+	    <property name="dataDictionaryService" ref="dataDictionaryService"/>
+	    <property name="dateTimeService" ref="dateTimeService"/>
+        <property name="parameterService" ref="parameterService"/>
```
nkimble: fixing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
nkimble: changing as requested

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: removing this initialization

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
-            LOG.error("Header validation should have received single parsed file. More or less than one parsed file was detected.");
+        LOG.info("batchDate=" + requestExtractFile.getBatchDate() + "=   recordCount=" + requestExtractFile.getRecordCount() + "=     totalApprovedAmount=" + requestExtractFile.getTotalApprovedAmount() + "=");
+        headerValidationPassed = (fileRowCountMatchesHeaderRowCount(requestExtractFile) &&
+                                  fileOnlyContainsOurEmployeeCustomerIndicator(requestExtractFile) &&
+                                  fileTotalApprovedAmountsMatchHeaderTotalApprovedAmount(requestExtractFile));
+        LOG.info("Header validation : " + ((headerValidationPassed) ? "PASSED" : "FAILED"));
+        return headerValidationPassed;
+    }
+
+    public void performRequestDetailLineValidation(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
```
nkimble: I disagree. The fact that the service is updating flags on a class attribute of the detail line is part of the implementation between this service and the business object holding those validation results.  All of that should be hidden from the caller of a public method.  The way that validation is performed in the future could change that should not affect any of those public method names.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,169 @@
...
+        if (requestDetailFileLineIsCashAdvance(detailFileLine)) {
+            boolean lineValidationPassed = false;
```
nkimble: removing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileServiceImpl.java
db483: Ok, I think I understand now. You want to retrieve just the file name from the full path string.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
db483: My concern here is that when we call this method to perform a validation and pass in a BO it is not clear that we are also changing the state of that BO.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
db483: ok

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
db483: got it

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
jhulslander: I spoke with Nancy, and creating ConcurBatchUtilityServiceImpl seems to make the most sense.

nkimble:
Thread and number of changes getting to long for reviewers to manage.
Closing this PR and will create a new one.

PR #341, opened 2017-03-28
KFSPTS-8205 payee address concur to pdp

jhulslander:
Marcia is still reviewing the results from a local run so this isn't ready for merging yet.


jhulslander:
Please review this change but don't merge it yet.  I gave some output file to marcia to review before we get into kfs-dev

jhulslander:
Marcia agreed the addresses are in the bank file as expected, so I think we can move this to dev after it passes code review

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -87,6 +96,55 @@ public void loadPayments(PaymentFileLoad paymentFile, LoadPaymentStatus status,
...
+                paymentGroup.setState(employee.getAddressStateProvinceCodeUnmasked());
+                paymentGroup.setZipCd(employee.getAddressPostalCodeUnmasked());
+                paymentGroup.setCountry(employee.getAddressCountryCodeUnmasked());
+            } else {
+                LOG.error("setPaymentAddressFieldsForConcurPayments, unable to get a person from the employee ID: " + paymentGroup.getPayeeId());
+            }
+        }
+    }
+
+    private Person buildPerson(String employeeId) {
```
db483: I think retrieve or get Person would be a better name for this method since it is just retrieving the Person from the DB and not building it.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -87,6 +96,55 @@ public void loadPayments(PaymentFileLoad paymentFile, LoadPaymentStatus status,
...
+        Person person = null;
+        if (StringUtils.isNotBlank(employeeId)) {
+            try {
+                person = getPersonService().getPersonByEmployeeId(employeeId);
+            } catch (Exception e) {
```
db483: I am not sure I understand why we want to catch Exception here. getPersonByEmployeeId does not throw any exception. If there is an exception then probably we have a bigger problem and we want this to fail instead of catching and logging.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -87,6 +96,55 @@ public void loadPayments(PaymentFileLoad paymentFile, LoadPaymentStatus status,
...
+                LOG.debug("setPaymentAddressFieldsForConcurPayments, found a concur customer, for employee: " + employee.getName());
+                paymentGroup.setLine1Address(employee.getAddressLine1Unmasked());
+                paymentGroup.setLine2Address(employee.getAddressLine2Unmasked());
+                paymentGroup.setLine3Address(employee.getAddressLine3Unmasked());
+                paymentGroup.setCity(employee.getAddressCityUnmasked());
+                paymentGroup.setState(employee.getAddressStateProvinceCodeUnmasked());
+                paymentGroup.setZipCd(employee.getAddressPostalCodeUnmasked());
+                paymentGroup.setCountry(employee.getAddressCountryCodeUnmasked());
+            } else {
+                LOG.error("setPaymentAddressFieldsForConcurPayments, unable to get a person from the employee ID: " + paymentGroup.getPayeeId());
```
db483: Is the pdp load going to succeed if there is no address info? And if it does, do we want it to? If the address is needed to send the check to a person probably we don't want this to succeed. I think we need to clarify this with the functionals.

jhulslander: We validate employee ID at creation of the PDP XML, so this should never happen, we should probably fail as there is probably something wrong.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
jhulslander: OK

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
jhulslander: See my note above about failing.  Concerning the catch, some people have problems with their KIM record, Sherry Wilson for example sometimes has a problem associated with some bad data.  When this happens our functionals and Nicole fix the data.

db483: If we just catch and log as error then they will not know about the issue unless not having the address in those fields will cause a  failure down the stream in PDP.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
db483: Does the possibility exist of something changing in KIM between the time we validate the SAE and the time the PDP xml is loaded?

jhulslander:

src/main/java/edu/cornell/kfs/pdp/service/CuPdpEmploueeService.java
```diff
@@ -0,0 +1,13 @@
...
+import org.kuali.kfs.pdp.businessobject.PaymentFileLoad;
+
+public interface CuPdpEmploueeService {
+    
+    /**
+     * Based on the customer in the payment file, this function will decide if the caller should process the payee as an employee
+     * @param paymentFile
+     * @return
+     */
+    boolean shouldProcessPayeeAsEmployee(PaymentFileLoad paymentFile);
```
jhulslander: I couldn't think of a good name, that works

jhulslander:

PR #342, opened 2017-03-28
KFSPTS-8215 fix concur authentication filter

db483:


src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
```diff
@@ -34,13 +34,18 @@ public void doFilter(ServletRequest servletRequest, ServletResponse response, Fi
...
         if (StringUtils.isBlank(authorizationHeader)) {          
             httpServletResponse.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+            LOG.debug("doFilter(): Concur authentication failed");
+            return;
         } else {
             final String encodedUserNameAndPasswordAsToken = authorizationHeader.replaceFirst(ConcurConstants.BASIC_AUTHENTICATION_SCHEME + " ", "");
 
             if (!getConcurAuthenticationService().isConcurTokenValid(encodedUserNameAndPasswordAsToken)) {
                 httpServletResponse.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+                LOG.debug("doFilter(): Concur authentication failed");
```
jhulslander: Do we care to know how the login failed (no auth token versus wrong auth token)?  If so, maybe we want different debug statements.  I would be inclined to think they might be errors that we would want in the log.

db483: Good point, I will add more info in the log. Thank you!

db483:
I updated the code to include more information about why authentication failed. This is ready for review again. Thank you.

src/main/java/edu/cornell/kfs/concur/rest/authentication/ConcurAuthenticationFilter.java
db483: Let me change them to error too.

db483:
Ready for review.

jhulslander:
Thanks, I will merge after the unit tests run

PR #357, opened 2017-04-10
KFSPTS-5549: cleanup test

bhutchinson:


src/test/java/edu/cornell/kfs/module/receiptProcessing/service/CuReceiptProcessingServiceImplNegativeTest.java
```diff
@@ -28,66 +26,66 @@
...
+        deleteDirectory(BATCH_DIRECTORY);
+        deleteDirectory(RECEIPT_DIR);
     }
 
-    @Override
-    protected void tearDown() throws Exception {
-        File batchDirectoryFile = new File(BATCH_DIRECTORY);
+    private void deleteDirectory(String batchDirectory) {
+        File batchDirectoryFile = new File(batchDirectory);
         batchDirectoryFile.delete();
```
cah292: I don't think this deletion operation will delete the related parent directories in the path. FileUtils has some convenience methods for deleting a directory an its sub-directories, so perhaps one of those should be used here.

bhutchinson: Yeah, let me look into that - I think we may be doing that in another test class.

bhutchinson:
Yes, that should probably be a relative path - let me look into that and the other suggested change, thanks!

bhutchinson:
I implemented the suggestions and also some more cleanup. This is ready for review again, thanks.

bhutchinson:
It's still being referenced by CuReceiptProcessingServiceImplPositiveTest. We may want to clean that up in the future, but it won't be as easy to remove the spring dependency on that one.

bhutchinson:
Thanks @cah292

PR #358, opened 2017-04-10
KFSPTS-7869: Concur Request Extract file to PDP feed file base code that creates a PDP file. Includes initial version of Concur Batch Utility Service.

nkimble:


nkimble:
This is the base code that just creates the PDP feed file from the Concur Request Extract file. In a previous PR for this code, it was requested that the Concur Batch Utility Service be created and included with these changes.  These code changes require related nonprod-sql PR#77 and should be merged at the same time.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
```diff
@@ -0,0 +1,179 @@
...
+        String parameterValue = getParameterService().getParameterValueAsString(CUKFSConstants.ParameterNamespaces.CONCUR, CUKFSParameterKeyConstants.ALL_COMPONENTS, parameterName);
+        return parameterValue;
+    }
+    
+    public void createDoneFile(String fullyQualifiedFileName) throws FileStorageException {
+        getFileStorageService().createDoneFile(fullyQualifiedFileName);
+    }
+    
+    public void removeDoneFile(String fullyQualifiedFileName) throws FileStorageException {
+        getFileStorageService().removeDoneFiles(Collections.singletonList(fullyQualifiedFileName));
```
jhulslander: similar thoughts here as above

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    public String formatPdpPayeeName(String lastName, String firstName, String middleInitial) {
+        String fullName = lastName + KFSConstants.COMMA + KFSConstants.BLANK_SPACE + firstName;
+        if (StringUtils.isNotBlank(middleInitial)) {
+            fullName = fullName + KFSConstants.BLANK_SPACE + middleInitial + KFSConstants.DELIMITER;
+        }
+
+        int payeeNameFieldSize = (getDataDictionaryService().getAttributeMaxLength(PaymentGroup.class, PdpConstants.PaymentDetail.PAYEE_NAME)).intValue();
```
jhulslander: would it make sense to just use Integer for this variable and use the int value in the if statement.  The line reads better without the extra parenthesis.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,168 @@
...
-            LOG.error("Header validation should have received single parsed file. More or less than one parsed file was detected.");
+    /**
+     * A request detail line is considered a cash advance when request entry detail type is set to the cash advance indicator.
+     *
+     * @param detailFileLine
+     * @return boolean
+     */
+    private boolean requestDetailFileLineIsCashAdvance(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if (StringUtils.isNotEmpty(detailFileLine.getRequestEntryExpenseType()) &&
+            detailFileLine.getRequestEntryExpenseType().equalsIgnoreCase(ConcurConstants.CASH_ADVANCE_INDICATOR)) {
```
jhulslander: You could simply this if statement by using StringUtils.equalsIgnoreCase(detailFileLine.getRequestEntryExpenseType(), (ConcurConstants.CASH_ADVANCE_INDICATOR)

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -3,28 +3,168 @@
...
+            detailFileLine.getValidationResult().setCashAdvanceLine(true);
         }
         else {
-            ConcurRequestExtractFile requestExtractFile = requestExtractFiles.get(0);
-            LOG.info("batchDate=" + requestExtractFile.getBatchDate() + "=   recordCount=" + requestExtractFile.getRecordCount() + "=     totalApprovedAmount=" + requestExtractFile.getTotalApprovedAmount() + "=");
-            headerValidationPassed = (fileRowCountMatchesHeaderRowCount(requestExtractFile) && fileApprovedAmountsMatchHeaderApprovedAmount(requestExtractFile));
+            detailFileLine.getValidationResult().setCashAdvanceLine(false);
+            detailFileLine.getValidationResult().addMessage(new String(getConfigurationService().getPropertyValueAsString(ConcurKeyConstants.CONCUR_REQUEST_EXTRACT_NOT_CASH_ADVANCE_DATA_LINE)));
```
jhulslander: Just for my own understanding, why the need for the newString() in here?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -93,4 +93,7 @@
         public static final String SAE_MISSING_OBJECT_CODES_SUB_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE = "SAE_MISSING_OBJECT_CODES_SUB_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE";
     }
 
+    public static final String CASH_ADVANCE_INDICATOR = "CASH ADVANCE";
```
cah292: Is this really a payment code, though? We need to be careful not to rename it to something that could get confused with the SAE payment codes.

jhulslander: OK, I thought this was a payment code, but if it isn't then totally agree.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
```diff
@@ -0,0 +1,179 @@
...
+    protected CUMarshalService cuMarshalService;
+    protected ParameterService parameterService;
+
+    public String getConcurParameterValue(String parameterName) {
+        String parameterValue = getParameterService().getParameterValueAsString(CUKFSConstants.ParameterNamespaces.CONCUR, CUKFSParameterKeyConstants.ALL_COMPONENTS, parameterName);
+        return parameterValue;
+    }
+    
+    public void createDoneFile(String fullyQualifiedFileName) throws FileStorageException {
+        getFileStorageService().createDoneFile(fullyQualifiedFileName);
```
nkimble: IMHO, having "source" in the parameter name would be more confusing and could imply it is the concur data file name.   What do you think about renaming the method to "createDoneFileFor" and leaving the parameter as fullyQualifiedFileName?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
nkimble: Similar suggestion as above: "removeDoneFileFor"    
Thoughts?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
jhulslander: yeah, that is better

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
nkimble: Changing as requested.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
```diff
@@ -38,8 +178,40 @@ private boolean fileRowCountMatchesHeaderRowCount(ConcurRequestExtractFile reque
...
+            LOG.error("There are no Request Detail lines in the file.");
+            return false;
+        }
+        else {
+            String ourCustomerProfile = getConcurBatchUtilityService().getConcurParameterValue(ConcurParameterConstants.CONCUR_CUSTOMER_PROFILE_GROUP_ID);
+            if (StringUtils.isNotEmpty(ourCustomerProfile) && ourCustomerProfileIsOnAllRequestDetailLines(requestExtractFile, ourCustomerProfile)) {
+                return true;
+            }
+            else {
+                LOG.error("File contains Request Detail lines that do not match the employee customer profile group of: " + getConcurBatchUtilityService().getConcurParameterValue(ConcurParameterConstants.CONCUR_CUSTOMER_PROFILE_GROUP_ID));
```
cah292: This parameter value is already being stored in a variable above, so you could use that here to simplify the logging statement.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
nkimble: This is the hard coded string that will be on a Concur data file line to indicate a data line represents a cash advance and should be further looked at for possible processing and inclusion in the request extract PDP file being created.  
Would prefixing with "REQUEST_EXTRACT_" help to clear things up?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestedCashAdvanceServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+        Collection<ConcurRequestedCashAdvance> concurRequestedCashAdvances = null;
+        Map<String, String> fieldValues = new HashMap<String, String>();
+        
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.EMPLOYEE_ID, concurRequestedCashAdvance.getEmployeeId());
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.REQUEST_ID, concurRequestedCashAdvance.getRequestId());
+        fieldValues.put(ConcurPropertyConstants.ConcurRequestedCashAdvance.PAYMENT_AMOUNT, concurRequestedCashAdvance.getPaymentAmount().toString());
+        
+        concurRequestedCashAdvances = businessObjectService.findMatching(ConcurRequestedCashAdvance.class, fieldValues);
+        
+        boolean isDuplicate = (CollectionUtils.isEmpty(concurRequestedCashAdvances)) ? false : true;
```
cah292: It would be simpler to either just negate the condition or use CollectionUtils.isNotEmpty(), so that you can avoid using the ternary operator here.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: Changing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImpl.java
nkimble: Good catch, probably don't.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurRequestExtractFileFixture.java
```diff
@@ -13,27 +13,54 @@
...
+
     public final String fileName;
     public final Date batchDate;
     public final Integer recordCount;
     public final KualiDecimal totalApprovedAmount;
     public final List<ConcurRequestExtractRequestDetailFileLine> requestDetails;
     
-    private ConcurRequestExtractFileFixture (String fileName, Date batchDate, Integer recordCount, KualiDecimal totalApprovedAmount) {
+    private ConcurRequestExtractFileFixture (String fileName, Date batchDate, Integer recordCount,
+                                             KualiDecimal totalApprovedAmount, ConcurRequestExtractRequestDetailFileLine[] detailLines) {
```
cah292: There are a few things that could help simplify this constructor and the setup of the enum constants:

* You can pass in the date as a String or a Joda DateTime and then convert it to a Date when you're ready to use it in that form. Depending on how you store it in the enum, this can help ensure that your enum constants stay immutable (since Java Date objects are mutable).
* Instead of passing in an Integer object, you can have the constructor accept a primitive int and then have the constructor body wrap it in an Integer.
* Instead of passing in a KualiDecimal, you can have the constructor accept a primitive double and then have the constructor body convert it to a KualiDecimal.
* If the request extract lines are mutable, I would suggest keeping them in their immutable enum form until you're ready to convert them to the needed POJO form. And if you need to store a List of them on this enum, I would strongly recommend using an unmodifiable List, to keep your enum constants truly immutable.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
jhulslander: since it is used when inspected the expenseType, maybe REQUEST_EXTRACT_EXPECT_TYPE_CASH_ADVANCE?

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurRequestExtractRequestDetailFileLineFixture.java
```diff
@@ -11,23 +11,26 @@
...
     public final Date batchDate;
     public final String requestId;
-    public final KualiDecimal requestAmount;
+    public final KualiDecimal totalApprovedAmount;
+    public final String employeeGroupId;
     public final List<ConcurRequestExtractRequestEntryDetailFileLine> requestEntryDetails;    
-    
-    private ConcurRequestExtractRequestDetailFileLineFixture(Date batchDate, String requestId, KualiDecimal requestAmount, int numberRequestEntryDetailLines) {
+
+    private ConcurRequestExtractRequestDetailFileLineFixture(Date batchDate, String requestId, KualiDecimal totalApprovedAmount, String employeeGroupId, int numberRequestEntryDetailLines) {
```
cah292: See my comments on the other enum constructor; many of my points there could help simplify the setup of this enum as well. Also, in your enum constants, you don't need the "new String()" calls unless you absolutely need new instances of the Strings (in which case it would be simpler to let the constructor body handle that instead).

src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurRequestedCashAdvance.java
```diff
@@ -0,0 +1,86 @@
+package edu.cornell.kfs.concur.batch.businessobject;
+
+import java.sql.Date;
+
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+public class ConcurRequestedCashAdvance extends PersistableBusinessObjectBase{
+    protected int concurRequestedCashAdvanceId;
```
nkimble: I am confused by this recommendation since
https://db.apache.org/ojb/docu/guides/jdbc-types.html  
indicates that the JDBC type of INTEGER should be mapped to Java type int.

cah292: In that case, do you think INTEGER is large enough for what you need, or should it be BIGINT for this field instead?

Just for reference, Kuali BOs use Long and BIGINT for the version number field.

nkimble: The maximum number that could be generated for this field with data types JDBC INTEGER and Java int are values with a minimum value of -2,147,483,648 and a maximum value of 2,147,483,647.  This table would only be dealing with the positive side of this range. With this table representing only cash advance payments that were requested and generated, I think we are are going to be ok with those two existing data types.

nkimble:
Thanks for the unit test suggestions.  I have implemented some of them for the second code review. Due to the fact that I strongly anticipate this area undergoing substantial modifications with the next user story's round of coding, I would like to defer those suggestions not yet implemented until then.
Research has been performed regarding the database table primary key concern and I think we are ok.  This round of code review changes have been made, unit tests have passed, and code is ready for re-review.

jhulslander:
This round of changes look good to me

src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurRequestedCashAdvance.java
cah292: Ok, thanks for checking.

PR #359, opened 2017-04-10
KFSPTS-8357: Fixed SAE Collector .done file creation and error report.

cah292:
This fixes the issue where the .done file is not being created for the generated Collector file. The fix involves updating the appropriate code to return just the file name and not the full file path.

Also, as per a recent discussion with Nancy, I've updated the code so that it will not report a header-level validation error when writing the Collector data to a flat file. If more report-related logging for such errors is needed, then that will be addressed in a future user story.

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCreateCollectorFileServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCreateCollectorFileServiceImpl.java
@@ -112,7 +112,7 @@ public class ConcurStandardAccountingExtractCreateCollectorFileServiceImpl
 
     protected String writeToCollectorFile(String originalFileName, CollectorBatch collectorBatch) {
         String collectorFileName = buildCollectorFileName(originalFileName);
-        String collectorFilePath = buildCollectorFilePath(collectorFileName);
+        String collectorFilePath = buildFullyQualifiedCollectorFileName(collectorFileName);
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCreateCollectorFileServiceImpl.java
```diff
@@ -110,26 +110,25 @@ protected int countCollectorFilesGeneratedFromSAEFilesToday() {
...
-                .append(collectorDirectoryPath)
-                .append(ConcurConstants.COLLECTOR_CONCUR_OUTPUT_FILE_NAME_PREFIX)
-                .append(saeFileNameWithoutExtension)
-                .append(GeneralLedgerConstants.BatchFileSystem.EXTENSION)
-                .toString();
+        return ConcurConstants.COLLECTOR_CONCUR_OUTPUT_FILE_NAME_PREFIX + saeFileNameWithoutExtension
+                + GeneralLedgerConstants.BatchFileSystem.EXTENSION;
+    }
+
+    protected String buildCollectorFilePath(String collectorFileName) {
```
jhulslander: should this function buildCollectorFileFullyQualifiedFileName?  We are using the "fullyQualified" naming convention when the file name includes the path.

cah292: Good catch, I will update the naming accordingly.

cah292:
I've made some minor changes based on the initial feedback. This is now ready for review again.

cah292:
retest this please

jhulslander:
Changes look good to me

PR #363, opened 2017-04-11
KFSPTS-5947: move test data files to src/test/resources

bhutchinson:


bhutchinson:
I think that may be a problem with github. When I look at the git log in Intellij, it is showing the correct locations. Also, as you said, the files' contents are the same, so I think this is ok. Thanks.

cah292:
Ok, thanks for checking.

PR #367, opened 2017-04-13
Kfspts 8219 error message handling

jhulslander:
This fix truncates the error message list to 2000 characters before database insertion.  We are also aggregating error messages so there isn't duplication.

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
+++ b/src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
@@ -3,6 +3,7 @@ package edu.cornell.kfs.concur.businessobjects;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.sys.KFSConstants;
 
@@ -51,7 +52,7 @@ public class ValidationResult {
     
     private boolean isNotDuplicateMessage(String message) {
         for (String currentMessage : getMessages()) {
-            if (StringUtils.equals(currentMessage, message)) {
+            if (StringUtils.equalsIgnoreCase(currentMessage, message)) {
                 return false;
             }
         }
@@ -59,7 +60,7 @@ public class ValidationResult {
     }
 
     public void addMessages(List<String> messagesToAdd) {
-        if (messagesToAdd != null) {
+        if (CollectionUtils.isNotEmpty(messagesToAdd)) {
             for (String messageToAdd : messagesToAdd) {
                 addMessage(messageToAdd);
             }
--- a/src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
+++ b/src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
@@ -3,6 +3,7 @@ package edu.cornell.kfs.concur.businessobjects;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.sys.KFSConstants;
 
 public class ValidationResult {
@@ -43,15 +44,25 @@ public class ValidationResult {
         if (messages == null) {
             messages = new ArrayList<String>();
         }
-        messages.add(message);
+        if (StringUtils.isNotBlank(message) && isNotDuplicateMessage(message)) {
+            messages.add(message);
+        }
+    }
+    
+    private boolean isNotDuplicateMessage(String message) {
+        for (String currentMessage : getMessages()) {
+            if (StringUtils.equals(currentMessage, message)) {
+                return false;
+            }
+        }
+        return true;
     }
 
     public void addMessages(List<String> messagesToAdd) {
-        if (messages == null) {
-            messages = new ArrayList<String>();
-        }
         if (messagesToAdd != null) {
-            messages.addAll(messagesToAdd);
+            for (String messageToAdd : messagesToAdd) {
+                addMessage(messageToAdd);
+            }
         }
     }
 
--- a/src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
@@ -9,6 +9,7 @@ import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.krad.service.BusinessObjectService;
 import org.kuali.kfs.krad.util.KRADConstants;
 
+import edu.cornell.kfs.concur.ConcurConstants;
 import edu.cornell.kfs.concur.ConcurPropertyConstants;
 import edu.cornell.kfs.concur.businessobjects.ConcurEventNotification;
 import edu.cornell.kfs.concur.rest.xmlObjects.ConcurEventNotificationDTO;
@@ -48,8 +49,15 @@ public class ConcurEventNotificationServiceImpl implements ConcurEventNotificati
         concurEventNotification.setInProcess(inProcess);
         concurEventNotification.setProcessed(processed);
         concurEventNotification.setValidationResult(validationResult);
-        concurEventNotification.setValidationResultMessage(validationResultMessages);
+        concurEventNotification.setValidationResultMessage(truncateValidationResultMessageToMaximumDatabaseFieldSize(validationResultMessages));
         saveConcurEventNotification(concurEventNotification);
+    }
+    
+    protected String truncateValidationResultMessageToMaximumDatabaseFieldSize(String validationResultMessages) {
+        if (validationResultMessages != null && validationResultMessages.length() > ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH) {
+            validationResultMessages = validationResultMessages.substring(0, ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH);
+        }
+        return validationResultMessages;
     }
     
     @Override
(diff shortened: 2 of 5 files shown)


src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
```diff
@@ -43,15 +44,25 @@ public void addMessage(String message) {
...
             messages = new ArrayList<String>();
         }
-        messages.add(message);
+        if (StringUtils.isNotBlank(message) && isNotDuplicateMessage(message)) {
+            messages.add(message);
+        }
+    }
+    
+    private boolean isNotDuplicateMessage(String message) {
+        for (String currentMessage : getMessages()) {
```
nkimble: This function is presuming that the method caller is verifying "messages" is not null prior to being called which is ok for these changes as that is happening but if this private method is used elsewhere in this class in the future that may pose an issue.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
```diff
@@ -43,15 +44,25 @@ public void addMessage(String message) {
...
+            if (StringUtils.equals(currentMessage, message)) {
```
nkimble: I do not believe that this will deal with string case of the message contents..... meaning I don't think it will see "Error Encountered" as a duplicate of "ERROR encountered".  Is that an issue?

src/test/java/edu/cornell/kfs/concur/businessobjects/ValidationResultTest.java
```diff
@@ -0,0 +1,54 @@
...
+        validationResult.addMessage(TEST_STRING_TWO);
+        validationResult.addMessage(TEST_STRING_ONE);
+        validationResult.addMessage(TEST_STRING_THREE);
+        validationResult.addMessage(TEST_STRING_TWO);
+        assertEquals("There should only be three lines", 3, validationResult.getMessages().size());
+    }
+    
+    @Test
+    public void verifyAddMessages() {
+        List<String> messagesToAdd = new ArrayList();
```
cah292: Missing generics on ArrayList construction.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
```diff
@@ -48,10 +49,17 @@ public void updateConcurEventNotificationFlagsAndValidationMessage(ConcurEventNo
...
         concurEventNotification.setValidationResult(validationResult);
-        concurEventNotification.setValidationResultMessage(validationResultMessages);
+        concurEventNotification.setValidationResultMessage(truncateValidationResultMessageToMaximumDatabaseFieldSize(validationResultMessages));
         saveConcurEventNotification(concurEventNotification);
     }
     
+    protected String truncateValidationResultMessageToMaximumDatabaseFieldSize(String validationResultMessages) {
+        if (validationResultMessages != null && validationResultMessages.length() > ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH) {
+            validationResultMessages = validationResultMessages.substring(0, ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH);
+        }
```
nkimble: These constant names may be confusing since they are only being used in event notification validation to size the items for the database and have no meaning for the rest of the Concur processing (request extract, collector, sae) validation messages.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
cah292: The "messages" field is initialized to an empty list in the default constructor, or to an explicit list in the other constructor. Unless other code is trying to pass null lists into the constructors or setters, then I think this code is fine as-is.

src/test/java/edu/cornell/kfs/concur/businessobjects/ValidationResultTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+public class ValidationResultTest {
+    
+    private ValidationResult validationResult;
+    
+    private static final String TEST_STRING_ONE = "test string 1";
+    private static final String TEST_STRING_TWO = "test string 2";
+    private static final String TEST_STRING_THREE = "test string 3";
```
cah292: As a follow-up to the question @nkimble had about case sensitivity, it might be helpful to have some mixed-case values to test with in here.

src/test/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImplTest.java
```diff
@@ -27,5 +29,43 @@ public void validateFindingNotificationId() {
...
+        String actualResults = concurEventNotificationServiceImpl.truncateValidationResultMessageToMaximumDatabaseFieldSize(overageMessage);
+        assertEquals("message should be the long message without the extra stuff", longMessage, actualResults);
+    }
+    
+    private String buildLongMessage() {
+        StringBuilder sb = new StringBuilder(ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH);
+        for (int i = 0; i<ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH; i++) {
+            sb.append("A");
+        }
+        return sb.toString();
```
cah292: Since you're just repeating the same value to build the long string, you could simplify this by calling StringUtils.repeat() instead.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
```diff
@@ -48,10 +49,17 @@ public void updateConcurEventNotificationFlagsAndValidationMessage(ConcurEventNo
...
         concurEventNotification.setProcessed(processed);
         concurEventNotification.setValidationResult(validationResult);
-        concurEventNotification.setValidationResultMessage(validationResultMessages);
+        concurEventNotification.setValidationResultMessage(truncateValidationResultMessageToMaximumDatabaseFieldSize(validationResultMessages));
         saveConcurEventNotification(concurEventNotification);
     }
     
+    protected String truncateValidationResultMessageToMaximumDatabaseFieldSize(String validationResultMessages) {
+        if (validationResultMessages != null && validationResultMessages.length() > ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH) {
+            validationResultMessages = validationResultMessages.substring(0, ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH);
```
cah292: You could simplify this whole method by calling StringUtils.left() instead.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
jhulslander: The messages are built by validation service, so we shouldn't have a case sensitivity issue, but it won't hurt to put the check in.

src/main/java/edu/cornell/kfs/concur/businessobjects/ValidationResult.java
```diff
@@ -43,15 +44,25 @@ public void addMessage(String message) {
...
+            }
+        }
+        return true;
     }
 
     public void addMessages(List<String> messagesToAdd) {
-        if (messages == null) {
-            messages = new ArrayList<String>();
-        }
         if (messagesToAdd != null) {
```
jhulslander: Nice, I will switch it

src/test/java/edu/cornell/kfs/concur/businessobjects/ValidationResultTest.java
jhulslander: I'll change, but I am confused, I thought you only need the Generics identifier on one of the two.

src/test/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImplTest.java
jhulslander: good idea

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
jhulslander: good idea, I will switch

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurEventNotificationServiceImpl.java
jhulslander: This is a pre-existing constant.  What would you recommend?

jhulslander:
I made the changes suggested.  I made use of the existing constant ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH .  I am not sure what a better name is for this constant.

nkimble:
Since the existing constant ConcurConstants.VALIDATION_RESULT_MESSAGE_MAX_LENGTH  contains the name of the data object, I am ok with leaving it as-is.

src/test/java/edu/cornell/kfs/concur/businessobjects/ValidationResultTest.java
cah292: You still need at least the empty "<>" on the other one, so that the compiler knows to derive the generic type information from the left-hand side. Otherwise, the right-hand side will just be using a raw type.

src/test/java/edu/cornell/kfs/concur/businessobjects/ValidationResultTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void verifyAddNullCollectionMessage() {
+        validationResult.addMessage(TEST_STRING_ONE);
+        validationResult.addMessages(null);
+        assertEquals("There should only be one line", 1, validationResult.getMessages().size());
+    }
+    
+    @Test
+    public void verifyAddEmptyCollectionMessage() {
+        validationResult.addMessage(TEST_STRING_ONE);
+        validationResult.addMessages(new ArrayList());
```
cah292: Missing generics on ArrayList. Adding at least an empty "<>" should allow the compiler to derive the correct generic type, provided that the addMessages() method signature has enough information.

jhulslander:
I fixed the mixing generics from the unit test,

jhulslander:
Thanks for the reviews Chad & Nancy, the unit tests completed, and it looks like both of you are OK with this PR now.  Can this be merged?

jhulslander:
thanks

PR #374, opened 2017-04-19
KFSPTS-8398 sae cash advance

jhulslander:
This is ready for review.  Chad, please pay close attention to the validation updates.  My changes work for PDP processing, but I am not sure if they work for Collector processing.


src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurStandardAccountingExtractDetailLine.java
```diff
@@ -31,6 +31,10 @@
     private Date reportEndDate;
     private String policy;
     private String expenseType;
+    private KualiDecimal cashAdvanceAmount;
+    private Date cashAdvanceIssueDate;
+    private String cashAdvanceCaKey;
```
cah292: I would suggest renaming this to just "cashAdvanceKey" for simplicity, along with the other places in this PR that reference this or have a similar naming pattern.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +165,104 @@ public Integer getPayeeNameFieldSize() {
...
+            
+            for (PdpFeedDetailEntry detailEntry : groupEntry.getDetail()) {
+                PdpFeedDetailEntry newDetailEntry = copyDetailEntry(detailEntry);
+                KualiDecimal transactionTotal = KualiDecimal.ZERO;
+                for (PdpFeedAccountingEntry accountingEntry : detailEntry.getAccounting()) {
+                    transactionTotal = transactionTotal.add(accountingEntry.getAmount());
+                    if (accountingEntry.getAmount().isGreaterThan(KualiDecimal.ZERO)) {
+                        newDetailEntry.getAccounting().add(copyAccountingEntry(accountingEntry));
+                    } else {
+                        LOG.debug("buildNewGroupEntries, not adding acounting entry: " + accountingEntry.getDebugInfo());
```
cah292: Minor typo in log message.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
```diff
@@ -189,13 +199,16 @@ private ConcurSAEDetailLineFixture(ConcurSAEDetailLineFixture baseFixture, Concu
                 overrideMap.getOrDefault(LineField.ORG_REF_ID, baseFixture.orgRefId),
                 journalDebitCredit,
                 journalAmount,
-                overrideMap.getOrDefault(LineField.REPORT_END_DATE, baseFixture.reportEndDate));
+                overrideMap.getOrDefault(LineField.REPORT_END_DATE, baseFixture.reportEndDate), 
+                overrideMap.getOrDefault(LineField.CASH_ADVANCE_CA_KEY, ""),
+                overrideMap.getOrDefault(LineField.REPORT_ENTRY_ID, ""));
```
cah292: These should default to the baseFixture field values instead of defaulting to "".

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedFileBaseEntryFixture.java
```diff
@@ -12,92 +15,239 @@
...
-        trailer.setDetailTotAmt(new KualiDecimal(96.95));
-        return trailer;
-    }
+public enum PdpFeedFileBaseEntryFixture {
 
-    private static List<PdpFeedGroupEntry> buildGroups() {
-        List<PdpFeedGroupEntry> groups = new ArrayList();
-        groups.add(buildGroup("Bimbo Foods Inc"));
-        return groups;
+    MARSHAL_TEST(HeaderEntry.BASIC_HEADER, TrailerEntry.BASIC_TRAILER, EnumSet.of(GroupEntry.MARSHAL_TEST)),
```
cah292: To simplify the setup of these enums, you could have the last constructor arg be a var-arg and then pass it into EnumSet.of() to create the Set.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCashAdvanceServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+import edu.cornell.kfs.concur.batch.service.ConcurStandardAccountingExtractCashAdvanceService;
+import edu.cornell.kfs.concur.businessobjects.ConcurAccountInfo;
+
+public class ConcurStandardAccountingExtractCashAdvanceServiceImpl implements ConcurStandardAccountingExtractCashAdvanceService {
+    
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurStandardAccountingExtractCashAdvanceServiceImpl.class);
+
+    @Override
+    public boolean isCashAdvanceLine(ConcurStandardAccountingExtractDetailLine line) {
+        return StringUtils.isNotEmpty(line.getCashAdvanceCaKey());
```
db483: would it be better to use isNotBlank here?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCashAdvanceServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+        String cashAdvanceReportEntryId = cashAdvanceLine.getReportEntryId();
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("findAccountingInfoForCashAdvanceLine, cashAdvanceReportEntryId: " + cashAdvanceReportEntryId);
+        }
+        for (ConcurStandardAccountingExtractDetailLine line : saeLines) {
+            if (StringUtils.equalsIgnoreCase(line.getReportEntryId(), cashAdvanceReportEntryId) && !isCashAdvanceLine(line)) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("findAccountingInfoForCashAdvanceLine, setting accountNumber to " + line.getAccountNumber() + " and chart to " + line.getChartOfAccountsCode());
+                }
+                info.setAccountNumber(line.getAccountNumber());
```
db483: do you need to also set subAccount, object code, sub object code, orgRefId and project code here?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
```diff
@@ -201,18 +205,28 @@ private boolean validateReportId(ConcurStandardAccountingExtractDetailLine line,
...
+            LOG.debug("validateAccountingLine, found a cash advance line, no need to validate");
+            return true;
+        }
+    }
+    
+    private ConcurAccountInfo buildOverriddenConcurAccountingInformation(ConcurStandardAccountingExtractDetailLine line) {
         String overriddenObjectCode = getParameterService().getParameterValueAsString(CUKFSConstants.ParameterNamespaces.CONCUR, 
                 CUKFSParameterKeyConstants.ALL_COMPONENTS, ConcurParameterConstants.CONCUR_SAE_PDP_DEFAULT_OBJECT_CODE);
-        ConcurAccountInfo overriddenConcurAccountingInformation = new ConcurAccountInfo(line.getChartOfAccountsCode(), line.getAccountNumber(), 
+        ConcurAccountInfo overriddenConcurAccountingInformation = new ConcurAccountInfo(line.getChartOfAccountsCode(),  line.getAccountNumber(),
```
db483: it looks like the only change here is adding an extra space after "line.getChartOfAccountsCode()," and I don't think that os needed

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurAccountInfo.java
```diff
@@ -9,6 +9,11 @@
     protected String objectCode;
     protected String subObjectCode;
     protected String projectCode;
+    protected String orgRefId;
+    
+    public ConcurAccountInfo() {
```
db483: I wonder if it would make sense to add another constructor that also takes in the orgRefId as a param.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +165,104 @@ public Integer getPayeeNameFieldSize() {
...
+    private void buildNewGroupEntries(PdpFeedFileBaseEntry pdpFeedFileBaseEntry, PdpFeedFileBaseEntry baseEntry) {
+        for (PdpFeedGroupEntry groupEntry : pdpFeedFileBaseEntry.getGroup()) {
+            PdpFeedGroupEntry newGroupEntry = copyGroupEntry(groupEntry);
+            boolean validNewGroup = false;
+            
+            for (PdpFeedDetailEntry detailEntry : groupEntry.getDetail()) {
+                PdpFeedDetailEntry newDetailEntry = copyDetailEntry(detailEntry);
+                KualiDecimal transactionTotal = KualiDecimal.ZERO;
+                for (PdpFeedAccountingEntry accountingEntry : detailEntry.getAccounting()) {
+                    transactionTotal = transactionTotal.add(accountingEntry.getAmount());
```
db483: is it possible for the accountingEntry amount here to be negative? if so should this line only be called  inside the if if (accountingEntry.getAmount().isGreaterThan(KualiDecimal.ZERO))?

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
jhulslander: OK, I will remove

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedFileBaseEntryFixture.java
jhulslander: I am not sure how that works, I may need some assistance.  I will move the sub enums into different fixture classes as suggested

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCashAdvanceServiceImpl.java
jhulslander: good catch

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCashAdvanceServiceImpl.java
jhulslander: I can add the other attributes.  I think it will only have the two I bring in, but if they start adding the other attributes, we will have them

src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurStandardAccountingExtractDetailLine.java
jhulslander: OK, I will change

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractValidationServiceImpl.java
jhulslander: hmm, that's odd, I will remove the extra space

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurAccountInfo.java
jhulslander: good idea

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: OK, I will fix accounting

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: This was by design.  Yes, it could be negative.  We want to sum all the transactions of the detail, not just the positive transaction we add to "Cleaned" detail.  If the sum of all the transactions for a detail section isn't positive, we don't want to include the detail section to the group.  This case means, the total cash advances are more than what the person spent.  So it will be handled by the collector process.  Is there something I could do here to make my intent more clear?

jhulslander:
I made the code review changes, this is ready for review.  Please hold off on merging this code though.  I want to do some more local validation first.

src/main/java/edu/cornell/kfs/concur/batch/xmlObjects/PdpFeedAccountingEntry.java
```diff
@@ -124,5 +124,13 @@ public KualiDecimal getAmount() {
     public void setAmount(KualiDecimal value) {
         this.amount = value;
     }
+    
+    public String getDebugInfo() {
```
nkimble: Shouldn't this method be named "toString()" to conform to Java standards? 
I saw this method "getDebugInfo()"in other SAE business objects and thought that it was odd it was not the toString function.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +165,104 @@ public Integer getPayeeNameFieldSize() {
...
+                    if (accountingEntry.getAmount().isGreaterThan(KualiDecimal.ZERO)) {
```
cah292: KualiDecimal has an isPositive() method, if you want to simplify this expression.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +165,104 @@ public Integer getPayeeNameFieldSize() {
...
+                        newDetailEntry.getAccounting().add(copyAccountingEntry(accountingEntry));
+                    } else {
+                        LOG.debug("buildNewGroupEntries, not adding accounting entry: " + accountingEntry.getDebugInfo());
+                    }
+                }
+                LOG.debug("buildNewGroupEntries, total transaction for " + newDetailEntry.getSourceDocNbr() + " detail: " + transactionTotal);
+                if (transactionTotal.isGreaterThan(KualiDecimal.ZERO)) {
```
cah292: Another area that could be simplified via KualiDecimal.isPositive()

src/main/java/edu/cornell/kfs/concur/batch/xmlObjects/PdpFeedAccountingEntry.java
jhulslander: I sit corrected, yes it should be toString, I will change this, and a few other places I have used this convention

src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurStandardAccountingExtractDetailLine.java
```diff
@@ -31,6 +31,10 @@
     private Date reportEndDate;
     private String policy;
     private String expenseType;
+    private KualiDecimal cashAdvanceAmount;
+    private Date cashAdvanceIssueDate;
```
cah292: Are cashAdvanceAmount and cashAdvanceIssueDate actually needed on this POJO? I don't see them actively used by non-logging code, and I don't think I need them for the Collector processing.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
```diff
@@ -189,13 +199,16 @@ private ConcurSAEDetailLineFixture(ConcurSAEDetailLineFixture baseFixture, Concu
...
+                overrideMap.getOrDefault(LineField.CASH_ADVANCE_KEY, baseFixture.cashAdvanceKey),
+                overrideMap.getOrDefault(LineField.REPORT_ENTRY_ID, baseFixture.reportEntryId));
     }
 
     private ConcurSAEDetailLineFixture(ConcurSAEFileFixture extractFile, ConcurEmployeeFixture employee,
             String reportId, String paymentCode, String chartOfAccountsCode,
             String journalAccountCode, String accountNumber, String subAccountNumber, String subObjectCode,
-            String projectCode, String orgRefId, String journalDebitCredit, double journalAmount, String reportEndDate) {
+            String projectCode, String orgRefId, String journalDebitCredit, double journalAmount, String reportEndDate,
+            String cashAdvanceCaKey, String reportEntryId) {
```
cah292: It would be a good idea to rename the cash advance arg to be consistent with the "cashAdvanceKey" naming on the detail POJO.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEFileFixture.java
```diff
@@ -37,7 +37,8 @@
     DOCUMENT_NUMBER_TEST(4, 200.00),
     EMPLOYEE_NAME_TEST(4, 200.00),
     CASH_AND_CARD_TEST(2, 105.00),
-    CANCELED_TRIP_TEST(2, -432.40);
+    CANCELED_TRIP_TEST(2, -432.40),
+    PDP_TEST(3, 1000);
```
cah292: Since you have four SAE detail lines for your PDP test, the first arg should be a 4 instead of a 3. (I might refactor this enum in the future so that the line count is derived instead of explicitly given here.)

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedDetailEntryFixture.java
```diff
@@ -0,0 +1,55 @@
...
+    DETAIL_TWO_TRANS_SUM_TO_ZERO(0, EnumSet.of(PdpFeedAccountingEntryFixture.TRANSACTION_AMOUNT_10, PdpFeedAccountingEntryFixture.TRANSACTION_AMOUNT_NEGATIVE_10)),
+    DETAIL_TWO_TRANS_SUM_TO_NEGATIVE(-5, EnumSet.of(PdpFeedAccountingEntryFixture.TRANSACTION_AMOUNT_5, PdpFeedAccountingEntryFixture.TRANSACTION_AMOUNT_NEGATIVE_10));
+    
+    public final String sourceDocNbr;
+    public final String invoiceNbr;
+    public final String invoiceDate;
+    public final double origInvoiceAmt;
+    public final String fsOriginCd;
+    public final String fdocTypCd;
+    public final EnumSet<PdpFeedAccountingEntryFixture> accountingEntries;
```
cah292: EnumSet is mutable. If you want your fixtures to be truly immutable, then I would suggest wrapping the EnumSet inside of a Collections.unmodifiableSet() (which can happen inside the constructor body for convenience).

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedFileBaseEntryFixture.java
```diff
@@ -12,92 +15,42 @@
...
+    FEED_ONE_TRANS_ZERO(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_ONE_TRANS_ZERO)),
+    FEED_ONE_TRANS_ZER_ONE_TRANS_POSITIVE(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_ONE_TRANS_ZERO_ONE_TRANS_POSITIVE)),
+    FEED_TWO_TRANS_SUM_TO_POSITIVE(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_TWO_TRANS_SUM_TO_POSITIVE)),
+    FEED_TWO_TRANS_SUM_TO_ZERO(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_TWO_TRANS_SUM_TO_ZERO)),
+    FEED_TWO_TRANS_SUM_TO_NEGATIVE(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_TWO_TRANS_SUM_TO_NEGATIVE)),
+    FEED_TWO_DETAILS_ONE_POSITIVE_ONE_ZERO(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_TWO_DETAILS_ONE_POSITIVE_ONE_ZERO)),
+    FEED_TWO_GROUPS_ONE_ZERO(EnumSet.of(PdpFeedGroupEntryFixture.GROUP_TWO_TRANS_SUM_TO_ZERO, PdpFeedGroupEntryFixture.GROUP_TWO_TRANS_SUM_TO_POSITIVE));
+    
+    public final PdpFeedHeaderEntryFixture headerEntry;
+    public final EnumSet<PdpFeedGroupEntryFixture> groupEntries;
```
cah292: Another area where it would be helpful to wrap the EnumSet inside of an immutable Set.

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedGroupEntryFixture.java
```diff
@@ -0,0 +1,75 @@
...
+    
+    public final String payeeName;
+    public final String payeeIdType;
+    public final String payeeId;
+    public final String payeeOwnershipCode;
+    public final String customerInstitutionIdentifier;
+    public final String paymentDate;
+    public final String combineGroupInd;
+    public final String bankCode;
+    public final EnumSet<PdpFeedDetailEntryFixture> detailEntries;
```
cah292: Another area that could benefit from an immutable Set.

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedGroupEntryFixture.java
```diff
@@ -0,0 +1,75 @@
...
+        this.payeeOwnershipCode = payeeOwnershipCode;
+        this.customerInstitutionIdentifier = customerInstitutionIdentifier;
+        this.paymentDate = paymentDate;
+        this.combineGroupInd = combineGroupInd;
+        this.bankCode = bankCode;
+        this.detailEntries = detailEntries;
+    }
+    
+    private PdpFeedGroupEntryFixture(String payeeName, String payeeIdType, String payeeId, String customerInstitutionIdentifier, EnumSet<PdpFeedDetailEntryFixture> detailEntries) {
+        this(payeeName, payeeIdType, payeeId, "xyz", customerInstitutionIdentifier, PdpFeedFileConstants.PAYMENT_DATE, PdpFeedFileConstants.COMBINE_GROUP_INDICATOR_YES,
```
cah292: It might be helpful to move the "xyz" literal to its own constant.

src/main/java/edu/cornell/kfs/concur/batch/businessobject/ConcurStandardAccountingExtractDetailLine.java
jhulslander: OK, I will remove these, I thought I heard in the meeting that you would need them, I guess I misheard

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: ah, good catch

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedDetailEntryFixture.java
jhulslander: OK

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedFileBaseEntryFixture.java
jhulslander: OK

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEFileFixture.java
jhulslander: good catch

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
jhulslander: ok

src/test/java/edu/cornell/kfs/concur/batch/fixture/PdpFeedDetailEntryFixture.java
jhulslander: After testing out, I had to change two things:
public final Collection<PdpFeedAccountingEntryFixture> accountingEntries;
And in the constructor:
this.accountingEntries = Collections.unmodifiableSet(accountingEntries);

cah292: The constructor line sounds correct, but for the field on the enum, you could use Set instead of Collection if you want to.

jhulslander:

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
```diff
@@ -122,7 +122,15 @@
     CANCELED_TRIP_TEST_LINE1(DEFAULT_CREDIT, ConcurSAEFileFixture.CANCELED_TRIP_TEST, -442.40,
             buildOverride(LineField.PAYMENT_CODE, ConcurConstants.PAYMENT_CODE_UNIVERSITY_BILLED_OR_PAID)),
     CANCELED_TRIP_TEST_LINE2(DEFAULT_DEBIT, ConcurSAEFileFixture.CANCELED_TRIP_TEST, 10.00,
-            buildOverride(LineField.PAYMENT_CODE, ConcurConstants.PAYMENT_CODE_UNIVERSITY_BILLED_OR_PAID));
+            buildOverride(LineField.PAYMENT_CODE, ConcurConstants.PAYMENT_CODE_UNIVERSITY_BILLED_OR_PAID)),
+    
+    PDP_TEST_CASH_ADVANCE_500(DEFAULT_CREDIT, ConcurSAEFileFixture.PDP_TEST, -500, buildOverride(LineField.CASH_ADVANCE_KEY, ConcurTestConstants.PDP_LINE_FIXTURE_CASH_ADVANCE_KEY),
+            buildOverride(LineField.REPORT_ENTRY_ID, ConcurTestConstants.PDP_LINE_FIXTURE_REPORT_ENTRY_ID), 
+            buildOverride(LineField.CHART_OF_ACCOUNTS_CODE, StringUtils.EMPTY), buildOverride(LineField.ACCOUNT_NUMBER, StringUtils.EMPTY)),
+    PDP_TEST_DEBIT_1_500(DEFAULT_DEBIT, ConcurSAEFileFixture.PDP_TEST, 50, buildOverride(LineField.REPORT_ENTRY_ID, ConcurTestConstants.PDP_LINE_FIXTURE_REPORT_ENTRY_ID)),
```
cah292: Is this amount meant to be 50 or 500? The name suggests the latter but the third argument suggests the former.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: This total is for the whole report.  Individual accounting lines that sum to zero are removed from the PDP file above.  The concern I was thinking about in my email doesn't happen wth way concur processes cash advance.
The SAE from 4/18 had some data for Sherry W. that that proved various iteration of cash advance lines working.  So I think this is correct, and I don't need to keep track of another transaction line total here.

cah292: Ok, thanks for clarifying.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEDetailLineFixture.java
jhulslander: good catch, I will fix

jhulslander:
I fixed the enum name as noted

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
cah292: There's one other thing I thought of; sorry for not mentioning it earlier. If the transaction total is positive but there are still some cash advances in there with negative values, do any adjustments need to be made on the positive detail entries as a result? (For example, would the reimbursable amount be reduced by the cash advance amount?) Or does the PDP file not need to worry about this? Or does the higher-level grouping prevent situations like this?

jhulslander: That was the crutch of the concern I was working with Sherry about this afternoon.  The way Concur handles Cash Advances and the summing of transactions in a Concur report / PDP Detail section, we should have this scenario handled correctly.

src/test/java/edu/cornell/kfs/concur/batch/fixture/ConcurSAEFileFixture.java
```diff
@@ -37,7 +37,8 @@
     DOCUMENT_NUMBER_TEST(4, 200.00),
     EMPLOYEE_NAME_TEST(4, 200.00),
     CASH_AND_CARD_TEST(2, 105.00),
-    CANCELED_TRIP_TEST(2, -432.40);
+    CANCELED_TRIP_TEST(2, -432.40),
+    PDP_TEST(4, 1000);
```
cah292: The amount total is not lining up with the detail lines from the other fixture. However, since this enum constant is not actively used by the unit test code that needs the correct value, it may be okay to leave it as-is for now, and then fix it later if problems arise.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
cah292: Ok, thanks for the confirmation.

db483: My concern here is that if the accountingEntry.getAmount() is negative it doesn't get added to the accounting entries on the newDetailEntry on line 198. I thought that only the entries with a zero amount were supposed to be omitted.

db483:
Changes look good. I only had one last question on the negative accounting entries.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
jhulslander: Before this code runs, previous code aggregates SAE file transactions into PDP transactions.  The negative SAE transactions get sumed, so cash advance amounts are subtracted to find what Cornell owes the traveler..  If there is a negative PDP transaction, that should get handled by collector since the payee owes Cornell, so we don't want to process that into PDP.

db483: Thank you for explaining this Jay, it makes sense now. 

Jay and I chatted about ways to improve this code so that this would be more clear. Jay is currently looking into re-working some things to add clarity.

jhulslander:
I tried to refactor removeNonReimbursableSectionsFromPdpFeedFileBaseEntry for clarity.  It feels better to me now.  This is ready for review again.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +167,121 @@ public Integer getPayeeNameFieldSize() {
...
+        }
+        if (originalDetailTransactionTotal.isPositive()) {
+            newGroupEntry.getDetail().add(newDetailEntry);
+        } else {
+            if (LOG.isDebugEnabled()) {
+                StringBuilder sb = new StringBuilder("addNewPaymentDetailToGroup SAE transactions summed to");
+                if (originalDetailTransactionTotal.isNegative()) {
+                    sb.append(" a negative amount in PDP, so it will be handled by collector.");
+                } else {
+                    sb.append(" a zerion amount in PDP, so no payment needs to be issued");
```
db483: there is a spelling error in this message

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +167,121 @@ public Integer getPayeeNameFieldSize() {
...
+        }
+    }
+    
+    private void addDetailEntriesToNewGroupEntry(PdpFeedGroupEntry newGroupEntry, PdpFeedGroupEntry originalGroupEntry) {
+        for (PdpFeedDetailEntry originalDetailEntry : originalGroupEntry.getDetail()) {
+            PdpFeedDetailEntry newDetailEntry = copyDetailEntry(originalDetailEntry);
+            KualiDecimal originalDetailTransactionTotal = KualiDecimal.ZERO;
+            for (PdpFeedAccountingEntry originalAccountingEntry : originalDetailEntry.getAccounting()) {
+                originalDetailTransactionTotal = originalDetailTransactionTotal.add(originalAccountingEntry.getAmount());
+                if (originalAccountingEntry.getAmount().isPositive()) {
```
db483: I think that it would be more clear if the if/else code was extracted in a separate method called: addOnlyPositiveAccountingEntriesForPDPProcessing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +167,121 @@ public Integer getPayeeNameFieldSize() {
...
+                    newDetailEntry.getAccounting().add(copyAccountingEntry(originalAccountingEntry));
+                } else {
+                    LOG.debug("addDetailEntriesToNewGroupEntry, not adding accounting entry: " + originalAccountingEntry.toString());
+                }
+            }
+            addNewPaymentDetailToGroup(newGroupEntry, newDetailEntry, originalDetailTransactionTotal);
+        }
+    }
+
+    private void addNewPaymentDetailToGroup(PdpFeedGroupEntry newGroupEntry, PdpFeedDetailEntry newDetailEntry, KualiDecimal originalDetailTransactionTotal) {
```
db483: I think it would be more clear if this method was called addOnlyDetailEntriesWithPositiveTotalsForPDPProcessing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +167,121 @@ public Integer getPayeeNameFieldSize() {
...
-        }
         return parameterValue;
     }
 
     public void setPayeeNameFieldSize(Integer payeeNameFieldSize) {
         this.payeeNameFieldSize = payeeNameFieldSize;
     }
     
+    @Override
+    public PdpFeedFileBaseEntry removeNonReimbursableSectionsFromPdpFeedFileBaseEntry(PdpFeedFileBaseEntry pdpFeedFileBaseEntry,
```
db483: maybe this method too could be called: createPdpFileBaseEntryThatDoesNotContainNonReimbursableSections since this method is not really altering the input entry but creating a new one

db483:
Hi Jay, I think the code looks better. I added some further suggestions for the method that creates the new pdpFileBaseEntry.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
cah292: Or, since this method is only adding up to one payment detail at a time, perhaps it would be better to rename it to "addNewPaymentDetailToGroupIfTotalIsPositive"

db483: good point Chad

cah292:
There's one other thing I wanted to mention. Because of the validation changes made in this PR, the Collector side of the processing may start generating bad Collector files since the SAE-to-Collector Cash Advance handling has not been implemented in develop yet. However, since there's really not much more to validate on the SAE-to-Collector side until the Cash Advance and/or personal expense processing is ready, I'm still fine with this PR moving forward once the other comments have been addressed. Just keep in mind that the Collector files may fail to load to the GL, or may fail to get generated by the batch job.

jhulslander:
I renamed the functions as requested, and fixed the fixture as noted as well.  This is ready for review again

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +167,126 @@ public Integer getPayeeNameFieldSize() {
...
+            KualiDecimal originalDetailTransactionTotal = KualiDecimal.ZERO;
+            for (PdpFeedAccountingEntry originalAccountingEntry : originalDetailEntry.getAccounting()) {
+                originalDetailTransactionTotal = originalDetailTransactionTotal.add(originalAccountingEntry.getAmount());
+                addOnlyPositiveAccountingEntriesForPDPProcessing(newDetailEntry, originalAccountingEntry);
+            }
+            addNewPaymentDetailToGroupIfTotalIsPositive(newGroupEntry, newDetailEntry, originalDetailTransactionTotal);
+        }
+    }
+
+    private void addOnlyPositiveAccountingEntriesForPDPProcessing(PdpFeedDetailEntry newDetailEntry,
```
cah292: This method is only adding up to one accounting entry at a time, even though the name implies that multiple entries could be added. It might make sense to rename this method to reflect that. Sorry for not noticing that in the suggestions for the previous round of review.

jhulslander:
I renamed the function as requested

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +160,126 @@ public Integer getPayeeNameFieldSize() {
...
     }
 
     public void setPayeeNameFieldSize(Integer payeeNameFieldSize) {
         this.payeeNameFieldSize = payeeNameFieldSize;
     }
     
+    @Override
+    public PdpFeedFileBaseEntry createPdpFileBaseEntryThatDoesNotContainNonReimbursableSections(PdpFeedFileBaseEntry pdpFeedFileBaseEntry, 
+            ConcurStandardAccountingExtractBatchReportData reportData) {
+        LOG.debug("Entering removeNonReimbursableSectionsFromPdpFeedFileBaseEntry");
```
db483: the log should reflect the new method name

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +160,126 @@ public Integer getPayeeNameFieldSize() {
...
+        if (originalAccountingEntry.getAmount().isPositive()) {
+            newDetailEntry.getAccounting().add(copyAccountingEntry(originalAccountingEntry));
+        } else {
+            LOG.debug("addOnlyPositiveAccountingEntryForPDPProcessing, not adding accounting entry: " + originalAccountingEntry.toString());
+        }
+    }
+
+    private void addNewPaymentDetailToGroupIfTotalIsPositive(PdpFeedGroupEntry newGroupEntry, PdpFeedDetailEntry newDetailEntry, KualiDecimal originalDetailTransactionTotal) {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("addNewPaymentDetailToGroup, total transaction for " + newDetailEntry.getSourceDocNbr() + " detail: " + originalDetailTransactionTotal);
```
db483: method name should be updated in the log message

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -162,16 +160,126 @@ public Integer getPayeeNameFieldSize() {
...
+        }
+        if (originalDetailTransactionTotal.isPositive()) {
+            newGroupEntry.getDetail().add(newDetailEntry);
+        } else {
+            if (LOG.isDebugEnabled()) {
+                StringBuilder sb = new StringBuilder("addNewPaymentDetailToGroup SAE transactions summed to");
```
db483: method name should be updated in the log message

jhulslander:
Good catch on the log messages, they have been fixed now

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountExtractPdpEntryServiceImpl.java
```diff
@@ -28,8 +24,10 @@
 import edu.cornell.kfs.concur.batch.xmlObjects.PdpFeedHeaderEntry;
 import edu.cornell.kfs.concur.batch.xmlObjects.PdpFeedPayeeIdEntry;
 import edu.cornell.kfs.concur.batch.xmlObjects.PdpFeedTrailerEntry;
+import edu.cornell.kfs.concur.businessobjects.ConcurAccountInfo;
 import edu.cornell.kfs.sys.CUKFSConstants;
 import edu.cornell.kfs.sys.CUKFSParameterKeyConstants;
+import freemarker.core._RegexBuiltins.split_reBI;
```
db483: what is this import for?

jhulslander:
Eclipse was weird for me Friday, code insight was getting me to the wrong libraries, it was an erroneous import that I missed on previous clean up.

db483:
Changes were approved and unit tests passed. Merging.

PR #380, opened 2017-04-24
KFSPTS-7379: Code to create a report for the  Request Extract to PDP feed file generation batch job.

nkimble:


nkimble:
Please do not merge until I comment back. 
Creating code review to keep this work moving forward. 
Meeting with SherryW on Tuesday morning to confirm report content.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -51,32 +53,73 @@ public boolean createPdpFeedFileForValidatedDetailFileLines(ConcurRequestExtract
         return pdpFileSuccessfullyCreated;
     }
 
-    private PdpFeedFileBaseEntry buildPdpFeedBaseEntry(ConcurRequestExtractFile requestExtractFile) {
+    private PdpFeedFileBaseEntry buildPdpFeedBaseEntry(ConcurRequestExtractFile requestExtractFile, ConcurRequestExtractBatchReportData reportData) {
```
jhulslander: this seems like a long function, can this be refactored? to make shorter methods and remove some duplication

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -51,32 +53,73 @@ public boolean createPdpFeedFileForValidatedDetailFileLines(ConcurRequestExtract
...
+        int totalCashAdvancesUseInExpenseReportCount = 0;
+        KualiDecimal totalCashAdvancesUseInExpenseReportAmount= KualiDecimal.ZERO;
+        int totalDuplicateCashAdvanceCount = 0;
+        KualiDecimal totalDuplicateCashAdvanceAmount = KualiDecimal.ZERO;
+        int totalNonCashAdvanceRecordsCount = 0;
+        KualiDecimal totalNonCashAdvanceRecordsAmount = KualiDecimal.ZERO;
+        int totalCashAdvancesNotProcessedDueToValidationErrorsCount = 0;
+        KualiDecimal totalCashAdvancesNotProcessedDueToValidationErrorsAmount = KualiDecimal.ZERO;
+        int totalProcessingAttemptedLineCount = 0;
+        KualiDecimal totalProcessingAttemptedLineAmount = KualiDecimal.ZERO;
```
jhulslander: I'm curious,  It looks like you are setting these values in this function, and then just update the corresponding values in the report data object at the end.  What does the local copy of these values buy us?  Would it make more sense just to use the values in the report data object?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractReportServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import edu.cornell.kfs.concur.ConcurConstants;
+import edu.cornell.kfs.concur.batch.report.ConcurBatchReportLineValidationErrorItem;
+import edu.cornell.kfs.concur.batch.report.ConcurBatchReportSummaryItem;
+import edu.cornell.kfs.concur.batch.report.ConcurRequestExtractBatchReportData;
+import edu.cornell.kfs.concur.batch.service.ConcurRequestExtractReportService;
+import edu.cornell.kfs.sys.service.ReportWriterService;
+
+public class ConcurRequestExtractReportServiceImpl implements ConcurRequestExtractReportService {
```
jhulslander: There are a couple of long functions in this class.  Can they be refactored per clean code.

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
```diff
@@ -96,4 +96,17 @@
...
+        public static final String NO_VALIDATION_ERROR_MESSAGES_TO_OUTPUT = "No validation errors to output.";
+        public static final String END_OF_REPORT_MESSAGE = "End of report.";
+        public static final String PREFIX_FIRST_PART_NOT_SET_IN_CONFIGURATION_FILE = "PREFIX_FIRST_PART_NOT_SET_IN_CONFIGURATION_FILE";
+        public static final String PREFIX_SECOND_PART_NOT_SET_IN_CONFIGURATION_FILE = "PREFIX_SECOND_PART_NOT_SET_IN_CONFIGURATION_FILE";
+        public static final String REQUEST_EXTRACT_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE = "REQUEST_EXTRACT_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE";
+        public static final String REQUEST_EXTRACT_SUMMARY_REPORT_SUB_TITLE_NOT_SET_IN_CONFIGURATION_FILE = "REQUEST_EXTRACT_SUMMARY_REPORT_SUB_TITLE_NOT_SET_IN_CONFIGURATION_FILE";
+        public static final String REQUEST_EXTRACT_REPORT_CONCUR_FILE_NAME_LABEL_NOT_SET_IN_CONFIGURATION_FILE = "REQUEST_EXTRACT_REPORT_CONCUR_FILE_NAME_LABEL_NOT_SET_IN_CONFIGURATION_FILE";
+        public static final String REQUEST_EXTRACT_VALIDATION_SUB_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE = "REQUEST_EXTRACT_VALIDATION_SUB_REPORT_TITLE_NOT_SET_IN_CONFIGURATION_FILE";
```
nkimble: I am not sure we want to do that.  These are just default markers to indicate in the actual report that the actual string label is not set the way it should be. Those actual string labels conform to the standard of the other foundation reports by being defined in the spring xml file where the report writer is defined. In this case in cu-spring-concur.xml.  If anything, the actual labels in the xml should be in the cu-application.properties file but that would mean the Concur report code would not conform to the foundation code.
If we change how these defaults and actual string labels are populated for the request extract report, we will need to have a new user story created to have the same change also made for the SAE PDP/Collector Concur report.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: The "longness" is due to the addition of the processing needed for the report data into a function that is mainly control flow. See my comment below.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: When I first coded the report data processing into this function, I had it coded as you suggested and it did not seem read as clearly which values were being updated and why.  If you and Daniela feel strongly about this method using the report data object values directly, I can refactor the code to remove the "updateReportData..." methods and just do that summing directly against the reportData attributes.  Let me know your preference.

jhulslander: I'll let Daniela give her opinion on the matter.  Would it help have helper methods in the report data object that add an amount to a total, and increment the related counter?

src/main/java/edu/cornell/kfs/concur/ConcurConstants.java
jhulslander: sounds reasonable to me then.  Lets leave as is.

nkimble:
Code review changes are completed. Local testing has been completed. Please consider this version of the code as mergable if your noted issues have been addressed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -172,6 +185,68 @@ private PdpFeedTrailerEntry buildPdpFeedTrailerEntry(int totalDetailRecordCount,
...
+        reportData.getCashAdvancesProcessedInPdp().setRecordCount(count);
+        KualiDecimal amount = reportData.getCashAdvancesProcessedInPdp().getDollarAmount();
+        reportData.getCashAdvancesProcessedInPdp().setDollarAmount(amount.add(detailFileLine.getRequestAmount()));
+    }
+
+    private void updateReportDataClonedCashAdvanceCountAndAmountTotals(ConcurRequestExtractBatchReportData reportData, ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        int count = reportData.getClonedCashAdvanceRequests().getRecordCount();
+        count++;
+        reportData.getClonedCashAdvanceRequests().setRecordCount(count);
+        KualiDecimal amount = reportData.getClonedCashAdvanceRequests().getDollarAmount();
```
db483: I think it might be cleaner to add a new method in the ConcurBatchReportSummaryItem class that adds the input amount to the DollarAmount.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -15,73 +16,115 @@
...
     
     @Test
     public void testHeaderAmountAndHeaderRowCountsMatch() {
+        LOG.info("testHeaderAmountAndHeaderRowCountsMatch");
         ConcurRequestExtractFile testFile = ConcurRequestExtractFileFixture.GOOD_FILE.createConcurRequestExtractFile();
-        assertTrue("Expected Result: Header amount SHOULD match sum of row amounts from file.", concurRequestExtractFileValidationService.requestExtractHeaderRowValidatesToFileContents(testFile));
+        ConcurRequestExtractBatchReportData reportData = new ConcurRequestExtractBatchReportData();
+        assertTrue("Expected Result: Header amount SHOULD match sum of row amounts from file.", concurRequestExtractFileValidationService.requestExtractHeaderRowValidatesToFileContents(testFile, reportData));
+        LOG.info(reportData.getHeaderValidationErrors());
+        System.out.println("");
```
db483: Is this println needed?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
```diff
@@ -15,73 +16,115 @@
...
+        reportData = null;
```
db483: is this null set needed?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -51,31 +53,42 @@ public boolean createPdpFeedFileForValidatedDetailFileLines(ConcurRequestExtract
...
-
-              recordCashAdvanceGenerationInDuplicateTrackingTable(detailFileLine, pdpDetailEntry.getSourceDocNbr(), requestExtractFile.getFileName());
-
-              totalPdpDetailRecordsCount++;
-              totalPdpDetailRecordsAmount = totalPdpDetailRecordsAmount.add(detailFileLine.getRequestAmount());
-          }
-          else if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isNotValidCashAdvanceLine()) {
-              LOG.info("buildPdpFeedBaseEntry: Cash Advance was detected but validation failed for:  " + KFSConstants.NEWLINE + detailFileLine.toString());
-          }
+                PdpFeedDetailEntry pdpDetailEntry = buildPdpFeedDetailEntry(detailFileLine, buildPdpFeedAccountingEntry(detailFileLine));
```
db483: I was wondering if this code could be re-written as a call to a method that builds the pdpFeedsBaseEntry and one that updates the report data. Can the two be separated in such a way that they can be done in 2 separate methods?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -172,6 +185,68 @@ private PdpFeedTrailerEntry buildPdpFeedTrailerEntry(int totalDetailRecordCount,
         return trailer;
     }
 
+    private void updateReportDataPdpFeedFileCountAndAmountTotals(ConcurRequestExtractBatchReportData reportData, ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        int count = reportData.getCashAdvancesProcessedInPdp().getRecordCount();
```
nkimble: Looking at doing this.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Looking at doing this.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: It gives white space when you run the unit tests so that it is easier to see each unit test in its entirety. I could change it to a LOG.info(""): instead if you prefer.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
nkimble: Following the paradigm of "tear down" in each unit test to ensure garbage collection is invoked as this data item is being created cleanly for each unit test.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: That is what I was attempting before which the reviewers did not like because of the local variables.  The state of each line of data from the file is held in file line's attributes.  Our options are:

(a) Have two separate methods with the exact same for-lopp and if-else-if structure that goes through all of the data lines of the file.  One would build the pdpFeedsBaseEntry and the other method would build the data for the report. I do not like this as we would not be actually reporting on the processing that has happened to create the pdpFeedsBaseEntry. We would be recreating the report data asynchronously afterward.

(b) Put all of the local counters back in and call a method that would set all the report data after the pdpFeedsBaseEntry was built.  Previous review comments indicated that was not an acceptable clean code approach but it gives you the counters for the processing that are separate from the reporting where a separate method can be called to generate the report data. This is my preference.

(c) Leaving it as. Not my preference either since those two things are are separate.

Let me know how to proceed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
jhulslander: Just offering a thought, how would it be if the functions like updateReportDataPdpFeedFileCountAndAmountTotals were reportData object?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractFileValidationServiceImplTest.java
db483: Log.info would probably be better. Or maybe the getHeaderValidationErrors could have a new line or space added at the end?

nkimble:
retest this please

nkimble:
Code review changes have been made and processing has been validated locally.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -51,35 +53,40 @@ public boolean createPdpFeedFileForValidatedDetailFileLines(ConcurRequestExtract
...
+            updateReportDataForDetailFileLineBeingProcessed(reportData, detailFileLine, totalPdpDetailRecordsCount, totalPdpDetailRecordsAmount);
         }
         pdpBaseEntry.setGroup(groupEntries);
         pdpBaseEntry.setTrailer(buildPdpFeedTrailerEntry(totalPdpDetailRecordsCount, totalPdpDetailRecordsAmount));
         pdpBaseEntry.setVersion(ConcurConstants.FEED_FILE_ENTRY_HEADER_VERSION);
         return pdpBaseEntry;
     }
 
+    private boolean isDetailFileLineValidCashAdvanceRequest(ConcurRequestExtractRequestDetailFileLine detailFileLine) {
+        if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isValidCashAdvanceLine()) {
```
jhulslander: Can we just return the boolean expression?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -172,6 +179,80 @@ private PdpFeedTrailerEntry buildPdpFeedTrailerEntry(int totalDetailRecordCount,
...
+            updateReportDataTravelRequestsOnlyCountAndAmountTotals(reportData, detailFileLine);
+            LOG.info("updateReportDataForDetailFileLineBeingProcessed: NOT a cash advance :  " + KFSConstants.NEWLINE + detailFileLine.toString());
+        }
+        else if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isNotValidCashAdvanceLine()) {
+            updateReportDataValidationErrorCountAndAmountTotals(reportData, detailFileLine);
+            updateReportDataWithFileLineValidationError(reportData, detailFileLine);
+            LOG.info("updateReportDataForDetailFileLineBeingProcessed: Cash Advance but validation failed :  " + KFSConstants.NEWLINE + detailFileLine.toString());
+        }
+        else {
+            LOG.info("updateReportDataForDetailFileLineBeingProcessed: ***PROBLEM*** detailFileLine Line NOT included in any report totals and this should NOT happen :  " + KFSConstants.NEWLINE + detailFileLine.toString());
```
jhulslander: I would make this a LOG.error message.  Is a log message sufficient for when this happens, do we need to do something else?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
```diff
@@ -172,6 +179,80 @@ private PdpFeedTrailerEntry buildPdpFeedTrailerEntry(int totalDetailRecordCount,
         return trailer;
     }
 
+    private void updateReportDataForDetailFileLineBeingProcessed(ConcurRequestExtractBatchReportData reportData, ConcurRequestExtractRequestDetailFileLine detailFileLine, int totalPdpDetailRecordsCount, KualiDecimal totalPdpDetailRecordsAmount) {
+        if (isDetailFileLineValidCashAdvanceRequest(detailFileLine)) {
+            updateReportDataPdpFeedFileCountAndAmountTotals(reportData, totalPdpDetailRecordsCount, totalPdpDetailRecordsAmount);
+            LOG.info("updateReportDataForDetailFileLineBeingProcessed: Cash advance included in PDP feed file :  " + KFSConstants.NEWLINE + detailFileLine.toString());
+        }
+        else if (detailFileLine.getValidationResult().isCashAdvanceLine() && detailFileLine.getValidationResult().isClonedCashAdvance()) {
```
jhulslander: I think it would be helpful to extract the boolean expressions in the if statements to methods to make this more readable

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractReportServiceImpl.java
```diff
@@ -0,0 +1,315 @@
...
+            getReportWriterService().writeFormattedMessageLine("--------------------------------------------------------");
+            getReportWriterService().writeFormattedMessageLine("The following header row validation errors were detected");
+            getReportWriterService().writeFormattedMessageLine("--------------------------------------------------------");
+            for (String errorString : reportData.getHeaderValidationErrors()) {
+                getReportWriterService().writeFormattedMessageLine(errorString);
+            }
+            getReportWriterService().writeNewLines(2);
+        }
+        else {
+            LOG.debug("writeHeaderValidationErrors, NO header validation errors detected");
```
jhulslander: Should this be included on the report?  Just asking.

src/main/java/edu/cornell/kfs/concur/batch/report/ConcurBatchReportSummaryItem.java
```diff
@@ -37,12 +37,22 @@ public void setRecordCount(int recordCount) {
         this.recordCount = recordCount;
     }
 
+    public void incrementRecordCount() {
+        int count = this.getRecordCount();
+        count++;
+        this.setRecordCount(count);
```
jhulslander: recordCount++; doesn't work?

src/main/java/edu/cornell/kfs/concur/batch/report/ConcurBatchReportSummaryItem.java
```diff
@@ -37,12 +37,22 @@ public void setRecordCount(int recordCount) {
...
         return dollarAmount;
     }
 
     public void setDollarAmount(KualiDecimal dollarAmount) {
         this.dollarAmount = dollarAmount;
     }
- 
+
+    public void addDollarAmount(KualiDecimal amountToAdd) {
+        this.setDollarAmount(this.getDollarAmount().add(amountToAdd));
```
jhulslander: dollarAmount = dollarAmount.add(amountToAdd); doesn't work?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Changing the code as requested but IMHO the code is clearer to understand with the if-statement. 
With returning just the result of the boolean expression, you need to look at everything in that expression to actually understand if true or false will be returned.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurCashAdvancePdpFeedFileServiceImpl.java
nkimble: Changing to LOG.error. 
At this point the code could not do anything to address the issue as this would be a condition that was not considered in the spec. I don't think that we would want to fail the batch job. I will make a note in the User Story that does the emailing of the report to include a check where the sum all the "counts" and "amounts" is compared to the corresponding totals values in the attribute references on line 212 below.  If either count or amount summation of the report attributes does not match exactly to those totals then flag an alert on the email of the report being sent.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestExtractReportServiceImpl.java
nkimble: I think it would be distracting noise to the users.

nkimble:
retest this please

jhulslander:
@nkimble this looks good to me.  Are you ready for this to be merged?

nkimble:
This code has been validated locally. Unit tests have passed. IMHO, it is ready to be merged.

PR #382, opened 2017-04-24
KFSPTS-8358: Implement SAE-to-Collector Cash Advance handling.

cah292:
Please do NOT merge this yet! I still need to make some bean-related changes and perform more local validation.

This is an initial draft for adding Cash Advance processing to the SAE-to-Collector code. Some sub-grouping functionality has also been added to simplify a few areas of existing code.

In the main grouping class, there's a few areas where I could potentially eliminate some duplication, but the only practical way would be to add a new interface that impacts the ConcurRequestedCashAdvance BO and the SAE detail line POJO. If you think that doing so would not be too risky or impacting, then I might make that update on the next commit.


db483:
retest this please

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -156,14 +222,30 @@ protected KualiDecimal calculateTotalAmountForLines(List<ConcurStandardAccountin
...
+        OriginEntryFull offsetEntry = buildOffsetOriginEntry(cashEntry, cashAmountToOffset);
         offsetEntry.setFinancialObjectCode(collectorHelper.getCashOffsetObjectCode());
         offsetEntry.setFinancialSubObjectCode(collectorHelper.getDashOnlyPropertyValue(KFSPropertyConstants.SUB_OBJECT_CODE));
         
         return Optional.of(offsetEntry);
     }
 
+    protected KualiDecimal calculateTotalAmountForCashAdvanceLinesReferencedByRegularLines(
+            List<ConcurStandardAccountingExtractDetailLine> regularDetailLines) {
+        return regularDetailLines.stream()
```
jhulslander: Can this statement be refactored for clarity?  Maybe creation of some well named variables would help.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -178,20 +260,56 @@ protected KualiDecimal calculateTotalAmountForLines(List<ConcurStandardAccountin
         return Optional.of(offsetEntry);
     }
 
+    protected Optional<OriginEntryFull> doNotBuildOffsetOriginEntry(
```
jhulslander: This is confusing, since it just returns an empty option, why are there input parameters.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -27,39 +32,73 @@
...
+            cashAdvanceLinesByReportEntryId.put(detailLine.getReportEntryId(), detailLine);
+        } else {
+            String accountingFieldsKey = buildAccountingFieldsKey(detailLine);
+            consolidatedRegularLines.computeIfAbsent(accountingFieldsKey, ConcurDetailLineSubGroupForCollector::new)
+                    .addDetailLine(detailLine);
+        }
+    }
+
+    protected boolean isCashAdvanceLine(ConcurStandardAccountingExtractDetailLine detailLine) {
+        return StringUtils.isNotBlank(detailLine.getCashAdvanceKey());
```
cah292: Good catch, I had started implementing the use of your new service method but I didn't completely change things over in this class. I'll make that change.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: I was hoping to use streams for simplifying this operation, but I'll see if I can adjust naming and other portions to make this more readable.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: This is being used as a method reference by earlier code. I'll update things accordingly to make the intent clearer.

cah292:
I've committed some changes based on code review feedback, as well as some additional fixes and improvements. As for the de-duplication mentioned in my first comment, I am not planning to implement that at this time, to avoid potentially breaking the Request Extract code; it can possibly be done as a future enhancement.

This is ready for review again, and can be merged once the reviewers are satisfied and the unit tests pass.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -79,47 +142,47 @@ protected String defaultToDashesIfBlank(String value, String propertyName) {
         return StringUtils.defaultIfBlank(value, collectorHelper.getDashOnlyPropertyValue(propertyName));
     }
 
-    public List<OriginEntryFull> buildOriginEntries() {
-        List<OriginEntryFull> originEntries = new ArrayList<>();
+    /**
+     * Generates OriginEntryFull BOs from the SAE lines and passes them to the given Consumer.
```
db483: This method has a lengthy comment to explain what it does. Perhaps the method could be refactored for better clarity and the comment removed?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -79,47 +142,47 @@ protected String defaultToDashesIfBlank(String value, String propertyName) {
...
+     * If there was an error condition that should cause all of the lines of the current Report ID
+     * to be skipped (such as a Cash Advance line missing a corresponding Requested Cash Advance BO),
+     * the SAE lines and their error messages will be passed to the given BiConsumer instead.
+     * 
+     * @param entryConsumer A Consumer that the generated OriginEntryFull BOs can be passed to.
+     * @param lineErrorReporter A BiConsumer that can accept an SAE line and its error message.
+     */
+    public void buildAndAddOriginEntries(Consumer<OriginEntryFull> entryConsumer,
+            BiConsumer<ConcurStandardAccountingExtractDetailLine, String> lineErrorReporter) {
```
db483: What is the advantage of using a BiConsumer here?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -255,4 +400,50 @@ protected KualiDecimal getSignedAmountFromOriginEntry(OriginEntryFull originEntr
...
+            linesForGroup.addAll(cashAdvanceLines);
+        }
+        
+        Collections.sort(linesForGroup, this::compareBySequenceNumberAsPositiveNumericString);
+        return linesForGroup;
+    }
+
+    protected int compareBySequenceNumberAsPositiveNumericString(
+            ConcurStandardAccountingExtractDetailLine line1, ConcurStandardAccountingExtractDetailLine line2) {
+        int lengthComparison = line1.getSequenceNumber().length() - line2.getSequenceNumber().length();
```
db483: is this needed? doesn't the String.compareTo method already take care of this?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollectorHelper.java
```diff
@@ -79,4 +97,18 @@ public String formatDate(Date value, String format) {
         return dateTimeService.toString(value, format);
     }
 
+    public ConcurRequestedCashAdvance getRequestedCashAdvanceByCashAdvanceKey(String cashAdvanceKey) {
+        Collection<ConcurRequestedCashAdvance> requestedCashAdvances = concurRequestedCashAdvanceService
+                .findConcurRequestedCashAdvanceByCashAdvanceKey(cashAdvanceKey);
+        if (CollectionUtils.size(requestedCashAdvances) == 1) {
```
db483: if more than one is found we just want to return null?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineSubGroupForCollector.java
```diff
@@ -0,0 +1,64 @@
...
+    public ConcurDetailLineSubGroupForCollector() {
+        this.cashLines = new ArrayList<>();
+        this.corporateCardLines = new ArrayList<>();
+    }
+
+    /**
+     * Convenience constructor that takes a Map-related key as input and then discards it;
+     * is only meant for use as a method reference to simplify certain Map operations.
+     */
+    public <T> ConcurDetailLineSubGroupForCollector(T mapKey) {
```
db483: this constructor is confusing. what is it's utility?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -27,39 +32,70 @@
  */
 public class ConcurDetailLineGroupForCollector {
 
+    protected static final String ORPHANED_CASH_ADVANCE_MESSAGE_FRAGMENT = "Cash Advance did not have a corresponding entry from the Request Extract.";
```
cah292: Yes, it would be good to move these to the application resources eventually. If you think this is critical enough to have it done now, then I can do it for this PR; otherwise, we can defer it to a future clean-up effort.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -27,39 +32,70 @@
...
+            String accountingFieldsKey = buildAccountingFieldsKey(detailLine);
+            consolidatedRegularLines.computeIfAbsent(accountingFieldsKey, ConcurDetailLineSubGroupForCollector::new)
+                    .addDetailLine(detailLine);
+        }
+    }
+
+    protected ConcurRequestedCashAdvance getExistingRequestedCashAdvanceByCashAdvanceKey(String cashAdvanceKey) {
+        ConcurRequestedCashAdvance requestedCashAdvance = collectorHelper.getRequestedCashAdvanceByCashAdvanceKey(cashAdvanceKey);
+        if (ObjectUtils.isNull(requestedCashAdvance)) {
+            // The null Requested Cash Advance is only flagged here; the addOriginEntriesForLines() method will do the error reporting.
```
cah292: Ok, I'll remove this comment and update the code for clarity.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: Ok, I'll work on refactoring this into multiple methods for clarity and less commenting.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: This was meant to simplify calling the ConcurStandardAccountingExtractCollectorBatchBuilder.reportUnprocessedLine() method, without having to pass a reference to the ConcurStandardAccountingExtractCollectorBatchBuilder that is calling this buildAndAddOriginEntries() method.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -128,7 +191,7 @@ protected void addOriginEntriesForCorporateCardLines(
      * @param offsetGenerator A BiFunction that may create an offset entry using the generated regular entry and the detailLines List.
      */
     protected void addOriginEntriesForLines(
-            List<OriginEntryFull> originEntries, List<ConcurStandardAccountingExtractDetailLine> detailLines,
+            Consumer<OriginEntryFull> entryConsumer, List<ConcurStandardAccountingExtractDetailLine> detailLines,
```
cah292: It allows for this class to call CollectorBatch.addOriginEntry(), without passing in the CollectorBatch object itself or creating an intermediate List that the calling code then has to insert. (It may not make sense to pass in the entire CollectorBatch, since this code only needs to invoke that one method on the object and the Consumer allows for that.)

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -156,14 +219,45 @@ protected KualiDecimal calculateTotalAmountForLines(List<ConcurStandardAccountin
     protected Optional<OriginEntryFull> buildOriginEntryForCashOffset(
             OriginEntryFull cashEntry, List<ConcurStandardAccountingExtractDetailLine> cashLines) {
         KualiDecimal cashAmount = getSignedAmountFromOriginEntry(cashEntry);
+        KualiDecimal cashAdvanceAmount = calculateTotalAmountForCashAdvanceLinesReferencedByRegularLines(cashLines);
+        KualiDecimal cashAmountToOffset = cashAmount.add(cashAdvanceAmount);
+        if (cashAmountToOffset.isZero()) {
+            return Optional.empty();
```
cah292: Using an Optional allows for returning a wrapper around zero or one other objects, to avoid simply returning null for when a value does not exist. If this should just return null for simplicity, I could do that instead.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -178,20 +272,57 @@ protected KualiDecimal calculateTotalAmountForLines(List<ConcurStandardAccountin
         return Optional.of(offsetEntry);
     }
 
+    // This is only meant for use as a method reference, as an argument to the addOriginEntriesForLines() method.
```
cah292: Ok, I'll remove this comment.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -255,4 +400,50 @@ protected KualiDecimal getSignedAmountFromOriginEntry(OriginEntryFull originEntr
         }
     }
 
+    protected void reportAllLinesForGroupAsFailuresDueToOrphanedCashAdvanceLines(
+            BiConsumer<ConcurStandardAccountingExtractDetailLine, String> lineErrorReporter) {
```
cah292: See my comment on your other BiConsumer question, since this BiConsumer is being used for the same purpose.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: I need to compare these Strings as if they were positive integers. With String.compareTo(), the Strings are compared first by character content and then by length; I need it to be the reverse for numeric comparsion. (For example, "10" would precede "2" if I used String.compareTo().) I could alternatively parse them into numbers and then compare them, but that could result in a lot of construct-and-discard operations.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollectorHelper.java
cah292: There should be no more than one ConcurRequestedCashAdvance object per cash advance key. The current implementation of this ConcurRequestedCashAdvanceService method is just returning the results of a BusinessObjectService.findMatching() operation, which returns a Collection.

Perhaps ConcurRequestedCashAdvanceService could be updated to return zero or one result instead, but that is beyond the scope of this PR and should be deferred to a future enhancement.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineSubGroupForCollector.java
cah292: A reference for this method was meant to be used as the last argument for the Map.computeIfAbsent() method, so that I wouldn't need a lambda expression to construct the argument. I'll see if I can clarify the intent of this or just replace its use case with a lambda expression instead.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
db483: Since it's only a couple of them I think it might be better to move them now. Thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
db483: ok, thank you for explaining this.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
db483: where is the addOriginEntry called?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
db483: I understand this now, thank you.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: Ok, I'll move these to the application resources.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollectorHelper.java
db483: thank you for explaining this. I agree with leaving it as it is.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: This Consumer (and hence addOriginEntry()) is being invoked in the addOriginEntriesForLines() method.

db483: thank you, makes sense now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
db483: thank you for explaining this. Let's leave it as is for now.

cah292:
I've made some more changes based on code review feedback. This is now ready for review again, but I need to do a quick local test before this can be merged. I'll add another comment once I've validated things accordingly.

As a result of some of the latest changes, I also modified some of the key constant and app resource classes/files that are also impacted by the other pending PR. I tried to place things in a way that should not cause merge conflicts.

cah292:
My local testing was successful. This can be merged once the reviewers are satisfied with the latest changes.

jhulslander:
The issues I noted about in my first code review were adjusted, it looks like Daniela's concerns were addressed, and I don't see anything new of concern here.  Should I merge this PR?

cah292:
Did Daniela mention something at the standup about moving the Concur code along? I can't remember.

jhulslander:
I think she did say we should go ahead, but I wanted to double check on this

jhulslander:
Just got pinged by Daniela, she gave the OK that if the PR looks fine to me, we can go ahead with this

cah292:
As long as you're satisfied that your findings and Daniela's findings have been addressed, then I'm fine with this code being merged. We can create a new clean-up user story if we end up discovering problems later.

cah292:
Can Daniela's change-request-related block be dismissed to allow the merge, or does she have to explicitly approve first?

jhulslander:
I can't merge this PR, it is blocked by Daniela's change request.  I am not sure the best way to go around this.  @bhutchinson pinging you in to see if you know

bhutchinson:
What if you click "Dismiss review" by @db483 requested changes?

jhulslander:
Per some research Chad did, dismissing Daniela's review is the cleanest, I shall do

jhulslander:
dismissing worked, I merged the PR

PR #388, opened 2017-04-28
KFSPTS-8521: fix name masking for IWNT and REQS

bhutchinson:


bhutchinson:
So, typically, this will just set the value to be the unmasked value for the initiator, which should be ok. In the case of batch, it is unmasking the requester's information and that is ok, per Cathy.

bhutchinson:
Thanks @cah292!

PR #396, opened 2017-05-09
KFSPTS-8305: Fixed editing of IWNT document description.

cah292:
This just updates the readOnly attribute on the IWNT doc description field so that its editability is being controlled properly. (The IWNT refactoring for 7.x had accidentally introduced a bug with that field.)


nkimble:
Unit tests have passed. Merging change.

PR #398, opened 2017-05-10
KFSPTS-8592: Add superuser tab to IWNT document.

cah292:
This PR adds the superuser tab to the IWNT document, which got missed during the 7.x conversion. When the IWNT is displayed in its normal/full view, the tab will be placed at the bottom just before the action buttons, like what we do on other documents. When the IWNT is displayed in its "wizard" mode, the tab will be placed beneath the notes/attachments tab on the third "wizard" step, to line up with what the functionals had requested for placement.

### DIFF
--- a/src/main/webapp/jsp/module/purap/IWant.jsp
+++ b/src/main/webapp/jsp/module/purap/IWant.jsp
@@ -83,6 +83,10 @@
 
         <purap:iWantNotes attachmentTypesValuesFinderClass="${documentEntry.attachmentTypesValuesFinderClass}"
                           defaultOpen="true"/>
+        
+        <c:if test="${not isRegularStep}">
+            <kul:superUserActions/>
+        </c:if>
     </c:if>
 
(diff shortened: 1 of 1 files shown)


src/main/webapp/jsp/module/purap/IWant.jsp
```diff
@@ -83,6 +83,10 @@
 
         <purap:iWantNotes attachmentTypesValuesFinderClass="${documentEntry.attachmentTypesValuesFinderClass}"
                           defaultOpen="true"/>
+        
+        <c:if test="${not isRegularStep}">
```
nkimble: Lines 52 and 126 use "!" instead of "not" for negation. Should this line also use that construct?

cah292: Yes, I agree it would be good to update this to "!" for consistency with the other two lines; I'll make the change. Thanks for checking.

cah292:
I've made changes based on feedback. This is now ready for review again.

nkimble:

PR #399, opened 2017-05-10
KFSPTS-6383: Fix capitalization entry's object type for Scrubber.

cah292:
The functionals do not want Scrubber-generated capitalization entries to always use the "AS" (Asset) object type. This change updates the Scrubber code to undo the explicit object code overwrite, to set it back to the one used by the base origin entry it was derived from.

The processCapitalization() method would have been the ideal spot for the fix, but doing so would involve pulling the entire contents of the superclass method into our subclass just to modify one line. I decided to override plantFundAccountLookup() instead, which is called by processCapitalization() but is much easier to plug the fix into. The plantFundAccountLookup() method does get called by one other superclass method, so I also added a check to make sure the new Scrubber entry is a capitalization one before updating the field.

Please let me know if you think it would be better to pull in the full processCapitalization() method contents instead, or if there's some better way to add this fix.

### DIFF
--- a/src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
+++ b/src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuScrubberProcessImpl.java
@@ -6,6 +6,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


cah292:
As per the related PTS ticket, we would like to have KualiCo fix this instead, since the issue is in base code. I'm closing this PR as a result.

PR #400, opened 2017-05-11
KFSPTS-8586: Change KFS parameter name from CONCUR_SAE_COLLECTOR_CASH_OFFSET_OBJECT_CODE to CONCUR_SAE_COLLECTOR_PAYMENT_OFFSET_OBJECT_CODE.

nkimble:


nkimble:
Both the parameter and the attribute that uses the parameter were renamed for consistency.

SQL PR#91 and this code PR are dependent on each other and must go in together.

Please check with Sandy before merging this PR to DEV or applying the SQL in PR#91 to DEV so that any validation of SAE files in DEV is not affected.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -320,7 +320,7 @@ protected void addOriginEntriesForCorpCardPersonalExpenseDebitLinesIfNecessary(C
         ConcurStandardAccountingExtractDetailLine firstDebitLine = debitLines.get(0);
         if (cashOffsetAdjustment.isNonZero()) {
             addOriginEntryForCorpCardPersonalExpense(
-                    entryConsumer, firstDebitLine, collectorHelper.getCashOffsetObjectCode(), cashOffsetAdjustment);
+                    entryConsumer, firstDebitLine, collectorHelper.getPaymentOffsetObjectCode(), cashOffsetAdjustment);
```
cah292: I think it would be a good idea to also rename the "cashOffsetAdjustment" variable here, as well as any other variables and methods in this class that have "cashOffset" in the name.

nkimble: OK.  I will rename the items in this class that have the term "cashOffset" in their names to the term "paymentOffset" taking into account proper camel casing where that term is used.

nkimble:
retest this please

nkimble:
retest this please

PR #402, opened 2017-05-15
KFSPTS-8726: Fix validation of SAE personal expense lines.

cah292:
Since we're overriding the object code and omitting the sub-object code on personal expense SAE lines anyway, we need to exclude the sub-object code from the validation of such lines. Otherwise, the validation of those lines may fail (due to the report/header sub-object code possibly not existing under those lines' object codes), causing them to be excluded from the Collector processing.

This fix allows such lines to be properly included on the Collector side, and should prevent them from being logged as errors under the "original" accounting data validation on the PDP side.


jhulslander:
retest this please

jhulslander:
I'm confused, don't we want to log a sub object code errors as non-stopping errors if the sub object code exists in Concur?

cah292:
The problem is that the Collector validation code is sharing some of the code that the PDP validation uses for logging errors with the original accounting data. Without this change, the Collector will fail the line's validation and exclude it, even though we're replacing the sub-object code with a blank value for our processing anyway.

I will add a comment to the user story, to get Sherry's input on what logging to do where the report/header sub-object is incompatible with the line's object code.

jhulslander:
Sounds good, if Sherry is OK with this logging change, I can merge in this code, let me know results

cah292:
Jay,

Based on the latest comments on the user story, Sherry is fine with the code not reporting the sub-object mismatch in such situations. This can be merged once you are satisfied with the code changes.

PR #411, opened 2017-05-22
KFSPTS-8224: Fix role updates from Access Security Model maint docs.

cah292:
The Security Model maintenance documents are currently performing KIM updates in a way that can result in various problems, such as optimistic locking exceptions and unique constraint violations. This fix introduces a new maintainable subclass that addresses these issues.

At some point, it would be good to have KualiCo fix these problems on their end, so that we can remove this customization in the future.


cah292:
retest this please

db483:
Hi Chad,

Has Cathy already talked to the product team? Is this something she wanted fixed before getting a fix from the product team? 

Thank you.

cah292:
I'm closing this PR for now, since the related user story is now in the "Blocked" sprint. If we decide to move forward with fixing this issue ourselves, then we can reopen this PR accordingly.

PR #412, opened 2017-05-23
KFSPTS-8339: Improve resource usage for Jersey client objects.

cah292:
There were some areas in the Concur code where we were not explicitly closing Jersey's Client and ClientReponse objects, and instead implicitly relying on their GC-related or getEntity()-related closing functionality. This fix changes the code to allow for such objects to be closed explicitly, and to keep the Client object alive longer because of its heavyweight setup.

In addition, a new helper superclass has been introduced to help simplify the management of these Jersey objects. This superclass may be useful for future integrations that rely on RESTful web services.


db483:
Hi Chad,

I think the previous version of the code was more simple and easy to follow. Perhaps we could keep that version of the code for now and use the abstracted version once we all become a bit more familiar with Java 8 features.

Thank you.

cah292:
Ok, I'll switch back to the older code for this PR once I've validated it locally. Thanks!

cah292:
I've reverted back to the previous clearer code, and made a couple additional tweaks. This is ready for review again.

PR #417, opened 2017-06-07
KFSPTS-8864 the size of the capital asset information list plus one i…

jhulslander:
…s not nexessarily the next line number


jhulslander:
setupMissingCapitalAssetInformation is a cornell specific function in GLLineService.  It calculated the next available capital asset line number based on the size of the capital asset information list.  The problem is that when a capital asset information entry is removed, the line numbers aren't updated.  Instead of using the size of the list, we are now finding what the maximum line number is, then working from that number.

src/main/java/org/kuali/kfs/module/cam/document/service/impl/GlLineServiceImpl.java
```diff
@@ -799,6 +801,18 @@ public void setupMissingCapitalAssetInformation(String documentNumber) {
         }
 
     }
+    
+    private int findMaximumCapitalAsseetLineNumber(Collection<CapitalAssetInformation> allCapitalAssetInformation) {
```
cah292: Minor typo in method name.

jhulslander:
I fixed the spelling mistake, this is ready for review again

PR #418, opened 2017-06-07
KFSPTS-6196: Add eVerify attribute to Ezra integration.

cah292:
This change allows us to update the Award's eVerify flag based on the value stored in Ezra.

There's also a SQL PR related to this. Both PRs need to go out together.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
@@ -280,7 +280,8 @@ public class EzraServiceImpl implements EzraService {
 	    criteria.put(AWARD_PROPOSAL_ID_FIELD, ezraAward.getAwardProposalId());
 	    
 	    Collection<Compliance> ezraCompliances = businessObjectService.findMatching(Compliance.class, criteria);
-	    if (ezraCompliances.size() == 1) {
+	    // There may be duplicate Compliance entries with the same data, but we only need to retrieve one.
+	    if (!ezraCompliances.isEmpty()) {
 	        return ezraCompliances.iterator().next();
 	    } else {
 	        return null;
--- a/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
@@ -9,6 +9,14 @@ import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
 import org.kuali.kfs.coa.businessobject.Organization;
+import org.kuali.kfs.coreservice.framework.parameter.ParameterService;
+import org.kuali.kfs.kns.document.MaintenanceDocument;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -53,6 +54,9 @@
 
     private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(EzraServiceImpl.class);
 
+    protected static final String PROJECT_ID_FIELD = "projectId";
+    protected static final String AWARD_PROPOSAL_ID_FIELD = "awardProposalId";
+
```
nkimble: Shouldn't these ID fields be defined as property constants?

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -258,12 +262,32 @@ protected Award createAward(Proposal proposal, Award oldAward, EzraProposalAward
...
 		return award;
 	}
 	
+	protected Compliance getEzraCompliance(EzraProposalAward ezraAward) {
+	    Map<String, Object> criteria = new HashMap<>();
+	    criteria.put(PROJECT_ID_FIELD, ezraAward.getProjectId());
+	    criteria.put(AWARD_PROPOSAL_ID_FIELD, ezraAward.getAwardProposalId());
+	    
+	    Collection<Compliance> ezraCompliances = businessObjectService.findMatching(Compliance.class, criteria);
+	    // There may be duplicate Compliance entries with the same data, but we only need to retrieve one.
```
nkimble: Will those duplicate Compliance entries all contain the same boolean data value for the everify flag?

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
cah292: Ok, I'll move these to a separate property-constants class.

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
cah292: Based on Brett's 06/07 comment on the user story, all the duplicates should be getting updated by the same update query, so they should all have the same eVerify flag value.

cah292:
Thanks for pointing out the KFSPTS-8624 note from Janet. I'll follow up with her on this user story, to confirm that she still wants this feature to move forward.

cah292:
I've committed a few more changes based on feedback.

Also, as per Janet's comment on the user story, it sounds like she does want the EZRA integration to overwrite the KFS value in many cases. Once I get a little more clarity from her on the forced-update situations, this will be ready for reviewing and merging again.

cah292:
As per Janet's latest comments on the user story, it sounds like the changes introduced by this PR will meet the expectations of the new desired EZRA job functionality.

nkimble:
Previous code review concerns have been addressed or answered. Unit tests have passed. Merging.

PR #423, opened 2017-06-29
KFSPTS-9067 add RCDV explanation to the created DV docs explanation

db483:


src/main/java/edu/cornell/kfs/fp/service/impl/RecurringDisbursementVoucherDocumentServiceImpl.java
```diff
@@ -272,6 +275,24 @@ private void saveDisbursementVouchers(List<DisbursementVoucherDocument> dvs, Rec
             }
         }
     }
+    
+    private String buildDVExplanation(RecurringDisbursementVoucherDocument recurringDV) {
+        String dvExplanation = CuFPConstants.RecurringDisbursementVoucherDocumentConstants.RECURRING_DV_EXPLANATION_TO_DV_NOTE_STARTER
+                + recurringDV.getDocumentNumber() + KFSConstants.DELIMITER + KFSConstants.BLANK_SPACE + recurringDV.getDocumentHeader().getExplanation();
```
jhulslander: Do you think it would be worth while adding some logic to only include the coma and space only if the recurring DV has an explanation to append?

I don't have a strong opinion on this, but when I start appending a number of String together, I think using a StringBuilder is more efficient, and slightly easier to read.

db483: Good point. I shall make the changes. Thank you.

jhulslander:
I spoke with Daniela, this PR isn't ready to be merged yet.  However, the code changes are what I was thinking about, so I think this can be merged after further local validation.

PR #426, opened 2017-07-17
KFSPTS-9312 Setting up locking job for CU implementation, and support…

jhulslander:
…ing a list of people who can access documents while KFS is locked

I pulled in RemoteModuleServiceBase.java and overrode the isLocked function, and added the two helper methods below it.  This was to support some people getting in while KFS is locked.

I also had to bring in the two new job descriptors into the spring file, so we lock the modules we are using.  If we attempt to lock a namespace that doesn't exist, an NPE is thrown.


src/main/java/edu/cornell/kfs/sys/CUKFSParameterKeyConstants.java
```diff
@@ -20,6 +20,8 @@
     public static final String PDP_CUSTOMER_MISSING_ADVICE_RETURN_EMAIL = "PDP_CUSTOMER_MISSING_ADVICE_RETURN_EMAIL";
 
     public static final String CONFIDENTIAL_ATTACHMENT_FILENAME_PATTERNS = "CONFIDENTIAL_ATTACHMENT_FILENAME_PATTERNS";
+    
+    public static final String OLTP_LOCK_OVERRIDE_LIST = "OLTP_LOCK_OVERRIDE_LIST";
```
db483: It would be good for this constant to reflect the fact that it is a list of net ids/user names. Maybe OLTP_LOCK_OVERRIDE_USERNAME_LIST?

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
```diff
@@ -0,0 +1,653 @@
...
+     *
+     * @see ModuleService#isLocked()
+     */
+    @Override
+    public boolean isLocked() {
+        LOG.debug("isLocked, entering Cornell customization");
+        boolean shouldLock = false;
+        ModuleConfiguration configuration = this.getModuleConfiguration();
+        if (configuration != null) {
+            String namespaceCode = configuration.getNamespaceCode();
```
db483: Since this method has grown in length significantly it might make sense to extract the code that checks if the module is locked into a separate method: isModuleLocked. Then this method could return something like: 
isModuleLocked && isUserLocked.

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
```diff
@@ -0,0 +1,653 @@
...
+        }
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("isLocked, shouldLock: " + shouldLock);
+        }
+        return shouldLock;
+    }
+    
+    protected boolean isUserInLockoutOverrideList() {
+        String[] overrideList = StringUtils.split(getOverrideUsersList(), ",");
+        String loggedInUserPrincipleName = GlobalVariables.getUserSession().getPerson().getPrincipalName();
```
db483: variable should be renamed to loggedInUserPrincipalName

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
```diff
@@ -0,0 +1,653 @@
...
+        Object[] filteredList = Arrays.stream(overrideList).filter(x -> x.equalsIgnoreCase(loggedInUserPrincipleName)).toArray();
```
db483: I think it might make more sense to use a parameter evaluator to check if the username is in the list. The list might need to be separated by ";" in that case..

db483:
I am curious what was the feedback you got from IU. Are they also allowing specific users during lockout?

It would be nice if we could somehow have a local customization instead of bringing in RemoteModuleServiceBase.java but I see there isn't an easy way to do that. I am also wondering if this might be something that KualiCo could add to base functionality.

Regarding the NPE, is this something that maybe could be fixed in the base code? Is IU using all modules? Did they also override what modules they lock?

Thank you.

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
jhulslander: sure, let me give that a try

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
jhulslander: good idea

src/main/java/org/kuali/kfs/krad/service/impl/RemoteModuleServiceBase.java
jhulslander: ok

jhulslander:
I haven't heard back from IU yet, but I will be sure and ask about the NPE on namespaces that aren't used.

the override users from lockout idea sounds like something KualiCo might want to add base functionality.  Let's ask her at the standup.

jhulslander:
I have made the code review changes.  I am still waiting for feedback from IU on the questions involved

db483:
The changes look good to me but will wait to merge until unit tests will pass and hopefully we also get feedback from IU so we don't have to ask out functionals to validate twice.

jhulslander:
retest this please

jhulslander:
The unit tests are passing now.

jhulslander:
Daniela, this is ready for review again.  I have eliminated the extra code that used the parameter we were originally going to try.  The base code permission  'Access Locked Module' worked as expect in local testing.  We also don't need the SQL that I originally included in this user story.  Nicole said she can create the new role for us and link it to the permission, no need for SQL to do it.

PR #427, opened 2017-07-18
KFSPTS-9360 fix unit tests

db483:


src/test/java/edu/cornell/kfs/sys/service/impl/CuUserProcurementProfileValidationServiceImplTest.java
```diff
@@ -16,7 +16,7 @@
 import edu.cornell.kfs.sys.businessobject.UserProcurementProfile;
 import edu.cornell.kfs.sys.service.UserProcurementProfileValidationService;
 
-@ConfigureContext(session = db18)
+@ConfigureContext(session = kad13)
```
cah292: In the user story, you mentioned that you had to change the user in the session because the old one no longer had the expected access. Should this test add some other KIM setup to make sure the user does indeed have the expected permission or role? Otherwise, we could potentially end up with this situation again in the near future.

db483: Good point. Do you have some suggestion on how that can be done? Thank you!

cah292: After a quick search of the base KFS unit test code, there could be a few things that could help with this. One is a UnitTestSqlDao helper object that can be used to run SQL on demand; I think SpringContext.getBean() can retrieve this if the test needs KFS running. Another thing that could help is the TestDataPreparator class, which can convert property file data into BOs or search criteria, and which can then be processed as needed.

Alternatively, you could modify or override the UserProcurementProfileValidationServiceImpl.canMaintainUserProcurementProfile method to work around the issue. However, depending on the workaround, that may defeat the purpose of the test.

cah292:
I added my recommendations on what tools could be used for a longer-term fix on the one test. If you think such a change should be deferred to a future PR, just let me know and I'll merge the current fix as-is.

src/test/java/org/kuali/kfs/sys/fixture/UserNameFixture.java
```diff
@@ -46,6 +46,7 @@
 ls35,
 // favorite account manager
 db18,
```
jhulslander: Since Dennis has retired, should we remove this user from our unit tests?

db483: I would probably make sense to remove it. It is however used by another unit test so that would need to be cleaned up as well.

se12: Please replace db18 with mgw3 (Mike Winters). Thanks!

db483:
Chad,

Thank you for your recommendations. I think we want our unit tests to be micro tests and remove the Spring dependency. I spent some time this morning trying to make this a microtest but because of the way the service is implemented that proved rather difficult. The service itself would need to be refactored. I think it would make more sense to get the code in as is and have this cleanup effort as a future jira.

Thank you.

db483:
retest this please

src/test/java/org/kuali/kfs/sys/fixture/UserNameFixture.java
db483: Thank you for the suggestion Sandy, I have replaced db18 with mgw3.

PR #428, opened 2017-07-19
KFSPTS-7312: Add page to manage Concur access tokens.

cah292:
This adds a new KFS page for managing the access and refresh tokens that we use for contacting Concur web services. Both this PR and the associated nonprod-sql PR need to go out together.

The related JAXB DTO has been updated to use the correct case-sensitive XML elements, and a new element has been added as well. The new element exists on request-token feedback but not on refresh-token feedback, but the DTO still works in both cases.

As noted in other team meetings, I was unable to use the Jersey REST client API to successfully perform the revoke-token calls, so I ended up using the Apache HttpClient API instead.

Also, there was some existing Jersey REST client usage that wasn't performing clean-up like in one of the other recent PR enhancements, but I left it as-is since I thought it was out of scope. If you think such closing should be included in the scope of these changes, please let me know.


src/main/java/edu/cornell/kfs/concur/ConcurKeyConstants.java
```diff
@@ -25,4 +25,11 @@
     public static final String CONCUR_SAE_NO_REPORT_EMAIL_SUBJECT = "message.concur.sae.no.report.email.subject";
     public static final String CONCUR_SAE_NO_REPORT_EMAIL_BODY = "message.concur.sae.no.report.email.body";
     public static final String CONCUR_INCOMPLETE_ADDRESS = "validation.error.concur.incomplete.address";
+    public static final String CONCUR_TOKEN_REQUEST_QUESTION = "message.concur.token.request.question";
```
db483: I realize now this file has not followed the general KFS convention of naming these constants starting with ERROR for errors MESSAGE for messages but perhaps we could start now?

src/main/java/edu/cornell/kfs/concur/ConcurUtils.java
```diff
@@ -65,4 +68,21 @@ public static boolean isConcurReportStatusAwaitingExternalValidation(String stat
         return ConcurConstants.EXPENSE_AWAITING_EXTERNAL_VALIDATION_STATUS_CODE.equalsIgnoreCase(statusCode) || ConcurConstants.REQUEST_AWAITING_EXTERNAL_VALIDATION_STATUS_CODE.equalsIgnoreCase(statusCode);
     }
 
+    /**
```
db483: I think we should eliminate this comment and rename the method to something more general like generateBase64EncodedString

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -21,20 +31,57 @@
...
+    private String concurRequestAccessTokenURL;
     private String concurRefreshAccessTokenURL;
+    private String concurRevokeAccessTokenURL;
     protected WebServiceCredentialService webServiceCredentialService;
 
     @Override
+    public void requestNewAccessToken() {
+        AccessTokenDTO newToken = buildRequestAccessTokenOutput();
+        
+        webServiceCredentialService.updateWebServiceCredentialValue(ConcurConstants.CONCUR_ACCESS_TOKEN, newToken.getToken());
```
db483: Should we ensure these operations are transactional? I think we either want all of them to succeed or none.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -68,6 +115,94 @@ protected ClientRequest buildRefreshAccessTokenClientRequest() {
     }
 
     @Override
+    public void revokeAccessToken() {
```
db483: I think we may want to change the revoke Access token to perform:
- revoke current access token &
- request new access token and save it in the database to be used

As far as I understand it, we can request multiple access tokens and the previous one will not be automatically revoked. Because of this I am wondering if it makes sense to even have the request new access token functionality available. Perhaps we could replace the separate request new and revoke with a a single revoke and request new. This can check if a token exists in the DB then revoke and request new. If one does not exists just request new.

I think we want to use the request new access token alone only when generate the token for the first time.
We would only want to use the revoke token when we think the current token may have been compromised or if we want to renew it regularly for security purposes since the refresh functionality does not update the token but just extend it's lifetime. 

The only scenario for using the revoke token alone is if at one point  we are no longer using Concur.

db483:
Regarding the existing Jersey REST client usage that wasn't performing clean-up like in one of the other recent PR enhancements, I agree that should be done on a separate jira. Do you know if one such jira already exists? If not, would you mind creating it in the Concur queue? Thank you!

src/main/java/edu/cornell/kfs/concur/ConcurKeyConstants.java
cah292: Ok, I'll update the names of the constants accordingly.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -21,20 +31,57 @@
...
-    private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurAccessTokenServiceImpl.class);
+    private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ConcurAccessTokenServiceImpl.class);
 
+    private String concurRequestAccessTokenURL;
     private String concurRefreshAccessTokenURL;
+    private String concurRevokeAccessTokenURL;
     protected WebServiceCredentialService webServiceCredentialService;
 
     @Override
+    public void requestNewAccessToken() {
```
cah292: Good point, I'll update the name.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
cah292: Sounds like a good idea. I'll see if I can make these methods or this service transactional.

src/main/java/edu/cornell/kfs/concur/ConcurUtils.java
cah292: Makes sense, I'll rename the method accordingly.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
cah292: Thanks for the suggestions. I agree that we don't need the revoke-only functionality, and that having a revoke-and-request-new action would be better.

Because of our current environment setups, though, we may need some way to request a new token without revoking the current one. Otherwise, we could run into issues because DEV and TEST are currently using the same token. Perhaps an appropriate question prompt can be used to give the user more control about whether to revoke the token before requesting a new one, or perhaps we need to redo our process for bootstrapping tokens in DEV and TEST.

cah292:
There's already a KFSPTS-9037 ticket for fixing the resource lifetime problems with the service that this PR is updating, so we're already tracking it.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
cah292: I've left this name as-is for now, due to the way that I refactored things in my latest revisions. The revoke-and-replace naming has been moved to a different method as a result. Let me know if you think this still needs renaming.

cah292:
We appear to be having some unit test database issues at the moment. The KDO-1680 ticket will need to be dealt with before we run the tests again for this PR.

db483:
Thank you Chad. I think revoking the tokens in DEV and TEST for validation is fine. We will want to update the KFSPTS-8501.sql script with the new values.

db483:
retest this please

db483:
I have restored the utest database and now there is only 1 test failing: AccountGlobalSearchLookupableHelperServiceImplTest

cah292:
retest this please

cah292:
The associated SQL has been updated based on functional feedback. This PR and the SQL PR are now ready for review again.

src/main/java/edu/cornell/kfs/concur/web/struts/ConcurManageAccessTokenAction.java
```diff
@@ -0,0 +1,80 @@
...
+    public ActionForward start(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+        return mapping.findForward(KFSConstants.MAPPING_BASIC);
+    }
+
+    public ActionForward replaceToken(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+        Object question = request.getParameter(KFSConstants.QUESTION_INST_ATTRIBUTE_NAME);
+        boolean revokeOldToken;
+        
+        if (ObjectUtils.isNull(question)) {
+            if (getConcurAccessTokenService().currentAccessTokenExists()) {
```
db483: Do we want to give the user the option to request a new access token without revoking the old one? How is the old one going to be revoked in that case? We are leaving an access token available that we are no longer controlling.

db483:
retest this please

src/main/java/edu/cornell/kfs/concur/web/struts/ConcurManageAccessTokenAction.java
cah292: The problem is that our current non-PROD databases are being auto-updated with a specific access token (as per our SQL scripts). Thus, if someone validates this in DEV or TEST and chooses to revoke that token, then the Concur jobs in DEV and TEST will stop working upon each refresh.

Would it make sense to have some kind of temporary SQL to alter or remove the DEV/TEST token stored in the database, so that the token from our existing scripts won't get revoked by mistake when validating the change? That would eliminate the need for this dialog.

db483: Hi Chad, that sounds like a reasonable solution to me. Thank you!

cah292:
I've made some changes accordingly to remove the question dialog, and have also rebased the branch due to its age. This is now ready for review again.

PR #430, opened 2017-07-21
KFSPTS-9360 fix another unit test

db483:


db483:
There was a change in PROD and hac95 is no longer a manager for accounts with org code 0121. I updated the user in the failing test with one that is a manager for accounts in or 0121.

PR #432, opened 2017-08-01
KFSPTS-9265 upgrade to 2017-06-01 version of financials

db483:


db483:
Hi Chad,

Please review the code changes related to the upgrade to 2017-06-01version of financials but DO NOT MERGE the changes yet. We need the previous upgrade moved to test before merging this to develop. Thank you!

src/main/java/edu/cornell/kfs/sys/batch/service/impl/CuAutoDisapproveDocumentsServiceImpl.java
```diff
@@ -47,14 +49,14 @@
     private static final String TAB = "\t";
 
     private RouteHeaderService routeHeaderService;
+    private PersonService personService;
+    
 
-    protected boolean processAutoDisapproveDocuments(String principalId, String annotation, Date documentCompareDate) {
+    protected boolean processAutoDisapproveDocuments(String principalId, String annotation) {
```
cah292: Since this is overriding a superclass method, I would suggest adding the "Override" annotation for clarity.

src/main/java/edu/cornell/kfs/sys/batch/service/impl/CuAutoDisapproveDocumentsServiceImpl.java
```diff
@@ -420,23 +319,25 @@ protected void autoDisapprovalYearEndDocument(Document document, String annotati
...
-			documentEligible = childDocumentType.getParentId().equals(parentDocumentType.getId());
-
-			if (documentEligible) {
-				break;
-			}
-		}
-
-		return documentEligible;
-	}
+    protected boolean checkIfDocumentEligibleForAutoDispproval(DocumentHeader documentHeader) {
```
cah292: Another method that could use the "Override" annotation.

src/main/java/edu/cornell/kfs/sys/batch/service/impl/CuAutoDisapproveDocumentsServiceImpl.java
```diff
@@ -486,9 +387,37 @@ protected boolean checkIfDocumentEligibleForAutoDispproval(DocumentHeader docume
...
+        }
+
+        return document;
+    }
 
     public void setRouteHeaderService(RouteHeaderService routeHeaderService) {
         this.routeHeaderService = routeHeaderService;
+    } 
+
+    protected PersonService getPersonService() {
```
cah292: The superclass already has a non-final setter for its private personService variable. If that setter is being used properly by base KFS, perhaps it would be better to remove this getter, and to instead override the setter to call the superclass method and also set this subclass's personService variable. (Or, if this getter is still needed, then perhaps it would be better to perform the setter override above and to remove the lazy-load section from this getter.)

db483:
Hi Chad,

I made the changes you suggested and this is ready for review again. Please do NOT merge yet,

Thank you!

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -203,6 +203,7 @@
 
     <bean id="autoDisapproveService" class="edu.cornell.kfs.sys.batch.service.impl.CuAutoDisapproveDocumentsServiceImpl" parent="autoDisapproveService-parentBean">
         <property name="routeHeaderService" ref="enDocumentRouteHeaderService" />
+        <property name="personService" ref="personService" />
```
cah292: I believe the parent bean is already injecting this service, so it shouldn't be necessary to inject it again here.

db483: Good catch! I will update that.

PR #434, opened 2017-08-03
KFSPTS-9308: Fix handling of returned CBCP personal charges

cah292:
This change fixes an issue where returns/credits of CBCP personal charges were not being handled properly in the SAE-to-PDP/Collector job. Some new DTO fields and service methods have been added to help identify such transactions so that they can be deducted from the personal charge amounts accordingly.

I also had to update part of the SAE-to-PDP code so that it would include the debit side of the returns in its processing, effectively allowing it to fully or partially cancel out the related personal charge as needed. I added a related unit test as well, to make sure the edited code was handling the include-in-PDP check properly. Please let me know if you find problems with the proposed solution or its associated unit test.

This new code is currently using hard-coded values for the payment types that we're expecting Concur to use. Please let me know if you think they should be parameterized instead.


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -92,14 +92,24 @@ public void addDetailLine(ConcurStandardAccountingExtractDetailLine detailLine)
...
+        Map<String, List<ConcurStandardAccountingExtractDetailLine>> consolidatedLinesMap = getConsolidatedPersonalExpensesMapForLine(detailLine);
+        consolidatedLinesMap.computeIfAbsent(accountingFieldsKey, (key) -> new ArrayList<>())
+                .add(detailLine);
+    }
+
+    protected Map<String, List<ConcurStandardAccountingExtractDetailLine>> getConsolidatedPersonalExpensesMapForLine(
+            ConcurStandardAccountingExtractDetailLine detailLine) {
         if (detailLine.getJournalAmount().isPositive()) {
-            consolidatedLinesMap = consolidatedCorpCardPersonalExpenseDebits;
+            if (collectorHelper.lineRepresentsReturnOfCorporateCardPersonalExpenseToUser(detailLine)) {
```
jhulslander: I am thinking reorganizing this a little might make it easier to read.  The if/else levels make it confusing to read and it is starting to get long.  Maybe something like
if (detailLine.getJournalAmount().isPositive()) {
  getConsolidatedPersonalExpensesMapForPositiveLine(line);
} else {
  getConsolidatedPersonalExpensesMapForNegativeLine(line)
}

and move the contents if the if else into the new functions?

cah292: Yes, that makes sense. I will rework this section of code accordingly to make it more readable.

cah292:
I've updated the code noted in the PR feedback to improve its readability. This is now ready for review again.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -100,12 +100,24 @@ protected void addCorpCardPersonalExpenseDetailLine(ConcurStandardAccountingExtr
...
-                return consolidatedCorpCardPersonalExpenseDebits;
-            }
-        } else if (collectorHelper.lineRepresentsReturnOfCorporateCardPersonalExpenseToUniversity(detailLine)) {
+            return getConsolidatedPersonalExpensesMapForDebitLine(detailLine);
+        } else {
+            return getConsolidatedPersonalExpensesMapForCreditLine(detailLine);
+        }
+    }
+
+    protected Map<String, List<ConcurStandardAccountingExtractDetailLine>> getConsolidatedPersonalExpensesMapForDebitLine(
```
jhulslander: I think this is better, but I just want to ask if we want to have the function named ..ForDebitLine when it could return credits?

cah292: The line passed to this method is expected to be a debit one, which was why I named the method as such (and similarly for the credit method). These methods are configured to return a mismatched map if the line is a return/credit of a personal expense, since that allows for properly deducting the returned amount from the current personal charges when the similar transactions get wrapped up.

PR #435, opened 2017-08-03
KFSPTS-9424: Fix missing formatter setup on Payee ACH Account ID.

cah292:
Our Payee ACH Account customizations accidentally excluded the setup of a special formatter on the ID field, and this absence was causing stack traces on PAAT doc searches as a result. This fixes the issue by adding the formatter config back in.


cah292:
retest this please

PR #439, opened 2017-08-09
Revert 428 KFSPTS-7312

jhulslander:


jhulslander:
This reverts the original commit from develop for 7312.  I had to resolve some conflicts (they were just institutional-struts-config) I believe this reverts the comit as expected, but please double check before merging this PR

PR #440, opened 2017-08-10
KFSPTS-7312 Concur Access Token Management screen

jhulslander:


src/main/java/edu/cornell/kfs/concur/service/impl/ConcurAccessTokenServiceImpl.java
```diff
@@ -61,12 +122,121 @@ protected ClientRequest buildRefreshAccessTokenClientRequest() {
...
+        boolean success = executeRevokeAccessTokenRequest();
+        if (success) {
+            setWebserivceCredentialValues(StringUtils.EMPTY, StringUtils.EMPTY, StringUtils.EMPTY);
+        } else {
+            throw new RuntimeException("Error revoking access token: Unsuccessful response from Concur");
+        }
+    }
+    
+    @Override
+    public void deleteTokensFromDatabase() {
```
db483: I think this method's name is misleading because it's not really deleting the entries from the database but rather reseting all values to empty

src/main/webapp/jsp/concur/manageAccessToken.jsp
```diff
@@ -0,0 +1,47 @@
...
+					<div style="font-weight: bold">Revoke the current token</div>
+					<div>
+						<html:submit property="methodToCall.revokeToken" styleClass="btn btn-default" value="Revoke Token" />
+					</div>
+				</div>
+			</c:when>
+		</c:choose>
+		
+		<div class="center" style="margin: 30px 0;">
+			<div style="font-weight: bold">Remove existing tokens from the KFS database.</div>
```
db483: I think this should say reset or set empty values since it is not really removing the entries from the database.

jhulslander:
I believe I have made the changes we talked about.

src/main/webapp/jsp/concur/manageAccessToken.jsp
```diff
@@ -0,0 +1,51 @@
...
+					</div>
+				</div>
+			</c:when>
+		</c:choose>
+
+		<c:choose>
+			<c:when test="${KualiForm.showResetTokenToEmptyStringButton}">
+				<div class="center" style="margin: 30px 0;">
+					<div style="font-weight: bold">Reset existing token to empty string in the KFS database.</div>
+					<div>(Do this before revoking tokens in devloocal, dev or test otherwise you'll need to update non-prod SQL refresh scripts.)</div>
```
db483: This has a typo. Also, it may make more sense to say do this in non-prod databases to test revoke access token

jhulslander:
I fixed the typo per your suggestion

PR #441, opened 2017-08-16
KFSPTS-5548: Update CuBatchExtractServiceImpl to inject services.

cah292:
This updates CuBatchExtractServiceImpl to rely only on injected services, so that SpringContext.getBean() does not need to be used there. Also, the superclass already has a protected variable for the ReconciliationService, so I did not need to add one to this class.

In addition, the associated unit test has been rewritten to be a micro-test, and to perform some other related cleanup. The associated fixtures were updated as well.

### DIFF
--- a/src/test/java/edu/cornell/kfs/module/cam/batch/service/impl/CuBatchExtractServiceImplTest.java
+++ b/src/test/java/edu/cornell/kfs/module/cam/batch/service/impl/CuBatchExtractServiceImplTest.java
@@ -1,81 +1,191 @@
 package edu.cornell.kfs.module.cam.batch.service.impl;
 
-import edu.cornell.kfs.module.cam.fixture.EntryFixture;
-import org.kuali.kfs.gl.businessobject.Entry;
-import org.kuali.kfs.krad.service.BusinessObjectService;
-import org.kuali.kfs.module.cam.batch.service.BatchExtractService;
-import org.kuali.kfs.module.purap.document.PaymentRequestDocument;
-import org.kuali.kfs.module.purap.document.VendorCreditMemoDocument;
-import org.kuali.kfs.sys.ConfigureContext;
-import org.kuali.kfs.sys.context.KualiTestBase;
-import org.kuali.kfs.sys.context.SpringContext;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
-@ConfigureContext
-public class CuBatchExtractServiceImplTest extends KualiTestBase {
-	
-	private BatchExtractService batchExtractService;
-	private CuBatchExtractServiceImpl cuBatchExtractServiceImpl;
-	private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(CuBatchExtractServiceImpl.class);
-	private BusinessObjectService businessObjectService;
-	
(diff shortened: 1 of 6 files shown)


src/test/java/edu/cornell/kfs/module/cam/fixture/EntryFixture.java
```diff
@@ -2,10 +2,13 @@
 
 import org.kuali.kfs.gl.businessobject.Entry;
 
+import edu.cornell.kfs.module.cam.CuCamsTestConstants;
+
 public enum EntryFixture {
 	
-	VCM_ONE("5319793", "CM"), VCM_TWO("5686500", "CM"), VCM_THREE("0","CM"),
-	PREQ_ONE("5773686","PREQ"), PREQ_TWO("5773687", "PREQ"), PREQ_THREE("0", "PREQ");
+	VCM_ONE(CuCamsTestConstants.DOC_5319793, "CM"), VCM_TWO(CuCamsTestConstants.DOC_5686500, "CM"), VCM_THREE("0", "CM"),
```
cah292: Ok, I'll come up with better names for these enum constants, though I may have to alter other unit tests as well if they rely on these.

src/test/java/edu/cornell/kfs/module/cam/batch/service/impl/CuBatchExtractServiceImplTest.java
```diff
@@ -1,81 +1,191 @@
...
+        cuBatchExtractServiceImpl.setDataDictionaryService(dataDictionaryService);
+    }
+
+    @Test
+    public void testFindCreditMemoDocument() {
+        VendorCreditMemoDocument vendorCreditMemoDocument = cuBatchExtractServiceImpl.findCreditMemoDocument(EntryFixture.VCM_TWO.createEntry());
+        assertNotNull("vendor credit memo should have been non-null", vendorCreditMemoDocument);
+        assertEquals("Wrong credit memo document was retrieved", CuCamsTestConstants.DOC_5686500, vendorCreditMemoDocument.getDocumentNumber());
+        
+        vendorCreditMemoDocument = cuBatchExtractServiceImpl.findCreditMemoDocument(EntryFixture.VCM_ONE.createEntry());
```
cah292: I think it's just retrieving a different CM document than the one above it. Given that this doesn't add much value, though, I'll refactor or remove this second retrieval accordingly.

src/test/java/edu/cornell/kfs/module/purap/fixture/VendorCreditMemoDocumentFixture.java
```diff
@@ -1,43 +1,62 @@
...
-import org.kuali.kfs.krad.service.DocumentService;
+
+import edu.cornell.kfs.module.cam.CuCamsTestConstants;
+import edu.cornell.kfs.module.purap.document.CuVendorCreditMemoDocument;
 
 public enum VendorCreditMemoDocumentFixture {
 
-	VENDOR_CREDIT_MEMO("Description", 0, 4291, "12345", "12-5-2014",
-			new KualiDecimal(100));
+	VENDOR_CREDIT_MEMO("Description", 0, 4291, "12345", "12-05-2014",
```
cah292: This fixture constant is used by CuPurchasingAccountsPayableModuleServiceImplTest, which is not in the scope for this user story. I only edited the date string for consistency with the new enum constants below it, and I do not believe the associated tests are actively using the date value on this particular constant. If you think this constant should be left as-is, let me know and I can undo the change.

db483: Got it, thank you for explaining it, no need to undo the change.

cah292:
I've made some more changes based on feedback. This is now ready for review again.

PR #442, opened 2017-08-17
KFSPTS-5547 inject services for CuAssetServiceImpl and update unit tests

db483:


src/main/resources/edu/cornell/kfs/module/cam/cu-spring-cam.xml
```diff
@@ -45,7 +45,10 @@
  	     
 	<bean id="assetDepreciationService"  parent="assetDepreciationService-parentBean" class="edu.cornell.kfs.module.cam.batch.service.impl.CuAssetDepreciationServiceImpl" />
  	<bean id="assetService"  parent="assetService-parentBean"
- 	     class="edu.cornell.kfs.module.cam.document.service.impl.CuAssetServiceImpl" />
+ 	     class="edu.cornell.kfs.module.cam.document.service.impl.CuAssetServiceImpl" >
+ 	     <property name="businessObjectService" ref="businessObjectService" />
+ 	     <property name="parameterService" ref="parameterService" />
```
cah292: The parent bean is already injecting parameterService and businessObjectService, so I don't think it's necessary to override the injection here.

src/test/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetServiceImplTest.java
```diff
@@ -1,42 +1,124 @@
 package edu.cornell.kfs.module.cam.document.service.impl;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
+import static org.junit.Assert.assertFalse;
```
cah292: If this static import isn't being used by the test, then I would suggest removing it.

db483:
Hi Chad,

I have added the requested changes.

Regarding the initialization of the services in the unit test, to me it seems more clear setting them up with the different values in each test as it makes it more clear what the conditions are for running the test and why it should succeed or fail based on the returned values.

PR #445, opened 2017-08-22
KFSPTS-9653: Fix auto-generated note when changing send-to-vendor flag

cah292:
This is an updated version of the KFSPTS-9313 fix. I've significantly narrowed its effects so that it only impacts the auto-generated note from cases where the "Send to Vendor" flag gets changed.

I've reworked this fix to incorporate certain note-updating behavior similar to that from the KFSPTS-4490 enhancement and from the base KualiDocumentActionBase.insertBONote() code.

Please let me know if you have questions or concerns about this version of the fix, or if you think someone else should review this as well.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
+++ b/src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
@@ -18,15 +18,16 @@
  */
 package org.kuali.kfs.module.purap.document.web.struts;
 
-import edu.cornell.kfs.module.purap.CUPurapConstants;
-import edu.cornell.kfs.module.purap.CUPurapKeyConstants;
-import edu.cornell.kfs.module.purap.document.web.struts.CuPurchaseOrderForm;
-import edu.cornell.kfs.module.purap.util.PurchasingFavoriteAccountLineBuilderBase;
-import edu.cornell.kfs.module.purap.util.PurchasingFavoriteAccountLineBuilderForDistribution;
-import edu.cornell.kfs.module.purap.util.PurchasingFavoriteAccountLineBuilderForLineItem;
-import edu.cornell.kfs.sys.businessobject.FavoriteAccount;
-import edu.cornell.kfs.sys.service.UserFavoriteAccountService;
-import edu.cornell.kfs.vnd.businessobject.CuVendorAddressExtension;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.struts.action.ActionForm;
@@ -37,6 +38,7 @@ import org.kuali.kfs.coreservice.framework.parameter.ParameterService;
 import org.kuali.kfs.integration.purap.CapitalAssetLocation;
 import org.kuali.kfs.integration.purap.CapitalAssetSystem;
 import org.kuali.kfs.integration.purap.ItemCapitalAsset;
+import org.kuali.kfs.kns.document.MaintenanceDocument;
 import org.kuali.kfs.kns.question.ConfirmationQuestion;
 import org.kuali.kfs.kns.service.DataDictionaryService;
 import org.kuali.kfs.kns.util.KNSGlobalVariables;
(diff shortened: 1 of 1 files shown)


src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
```diff
@@ -1647,9 +1655,11 @@ private void createReasonNote(ActionForm form) {
     	KualiDocumentFormBase kualiDocumentFormBase = (KualiDocumentFormBase) form;
         try {
             Note noteObj = SpringContext.getBean(DocumentService.class).createNoteFromDocument(kualiDocumentFormBase.getDocument(), ((PurchasingFormBase)kualiDocumentFormBase).getReasonToChange());
+            populateIdentifierOnNoteAndExtension(noteObj);
             kualiDocumentFormBase.getDocument().addNote(noteObj);
-         //   SpringContext.getBean(NoteService.class).save(noteObj);
-            SpringContext.getBean(NoteService.class).saveNoteList(kualiDocumentFormBase.getDocument().getNotes());
+            if (isNoteReadyToBeSaved(kualiDocumentFormBase.getDocument(), noteObj)) {
```
jhulslander: I am not sure if we need the isNoteReadyToBeSaved logic here.  Since we are in the createReasonNote function, and it looks like populateIdentifierOnNoteAndExtension takes care of the condition that caused the root problem of this Jira.  When would we not want to save the note here?

se12: I think since this was reverted from the most recent patch that another developer should also review, not because Jay was the original developer for kfspts-9313 but because this wasn't found until in production due to the 'edge case' that caused this to still not work correctly.  I think it is best practice to have at least two developers review this correction.

cah292: There are other types of documents besides REQS that could be using this functionality, and as per the "getPersistedBoNotesNotes" method, the notes being checked for requiring reason-note generation are not necessarily from the current document (if it's not a REQS). If there are other documents besides REQS that can alter the "Send to Vendor" flag while still in INITIATED status, then I think we need some form of the "isNoteReadyToBeSaved" logic. Otherwise, we can remove this extra check.

As per Sandy's suggestion, I will also add a second reviewer to this PR to get feedback on these code changes and this question.

db483: Chad,

Is the isNoteReadyToBeSaved trying to catch something that is missing in isCreatingReasonNote?

What other documents are using "Send to Vendor" flag?

cah292: The reason-note code is available to all PURAP documents, but according to our note tag customizations, only REQS and POs are actively using it.

The isNoteReadyToBeSaved method is trying to incorporate some of the should-save-note checks from the base insertBONote method for document actions. Of those checks, though, I think only the is-not-INITIATED-status check is the only one that would be applicable here (since this action class is for transactional docs only, and such docs typically have a note target already).

If there's cases where certain other PO-related docs (like POA) can change the "Send to Vendor" flag on the base PO doc's notes while in INITIATED status, then we may need to keep at least the is-not-INITIATED-status check. Otherwise, it should be safe to remove the isNoteReadyToBeSaved method altogether.

db483: Thank you Chad.

It seems that the isNoteReadyToBeSaved is checking the document to see if it is in a state where a note for it can be saved and not necessarily if the note itself is in a state where it can be saved. Maybe the method name could be improved.

I also noticed that the new code is only saving the reason not and the previous code saved all the notes:
 SpringContext.getBean(NoteService.class).saveNoteList(kualiDocumentFormBase.getDocument().getNotes());

Was the saving of all notes unnecessary?

This particular customization for the send to vendor flag could be improved for clarity and for easier maintenance in the future. However that would be a significant effort and is out of the scope of this particular fix.

I think that if you can find a better name for the isNoteReadyToBeSaved method and ensure saving all notes don't need to be saved then we can promote this fix to DEV for validation.

cah292: Ok, I'll come up with a better name for that method.

As for the all-note-save operation in the old version of the fix, that may have been there just to try and fix this problem retroactively for older docs, which we don't need as per Cathy's responses from a recent standup. Our other customizations to PURAP code should be able to handle updating the note extended attributes on all other notes as needed.

cah292:
I've renamed the new can-save-note method for clarity, as requested. This is now ready for review again.

Not sure if the unit tests will succeed because of the current Jenkins restart that's going on. We may have to re-trigger the tests accordingly once Jenkins is back up.

db483:
Changes look good to me. 

Jay, do you have any remaining questions or concerns? If no, I will merge after unit tests pass. Thank you.

db483:
retest this please

cah292:
Jay, please review this again when you get a chance. If you're satisfied with the changes and the latest unit test run, then this should be ready to be merged since Daniela has already approved the changes.

PR #447, opened 2017-08-24
KFSPTS-9617 add changes to accept multiple values for the CONCUR_CUSTOMER_PROFILE_GROUP_ID system parameter

db483:


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurEmployeeInfoValidationServiceImpl.java
```diff
@@ -95,6 +100,15 @@ public String getAddressValidationMessageIfCheckPayment(String employeeId) {
             }
         }
         return validationMessage;
+    }   
+
+    @Override
+    public boolean isEmployeeGroupIdValid(String employeeGroupId) {
+        if(StringUtils.isNotBlank(employeeGroupId)){
+            return parameterEvaluatorService.getParameterEvaluator(CUKFSConstants.ParameterNamespaces.CONCUR,
```
jhulslander: Is this case sensitive?  Cathy wanted the validation to not be case sensitive.

db483: It is not case insensitive. Unfortunately the parameter evaluators are case sensitive and I am not sure I want to change this to use customized code.

PR #448, opened 2017-08-25
Revert "KFSPTS-9265 upgrade to 2017-06-01 version of financials (#432)"

db483:
This reverts commit fff594091a95a8c723e2f16a08d4eb4a4aff2a70.


db483:
I need to revert these changes from develop to address some issues that have been found during validation.

PR #449, opened 2017-08-28
KFSPTS-9265 upgrade to 2017-06-01 version of financials

cah292:
This is a corrected version of the code originally from #432, which now updates and preserves our customizations related to generating PDP notes from DV documents.

Some of the PDP-note-related code has been relocated to a new PaymentSourceHelperService override, due to recent base KFS changes that have made it practical to move our check-stub-related customization back into that service.


src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
```diff
@@ -137,6 +137,18 @@ public boolean isDisbursementActionAllowed() {
     }
 
     /**
+     * Wrapper for addNote, loops through the list of notes passed in and calls that.
+     *
+     * @param pnts list of notes to add
+     */
+    public void addNotes(List<PaymentNoteText> pnts) {
+        pnts.stream().forEach(n -> {
```
jhulslander: perhaps a better variable name, paymentNoteTextes?

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
```diff
@@ -137,6 +137,18 @@ public boolean isDisbursementActionAllowed() {
...
+            LOG.debug("addNotes() Creating check stub text note: " + n.getCustomerNoteText());
```
jhulslander: Since this is doing a string concatenation, should this be in an isDebugEnabled if block?

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
```diff
@@ -204,19 +216,13 @@ public void addNote(PaymentNoteText pnt) {
...
             notes.add(pnt);
         } else {
-            LOG.warn("Did not add note to payment detail build from Document #: "+(!StringUtils.isBlank(custPaymentDocNbr) ? custPaymentDocNbr : "")+" because note was empty");
-        }
-    }
-
-    public void addNotes(List<PaymentNoteText> pnts) {
-        for(PaymentNoteText pnt : pnts) {
-            addNote(pnt);
+            LOG.warn("Did not add note to payment detail build from Document #: " + (!StringUtils.isBlank(custPaymentDocNbr) ? custPaymentDocNbr : "") + " because note was empty");
```
jhulslander: we don't often use log warnings, should this be an error?

src/main/java/edu/cornell/kfs/sys/batch/service/impl/CuAutoDisapproveDocumentsServiceImpl.java
```diff
@@ -486,9 +384,36 @@ protected boolean checkIfDocumentEligibleForAutoDispproval(DocumentHeader docume
         
         return finalList;
     }
+    
+    /**
+     * This method finds the document for the given document header id
+     *
+     * @param documentHeaderId
+     * @return document The document in the workflow that matches the document header id.
+     */
```
jhulslander: I think we can eliminate this comment

cah292: Ok, I'll remove the comment.

src/main/java/edu/cornell/kfs/sys/document/service/impl/CuPaymentSourceHelperServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+import org.kuali.kfs.sys.service.impl.KfsParameterConstants;
+import org.kuali.rice.core.api.util.type.KualiInteger;
+
+import edu.cornell.kfs.fp.document.CuDisbursementVoucherConstants;
+
+public class CuPaymentSourceHelperServiceImpl extends PaymentSourceHelperServiceImpl {
+    
+    @Override
+    public List<PaymentNoteText> buildNotesForCheckStubText(String text, int previousLineCount) {
+        PaymentNoteText pnt = null;
```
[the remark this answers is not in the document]
cah292: This code is a customized version of that from PaymentSourceHelperServiceImpl.buildNotesForCheckStubText, but I'll see if I can clean it up.

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
cah292: This is an overlay of a base KFS file, and this change just makes the method conform to how it appears in the base PaymentDetail file. If you think it could still use cleanup, then I'm fine with renaming the variable.

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
cah292: This is how the line appears in the base PaymentDetail file, but we can add the extra check if you think it's needed. (I don't think the "getCustomerNoteText" call is a very expensive operation anyway.)

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
cah292: This is how the line appears in the base PaymentDetail file, so I don't think we need to change the logging level in this case.

cah292:
FYI, I started writing a simple unit test for the code that we moved to CuPaymentSourceHelperServiceImpl, and I discovered a bug that could result in long check stub words/URLs being partially omitted. I'll include a fix for that in my refactoring as well.

src/main/java/edu/cornell/kfs/sys/document/service/impl/CuPaymentSourceHelperServiceImpl.java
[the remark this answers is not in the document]
jhulslander: It sounded  from the standup that is existing customization that got moved into a new class.  I am fine leaving as is.  Let me know if you want to pursue the cleanup effort.

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
jhulslander: lets leave as is

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
jhulslander: lets leave as is

src/main/java/org/kuali/kfs/pdp/businessobject/PaymentDetail.java
jhulslander: OK, lets leave as is

src/main/java/edu/cornell/kfs/sys/document/service/impl/CuPaymentSourceHelperServiceImpl.java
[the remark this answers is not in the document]
cah292: Based on our discussion at the standup, it sounds like it would be best to leave this as-is for now, since it's basically a cut-and-paste of an existing customization. If it's okay with you and with Daniela, we can move forward with this code as-is, and then perform cleanup and bug fixing in a future PR.

jhulslander:
It sounds like the items I mentioned in this PR aren't really worth pursuing right now (they were existing code that has been running a while and has just been moved, or part of an overlay).  I am fine merging this PR.  @db483 can you weigh in on this before merging?

db483:
Merging sounds good to me. Thank you.

PR #451, opened 2017-08-31
Revert "KFSPTS-9617 add changes to accept multiple values for the CON…

db483:
…CUR_CUSTOMER_PROFILE_GROUP_ID system parameter (#447)"

This reverts commit f8f43d8d07f31d7dafc24f1ba73b6e207ec4d6d9.


db483:
I need to revert this to implement a case insensitive solution. Thank you.

PR #454, opened 2017-09-07
KFSPTS-8767: Add handling of ATM cash advances.

cah292:
This is the final version of the ATM cash advance handling feature for SAE, which is based on the initial work from the KFSPTS-8767-atmCash branch.

I also included some refactoring of the existing SAE-to-Collector code, due to the large number of sub-group types that need to be tracked for consolidating SAE lines. Please let me know if you find any problems with that refactoring, or if you notice any problems with the PDP side of the new feature.

In addition, there's a related nonprod-sql PR that needs to be merged in conjunction with this one.


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -146,8 +216,20 @@ protected String buildAccountingFieldsKey(ConcurStandardAccountingExtractDetailL
                 StringUtils.defaultIfBlank(detailLine.getOrgRefId(), StringUtils.EMPTY));
     }
 
-    protected String buildAccountingFieldsKeyForCashAdvance(
-            ConcurStandardAccountingExtractDetailLine detailLine, ConcurRequestedCashAdvance requestedCashAdvance) {
+    protected String buildAccountingFieldsKeyForAtmCashAdvance(ConcurStandardAccountingExtractDetailLine detailLine) {
+        return String.format(ACCOUNTING_FIELDS_KEY_FORMAT,
```
jhulslander: There are a number of these building an accounting field key functions with similar logic.  I think this could be refactored to have one function do the string format, and the others call it.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -470,23 +626,52 @@ protected OriginEntryFull buildOffsetOriginEntry(OriginEntryFull baseEntry, Kual
...
-                    + detailLine.getCashAdvanceKey());
+    protected OriginEntryFull buildCashAdvanceOriginEntry(ConcurStandardAccountingExtractDetailLine detailLine,
+            KualiDecimal amount) {
+        OriginEntryFull originEntry;
+        if (collectorHelper.isAtmCashAdvanceLine(detailLine)) {
+            originEntry = buildOriginEntryFromDetailLineForAtmCashAdvance(detailLine);
+        } else {
+            ConcurRequestedCashAdvance requestedCashAdvance = requestedCashAdvancesByCashAdvanceKey.get(detailLine.getCashAdvanceKey());
+            if (requestedCashAdvance == null) {
+                throw new IllegalStateException(
```
jhulslander: Do you think it would be helpful to add a LOG.error here?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
```diff
@@ -522,20 +704,32 @@ protected OriginEntryFull buildCorpCardPersonalExpenseOriginEntry(
         return originEntry;
     }
 
+    protected OriginEntryFull buildAtmFeeDebitOriginEntry(ConcurStandardAccountingExtractDetailLine detailLine, KualiDecimal amount) {
+        OriginEntryFull originEntry = new OriginEntryFull();
```
jhulslander: there are a few build origin entry functions here.  Can these be refactored to share functionality?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurSAELineType.java
```diff
@@ -0,0 +1,15 @@
+package edu.cornell.kfs.concur.batch.service.impl;
+
+/**
+ * Helper enum for categorizing various types of SAE lines.
```
jhulslander: I don't think this comment is needed.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurSAELineType.java
```diff
@@ -0,0 +1,15 @@
+package edu.cornell.kfs.concur.batch.service.impl;
+
+/**
+ * Helper enum for categorizing various types of SAE lines.
+ */
+public enum ConcurSAELineType {
```
jhulslander: Since this appears to be just used in Collector processing, would it be better to call this ConcurSAECollectorLineType, or something so it's name suggest it is just for Collector, and not PDP?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: Ok, I'll refactor things so that the String.format() call is in a shared method.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: Farther up the call stack in the ConcurStandardAccountingExtractCollectorBatchBuilder class, there's a try-catch block that will log exceptions that are thrown by this class. Thus, I don't think an extra log statement is needed at this level.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurDetailLineGroupForCollector.java
cah292: Most of the shared OriginEntryFull setup is in the "configureOriginEntryGeneratedFromLine" method; the other accounting data setup may not be easy to consolidate because of the varying requirements for the entries being generated. However, after looking further, I think the "new OriginEntryFull()" call can also be consolidated into that helper method, so I'll try to refactor that accordingly.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurSAELineType.java
cah292: Ok, I'll remove the comment or replace if with a better description that describes what this enum is currently used for.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurSAELineType.java
cah292: Good point, I'll rename this accordingly to avoid confusion with the PDP side.

cah292:
I've made some changes based on the initial feedback. I also refactored some of the OriginEntryFull construction to minimize duplication, and to use code-as-data techniques for some of the entry setup.

PR #456, opened 2017-09-20
KFSPTS-9715: Upgrade to 2017-07-13 version of financials

cah292:
NOTE: This patch also has some related SQL changes in the nonprod-sql repo.

The unit tests may fail, due to impacting database changes that were made with this patch. Once we're closer to the point of moving these changes to develop, we can apply the SQL changes to UTEST and re-run the tests.


cah292:
This is ready for review, but please do NOT merge yet. The SQL changes need to be applied to UTEST in order for the unit tests to pass, and I don't want to apply the SQL there until the code has been reviewed in advance.

cah292:
The SQL has been applied to UTEST, so this should be ready to merge if the unit tests pass.

cah292:
retest this please

PR #460, opened 2017-10-10
KFSPTS-9849: Update KIM JPA metadata to fix bounded person searches

cah292:
NOTE: There are three PRs for this change across these projects: cornell_customizations, cynergy-server-customizations, and cu-kfs.

Also, I recently added a comment on the related user story pertaining to the search warning message. Please wait for a response from me or Cathy on that before merging.

I'm not sure if the unit tests will pass yet, depending on whether the application can still boot up if the new config property references a nonexistent customizer (since the cornell_customizations PR needs merging to move that customizer into the codebase).

### DIFF
--- a/src/main/resources/institutional-config.properties
+++ b/src/main/resources/institutional-config.properties
@@ -117,6 +117,7 @@ cynergy.cuwal.useHttpHeaders=true
 cynergy.antivirus.service=antiVirusService
 cynergy.encryption.busEncryption=true
 cynergy.keystore.reloadSeconds=600
+rice.krad.jpa.kim.eclipselink.descriptor.customizer.org.kuali.rice.kim.impl.identity.entity.EntityBo=edu.cornell.cynergy.kim.impl.identity.entity.EntityBoDescriptorCustomizer
 csrf.enabled=false
 tax.format.1042s.default=classpath:edu/cornell/kfs/tax/batch/Default1042STaxOutputDefinition.xml
 tax.format.1042s.summary.default=classpath:edu/cornell/kfs/tax/batch/DefaultTransactionRowOutputDefinition.xml


src/main/resources/institutional-config.properties
```diff
@@ -117,6 +117,7 @@ cynergy.cuwal.useHttpHeaders=true
 cynergy.antivirus.service=antiVirusService
 cynergy.encryption.busEncryption=true
 cynergy.keystore.reloadSeconds=600
+rice.krad.jpa.kim.eclipselink.descriptor.customizer.org.kuali.rice.kim.impl.identity.entity.EntityBo=edu.cornell.cynergy.kim.impl.identity.entity.EntityBoDescriptorCustomizer
```
jhulslander: I'm curious why we need to do this.  It looks like this same value is being set in cynergy-server-defaults-config.xml .  I don't see rice.krad.jpa.kim.eclipselink.descriptor.customizer.org.kuali.rice.kim.document.IdentityManagementRoleDocument reset here.

cah292: We currently don't have a common institutional config file that is shared by both Cynergy and KFS, so I had to duplicate this property on both the standalone Cynergy and KFS sides. If we have more properties like this in the future, then we could look into adding such a file to cornell_customizations.

The IdentityManagementRoleDocument property is not on the KFS side since that's a standalone-Rice document that doesn't need its metadata customized in KFS. I'm adding this EntityBo property on both sides in case KFS needs to perform bounded searches on person data (since "Embedded" KIM run mode may cause it to run the query on the KFS side).

jhulslander: thanks, that makes sense

cah292:
In order for the unit tests to pass, I believe the related cornell_customizations PR has to be merged and the "develop" Cynergy code has to be rebuilt.

jhulslander:
retest this please

PR #462, opened 2017-10-18
KFSPTS-9913: Add document XML POJOs for AWS billing

cah292:
These are the DI JAXB POJOs that are needed for AWS Billing, and which have been structured based on the example document(s) in Confluence. The test cases are also based on the Confluence examples.

If you notice any errors or omissions in the POJOs, or if you think that extra or stricter unit tests are needed at this time, please let me know.


src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentAccountingLine.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.xml.bind.annotation.adapters.NormalizedStringAdapter;
+import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.rice.core.api.util.type.KualiDecimal;
+
+import edu.cornell.kfs.sys.xmladapters.KualiDecimalXmlAdapter;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = StringUtils.EMPTY, propOrder = {
```
jhulslander: I am not sure if we want to enforce an order here, or on the other POJOs

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
```diff
@@ -0,0 +1,159 @@
...
+    @XmlElement(name = "Explanation", namespace = StringUtils.EMPTY, required = false)
+    protected String explanation;
+
+    @XmlElement(name = "OrganizationDocumentNumber", namespace = StringUtils.EMPTY, required = false)
+    @XmlJavaTypeAdapter(NormalizedStringAdapter.class)
+    protected String organizationDocumentNumber;
+
+    @XmlElementWrapper(name = "AccountingListFrom", namespace = StringUtils.EMPTY, required = false)
+    @XmlElement(name = "Accounting", namespace = StringUtils.EMPTY, required = false)
+    protected List<AccountingXmlDocumentAccountingLine> sourceAccountingLines;
```
jhulslander: should this be fromAccounitngLines?
I would expect we will add more list named well for particular documents as we implement more types

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @XmlElementWrapper(name = "AccountingListTo", namespace = StringUtils.EMPTY, required = false)
+    @XmlElement(name = "Accounting", namespace = StringUtils.EMPTY, required = false)
+    protected List<AccountingXmlDocumentAccountingLine> targetAccountingLines;
```
jhulslander: should this be toAccountingLines

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentAccountingLine.java
cah292: Ok, I'll remove these ordering restrictions.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
cah292: Ok, I'll check the naming on this and change it accordingly.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
cah292: Ok, I'll check the naming on this and change it accordingly.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
cah292: The accounting line lists are referred to as source/target in the KFS code, even if the UI is displaying them as from/to on the DI. (Also, it seems Internal Billing documents are displaying income/expense on the UI in spite of using source/target behind-the-scenes.)

I think it makes more sense to call them source/target lines in the new objects. If you think that the XML wrapper elements should be renamed to source/target, or if you think we should keep the from/to elements and add new ones later (source/target, income/expense, etc.), please let me know.

cah292:
I've added some more changes based on the feedback. If you think that the account-related POJO list names and XML wrapper element names are okay as-is, then this is ready for review again.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentEntry.java
jhulslander: I think I agree that using the source target naming makes sense.  However, since we will have external customers using this, I am worried about the names in the XML for transaction lists not matching the names they see on the documents in KFS.  @db483 I'd like your take on the matter

jhulslander: it occurs to me,  we should keep the tag and pojo field name the same, so if we ant to keep the pojo using source/target, we should change the XML to be consistent.  It is brand new XML, so we can change it without any repercussions.

cah292: Good point, I'll change the XML accordingly.

jhulslander: I think we need to confirm what naming convention we want

cah292:
As per the discussion at today's sprint retrospective, I have renamed the accounting line from/to wrapper elements to instead use source/target accounting line names. This is now ready for review again.

NOTE: Many of the wrapper elements' names have a suffix of "List"; is that okay, or would it be better to use plural words instead? (For example, should we use "SourceAccountingLines" instead of "SourceAccountingLineList"?)

cah292:
retest this please

jhulslander:
I think the new convention works, thanks

cah292:
Thanks, Jay.

Also, do you think that the backup link wrapper needs to be updated to the "List"-suffix naming? Or do you think its current name is fine for now?

jhulslander:
I'm inclined to leave it as is for now.  The backup Link section will probably need to have another attribute added after we confirm how we will be download the files.  I suspect we are going to have an attribute to set the authentication mechanism needed to download the file.

jhulslander:
retest this please

cah292:
retest this please

jhulslander:
retest this please

jhulslander:
retest this please

jhulslander:
retest this please

PR #464, opened 2017-10-24
KFSPTS-10100: Fix ICR ID validation issue on KFS docs

cah292:
The 08-10 KualiCo patch reorganized and revised various data dictionary beans, and this caused our customization for ICR ID alphanumeric validation to not affect all of the needed objects. (We overrode the ICR Rate ID property bean in our existing code, but the 08-10 patch caused other BOs' DD config to not extend from this bean.) I was able to fix this in my local environment by adding some more bean overrides to add the customization back in where needed.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/A21SubAccount.xml
+++ b/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/A21SubAccount.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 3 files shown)


cah292:
No, these changes currently do not fix the labeling issue that was reported on Account Global docs. It does look like 08-10 KualiCo changes for the ICR beans also caused that labeling issue, so I'll add a fix for that and let you know when that's ready. Thanks for checking!

cah292:
I've updated the fix so that it also corrects the ICR ID label on the Account Global doc. Also, based on the bean setup for Sub-Account Global, I believe that document already has the correct ICR ID setup.

This is now ready for review again.

PR #468, opened 2017-10-27
KFSPTS-10124: Place ICR as the beginning characters on System generated ICR entries.

nkimble:


src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
```diff
@@ -52,4 +60,161 @@ protected KualiDecimal getPercentage(KualiDecimal amount, BigDecimal percent) {
...
+     * CUMod to prefix the "Indirect Cost Recovery System Generated" edoc number with "ICR"
+     * 
+     * Generate a transfer transaction and an offset transaction
+     *
+     * @param et                         an expenditure transaction
+     * @param generatedTransactionAmount the amount of the transaction
+     * @param runDate                    the transaction date for the newly created origin entry
+     * @param group                      the group to save the origin entry to
+     */
+    protected final static String ICR_EDOC_PREFIX = "ICR";
```
cah292: I would suggest moving this constant towards the beginning of the class to make it easier to find, and to also reorder the declaration to be "static final" instead of "final static".

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
```diff
@@ -52,4 +60,161 @@ protected KualiDecimal getPercentage(KualiDecimal amount, BigDecimal percent) {
...
+            warnings.add(new Message("Infinite recursive encumbrance error " + et.getChartOfAccountsCode() + " " + et.getAccountNumber() + " " + et.getSubAccountNumber() + " " + et.getObjectCode() + " " + et.getSubObjectCode(), Message.TYPE_WARNING));
+            reportWriterService.writeError(et, warnings);
+            return;
+        }
+
+        e.setFinancialDocumentTypeCode(parameterService.getParameterValueAsString(PosterIndirectCostRecoveryEntriesStep.class, KFSConstants.SystemGroupParameterNames.GL_INDIRECT_COST_RECOVERY));
+        e.setFinancialSystemOriginationCode(parameterService.getParameterValueAsString(KfsParameterConstants.GENERAL_LEDGER_BATCH.class, KFSConstants.SystemGroupParameterNames.GL_ORIGINATION_CODE));
+        SimpleDateFormat sdf = new SimpleDateFormat(DATE_FORMAT_STRING);
+        /*CUMod - start*/
+        StringBuffer docNbr = new StringBuffer(ICR_EDOC_PREFIX);
```
cah292: Unless you have a need for thread safety or for working with the limitations of other APIs, it would be better to use StringBuilder instead of StringBuffer here.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
```diff
@@ -52,4 +60,161 @@ protected KualiDecimal getPercentage(KualiDecimal amount, BigDecimal percent) {
...
+        docNbr.append(sdf.format(runDate));
+        e.setDocumentNumber(docNbr.toString());
+        LOG.debug("CuPosterServiceImpl.generateTransactions: setDocumentNumber=" + docNbr.toString() + "=");
```
cah292: Do you need the extra "=" sign at the end of the log text? If not, then I would suggest removing it.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
```diff
@@ -52,4 +60,161 @@ protected KualiDecimal getPercentage(KualiDecimal amount, BigDecimal percent) {
...
+        //e.setDocumentNumber(sdf.format(runDate));
```
cah292: I think this comment could be removed, since your other comments already highlight the start and stop of the mod so that others know where to look to see the original code.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
nkimble: My understanding was that this fix was going to be suggested as a KualiCo mod. Therefore, my intention of putting the constant here was to keep it close to the actual method where it was being used so it was easier to find during packaging of that work.

nkimble: I'll flip the "static final".

cah292: Ok, keeping the constant in this position is okay with me, then. And thanks for moving forward with the "static final" flip.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
nkimble: The base PosterServiceImpl class uses StringBuffer so I was keeping the code consistent.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
nkimble: My intent of the extra equal sign is to find null or blank Strings that we sometimes get as bugs.  Those should not occur here but I wold prefer to leave it in unless you are really opposed to it.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
nkimble: I'll remove the original line of code.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
cah292: In that case, using StringBuffer in this spot is okay for now, but I'm also fine with you switching to StringBuilder if you want to. Hopefully KualiCo can clean up their code in the future to use StringBuilder in most cases.

src/main/java/edu/cornell/kfs/gl/batch/service/impl/CuPosterServiceImpl.java
cah292: What about using single or double quotes around the printed doc number? I think that would make the formatting clearer, and would make it easier to spot empty or blank values.

nkimble: I'll change the "=" to single quotes.

cah292:
retest this please

PR #469, opened 2017-10-31
KFSPTS-10067: Upgrade to 2017-09-07 version of financials

cah292:
NOTE: There are PRs in cu-kfs and nonprod-sql for this change. Please make sure both PRs are good to go before merging.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
+++ b/src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
@@ -1152,12 +1152,10 @@ public class PurchasingActionBase extends PurchasingAccountsPayableActionBase {
         String errorPath = PurapConstants.CAPITAL_ASSET_TAB_ERRORS;
         // validate entry is selected for each field
         if (StringUtils.isEmpty(document.getCapitalAssetSystemTypeCode())) {
-            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_MISSING, "Capital Assets System Type Code");
-        }
-        else if (StringUtils.isEmpty(document.getCapitalAssetSystemStateCode())) {
-            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_MISSING, "Capital Assets System State Code");
-        }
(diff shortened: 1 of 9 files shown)


src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
```diff
@@ -1152,12 +1152,10 @@ public ActionForward selectSystem(ActionMapping mapping, ActionForm form, HttpSe
...
         if (StringUtils.isEmpty(document.getCapitalAssetSystemTypeCode())) {
-            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_MISSING, "Capital Assets System Type Code");
-        }
-        else if (StringUtils.isEmpty(document.getCapitalAssetSystemStateCode())) {
-            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_MISSING, "Capital Assets System State Code");
-        }
-        else {
+            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_CUSTOM, "Capital Asset System Type and Capital Asset System State are both required to proceed");
+        } else if (StringUtils.isEmpty(document.getCapitalAssetSystemStateCode())) {
+            GlobalVariables.getMessageMap().putError(errorPath, KFSKeyConstants.ERROR_CUSTOM, "Capital Asset System Type and Capital Asset System State are both required to proceed");
```
jhulslander: It looks like this is adding details to what is required, should the error "Capital Asset System Type Code and Capital Asset System State are both required to proceed" since "Code" was in the original error message?  Both errors are the same string, so should this be a local variable and reused, or perphaps included in or application resources property file?

cah292: This is an overlay of a base KFS class, and this is how the change has been set up in base code. If you think it's still worth trying to refactor this for simplicity or clarity, please let me know.

jhulslander: OK, lets leave as is, I wasn't sure so I asked

PR #472, opened 2017-11-13
Revert "KFSPTS-9312 Setting up locking job for CU implementation, and…

cah292:
… support… (#426)"

This reverts commit cb36587d15dd47d849a14a8fa28b9776788b3ec9.


cah292:
I need to revert this, so that I can consolidate my changes and these changes into a single commit, and to also revise these bean changes as needed.

PR #478, opened 2017-12-04
KFSPTS-9131 Creating Corporate Billed Corporate Paid document type and batch job

jhulslander:
I believe this is ready for review again.  It looks like there are separate BOs and tables for the various CBCP processes.  The file formats of PCDO and CBCP are similar so the code that parses the files are shared.


jhulslander:

src/main/resources/edu/cornell/kfs/fp/document/datadictionary/CorporateBilledCorporatePaidDocument.xml
```diff
@@ -0,0 +1,25 @@
...
+ http://www.opensource.org/licenses/ecl2.php
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+  <bean id="corporateBilledCorporatePaidDocument" parent="ProcurementCardDocument-parentBean">
```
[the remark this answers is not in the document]
jhulslander: Probably would be better to use the non-parent, but I don't think it's negatively impacting anything.  If Daniela has concerns, I will include this in what she finds.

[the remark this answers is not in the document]
cah292: Ok, that's fine with me. Thanks!

src/main/java/edu/cornell/kfs/fp/batch/CorporateBilledCorporatePaidFlatInputFileType.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+    
+    @Override
+    public String getFileTypeIdentifer() {
+        return CuFPConstants.CORPORATE_BILLED_CORPORATE_PAID_FLAT_INPUT_FILE_TYPE;
+    }
+    
+    @Override
+    public String getFileName(String principalName, Object parsedFileContents, String userIdentifier) {
+        String fileName = "cbcp_" + principalName;
```
db483: should the defined file prefix be used instead of hard coded "cbcp_"?
At the same time I can see that the code this is based of does the same thing.

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
```diff
@@ -1101,5 +1101,391 @@
...
+	    <field-descriptor name="unitOfMeasure" column="UOM" jdbc-type="VARCHAR" />
+	    <field-descriptor name="unitCost" column="UNIT_COST" jdbc-type="DECIMAL" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbKualiDecimalFieldConversion" />
+	    <field-descriptor name="typeOfSupply" column="TYPE_OF_SUPPLY" jdbc-type="VARCHAR" />
+	    <field-descriptor name="serviceIdentifier" column="SRVC_ID" jdbc-type="VARCHAR" />
+	    <field-descriptor name="messageIdentifier" column="MSG_ID" jdbc-type="VARCHAR" />
+	    <field-descriptor name="itemSequenceNumber" column="ITM_SEQ_NUM" jdbc-type="INTEGER" />
+	    <field-descriptor name="lineItemDetailIndicator" column="LN_ITM_DTL_IND" jdbc-type="INTEGER" />
+	  </class-descriptor>
+	  
+	  <class-descriptor class="edu.cornell.kfs.fp.businessobject.CorporateBilledCorporatePaidTransaction" table="CU_FP_CBCP_TRN_MT">
```
db483: formatting is a bit off in certain places in this file

src/main/webapp/jsp/fp/ProcurementCard.jsp
```diff
@@ -0,0 +1,71 @@
+<%--
```
db483: I think it would be better to define a new jsp for the CBCP doc. We want to avoid customizing base code. It will also be confusing down the road to maintain this - one would expect a CorporateBilledCorporatePaid.jsp for CBCP instead of a ProcurementCard.jsp

src/main/java/edu/cornell/kfs/fp/batch/service/CuProcurementCardLoadTransactionsService.java
```diff
@@ -0,0 +1,7 @@
+package edu.cornell.kfs.fp.batch.service;
+
+import org.kuali.kfs.fp.batch.service.ProcurementCardLoadTransactionsService;
+
+public interface CuProcurementCardLoadTransactionsService extends ProcurementCardLoadTransactionsService {
```
db483: I am not sure I understand why this CuProcurementCardLoadTransactionsService is needed and not a CorporateBilledCorporatePaidLoadTransactionService instead. Then the loadProcurementCardFile could also be renamed to loadCorporateBilledCorporatePaidFile.

jhulslander: This actually isn't used in CBCP processing, it is needed for unit testing in ProcurementCardCreateDocumentServiceImplTest and ProcurementCardLoadFlatTransactionsServiceImplTest.  These aren't micro tests, and load spring.  
The original tests got ProcurementCardLoadTransactionsService from Spring and cast it to ProcurementCardLoadFlatTransactionsServiceImpl so the test could call loadProcurementCardFile.  That cast to ProcurementCardLoadFlatTransactionsServiceImpl was failing after my changes.  I suspect it had something to do with there being multiple services that implement ProcurementCardLoadTransactionsService now.

db483: That is strange, but it seems like a workaround rather than a fix. You've been changing this implementation since that unit test failure, would the unit tests still fail now if you removed this interface?

jhulslander: I can try

jhulslander:
OK, this is ready for review again

src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardCreateDocumentServiceImplTest.java
```diff
@@ -26,7 +26,7 @@
     private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ProcurementCardSummaryFeedService.class);
     
     private UnitTestSqlDao unitTestSqlDao;
-
```
db483: since this file has only formatting changes I don't think it should be included in this commit

src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardLoadFlatTransactionsServiceImplTest.java
```diff
@@ -24,9 +24,9 @@
     private static org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(ProcurementCardSummaryFeedService.class);
     
     private UnitTestSqlDao unitTestSqlDao;
-
```
db483: since this file has only formatting changes I don't think it should be included in this commit

src/main/resources/edu/cornell/kfs/fp/cu-spring-fp.xml
```diff
@@ -119,6 +120,13 @@
         </property>
     </bean>
     
+    <bean id="procurementCardLoadFlatTransactionsServiceImpl"  parent="procurementCardLoadFlatTransactionsServiceImpl-parentBean"/>
```
db483: this is something that we'll want removed, potentially with a future story

src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardCreateDocumentServiceImplTest.java
jhulslander: OK, I will see if I can this out of the comit

src/test/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardLoadFlatTransactionsServiceImplTest.java
jhulslander: OK, I will see if I can this out of the commit

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
```diff
@@ -1101,5 +1101,478 @@
     <field-descriptor name="objectId" column="OBJ_ID" jdbc-type="VARCHAR" index="true"/>
     <field-descriptor name="noteText" column="ACH_NTE_TXT" jdbc-type="VARCHAR"/>
   </class-descriptor>
+  
+  <class-descriptor
+	class="edu.cornell.kfs.fp.document.CorporateBilledCorporatePaidDocument"
+	table="CU_FP_CBCP_DOC_T">
```
cah292: If you want these descriptors to use tabs instead of spaces, then please update the "class-descriptior" tag opening above to also use tabs.

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
```diff
@@ -1101,5 +1101,478 @@
...
+	<field-descriptor name="nextTargetLineNumber" column="FDOC_NXT_LINE_NBR"
+		jdbc-type="INTEGER" />
+	<field-descriptor name="autoApprovedIndicator" column="AUTO_APPROVED_IND"
+		jdbc-type="VARCHAR"
+		conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbCharBooleanConversion" />
+
+	<reference-descriptor name="documentHeader"
+		class-ref="org.kuali.kfs.sys.businessobject.FinancialSystemDocumentHeader"
+		auto-retrieve="true" auto-update="object" auto-delete="object" proxy="true">
+		<foreignkey field-ref="documentNumber" />
```
cah292: If you're splitting up the opening tags along multiple lines, I would suggest using a different level of indentation than that of the sub-elements. This would make the XML easier to read, and would be more consistent with how we indent our code.

jhulslander:
OK, I removed the two unit test classes from this PR

jhulslander:
I fixed the formatting notes, ready again

jhulslander:
retest this please

cah292:
I just received word from Daniela that these changes can move forward once the unit tests pass. I will dismiss Daniela's request for changes since the updates have been made already.

PR #480, opened 2017-12-07
KFSPTS-10391: Fix country handling on Payment Detail inquiry

cah292:
This fixes a country display error on the Payment Detail inquiry in the 09/07 patch, so that the country-code-to-name conversion will be skipped if the value represents a country name or other value that cannot be converted.

This is only a temporary fix; we should remove it once KualiCo fixes it and we bring in a patch containing that fix.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/CUPdpPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/pdp/CUPdpPropertyConstants.java
@@ -4,6 +4,14 @@ public final class CUPdpPropertyConstants {
 
     public static final String PAYEE_PRINCIPAL_NAME = "payeePrincipalName";
 
+    public static final class PaymentDetail {
+        public static final String PAYMENT_COUNTRY = "paymentGroup.country";
+        
+        private PaymentDetail() {
+            throw new UnsupportedOperationException("do not call constructor");
+        }
+    }
+
     private CUPdpPropertyConstants() {
         throw new UnsupportedOperationException("do not call");
     }
--- a/src/main/java/edu/cornell/kfs/pdp/CUPdpPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/pdp/CUPdpPropertyConstants.java
@@ -3,14 +3,7 @@ package edu.cornell.kfs.pdp;
 public final class CUPdpPropertyConstants {
 
     public static final String PAYEE_PRINCIPAL_NAME = "payeePrincipalName";
-
-    public static final class PaymentDetail {
-        public static final String PAYMENT_COUNTRY = "paymentGroup.country";
-        
-        private PaymentDetail() {
-            throw new UnsupportedOperationException("do not call constructor");
-        }
-    }
+    public static final String PAYMENT_COUNTRY = "paymentGroup.country";
 
     private CUPdpPropertyConstants() {
         throw new UnsupportedOperationException("do not call");
--- a/src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
+++ b/src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
@@ -1,0 +1,45 @@
+package edu.cornell.kfs.pdp.businessobject.inquiry;
+
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.kns.web.ui.Field;
+import org.kuali.kfs.kns.web.ui.Section;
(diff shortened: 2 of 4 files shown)


src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
```diff
@@ -0,0 +1,45 @@
...
+     * or when it contains a country name instead of a country code.
+     * 
+     * @see org.kuali.kfs.pdp.businessobject.inquiry.PaymentDetailInquirable#convertCountryForDisplay(org.kuali.kfs.kns.web.ui.Section)
+     */
+    @Override
+    protected void convertCountryForDisplay(Section section) {
+        section.getRows().stream()
+                .flatMap((row) -> row.getFields().stream())
+                .filter(this::isPaymentGroupCountryField)
+                .findFirst()
```
jhulslander: do we only need to update the display for the first field, and not the rest?

src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
```diff
@@ -0,0 +1,45 @@
...
+
+    protected boolean isPaymentGroupCountryField(Field field) {
+        return StringUtils.equalsIgnoreCase(CUPdpPropertyConstants.PaymentDetail.PAYMENT_COUNTRY, field.getPropertyName());
+    }
+
+    protected void convertCountryForDisplayIfNecessary(Field field) {
+        String propertyValue = field.getPropertyValue();
+        if (StringUtils.isNotBlank(propertyValue)) {
+            Country country = getCountryService().getCountry(propertyValue);
+            if (ObjectUtils.isNotNull(country)) {
```
jhulslander: Is it an error condition of a country can't be found?

src/main/java/edu/cornell/kfs/pdp/CUPdpPropertyConstants.java
```diff
@@ -4,6 +4,14 @@
 
     public static final String PAYEE_PRINCIPAL_NAME = "payeePrincipalName";
 
+    public static final class PaymentDetail {
+        public static final String PAYMENT_COUNTRY = "paymentGroup.country";
```
jhulslander: Why does PAYMENT_COUNTRY need to be in an internal class, and not just in CUPdpPropertyConstants?

cah292: I was trying to be consistent with how similarly-named properties were being grouped in the PdpPropertyConstantsClass. If you think this adds too much clutter, then I'll remove this nested class and move the constant to the parent class.

src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
cah292: There should theoretically be only one field with the "paymentGroup.country" property name. The superclass's version of this method also short-circuits after updating the first match, so I was trying to be consistent with the superclass's logic as much as possible.

src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
cah292: This code is related to just the display of the payment information on an inquiry screen, not its validation. The other processes should be the ones that check for errors when an invalid country is displayed, or if a country code was expected instead of a country name.

We could add a feature that displays an additional error-related message if the country value is a code but doesn't map to an actual country. However, we would need to confirm with the functionals if that's truly the behavior they want, and that would likely be in the scope of a future enhancement unless the functionals say otherwise.

src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
jhulslander: OK, lets go with as is then

src/main/java/edu/cornell/kfs/pdp/businessobject/inquiry/CuPaymentDetailInquirable.java
jhulslander: OK, I think this implementation is fine then

cah292:
I've simplified the constants class as per the initial feedback. This is now ready for review again.

PR #484, opened 2017-12-12
KFSPTS-9458 update code to match new EZRA field lengths

db483:


src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -365,8 +365,8 @@ public Agency createAgency(Long sponsorId) {
 		Agency agency = new Agency();
 		agency.setAgencyNumber(sponsorId.toString());
 		agency.setReportingName(sponsor.getSponsorLabel());
-		if (ObjectUtils.isNotNull(sponsor.getSponsorName()) && sponsor.getSponsorName().length() > 50)
-			agency.setFullName(sponsor.getSponsorName().substring(0,49));
```
cah292: Should the second substring() argument be 150 instead of 149? I know the old code had 49 instead of 50, but the ending index for substring() calls is an exclusive index, not an inclusive one.

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -365,8 +365,8 @@ public Agency createAgency(Long sponsorId) {
...
+		if (ObjectUtils.isNotNull(sponsor.getSponsorName()) && sponsor.getSponsorName().length() > 150)
+			agency.setFullName(sponsor.getSponsorName().substring(0,149));
```
cah292: Should the second substring() argument be 150 instead of 149? I know the old code had 49 instead of 50, but the ending index for substring() calls is an exclusive index, not an inclusive one.

db483: Good point, I will update that. Thank you!

db483:
I have added the suggested changes.

db483:
Good question Chad, I thought that was discussed as part of the analysis and it was determined as not needed. Are the affected KFS fields expected to have the UI max length updated as well? Thank you.

cah292:
Yes, I believe the associated data dictionary attributes will also need the length increases, so that the descriptions can be displayed or edited appropriately on the UI. Sorry if this wasn't clear in the requirements.

db483:
Thank you Chad, I will update those as well!

db483:
Hi Chad, this is ready for review again. Thank you!

PR #490, opened 2018-01-04
Remove unused Budget Construction constants and properties

ajd299:
Please review these changes


ajd299:
I found no references to this property in the code to the AccountReportsNotBudgetableAccount property and therefore have removed the property.

PR #491, opened 2018-01-12
 KFSPTS-10631: Fix sub-object validation issue in Scrubber 

cah292:
The 2017-09-14 KualiCo patch introduced a bug that allows invalid sub-objects to make it through the GL and LD Scrubber jobs. KualiCo fixed it in the 2018-01-11 patch, so this PR incorporates that fix into our code until we can upgrade to that KualiCo patch or later.

Because of the ScrubberValidator class hierarchies I was dealing with, I relied on Java 8 "default" methods to minimize the amount of code duplication when implementing this. If you think it would be better to just have the corrected method duplicated in both validator classes, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/gl/service/CuSharedScrubberValidatorFixes.java
+++ b/src/main/java/edu/cornell/kfs/gl/service/CuSharedScrubberValidatorFixes.java
@@ -1,0 +1,46 @@
+package edu.cornell.kfs.gl.service;
+
+import org.kuali.kfs.coa.businessobject.SubObjectCode;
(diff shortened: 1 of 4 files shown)


cah292:
I was originally trying to avoid resorting to an overlay. After further consideration, though, I agree that an overlay would make the most sense, since this is only going to be a temporary change that will be removed in a few months anyway.

I may end up closing this PR to create a new one that has just the overlay in it.

cah292:
I'm closing this PR so that I can create one for the overlay-based fix.

PR #493, opened 2018-01-16
KFSPTS-10642: Transactional improvements for PDP cancel/paid job

cah292:
This modifies the PDP Cancel/Paid Job to process each payment update within its own transaction, allowing for the job to resume where it left off if it somehow terminates unexpectedly.


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuProcessPdpCancelPaidServiceImpl.java
```diff
@@ -4,77 +4,60 @@
...
-import org.kuali.kfs.sys.document.PaymentSource;
-import org.kuali.rice.kew.api.exception.WorkflowException;
-import org.kuali.kfs.krad.util.ObjectUtils;
+import org.springframework.transaction.annotation.Transactional;
 
-import edu.cornell.kfs.integration.purap.CuPurchasingAccountsPayableModuleService;
-import edu.cornell.kfs.pdp.businessobject.PaymentDetailExtendedAttribute;
+import edu.cornell.kfs.pdp.batch.service.ProcessPdpCancelPaidHelperService;
 
+@Transactional
```
db483: I am not sure why we need this annotation at the service level?

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/ProcessPdpCancelPaidHelperServiceImpl.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        paymentGroupService.processCancelledGroup(paymentDetail.getPaymentGroup(), processDate);
+    }
+
+    /**
+     * Default implementation uses most of the "while" loop contents from the ProcessPdpCancelPaidServiceImpl.processPdpPaids method,
+     * with additional CU-related changes as needed.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
db483: I have looked a bit into Propagation.REQUIRES_NEW and it looks like it might cause some performance overhead. This job already has some performance issues when we have large amounts of data to process. We should also do some testing to see how performance is affected.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/ProcessPdpCancelPaidHelperServiceImpl.java
```diff
@@ -0,0 +1,149 @@
...
+                LOG.warn("processPdpPaids() Unknown document type (" + documentTypeCode + ") for document ID: " + documentNumber);
+                return;
+            }
+        }
+
+        paymentGroupService.processPaidGroup(paymentDetail.getPaymentGroup(), processDate);
+    }
+
+    // Copied from ProcessPdpCancelPaidServiceImpl
+    protected PaymentSourceToExtractService<PaymentSource> getPaymentSourceToExtractService(PaymentDetail paymentDetail) {
```
db483: This will make upgrades difficult. If anything changes in the base method in ProcessPdpCancelPaidServiceImpl that this was copied from we may miss those changes.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuProcessPdpCancelPaidServiceImpl.java
cah292: I was trying to be consistent with the ProcessPdpCancelPaidServiceImpl superclass, which also has this annotation at the class/service level.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/ProcessPdpCancelPaidHelperServiceImpl.java
cah292: In that case, I could try changing this so that the REQUIRES_NEW processing will handle batches of a specific size, rather than processing each document individually. That should help improve efficiency without rolling back the entire set of updates on each job failure.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/ProcessPdpCancelPaidHelperServiceImpl.java
cah292: Ok, I will see if I can refactor the code to avoid as much of this duplication as possible, while still making the new transaction management possible.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuProcessPdpCancelPaidServiceImpl.java
cah292: Due to the refactoring changes I'm currently making, I will remove this class-level annotation for now and see if things still work.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/ProcessPdpCancelPaidHelperServiceImpl.java
cah292: After further discussion with Daniela, I will leave the per-payment transaction setup in place until we can get a better idea of how much of a performance impact it causes. (This may have to wait until the code moves to DEV, given the inherent slowness issues when running KFS locally.) If it ends up causing too much of a slow-down, then I can come up with a batch-processing-based solution.

cah292:
I've made some more changes to minimize code duplication and to address some of the other issues brought up. This is now ready for review again.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuProcessPdpCancelPaidServiceImpl.java
```diff
@@ -4,77 +4,151 @@
...
+                LOG.warn("processPdpPaids() Unknown document type (" + documentTypeCode + ") for document ID: " + documentNumber);
+                return;
+            }
         }
-   }
+
+        paymentGroupService.processPaidGroup(paymentDetail.getPaymentGroup(), processDate);
+    }
+
+    protected CuProcessPdpCancelPaidService getProxiedProcessPdpCancelPaidService() {
```
db483: Hi Chad,
I am not sure I understand why this method is needed?

cah292: If the service calls its new processPdpCancel() or processPdpPaid() methods internally, then it won't observe the custom "Transactional" annotation setup on those methods. In order for the appropriate transactional behavior to occur, I have to call the methods on the Spring-generated proxy that wraps around this service instance.

db483: Thank you Chad.

PR #494, opened 2018-01-17
Do not merge (for instructional purposes)

ajd299:
I've tried to make the code as clean and minimal as possible while conforming to the practices of the development team such as using properties and constants instead of hardcoding values in Java.


src/main/java/edu/cornell/kfs/fp/CuFPConstants.java
```diff
@@ -171,4 +171,8 @@ public static ScheduleTypes fromName(String name) {
         public static final String TRANSACTION_DESCRIPTION_STARTER = "AWS CHARGES ";
     }
 
+    public static class AccountFundsUpdateConstants {
+        public final static Integer MAX_TOTAL_ALLOWED = 10000;
```
db483: It was not part of the requirements, but just an idea - this is a good candidate for a system parameter as the functionals may change their mind about the max allowed value and it is easier to control that using a system parameter

src/main/java/edu/cornell/kfs/fp/document/validation/impl/TotalAmountNonZeroValidation.java
```diff
@@ -0,0 +1,39 @@
...
+public class TotalAmountNonZeroValidation extends GenericValidation {
+    private AccountingDocument accountingDocumentForValidation;
+
+    public boolean validate(AttributedDocumentEvent event) {
+        KualiDecimal sourceTotalAmount = accountingDocumentForValidation.getSourceTotal();
+        KualiDecimal targetTotalAmount = accountingDocumentForValidation.getTargetTotal();
+
+        boolean ret = true;
+
+        if(sourceTotalAmount == null || !sourceTotalAmount.isNonZero()){
```
db483: Please use ObjectUtils to check for nulls.

db483: perhaps it would make more sense to use the isZero method?

src/main/java/edu/cornell/kfs/fp/document/web/struts/AccountFundsUpdateForm.java
```diff
@@ -0,0 +1,19 @@
...
+import org.kuali.kfs.sys.web.struts.KualiAccountingDocumentFormBase;
+
+public class AccountFundsUpdateForm extends KualiAccountingDocumentFormBase {
+    public AccountFundsUpdateForm() {
+        super();
+    }
+
+    @Override
+    protected String getDefaultDocumentTypeName() {
+        return "AFU";
```
db483: the doc type should be moved into a constant in a constants file

src/main/java/edu/cornell/kfs/fp/document/validation/impl/TotalAmountNonZeroValidation.java
ajd299: Agreed, as a best practice it is always better to avoid using the ! operator when feasible. In ruby there is an unless(<expression>) statement that is the opposite of an if condition, meaning that it will always execute the code within unless the condition is met.

ajd299:
I have modified the code and added a system parameter via database insert to address your feedback. Please let me know if the changes are viable.

db483:
Andrew,

Your changes look good. I see you also made some changes for the IB doc? I think that is still work in progress, correct?

Thank you!

ajd299:
These changes should be ready for code review now. I made some final cleanups yesterday

ajd299:
Closing this also. Let me know if that's not alright.

PR #495, opened 2018-01-18
KFSPTS-10604: Allow expired accounts on POA documents

cah292:
This PR updates POAs to allow for specifying expired accounts, as per the associated user story.

Note that, due to the way the code was structured, the CuPurchaseOrderAmendmentAccountValidation class had to copy nearly all of the code from the PurchaseOrderAmendmentAccountValidation parent class in order to make the necessary changes. If you think it would be better to just overlay PurchaseOrderAmendmentAccountValidation instead, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuPurchaseOrderAmendmentAccountValidation.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuPurchaseOrderAmendmentAccountValidation.java
@@ -1,0 +1,111 @@
+package edu.cornell.kfs.module.purap.document.validation.impl;
+
+import java.util.List;
+
+import org.kuali.kfs.krad.util.GlobalVariables;
+import org.kuali.kfs.module.purap.PurapConstants;
+import org.kuali.kfs.module.purap.PurapKeyConstants;
+import org.kuali.kfs.module.purap.businessobject.PurApAccountingLine;
+import org.kuali.kfs.module.purap.businessobject.PurApItem;
+import org.kuali.kfs.module.purap.document.PurchaseOrderDocument;
+import org.kuali.kfs.module.purap.document.service.PurchaseOrderService;
+import org.kuali.kfs.module.purap.document.validation.impl.PurchaseOrderAmendmentAccountValidation;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.kfs.sys.document.validation.event.AttributedDocumentEvent;
+
+public class CuPurchaseOrderAmendmentAccountValidation extends PurchaseOrderAmendmentAccountValidation {
+
+    /**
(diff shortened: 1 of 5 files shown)


src/main/resources/edu/cornell/kfs/module/purap/document/validation/configuration/PurchaseOrderAmendmentValidation.xml
```diff
@@ -109,4 +109,53 @@
   			</list>
   		</property>
     </bean>
+
+    <bean id="PurchaseOrderAmendment-addAccountingLineDefault-failFastValidation" abstract="true" parent="CompositeValidation" scope="prototype">
```
db483: I cannot find what changed here compared to baseline?

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuPurchaseOrderAmendmentAccountValidation.java
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                }
+            }
+        }
+
+        return valid;
+    }
+
+    // Copied and tweaked this superclass method, and increased its visibility.
+    protected boolean isItemChanged(PurApItem poaItem, List<PurApItem> poItems) {
```
db483: I am not sure there is a better solution for this. We can put in a request with KualiCo to increase vivibility for this method in baseline so that we can remove it from the customized file.

src/main/resources/edu/cornell/kfs/module/purap/document/validation/configuration/PurchaseOrderAmendmentValidation.xml
cah292: I changed one of the sub-bean references from "Purchasing-accountingLineDataDictionaryValidation" to the new "PurchaseOrderAmendment-accountingLineDataDictionaryValidation" bean in this PR. Let me know if you think I need XML comments to mark that change.

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuPurchaseOrderAmendmentAccountValidation.java
cah292: I agree that getting KualiCo to increase the method visibility would be the best option in this case. I will leave this code as-is, then, unless there are other issues discovered with it.

src/main/resources/edu/cornell/kfs/module/purap/document/validation/configuration/PurchaseOrderAmendmentValidation.xml
db483: no need to update. Thank you for pointing that out!

PR #496, opened 2018-01-18
KFSPTS-10544 fix key for employee not found error message

db483:
There key constant for the employee not found error message did not exist in the CU-ApplicationResources.properties file. This caused a null pointer exception when generating the report and the email notification not to be sent. I have fixed this by using the correct key for the error message.


db483:
retest this please

PR #507, opened 2018-01-31
KFSPTS-9917 adding backup documentation to create accounting document job

jhulslander:


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingDocumentGeneratorBase.java
```diff
@@ -111,6 +124,38 @@ protected Note buildDocumentNote(AccountingXmlDocumentNote xmlNote) {
         note.setNotePostedTimestampToCurrent();
         return note;
     }
+    
+    protected void populateDocumentAttachments(T document, AccountingXmlDocumentEntry documentEntry) {
+        documentEntry.getBackupLinks().stream()
+            .filter(link -> StringUtils.isNotBlank(link.getLinkUrl()))
+            .map(link -> buildDocumentNoteAttachment(document, link))
+            .forEach(document :: addNote);
```
cah292: I would suggest removing the spaces around the double-colon, for consistency with how our other method references are formatted.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        try { 
+            Invocation request = buildClientRequest(accountingXmlDocumentBackupLink.getLinkUrl(), creds);
+            formFile = downloadFileFromWebResource(accountingXmlDocumentBackupLink.getFileName(), request);
+        } catch (URISyntaxException | IOException e) {
+            LOG.error("downloadFile, unable to download file.", e);
+        }
+        return formFile;
+    }
+
+    protected Collection<WebServiceCredential> getWebServiceCredtials(AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink) {
```
cah292: Typo in method name.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        return builder.buildGet();
+    }
+
+    protected File downloadFileFromWebResource(String fileName, Invocation request) throws IOException {
+        File downloadfile = new File(fileName);
+        Response response = request.invoke();
+        if (response.getStatus() == Response.Status.OK.getStatusCode()) {
+            InputStream is = response.readEntity(InputStream.class);
+            byte[] byteArray = IOUtils.toByteArray(is);
+            FileOutputStream fos = new FileOutputStream(downloadfile);
```
cah292: Do you need to write a temporary file for this step? If you can keep the contents as just a byte array, then you can pass a ByteArrayInputStream into the AttachmentService when you create the actual attachment. That way, the AttachmentService can use ClamAV to scan the attachment for viruses before it gets written to disk.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+            fos.write(byteArray);
+            fos.flush();
+            fos.close();
+            IOUtils.closeQuietly(is);
```
cah292: Please use a "finally" block for making the "closeQuietly" call.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+            String attachmentType = null;
+            
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("createAttachmentFromBackupLink, uploadFileName: " + uploadFileName + " mimeType: " + mimeType
+                    + " fileSize: " + fileSize);
+            }
+            InputStream fileContents = new FileInputStream(formFile);
+            Attachment attachment = attachmentService.createAttachment(document, uploadFileName, mimeType, fileSize,
+                    fileContents, attachmentType);
+            formFile.delete();
```
cah292: If you need this temporary file to create the attachment, I would suggest performing the deletion in a "finally" block in case an exception occurs above.

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -527,3 +527,4 @@ message.concur.token.reset.success=The current Concur access token has been rese
 error.concur.token.retet.production=Resetting the access token to an emty string can not be done in a production environment.
 error.cbcp.batch.upload.email.subject=Error occurred during Corporate Billed Corporate Paid batch upload process.
 error.cbcp.batch.upload.email.body.starter=Errors occured during the Corporate Billed Corporate Paid upload process.
+error.create.accounting.document.attachment.download=Could not download the attachment from
```
cah292: I would recommend adding a placeholder to this property, for consistency with similar messages in this file. If you need to resolve the placeholder programmatically, you can use the MessageFormat.format() method to do so.

src/test/java/edu/cornell/kfs/fp/batch/xml/fixture/AccountingXmlDocumentBackupLinkFixture.java
```diff
@@ -3,21 +3,33 @@
...
+    DFA_INDEX_PAGE("https://www.dfa.cornell.edu/index.cfm", "DFA index page", "index.cfm"),
+    AWS_BILLING_INVOICE("https://kfsaws-support.cd.cucloud.net/v1_0/aws/billing/invoice?year=2017&month=01&account=999999999999", "AWS Billing Invoice", 
+            "invoice.pdf", "AWS-Bill");
 
-    public final String linkUrl;
-    public final String description;
+    public String linkUrl;
+    public String description;
+    public String fileName;
+    public String groupCode;
```
cah292: To avoid the risk of accidentally editing the enum fields at runtime, I would suggest declaring these fields as final. Then you can adjust the four-arg constructor to initialize all four fields, and have the three-arg constructor invoke the four-arg one using an appropriate default or null/empty value for the fourth arg.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+            return attachment;
+        } else {
+            LOG.error("createAttachmentFromBackupLink, the form file is NULL");
+        }
+        
+        throw new IOException ("Unable to download attachment: " + accountingXmlDocumentBackupLink.getLinkUrl());
```
cah292: Please remove the space preceding the opening parentheses.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    protected File downloadFile(AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink) {
+        Collection<WebServiceCredential> creds = getWebServiceCredtials(accountingXmlDocumentBackupLink);
+        File formFile = null;
+        try { 
+            Invocation request = buildClientRequest(accountingXmlDocumentBackupLink.getLinkUrl(), creds);
```
cah292: Should we add some kind of check to make sure the URL is pointing to a legit location, and that the credential group corresponds to what is expected for that URL? Otherwise, we could end up sending web credentials to malicious endpoints, or could send the wrong set of credentials to the endpoint.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: good idea

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingDocumentGeneratorBase.java
jhulslander: can do

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: ok

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: ok

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: ok

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: ok

src/main/resources/CU-ApplicationResources.properties
jhulslander: can you direct me to an example of what you mean hear, I am not sure I understand

src/test/java/edu/cornell/kfs/fp/batch/xml/fixture/AccountingXmlDocumentBackupLinkFixture.java
jhulslander: ok

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: interesting idea.  Perhaps we should standardize the group code needs to be part of the domain.  For example instead of AWS-Bill, we could use the group code "kfsaws-support.cd.cucloud.net" and look for the group code in the backup link URL.  Do you think that would meet this need?

jhulslander: If not, what are your thoughts on this?

src/main/resources/CU-ApplicationResources.properties
cah292: For an example, see the "error.required" property (in the Rice-side KRAD ApplicationResources.properties file) and the ConcurAccountValidationServiceImpl.checkRequiredAccountInfo() method.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
cah292: Having the group code named as the domain would be one way of doing it. Another possibility would be to create some kind of mapping from domains to group codes. The latter could be maintained using a parameter or, alternatively, through the use of other web credential objects with the domain name as part or all of the key.

jhulslander: Daniela and I talked about this, and we are going to add another credential key "credential_base_URL#" and have the value be the valid domain.

jhulslander:

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,154 @@
...
+                    LOG.debug("isLinkUrlValidForGroupCode, found a CREDENTIAL_BASE_URL in the credentials table that is in the back up link URL");
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    protected boolean isCredentialBaseURLInBackupLinkURL(AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink, WebServiceCredential cred) {
+        return StringUtils.containsIgnoreCase(accountingXmlDocumentBackupLink.getLinkUrl(), cred.getCredentialValue());
```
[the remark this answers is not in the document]
jhulslander: my thought was for it to not include the https:// so I don't think  we want to do a starts with check

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingDocumentGeneratorBase.java
```diff
@@ -111,6 +123,39 @@ protected Note buildDocumentNote(AccountingXmlDocumentNote xmlNote) {
...
+
+    protected void addAttachmentToNote(T document, AccountingXmlDocumentBackupLink backupLink, Note note) {
+        try {
+            Attachment attachment = accountingXmlDocumentDownloadAttachmentService.createAttachmentFromBackupLink(document, backupLink);
+            note.setAttachment(attachment);
+        } catch (IOException e) {
+            LOG.error("addAttachmentToNote, unable to create attachment: " + e.getMessage());
+            String message = MessageFormat.format(
+                    configurationService.getPropertyValueAsString(CuFPKeyConstants.ERROR_CREATE_ACCOUNTING_DOCUMENT_ATTACHMENT_DOWNLOAD), 
+                    KFSConstants.BLANK_SPACE + backupLink.getLinkUrl());
```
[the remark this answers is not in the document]
jhulslander: good catch

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    protected boolean doesCredentialKeyReferenceAValidURLForGroupCode(WebServiceCredential cred) {
```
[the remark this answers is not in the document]
jhulslander: that sounds good

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        return StringUtils.containsIgnoreCase(cred.getCredentialKey(), CuFPConstants.CREDENTIAL_BASE_URL);
```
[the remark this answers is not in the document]
jhulslander: yes, this should be starts with

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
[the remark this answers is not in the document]
cah292: Ok, sounds good. If we do end up needing stricter matching later, we can do that in a future enhancement.

jhulslander:

PR #514, opened 2018-02-08
KFSPTS-9312: Fix module locking and parameter cache clearing

cah292:
This reinserts the module locking changes from the #473 PR, but without the temporary workaround beans that were causing problems with newer KualiCo releases. This is necessary because of the merge conflict resolution that was done when moving the KualiCo 09/07/2017 patch changes to the master branch (since those changes removed these temp beans).


jhulslander:
@cah292 something may be awry, when I click the squash merge button I get a commit message that mentions a revert for 9312.  Can you double check this? I am not seeing anything wrong, but would like another set of eyes to look at it before I do the merge.

cah292:
That should be correct, since I reapplied the changes onto this feature branch by doing a revert of the #511 revert. I resolved the needed merge conflicts when preparing the #511 reversion, so using this PR to undo that will create a commit with the proper changes in it.

PR #518, opened 2018-02-12
KFSPTS-10764: Upgrade to 2017-10-19 version of financials

cah292:
NOTE: There are PRs for this in both cu-kfs and nonprod-sql. Please make sure both are good to go before merging them.

Some of the overlays have also been updated to account for cleanup and other improvements in the base KFS files. Please let me know if you find any problematic changes in those overlays.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
         <skipTests>true</skipTests>
         <test.excludes>**/*Test.java.bogus</test.excludes>
 
-        <kfs.version>2017-09-21</kfs.version>
+        <kfs.version>2017-10-19</kfs.version>
         <rice.version>2.6.1</rice.version>
         <cynergy.version>2.6.1</cynergy.version>
         <cynergy.version.suffix>current</cynergy.version.suffix>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -351,6 +351,10 @@
                     <groupId>log4j</groupId>
                     <artifactId>log4j</artifactId>
                 </exclusion>
+                <exclusion>
```
db483: How did you determine the exclusions in the pom.xml?

cah292: I looked at what changes were made to the financials POM for the 10-05 and 10-19 releases, and then added similar changes to our POM based on those. The exclusions here are based on those introduced with the two KualiCo patches for the related dependencies.

KualiCo also introduced the use of a new "duplicate-finder-maven-plugin", but that has not been included in this PR due to its potential impacts on our build. If you think we need to add that, I would suggest doing it in a separate user story.

db483: Thank you for explaining this Chad. The changes look good.

PR #521, opened 2018-02-14
KFSPTS-10917: Make Cred Group Code required on backup links

cah292:
The runtime-only changes are fairly minor for this, but I had to do a significant amount of unit test rework so that the affected code could be tested. If you think that the rework is too extensive for what this PR is trying to fix, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
@@ -81,8 +81,8 @@ public class AccountingXmlDocumentDownloadAttachmentServiceImpl extends Disposab
     
     protected boolean isLinkUrlValidForGroupCode(AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink, Collection<WebServiceCredential> credentials) {
         if (CollectionUtils.isEmpty(credentials)) {
-            LOG.debug("isLinkUrlValidForGroupCode, no credential values for group code, so is valid");
-            return true;
+            LOG.debug("isLinkUrlValidForGroupCode, no credential values for group code, so is invalid");
+            return false;
         } else {
             for (WebServiceCredential cred : credentials) {
                 if (isCredentialUsedForValidatingBackupLinkURL(cred) && isCredentialBaseURLInBackupLinkURL(accountingXmlDocumentBackupLink, cred)) {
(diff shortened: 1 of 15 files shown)


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -36,6 +36,11 @@
     @Override
     public Attachment createAttachmentFromBackupLink(Document document,
             AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink) throws IOException {
+        if (StringUtils.isBlank(accountingXmlDocumentBackupLink.getCredentialGroupCode())) {
+            LOG.error("createAttachmentFromBackupLink, the Credential Group Code is blank");
+            throw new IOException("Unable to download attachment with blank Credential Group Code: " + accountingXmlDocumentBackupLink.getLinkUrl());
+        }
+
```
jhulslander: I think we should also update the isLinkUrlValidForGroupCode.  If there are webservice credential group codes for for a group code, the function returns true.  I think iti should return false.  The LOG message in that code area should probably be an error now as well.

cah292: I want to clarify what you meant by your requested isLinkUrlValidForGroupCode change. Do you want the method to return false when a group code does NOT have any web service credentials?

jhulslander: Yes, I think that would be appropriate, and update the logging statement that already exists

cah292:
I've made updates to the code based on the initial feedback. This is now ready for review again.

I do have concerns about whether the URL check in the isCredentialBaseURLInBackupLinkURL() method is strict enough, but I think that's beyond the scope of this PR. Let me know if you think it would be worth creating another user story to look into that.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -76,8 +81,8 @@ public Attachment createAttachmentFromBackupLink(Document document,
     
     protected boolean isLinkUrlValidForGroupCode(AccountingXmlDocumentBackupLink accountingXmlDocumentBackupLink, Collection<WebServiceCredential> credentials) {
         if (CollectionUtils.isEmpty(credentials)) {
-            LOG.debug("isLinkUrlValidForGroupCode, no credential values for group code, so is valid");
-            return true;
+            LOG.debug("isLinkUrlValidForGroupCode, no credential values for group code, so is invalid");
```
jhulslander: should this be a LOG.error?

cah292: Good catch, I agree that it makes sense to change the log level of this statement. I will make that change.

cah292:
I've updated the logging statement as requested. This is ready for review again.

cah292:
retest this please

PR #527, opened 2018-02-23
KFSPTS-11135: PVEN bank address should not be required data from PMW but when provided its ISO country should translate to a single FIPS country.

nkimble:


src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -229,7 +229,7 @@ private boolean allPmwNewVendorIsoCountriesMapToSingleFipsCountry(PaymentWorksVe
             isoCountryCodeTranslatesToSingleFipsCountryCode(stgNewVendorRequestDetailToProcess.getRequestingCompanyTaxCountry(), errorMessages) &&
             isoCountryCodeTranslatesToSingleFipsCountryCode(stgNewVendorRequestDetailToProcess.getCorpAddressCountry(), errorMessages) &&
             isoCountryCodeTranslatesToSingleFipsCountryCode(stgNewVendorRequestDetailToProcess.getRemittanceAddressCountry(), errorMessages) &&
-            isoCountryCodeTranslatesToSingleFipsCountryCode(stgNewVendorRequestDetailToProcess.getBankAddressCountry(), errorMessages)) {
+            bankIsoCountryCodeIsNotProvidedOrTranslatesToSingleFipsCountryCode(stgNewVendorRequestDetailToProcess.getBankAddressCountry(), errorMessages)) {
```
db483: should allCountryChecksPassed be passed the result of bankIsoCountryCodeIsNotProvidedOrTranslatesToSingleFipsCountryCode?

nkimble: Short answer, Yes.
Explanation: Method "canPaymentWorksNewVendorRequestProcessingContinueForVendor" is the first one in this method call chain. It is performing the high level checks on the data received from PMW and stops processing for a vendor when the child methods it calls encounter a data condition that would later fail processing. What I am trying to do here is weed out bad data quickly and save processing cycles. In this case, should a vendor provide a bank address and the country that was provided in that address did not convert to a single FIPS country, processing would fail down stream. So, returning the result here stops processing on this vendor faster due to bad data.

PR #529, opened 2018-02-27
KFSPTS-11109 add user friendly message for xml error

db483:


src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -81,7 +82,7 @@ protected void processAccountingDocumentFromXml(String fileName) {
             LOG.info("processAccountingDocumentFromXml: Finished processing accounting document XML file: " + fileName);
         } catch (Exception e) {
             reportItem.setXmlSuccessfullyLoaded(false);
-            reportItem.setReportItemMessage(e.getMessage());
+            reportItem.setReportItemMessage(configurationService.getPropertyValueAsString(CuFPKeyConstants.REPORT_CREATE_ACCOUNTING_DOCUMENT_XML_PROCESING_ERROR));
```
cah292: Is this error message meant to replace the stack trace message, or to be in addition to the stack trace message?

db483: The next line will log the exception in the log. I am not sure that the report needs those details. What do you think?

cah292: That sounds good to me, but it might be a good idea to confirm with the functionals first. Their current JIRA comments mention adding the new message instead of replacing the old message; some additional clarity would be helpful.

db483:
Chad, I have updated the code per functional feedback. Please review again at your convenience. Thank you.

db483:
retest this please

src/main/java/edu/cornell/kfs/fp/CuFPKeyConstants.java
```diff
@@ -29,4 +29,5 @@
     public static final String REPORT_CREATE_ACCOUNTING_DOCUMENT_SHARED_DOCUMENT_TYPE = "report.create.accounting.document.shared.document.type";
     public static final String REPORT_CREATE_ACCOUNTING_DOCUMENT_SHARED_DOCUMENT_DESCRIPTION = "report.create.accounting.document.shared.document.description";
     public static final String REPORT_CREATE_ACCOUNTING_DOCUMENT_SHARED_DOCUMENT_EXPLANATION = "report.create.accounting.document.shared.document.explanation";
+    public static final String REPORT_CREATE_ACCOUNTING_DOCUMENT_XML_PROCESING_ERROR = "report.create.accounting.document.xml.processing.error";
```
cah292: Small typo in the name of this constant; should be "...PROCESSING..." instead of "...PROCESING..."

db483:
Thank you for catching that typo Chad. I have updated the code.

PR #531, opened 2018-02-28
 KFSPTS-11122 add sperate error messages for each type of attachment error; add test case for credential group code with blank value

db483:


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -34,34 +35,37 @@
...
-            Attachment attachment = attachmentService.createAttachment(document, uploadFileName, mimeType, fileSize,
-                    inputStream, attachmentType);
-            return attachment;
-        } else {
-            LOG.error("createAttachmentFromBackupLink, the form file is NULL");
-        }
-        
-        throw new IOException("Unable to download attachment: " + accountingXmlDocumentBackupLink.getLinkUrl());
+        } catch (IOException e) {
+            LOG.error("Unable to download attachment: " + accountingXmlDocumentBackupLink.getLinkUrl());
```
cah292: This logging statement should start with the name of the enclosing method, for consistency with how we've been setting up other such log messages. Also, I think this logging call should pass in the exception as the second argument, so that we can view its stack trace in the logs.

jhulslander:
I think downloadByteArray() also needs to be updated.  Where it throws an IOException, it should throw a validation exception that the URL isn't valid for the group code.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
db483: Thank you Chad, I will update the code accordingly.

db483:
Chad, 

Regarding the unit tests behavior, can we have that updated as part of another user story?

db483:
Jay,

Good point on the downloadByteArray() method, I will update that too. Thank you.

cah292:
Yes, the updates to the unit test behavior can be done in a different user story. Thanks!

db483:
I have updated the code based on code review suggestions. This is now ready for review again. Thank you!

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -72,7 +76,8 @@ public Attachment createAttachmentFromBackupLink(Document document,
                 return downloadByteArrayFromWebResource(accountingXmlDocumentBackupLink.getFileName(), request);
             } else {
                 LOG.error("downloadByteArray, the group code isn't valid for the link URL");
-                throw new IOException("The group code isn't valid for the link URL");
+                throw new ValidationException("The group code: " + accountingXmlDocumentBackupLink.getCredentialGroupCode()
+                                + " isn't valid for the link URL: " + accountingXmlDocumentBackupLink.getLinkUrl());
             }
         } catch (URISyntaxException e) {
             throw new IOException(e.getMessage());
```
cah292: Should this be changed to throw a ValidationException instead? Otherwise, the calling code won't print the message of the IOException when it catches it.

db483: This IO exception will be caught on line 65 and then logged:
LOG.error("createAttachmentFromBackupLink, Unable to download attachment: " + accountingXmlDocumentBackupLink.getLinkUrl(), e);

cah292: The line after that log statement will just throw a ValidationException with a generic "Unable to download attachment" message. Is that enough information for someone reading the report file? Or is a more descriptive message needed to indicate that the URL's syntax was invalid?

db483:
Chad, I have added the requested change. Please review again at your convenience. Thank you.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -34,34 +35,37 @@
...
+                            + " fileSize: " + fileSize);
+                }
+                InputStream inputStream = new ByteArrayInputStream(formFile);
+
+                Attachment attachment = attachmentService.createAttachment(document, uploadFileName, mimeType, fileSize, inputStream, attachmentType);
+                return attachment;
+
+            } else {
+                LOG.error("createAttachmentFromBackupLink, the form file is NULL");
+                throw new ValidationException("The form file is NULL: " + accountingXmlDocumentBackupLink.getLinkUrl());
```
cah292: Did you see Jay's comment on this line? (I received an email for it, but for some reason I don't see the comment in the GitHub UI.) Jay thought that the generic cannot-download message would be better for this spot, given that this is something users will see in the report. At the very least, should this be changed to mention something other than "form file" if that would confuse the users?

db483: hmm, I did not see it but good point. Would this work better: 
throw new ValidationException("Unable to download attachment: " + accountingXmlDocumentBackupLink.getLinkUrl());

cah292: Yes, I believe a message like that would work better in this spot.

db483: I have updated the code. I think the logged message should stay in it's current form. thank you.

jhulslander: :-)

PR #533, opened 2018-03-02
KFSPTS-11151 upgrade to the 12/14/2017 version of financials

db483:


db483:
Hi Chad,

Excellent question. We do not use email signing, that was a requirement that was implemented for IU. If the property is not present then it is considered false by default, so email signing is not used. The property is not needed in the config file in this case but we could add it with a false value. As far as I can tell from the existing properties we only have the properties that we use in the properties files. Please let me know if you think this should still be added. Thank you.

cah292:
Thanks for checking, Daniela. I'm fine with not adding config changes at this time, based on your analysis. If we end up needing changes in the future, then we can address that in a separate ticket.

PR #536, opened 2018-03-06
KFSPTS-10762 Prevent creating requisition from e-shop with duplicate account strings

ajd299:


src/test/java/edu/cornell/kfs/module/purap/fixture/CuB2BShoppingCartFixture.java
```diff
@@ -60,4 +60,14 @@ public CuB2BShoppingCart createB2BShoppingCart() {
 
 		return cart;
 	}
+
+	public CuB2BShoppingCart createB2BShoppingCartWithDuplicateItems() {
```
jhulslander: Instead of doing this, I would suggest making a new in this enum and a new item in B2BShoppingCartItemFixture that has duplicate accounts in.  That way you have an enum item that meats your testing needs.  This function may add confusion in the future.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +192,60 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+                if (!existingAccounts.contains(purApAccountingLine.toString())) {
+                    existingAccounts.add(purApAccountingLine.toString());
+                }
+                else {
+                    GlobalVariables.getMessageMap().putError(PurapConstants.ITEM_TAB_ERROR_PROPERTY, PurapKeyConstants.ERROR_ITEM_ACCOUNTING_NOT_UNIQUE,
+                            requisitionItem.getItemIdentifierString());
+                    String errorMessage = getDuplicateAccountErrorMessage(purApAccountingLine, requisitionItem);
+                    LOG.error(errorMessage);
+                    cuB2BShoppingErrorEmailService.sendDuplicateRequisitionAccountErrorEmail(requisitionDocument, errorMessage);
+                    return false;
```
jhulslander: We may have cases where more than one item have duplicate accounts used.  I think it would be a good idea to report on all cases of duplicate accounts, not just the first.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +192,60 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+        }
+
+        if (Collections.isEmpty(requisitions)) {
+            throw new B2BShoppingException("Invalid e-shop Request. No requisitions were created.");
         }
+
         return requisitions;
     }
 
+    public boolean checkRequisitionAccountsAreUnique(RequisitionDocument requisitionDocument) {
```
db483: Andrew,

Is this the same validation that occurs when a user manually creates a Requisition in KFS? Is there any way to plug in that validation here so that we can make sure the exact same validation occurs in both cases? Thank you.

ajd299: I believe the validation is in the base code PurchasingAccountsPayableUniqueAccountingStringsValidation.validate. I could injecting that validation bean into the CuB2bShoppingServiceImpl and passing in a dummy AttributedDocumentEvent parameter?

jhulslander: That should work.  Looks like the object is already registered in Spring: PurchasingAccountsPayable-uniqueAccountingStringsValidation .  We should switch to use this validation routine as Daniela suggested.

ajd299:
- Can you take a look at my last commit and see if that is what you had in mind for the fixtures? 
- I am using SpringContext.getBean to get the validator. Is that the best way to retrieve it? 
- I'm passing null into PurchasingAccountsPayableUniqueAccountingStringsValidation.validate. Is there a better alternative?

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +190,48 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+
+        List<RequisitionItem> requisitionItems = requisitionDocument.getItems();
+        boolean requisitionAccountsUnique = true;
+        for (RequisitionItem requisitionItem : requisitionItems) {
+
+            PurchasingAccountsPayableUniqueAccountingStringsValidation uniqueRequisitionAccountValidator =
```
jhulslander: I would move PurchasingAccountsPayableUniqueAccountingStringsValidation to a member variable, and use getters and setters.  In the getter, if the member variable is NULL, us springcontext to set it.  That way, it is easier to test

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +190,48 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+                    SpringContext.getBean(PurchasingAccountsPayableUniqueAccountingStringsValidation.class);
+            uniqueRequisitionAccountValidator.setItemForValidation(requisitionItem);
+
+            if (!uniqueRequisitionAccountValidator.validate(null)) {
```
jhulslander: instead of using a NULL here, can you create a real event object?  I am just thinking about avoiding future NPEs if kualico changes their code in PurchasingAccountsPayableUniqueAccountingStringsValidation

ajd299:
Please re-review as I have committed the following changes:
- Added a test for multiple items each containing multiple accounts
- Passing in AttributedRouteEvent to the validate method
- moved the validator to member variable with a getter/setter

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -7,9 +7,11 @@
 import java.util.List;
 import java.util.Map;
 
+import io.jsonwebtoken.lang.Collections;
```
jhulslander: I don't think you want this library.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +193,47 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+                purapService.saveDocumentNoValidation(req);
 
-            // add requisition to List
-            requisitions.add(req);
+                // add requisition to List
+                requisitions.add(req);
+            }
+        }
+
+        if (Collections.isEmpty(requisitions)) {
```
jhulslander: I think you want CollectionUtils for the empty check

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -183,15 +193,47 @@ public List createRequisitionsFromCxml(B2BShoppingCart message, Person user) thr
...
+        uniqueRequisitionAccountValidator = getUniqueRequisitionAccountValidator();
+
+        List<RequisitionItem> requisitionItems = requisitionDocument.getItems();
+        for (RequisitionItem requisitionItem : requisitionItems) {
+            uniqueRequisitionAccountValidator.setItemForValidation(requisitionItem);
+
+            AttributedDocumentEvent attributedDocumentEvent = new AttributedRouteDocumentEvent(requisitionDocument);
+            if (!uniqueRequisitionAccountValidator.validate(attributedDocumentEvent)) {
+                requisitionAccountsUnique = false;
+                handleDuplicateAccountError(requisitionDocument, requisitionItem);
```
jhulslander: it looks like if a requisition has multiple items with duplicate account conditions, the user will get multiple email.  I think we want to collect all the error messages, and send just one email per requisition.

src/test/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImplTest.java
```diff
@@ -36,6 +38,42 @@ public void testCreateRequisitionsFromCxml() throws Exception {
 		assertNotNull(requisitions);
 	}
 
+	public void testCreateRequisitionsFromCxmlWithDuplicateItems() throws Exception {
```
jhulslander: it looks like there are some formatting issues here

src/test/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImplTest.java
```diff
@@ -36,6 +38,42 @@ public void testCreateRequisitionsFromCxml() throws Exception {
...
+        List<CuRequisitionDocument> cuRequisitionDocuments = b2bShoppingService.createRequisitionsFromCxml(cart, UserNameFixture.ccs1.getPerson());
+        List<RequisitionItem> requisitionItems = cuRequisitionDocuments.get(0).getItems();
+        assertTrue("Expected more than one RequisitionItem", requisitionItems.size() > 1);
+
+        for (RequisitionItem requisitionItem : requisitionItems){
+            List<PurApAccountingLine> requisitionAccounts = requisitionItem.getSourceAccountingLines();
+            requisitionAccounts.add(requisitionAccounts.get(0));
+        }
+
+        assertFalse(b2bShoppingService.checkRequisitionAccountsAreUnique(cuRequisitionDocuments.get(0)));
```
jhulslander: Is there a way to update this test to be sure both validation errors are caught?

ajd299:
Daniela found that there may be a better way of solving this issue. We are re-investigating how exactly this issue is occurring. We need to determine why the validation is passing when a requisition is submitted in KFS (that originates in e-shop). It may have something to do with the sequence number of the accounting line.

jhulslander:
Since it looks like we are going to ask Kuali Co to fix this, and not do this solution.  We should close this PR.

PR #537, opened 2018-03-07
KFSPTS-11108: Add answers to six goods and services questions as a No…

nkimble:
…te on PVEN while restricting the size of all Notes on PVEN to max number of characters allowed.


nkimble:
This fix contains the original code from the previous goods and services fix with the addition of a check for the size of the note contents. Other places where that note character limit could have been breached were also refactored to prevent stack traces when those note items were entered.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksBatchUtilityServiceImpl.java
```diff
@@ -70,7 +72,21 @@ public String getFileContents(String fileName) {
     }
 
     @Override 
-    public Note createNote(String noteText) {
+    public KfsVendorDataWrapper createNoteRecordingAnyErrors(KfsVendorDataWrapper kfsVendorDataWrapper, String noteText, String noteErrorDescriptor) {
+        if (noteTextSizeIsWithinLimit(noteText)) {
+            Note note = createNote(noteText);
+            kfsVendorDataWrapper.getVendorNotes().add(note);
+        } else {
+            LOG.error("createNoteRecordingAnyErrors: The " + noteErrorDescriptor + " Note contained " + noteText.length() + "characters and it can only have " + PaymentWorksConstants.NOTE_TEXT_DEFAULT_MAX_LENGTH);
```
jhulslander: do you think it would be helpful to add the note text to the log?

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksBatchUtilityServiceImpl.java
```diff
@@ -70,7 +72,21 @@ public String getFileContents(String fileName) {
...
+        }
+        return kfsVendorDataWrapper;
+    }
+    
+    private boolean noteTextSizeIsWithinLimit(String noteText) {
+        return (noteText.length() <= PaymentWorksConstants.NOTE_TEXT_DEFAULT_MAX_LENGTH);
```
jhulslander: I am wondering if this should include a null check here as well.  StringUtils.length(noteText), will return zero if noteText is null, otherwise return the length of the string.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksBatchUtilityServiceImpl.java
nkimble: Some of the note context is restricted by user security so even though it would help in debugging, I do not believe we should include it.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksBatchUtilityServiceImpl.java
nkimble: Adding the null check.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksBatchUtilityServiceImpl.java
jhulslander: OK

nkimble:
Requested changes made. Ready for rereview.

nkimble:
retest this please

jhulslander:
Changes look OK, I will merge after the unit tests pass

PR #538, opened 2018-03-09
KFSPTS-10760: Add ability to create IB docs from XML

cah292:
IB documents have an extra items list that DI documents don't have, so I added such item lists to the create-acct-doc XML.

In addition, I did a fair amount of unit test refactoring for the create-acct-doc tests, to fix some bugs and to make it much simpler to add new document generators in the future.


src/main/java/edu/cornell/kfs/fp/batch/service/impl/InternalBillingDocumentGenerator.java
```diff
@@ -0,0 +1,63 @@
...
+public class InternalBillingDocumentGenerator
+        extends AccountingDocumentGeneratorBase<InternalBillingDocument> {
+
+    public InternalBillingDocumentGenerator() {
+        super();
+    }
+
+    public InternalBillingDocumentGenerator(
+            Supplier<Note> emptyNoteGenerator, Supplier<AdHocRoutePerson> emptyAdHocRoutePersonGenerator) {
+        super(emptyNoteGenerator, emptyAdHocRoutePersonGenerator);
```
jhulslander: since these constructors are just calling their super, are they needed here?

src/test/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentPojoTest.java
```diff
@@ -134,6 +155,16 @@ private void assertAccountingLineWasUnmarshalledCorrectly(
         assertEquals("Wrong line amount", expectedLine.getAmount(), actualLine.getAmount());
     }
 
+    private void assertItemWasUnmarshalledCorrectly(
+            AccountingXmlDocumentItem expectedItem, AccountingXmlDocumentItem actualItem) {
+        assertEquals("Wrong service date", expectedItem.getServiceDate(), actualItem.getServiceDate());
```
jhulslander: I noticed you are doing this check in two different test classes/  Do you think it would be better to implement equals in AccountingXmlDocumentItem?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDistributionOfIncomeAndExpenseDocumentGenerator.java
```diff
@@ -20,8 +23,17 @@ public CuDistributionOfIncomeAndExpenseDocumentGenerator(
...
+    public Class<? extends CuDistributionOfIncomeAndExpenseDocument> getDocumentClass() {
         return CuDistributionOfIncomeAndExpenseDocument.class;
     }
 
+    @Override
+    protected void performCustomValidation(
+            CuDistributionOfIncomeAndExpenseDocument document, AccountingXmlDocumentEntry documentEntry) {
+        super.performCustomValidation(document, documentEntry);
+        if (CollectionUtils.isNotEmpty(documentEntry.getItems())) {
+            throw new ValidationException("Item lines are not permitted on DI documents");
```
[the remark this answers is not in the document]
cah292: Good point, I'll check with Cathy on this in the main user story to get her input.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/InternalBillingDocumentGenerator.java
cah292: Yes, these are needed so that this class knows which superclass constructor to call in each case, and to allow for properly invoking the two-arg constructor when needed. Without an explicit constructor, only the superclass's no-arg constructor (if available and visible) will be used when invoking the implicit no-arg constructor on this class.

src/test/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentPojoTest.java
cah292: One test class is validating the XML POJO items, the other class is validating the InternalBillingItem BOs. There may not be a practical way to create some common property-comparing code for both properties, unless the XML POJO's class hierarchy is changed or some reflection-based comparison is used instead. (InternalBillingItem doesn't appear to have an equals() method override of its own.)

src/main/java/edu/cornell/kfs/fp/batch/service/impl/InternalBillingDocumentGenerator.java
jhulslander: OK, thanks for clarifying

src/test/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentPojoTest.java
jhulslander: Sorry about the confusion, I see the difference now

cah292:
I've updated the code so that if a particular document doesn't use a certain section of the XML, then that section will just get ignored when creating the document.

PR #539, opened 2018-03-12
 KFSPTS-6018 moving some utility classes from org.kuali to edu.cornell

jhulslander:


jhulslander:
The user story specified to the Cornell package. @db483  Do you have a take on Chad's note?

jhulslander:
I spoke with Daniela on this, we think moving the to Cornell package is fine for now.  

She will check with Bryan and Sarah on this, and will create another user story if they decide to move them to the rSmart package.

cah292:
Ok, thanks for checking!

PR #543, opened 2018-03-19
KFSPTS-11307 update unit tests to use Mockito instead of EasyMock

db483:


src/test/java/edu/cornell/kfs/sys/document/web/struts/CuFinancialMaintenanceDocumentActionTest.java
```diff
@@ -11,8 +11,8 @@
 import java.util.List;
 import java.util.Set;
 
-import org.easymock.EasyMock;
-import org.easymock.IMockBuilder;
+import javafx.beans.binding.When;
```
jhulslander: I don't think this import is used

db483: Good find, I will remove it.

db483:
I've made the requested change. This is ready for review again.

PR #545, opened 2018-03-21
KFSPTS-11309 KFS upgrade to Feb 15 2018

jhulslander:
This is ready to review now.  There are related PRs in non prod SQL and kfs-upgrades.  We need to wait to merge this for now.  Also, I ran into some problems with the SQL, that I believe I have corrected, but I want to run it fresh to make sure it is good.


src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
```diff
@@ -362,10 +358,10 @@ protected void renderHideDetails(PageContext pageContext, Tag parentTag) throws
     }
 
     /**
-     * Oy, the big one...this one actually renders instead of returning the HTML in a String. This is because it's kind of complex
-     * (and a likely target for future refactoring)
+     * Oy, the big one...this one actually renders instead of returning the HTML in a String. This is because it's kind 
+     * of complex(and a likely target for future refactoring)
```
cah292: This line doesn't appear to have all of the comment-formatting changes from the new KualiCo release.

src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
```diff
@@ -424,12 +427,16 @@ protected void renderUploadCell(PageContext pageContext, Tag parentTag) throws J
...
+                cancelButtonTag.setAlt(
+                        "Cancel import of " + accountingLineGroupDefinition.getGroupLabel() + " accounting lines");
+                cancelButtonTag.setTitle(
+                        "Cancel import of " + accountingLineGroupDefinition.getGroupLabel() + " accounting lines");
+                cancelButtonTag.setOnclick(
+                        getHideImportName() + "(\\\'" + showLink + "\\\',\\\'" + uploadDiv + "\\\');return false;");
                 cancelButtonTag.doStartTag();
                 cancelButtonTag.doEndTag();
 
+
```
cah292: I don't think this extra space is needed.

src/main/java/org/kuali/kfs/vnd/businessobject/VendorDetail.java
```diff
@@ -559,7 +543,7 @@ public String getVendorCommoditiesAsString() {
             } else {
                 first = false;
             }
-            sb.append(vcc.getCommodityCode().getCommodityDescription());
+            sb.append(vcc.getCommodityCode().getPurchasingCommodityCode());
```
cah292: I believe this is a line that we customized for the KFSPTS-3487 enhancement. This should not be changed back to the KualiCo version of the line unless the customization is no longer needed.

pom.xml
```diff
@@ -22,8 +22,8 @@
         <skipTests>true</skipTests>
         <test.excludes>**/*Test.java.bogus</test.excludes>
 
-        <kfs.version>2017-12-14</kfs.version>
-        <rice.version>2.6.1</rice.version>
+        <kfs.version>2018-02-15</kfs.version>
+        <rice.version>2.6.2</rice.version>
         <cynergy.version>2.6.1</cynergy.version>
```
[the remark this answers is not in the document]
jhulslander: We would need to change that after the cynergy upgrade is ready, correct?

src/main/java/org/kuali/kfs/gl/service/impl/ScrubberValidatorImpl.java
```diff
@@ -1,7 +1,7 @@
 /**
  * The Kuali Financial System, a comprehensive financial management system for higher education.
  *
- * Copyright 2005-2017 Kuali, Inc.
+ * Copyright 2005-2018 Kuali, Inc.
```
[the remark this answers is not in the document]
jhulslander: Let me try, I had originally removed it, but ran into some problems, but I don't think those problems were related to this.

src/main/java/org/kuali/kfs/gl/web/struts/BalanceInquiryLookupAction.java
```diff
@@ -416,16 +444,17 @@ protected Collection performMultipleValueLookup(MultipleValueLookupForm multiple
                 if (segmentedPropertyNames.contains(propertyName)) {
                     String propertyValue = StringUtils.replace(column.getPropertyValue(), ",", "");
                     
-                    // If there is a negative value, we need to convert it from (##.##) to -##.##
-                    if(StringUtils.contains(propertyValue, "(")) {
-                      propertyValue = StringUtils.replace(propertyValue, "(", "-");
-                      propertyValue = StringUtils.replace(propertyValue, ")", "");
+                    //Cornell Customization If there is a negative value, we need to convert it from (##.##) to -##.##
```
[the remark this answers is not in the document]
jhulslander: I will create a user story in the backlog to address this

src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
jhulslander: I double checked, and this is how the comment is setup, when I look at the KualiCo version  of GroupTitleLineRenderer.java from Feb 15

src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
jhulslander: removed

src/main/java/org/kuali/kfs/vnd/businessobject/VendorDetail.java
jhulslander: good catch, I shall change it back

src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
cah292: According to the GitHub history for the Feb 15 patch, this line has a space before the parentheses and a period at the end. Not sure what's causing the discrepancy you're seeing.

Given that this line is just a comment and not actual code, I'm okay with leaving it as-is if needed.

pom.xml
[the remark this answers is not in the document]
jhulslander: ok

src/main/java/org/kuali/kfs/sys/document/web/renderers/GroupTitleLineRenderer.java
jhulslander: interesting, I am using eclipse to open the class file from 2/15 jar

jhulslander:
I implemented some of the changes node in the code review

pom.xml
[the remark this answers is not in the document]
cah292: I've pushed my current 2.6.2 changes to the KFSPTS-11127 branches of the relevant Cynergy projects, if you want to start testing the use of Cynergy 2.6.2 by building and installing the branch code locally.

Also, now that this cu-kfs PR is available, I should be able to move forward on my remaining Cynergy 2.6.2 local testing so that we can get that wrapped up as well.

jhulslander:
Thanks Chad, please also take a look at the PR for the SQL changes.  I kept in the database change log insert statements

cah292:
While reviewing the script report PR, I noticed that we may be able to remove our CuPaymentDetailInquirable override, due to a related financials fix from this release. Please either remove that override as part of this work, or create a follow-up ticket to potentially remove it.

jhulslander:
I investigated removing CuPaymentDetailInquirable.  We have a test class CuPaymentDetailInquirableTest.  KualiCo base code doesn't pass all the test cases.  If memory serves, testBlankCountryCodeIsNotConverted failed, but the other succeed.  So I think removing CuPaymentDetailInquirable would take some further analysis to determine if the test case testBlankCountryCodeIsNotConverted is still required.  Please let me know if you think we want a user story to do this further investigation.

cah292:
Based on the updated code, it looks like KualiCo is doing a non-empty check, whereas our fix is doing a non-blank check. That could explain why that specific test case isn't working in base code.

Deferring the potential removal of that fix to another user story is fine with me, if you would prefer to do it that way.

cah292:
Thanks for creating the follow-up ticket, Jay!

Once we get Cynergy 2.6.2 built and deployed to Nexus, then we can re-run this PR's unit tests and get things merged if they pass.

cah292:
retest this please

PR #554, opened 2018-04-11
KFSPTS-11657: PMW changes to track approve-disapprove action on PVEN.…

nkimble:
… Code is the fix for the reversion performed in KFSPTS-11644.

### DIFF
--- a/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
@@ -48,7 +48,7 @@ public class CuVendorMaintainableImpl extends VendorMaintainableImpl {
     private static final String PROC_METHODS_MULTISELECT_FIELD_NAME = "extension.procurementMethodsArray";
     private static final String MULTISELECT_FIELD_PATH_PREFIX = "dataObject.";
     
-    protected PaymentWorksBatchUtilityService paymentWorksBatchUtilityService;
+    protected transient PaymentWorksBatchUtilityService paymentWorksBatchUtilityService;
     
     @Override
     public void saveBusinessObject() {
--- a/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
@@ -272,7 +272,7 @@ public class CuVendorMaintainableImpl extends VendorMaintainableImpl {
     }
     
     private boolean recordPvenRouteStatusChangeForPaymentWorksIsEnabled() {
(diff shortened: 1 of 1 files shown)


nkimble:
The changes in this file are exactly the same as what was in the original commit performed in JIRA KFSPTS-9545 with the following difference: 
~Line 242 "super.doRouteStatusChange(header);" was removed.

cah292:
Did you mean the this.saveBusinessObject() call was removed? That's what I'm seeing in the older version of the feature.

nkimble:
Yes.You are correct. 
"this" instead of "super" as stated in my previous comment. 
Sorry for the confusion.

src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
```diff
@@ -225,4 +229,81 @@ private void setupMultiselectField(MaintenanceDocument document, MaintenanceDocu
...
+            processKfsVendorEditActionApproval(kfsDocumentNumber, vendorDetail);
+        }
+    }
+    
+    private boolean isExistingPaymentWorksVendor(String kfsDocumentNumber) {
+        return getPaymentWorksBatchUtilityService().foundExistingPaymentWorksVendorByKfsDocumentNumber(kfsDocumentNumber);
+    }
+    
+    private boolean recordPvenRouteStatusChangeForPaymentWorksIsEnabled() {
+        return (getPaymentWorksBatchUtilityService().retrievePaymentWorksParameterValue(PaymentWorksParameterConstants.PMW_INTEGRATION_IS_ACTIVE_IND).equalsIgnoreCase(KFSConstants.ParameterValues.YES) ? true : false);
```
cah292: You're already getting a boolean result from this check, so adding a ternary operation of "? true : false" is redundant.

nkimble: You are correct. Removing the ternary operator.

nkimble:
Suggested changes has been made. This is ready to be reviewed again.

src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
```diff
@@ -46,6 +48,8 @@
     private static final String PROC_METHODS_MULTISELECT_FIELD_NAME = "extension.procurementMethodsArray";
     private static final String MULTISELECT_FIELD_PATH_PREFIX = "dataObject.";
     
+    protected PaymentWorksBatchUtilityService paymentWorksBatchUtilityService;
```
cah292: Should this service be declared as transient?

nkimble:
Suggested change was made for adding transient. This is ready t be reviewed again.

PR #555, opened 2018-04-12
KFSPTS 11582 Increase the Length of the Account Extension Fields in the Lookup Results 

ajd299:
Please let me know if there is a number that would be better for the length.


ajd299:
Increased the maxLength to 2048.

PR #557, opened 2018-04-16
KFSPTS-11710: Filter invalid namespace prefixes from CXML

cah292:
This is just a temporary workaround for an issue that was discovered during the 03/15 patch validation. We will tentatively be removing it in the near future, once we've corrected the CU application that is adding the bad namespace prefixes.

Please make sure the unit tests pass, since I modified some existing CXML testing code to be compatible with micro-testing.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
@@ -1689,6 +1689,7 @@ public class CuElectronicInvoiceHelperServiceImpl extends ElectronicInvoiceHelpe
     /**
      * Temporary override to work around an issue where input files with unsupported "xmlns:*" attributes
      * could not be parsed successfully, but otherwise behaves the same as in the superclass.
+     * The only change was an added call to a new "removeUnsupportedXmlnsAttributes" method.
      * 
      * @see org.kuali.kfs.module.purap.service.impl.ElectronicInvoiceHelperServiceImpl#addNamespaceDefinition(
      * org.kuali.kfs.module.purap.businessobject.ElectronicInvoiceLoad, java.io.File)
@@ -1770,8 +1771,9 @@ public class CuElectronicInvoiceHelperServiceImpl extends ElectronicInvoiceHelpe
     }
 
     protected boolean isUnsupportedXmlnsAttribute(Attr attribute) {
-        return StringUtils.startsWith(attribute.getName(), XMLNS_ATTRIBUTE_PREFIX)
-                && !ALLOWED_XMLNS_ATTRIBUTES_PATTERN.matcher(attribute.getName()).matches();
+        String attributeName = StringUtils.lowerCase(attribute.getName());
+        return StringUtils.startsWith(attributeName, XMLNS_ATTRIBUTE_PREFIX)
+                && !ALLOWED_XMLNS_ATTRIBUTES_PATTERN.matcher(attributeName).matches();
     }
 
 }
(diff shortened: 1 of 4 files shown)


src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -1666,4 +1686,92 @@ protected void addRejectReasonsToNote(String rejectReasons, ElectronicInvoiceRej
         }
     }
 
+    /**
+     * Temporary override to work around an issue where input files with unsupported "xmlns:*" attributes
+     * could not be parsed successfully, but otherwise behaves the same as in the superclass.
```
jhulslander: It looks like the function call to removeUnsupportedXmlnsAttributes is the only thing different from the parent's implementation.  I would suggest noting that in the comment.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -1666,4 +1686,92 @@ protected void addRejectReasonsToNote(String rejectReasons, ElectronicInvoiceRej
...
+            attributes[i] = (Attr) attributesMap.item(i);
+        }
+        
+        Arrays.stream(attributes)
+                .filter(this::isUnsupportedXmlnsAttribute)
+                .forEach(element::removeAttributeNode);
+    }
+
+    protected boolean isUnsupportedXmlnsAttribute(Attr attribute) {
+        return StringUtils.startsWith(attribute.getName(), XMLNS_ATTRIBUTE_PREFIX)
```
jhulslander: should this be starts with ignore case?

src/test/java/edu/cornell/kfs/module/purap/fixture/CuElectronicInvoiceHelperServiceFixture.java
```diff
@@ -48,9 +70,18 @@ private static String getXMLChunk(){
...
                             append(getCXMLDate(true)).append("\n");
             
-            xmlChunk.append("version=\"1.2.014\" xml:lang=\"en\"\n"); 
-            xmlChunk.append("xmlns=\"http://www.kuali.org/kfs/purap/electronicInvoice\"\n"); 
-            xmlChunk.append("xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\">\n");
+            xmlChunk.append("version=\"1.2.014\" xml:lang=\"en\"\n");
+            
+            if (StringUtils.isNotBlank(xmlnsAttributeName)) {
+                xmlChunk.append(String.format(ATTRIBUTE_NAME_VALUE_FORMAT, xmlnsAttributeName, CuPurapTestConstants.EINVOICE_NAMESPACE_URL));
+                xmlChunk.append("\n");
```
jhulslander: this is in a constant kfsconstants.newline

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
cah292: Good point, I will update the comment accordingly.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
cah292: Good catch, I will update this to be case-insensitive.

src/test/java/edu/cornell/kfs/module/purap/fixture/CuElectronicInvoiceHelperServiceFixture.java
cah292: Ok, I will update this to use the constant instead.

cah292:
I've made the changes requested by the original feedback. This is now ready for review again.

PR #558, opened 2018-04-17
KFSPTS-11691 upgrade to 4/12/2018 version of financials

db483:


db483:
retest this please

db483:
retest this please

db483:
Good question Chad. Java Melody is already disabled in our master branch. It has been enabled in develop a while back to test the cache clearing but at this point we can just remove that change from develop as I do not expect we will be able to use java melody for the cache clearing functionality. I will create a separate user story to revert this change from develop.

PR #562, opened 2018-04-18
KFSPTS-11657: Track PVEN approve-disapprove actions in KFS PMW stagingg table for newly created vendors.

nkimble:
This change fixes the optimistic lock exception encountered when data is entered in the PVEN header explanation field while also tracking the header generated id - vendor detailed assigned id that is created when a PVEN edoc is approved in KFS.

### DIFF
--- a/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
@@ -266,6 +266,9 @@ public class CuVendorMaintainableImpl extends VendorMaintainableImpl {
             LOG.debug("performPaymentWorksApprovalProcessingForVendor: edit action");
             processKfsVendorEditActionApproval(kfsDocumentNumber, vendorDetail);
         }
+        else {
+            LOG.info("performPaymentWorksApprovalProcessingForVendor: No PVEN Approval processing performed for PMW because not existing PMW vendor -OR- KFS PVEN maintenance action not currently being tracked for PMW.");
+        }
     }
     
     private boolean isMaintenanceActionNewOrNewWithExisting(String maintenanceAction) {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
```diff
@@ -225,4 +229,84 @@ private void setupMultiselectField(MaintenanceDocument document, MaintenanceDocu
...
+        }
+        else if (isMaintenanceActionNewOrNewWithExisting(maintenanceAction)) {
+            LOG.debug("performPaymentWorksApprovalProcessingForVendor: new action resulting from AVf or hand keying");
+            this.saveBusinessObject();
+            processKfsVendorNewActionApproval(kfsDocumentNumber, vendorDetail);
+        }
+        else if (StringUtils.equals(KRADConstants.MAINTENANCE_EDIT_ACTION, maintenanceAction)) {
+            LOG.debug("performPaymentWorksApprovalProcessingForVendor: edit action");
+            processKfsVendorEditActionApproval(kfsDocumentNumber, vendorDetail);
+        }
```
cah292: Is a plain "else" block needed here to at least log the cases where no PaymentWorks processing is needed?

src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
```diff
@@ -225,4 +229,84 @@ private void setupMultiselectField(MaintenanceDocument document, MaintenanceDocu
...
+    }
+    
+    private void processKfsVendorEditActionApproval(String kfsDocumentNumber, VendorDetail vendorDetail) {
+        getPaymentWorksBatchUtilityService().registerKfsPvenApprovalForKfsEditedVendor(kfsDocumentNumber, vendorDetail);
+    }
+    
+    private void performPaymentWorksDisapprovalCancelProcessingForVendor(VendorDetail vendorDetail, String kfsDocumentNumber, String maintenanceAction) {
+        if (isExistingPaymentWorksVendor(kfsDocumentNumber) && (isMaintenanceActionNewOrNewWithExisting(maintenanceAction))) {
+            getPaymentWorksBatchUtilityService().registerKfsPvenDisapprovalForExistingPaymentWorksVendor(kfsDocumentNumber, vendorDetail);
+        }
```
cah292: Is a plain "else" block also needed here to at least log the cases where no PaymentWorks processing is needed?

src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
nkimble: Good suggestion. I have added as an info message since we would want to see it in the log when it is encountered and this condition should not be hit much so it should not generate a lot of noise.

src/main/java/edu/cornell/kfs/vnd/document/CuVendorMaintainableImpl.java
nkimble: Same message as above.

nkimble:
Code review suggestions implemented and ready for re-review.

PR #564, opened 2018-04-25
KFSPTS-11759 KFSPTS-10229 customize cgb/ar pages to use CuAward instead of Award

db483:
This is a fix we have had in the CGB environment and is needed in develop for turning on CGB in DEV.

### DIFF
--- a/src/main/webapp/WEB-INF/tags/module/ar/contractsGrantsInvoiceSummaryResultContent.tag
+++ b/src/main/webapp/WEB-INF/tags/module/ar/contractsGrantsInvoiceSummaryResultContent.tag
@@ -2,7 +2,7 @@
 
     The Kuali Financial System, a comprehensive financial management system for higher education.
 
-    Copyright 2005-2017 Kuali, Inc.
+    Copyright 2005-2018 Kuali, Inc.
 
     This program is free software: you can redistribute it and/or modify
     it under the terms of the GNU Affero General Public License as published by
--- a/src/main/webapp/WEB-INF/tags/module/ar/contractsGrantsInvoiceSummaryResultContent.tag
+++ b/src/main/webapp/WEB-INF/tags/module/ar/contractsGrantsInvoiceSummaryResultContent.tag
@@ -1,0 +1,45 @@
+<%--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
+    Copyright 2005-2017 Kuali, Inc.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


src/main/webapp/WEB-INF/tags/module/ar/contractsGrantsInvoiceSummaryResultContent.tag
```diff
@@ -0,0 +1,45 @@
+<%--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
+    Copyright 2005-2017 Kuali, Inc.
```
cah292: License header is out-of-date.

src/main/webapp/jsp/module/ar/FederalFinancialReport.jsp
```diff
@@ -0,0 +1,116 @@
+<%--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
+    Copyright 2005-2017 Kuali, Inc.
```
cah292: License header is out-of-date.

src/main/webapp/jsp/module/ar/FederalFinancialReport.jsp
```diff
@@ -0,0 +1,116 @@
...
+
+            <tr>
+                <th>
+                    <label>
+                        <kul:htmlAttributeLabel attributeEntry="${award.proposalNumber}" readOnly="true"/>
+                    </label>
+                </th>
+                <td>
+                    <kul:htmlControlAttribute attributeEntry="${award.proposalNumber}" property="proposalNumber"/>
+                    <kul:lookup boClassName="org.kuali.kfs.module.cg.businessobject.Award"/>
```
cah292: Should this lookup reference CuAward as well?

db483: I've checked and the lookup works fine. We also did not have a report of a problem with this. I am not sure it needs to be updated.

db483:
Looks like the license headers have changed since this change was put in place. I have updated them. Thank you!

src/main/webapp/jsp/module/ar/FederalFinancialReport.jsp
```diff
@@ -0,0 +1,116 @@
...
+          headerTitle="Federal Financial Report Generation"
+          docTitle="Federal Financial Report Generation" renderMultipart="true"
+          transactionalDocument="false" htmlFormAction="arFederalFinancialReport"
+          errorKey="foo">
+
+    <div id="lookup">
+        <div class="main-panel">
+            <div class="headerarea-small"></div>
+            <table class="standard" style="margin: 20px auto 0 auto;" summary="Federal Financial Report Generation"
+            ">
```
cah292: This quotation mark doesn't have a corresponding match. The base file from KualiCo already has this problem, so I'm fine with moving this forward as-is and then we can create a ticket to have KualiCo fix it.

PR #565, opened 2018-04-25
KFSPTS-10559: Add PMW token refresh feature

cah292:
I've added Nancy as a co-reviewer on this. Please let me know if this feature does not conform to the requirements. I also moved the PaymentWorks URL property to a web service credential; please let me know if that property should not be moved there.

I did manage to find a way to unit test this new code. I originally tried using Jersey's testing framework, but I couldn't get it to work as expected, so I went with the Apache HttpClient test classes instead (which are also used by a unit test in cornell_customizations). If you notice anything in the unit test setup that could pose problems for our other tests, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
+++ b/src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
@@ -105,19 +105,6 @@ public class PaymentWorksConstants {
         public static final String NEW_VENDOR_REQUESTS_PAYEE_ACH_REPORT_NAME = "New Vendor Requests Create KFS Payee ACH Account";
     }
     
-    public static final class PaymentWorksCredentialKeys {
-        public static final String PAYMENTWORKS_API_URL = "paymentworks.api.url";
-        public static final String PAYMENTWORKS_USER_ID = "paymentworks.user.id";
-        public static final String PAYMENTWORKS_AUTHORIZATION_TOKEN = "paymentworks.authorization.token";
-    }
-    
-    public static final class PaymentWorksTokenRefreshConstants {
-        public static final String STATUS_FIELD = "status";
-        public static final String AUTH_TOKEN_FIELD = "auth_token";
-        public static final String DETAIL_FIELD = "detail";
-        public static final String STATUS_OK = "ok";
-    }
-    
     public enum PaymentWorksNewVendorRequestStatusType {
         PENDING(0, "0", "Pending"),
         APPROVED(1, "1", "Approved"),
--- a/src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
+++ b/src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
@@ -110,6 +110,13 @@ public class PaymentWorksConstants {
         public static final String PAYMENTWORKS_AUTHORIZATION_TOKEN = "paymentworks.authorization.token";
     }
     
+    public static final class PaymentWorksTokenRefreshConstants {
+        public static final String STATUS_FIELD = "status";
+        public static final String AUTH_TOKEN_FIELD = "auth_token";
+        public static final String DETAIL_FIELD = "detail";
+        public static final String STATUS_OK = "ok";
+    }
+    
     public enum PaymentWorksNewVendorRequestStatusType {
         PENDING(0, "0", "Pending"),
         APPROVED(1, "1", "Approved"),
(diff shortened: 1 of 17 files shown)


src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -28,17 +34,18 @@
...
-
 public class PaymentWorksWebServiceCallsServiceImpl implements PaymentWorksWebServiceCallsService, Serializable {
     private static final long serialVersionUID = -4282596886353845280L;
     private static final org.apache.log4j.Logger LOG = org.apache.log4j.Logger.getLogger(PaymentWorksWebServiceCallsServiceImpl.class);
-    
-    private String paymentWorksUrl;
-    private String paymentWorksAuthorizationToken;
+
+    private static final String REFRESH_TOKEN_URL_FORMAT = "%susers/%s/refresh_auth_token/";
+    private static final String EMPTY_JSON_WRAPPER = "{}";
```
nkimble: Is there a reason why these two constants are defined here instead of in PaymentWorksWebServiceConstants?

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
```diff
@@ -104,6 +105,19 @@
         public static final String NEW_VENDOR_REQUESTS_PAYEE_ACH_REPORT_NAME = "New Vendor Requests Create KFS Payee ACH Account";
     }
     
+    public static final class PaymentWorksCredentialKeys {
+        public static final String PAYMENTWORKS_API_URL = "paymentworks.api.url";
+        public static final String PAYMENTWORKS_USER_ID = "paymentworks.user.id";
+        public static final String PAYMENTWORKS_AUTHORIZATION_TOKEN = "paymentworks.authorization.token";
+    }
```
nkimble: Wouldn't PaymentWorksWebServiceConstants be a better place to have this class defined as these items pertain to that functionality?

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
```diff
@@ -104,6 +105,19 @@
...
+    
+    public static final class PaymentWorksTokenRefreshConstants {
+        public static final String STATUS_FIELD = "status";
+        public static final String AUTH_TOKEN_FIELD = "auth_token";
+        public static final String DETAIL_FIELD = "detail";
+        public static final String STATUS_OK = "ok";
+    }
```
nkimble: Same as above, wouldn't PaymentWorksWebServiceConstants be a better location for these items?

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -261,21 +268,104 @@ private Invocation buildXmlClientRequest(Client client, URI uri) {
...
+        }
+        
+        JsonNode statusNode = rootNode.findValue(PaymentWorksTokenRefreshConstants.STATUS_FIELD);
+        if (ObjectUtils.isNull(statusNode)) {
+            LOG.error("checkForSuccessfulTokenRefreshStatus(): Did not receive a refresh status from PaymentWorks");
+            throw new RuntimeException("Token refresh failed: Did not receive a refresh status from PaymentWorks");
+        } else if (!StringUtils.equalsIgnoreCase(PaymentWorksTokenRefreshConstants.STATUS_OK, statusNode.textValue())) {
+            LOG.error("checkForSuccessfulTokenRefreshStatus(): Unexpected status from PaymentWorks response: " + statusNode.textValue());
+            throw new RuntimeException("Token refresh failed: Received an unexpected refresh status from PaymentWorks");
+        }
```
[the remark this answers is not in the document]
cah292: Yes, that sounds like a good idea. I will add that.

nkimble:
Daniela,
Yes, the approach of having the SQL insert the user ID and auth token should mitigate that problem. Note: We would not want that SQL in the repository and the SQL would need to be generated for each environment prior to the merge. Also, the first run of both batch jobs should be monitored by a developer to ensure successful execution after this change goes in.
Thanks for the suggestion.

cah292:
I have implemented the requested changes. This is ready for review again.

PR #567, opened 2018-04-30
PaymentWorksVendor Business Object xml changes to load DD

ajd299:
- Turned Load DataDictionary On in PaymentWorks Spring config
- Added property name to 4 attributes
- Renamed 2 attributes
- Renamed 7 inquiry entries, removed 4 inquiry entries


ajd299:
@nkimble please review these changes and see if they are compatible with what we are currently working on for PaymentWorks. After loading these changes (in temp and localdev) I was able to see Paymentworks Vendor and PaymentWorks Field Mapping in the Menu and use lookups on both while impersonating kme44.

nkimble:
Andrew,
I want to make sure these changes will run with the code presently in KFS-DEV and I am not able to do prove that out locally.  I think these changes are out of sync with the token refresh changes that have been applied to KFS-DEV. Could you please close this review, rebase the code with the develop branch, and then create a new review?   Thanks.

PR #574, opened 2018-05-14
KFSPTS-11615 remove BRs for customer name, change customer number to be numeric

db483:


src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuCustomerLoadServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import org.kuali.kfs.module.ar.batch.service.impl.CustomerLoadServiceImpl;
+
+public class CuCustomerLoadServiceImpl extends CustomerLoadServiceImpl {
+    
+    @Override
+    protected boolean validateSingle(MaintenanceDocument maintDoc, CustomerLoadBatchErrors batchErrors, String customerName) {
+        boolean result = true;
+
+        //  get an instance of the business rule
+        CuCustomerRule rule = new CuCustomerRule();
```
jhulslander: Why do we need CuCustomerLoadServiceImpl?  Looks like the only difference is we are using CuCustomerRule instead of CustomerRule.  Since CuCustomerRule extends CustomerRule, it should caste correctly in the parent, right?

src/main/java/edu/cornell/kfs/module/ar/document/validation/CuCustomerRule.java
```diff
@@ -0,0 +1,12 @@
+package edu.cornell.kfs.module.ar.document.validation;
+
+import org.kuali.kfs.module.ar.document.validation.impl.CustomerRule;
+
+public class CuCustomerRule extends CustomerRule {
+    
+    @Override
+    public boolean checkNameIsValidLength(String customerName) {
+        return true;
```
jhulslander: Should we check the max length against the data dictionary for the name field?

src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuCustomerLoadServiceImpl.java
db483: Since the BR class is not retrieved from DD I am not sure how it would know to cast to CuCustomerRule. Perhaps I am missing something?

src/main/java/edu/cornell/kfs/module/ar/document/validation/CuCustomerRule.java
db483: Max length is already checked with the DD checks and is not needed in this method.

jhulslander:
Concerning about my note about not needing a class, you're right, I was confused when I posted that comment.  I think this is good to merge. Since you made a change since my initial review, please let me know when you feel this is ready again.

db483:
Thank you Jay. This is ready to be merged.

PR #578, opened 2018-05-17
KFSPTS-11604: Add BA docs to create-acct-doc feature

cah292:


cah292:
Please let me know if you think this should be reassigned to (or co-reviewed by) another developer who is more familiar with the create-acct-doc feature that this PR updates.

db483:
Chad,

I have reviewed the code and changes look good to me.

Please feel free to assign this for review to a second reviewer that is more familiar with the batch job as well.

Thank you.

cah292:
I've added Jay as a co-reviewer on this.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuBudgetAdjustmentDocumentGenerator.java
```diff
@@ -0,0 +1,96 @@
...
+        if (ObjectUtils.isNull(fiscalYear)) {
+            throw new RuntimeException("Fiscal year cannot be null");
+        }
+        
+        @SuppressWarnings("unchecked")
+        List<FiscalYearFunctionControl> allowedYears = fiscalYearFunctionControlService.getBudgetAdjustmentAllowedYears();
+        boolean fiscalYearAllowed = allowedYears.stream()
+                .anyMatch((fyFunctionControl) -> fiscalYear.equals(fyFunctionControl.getUniversityFiscalYear()));
+        if (!fiscalYearAllowed) {
+            throw new RuntimeException("Budget Adjustments are not allowed for fiscal year: " + fiscalYear);
```
jhulslander: Should this be a ValidationException, so it can be handled by the report correctly?

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentAccountingLine.java
```diff
@@ -54,6 +56,58 @@
     @XmlJavaTypeAdapter(KualiDecimalXmlAdapter.class)
     protected KualiDecimal amount;
 
+    @XmlElement(name = "base_amount", namespace = StringUtils.EMPTY, required = false)
```
jhulslander: I'm wondering if we should add currentBudgetAdjustmentAmount?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuBudgetAdjustmentDocumentGenerator.java
cah292: Ok, I will update the exception type accordingly.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentAccountingLine.java
cah292: I was trying to reuse the existing POJO "amount" field for holding the value of the current BA amount field. And according to the code in BudgetAdjustmentDocument, it seems the regular amount is set to the value of the current BA amount when preparing to take action on the document.

If you think the naming conventions would be better if we had a "current_amount" field instead, please let me know.

jhulslander: I don't have a strong opinion, but wanted to raise the concern.  Since BudgetAdjustmentDocument is reusing amount in the background, I am fine leaving this as is

cah292:
I've made some minor updates based on the recent feedback. This is now ready for review again.

jhulslander:
retest this please

PR #582, opened 2018-06-01
 KFSPTS-9100 new document for multiple object code activate/inactivate

jhulslander:
This is NOT ready to be merged, just creating a PR to help my own review


jhulslander:
Daniela, this isn't ready to be merged nor is it ready for a full review.  I did want to get your feedback on a couple of things.

1) Referring to my changes CuFinancialMaintenanceDocumentAction.java.  I think it makes sense to have a new action class CuObjectCodeActivationGlobalDocumentAction which would extend CuFinancialMaintenanceDocumentAction.  However, I couldn't get it to work, I am not sure what I was missing.  
The things I did were:
1.a create the action class, add a new action element to institution-struts-config, 
1.b update the doc type XML to include a docHandler element using the URL, 
1.c ingest the new doc type XML.  
After that, the menu still referenced the old maintainencedocument.do URL, and not the new one.  In the doc search, newly created documents had the new doc handlers, but KFS erred trying  to open those new documents, it is like the new struts action wasn't being respected.  I am not sure how to do this.

2) I created a new object CuObjectCodeGlobalSearch, the OJB mappings to the balance table needs to be adjusted.  I suspect the problem lies with the join between the two objects isn't based on an actual database constrained key.

I just wanted to get your take on my approach.

jhulslander:
I resolved the problem with the join, so I just would like some feedback about the the struts action.

PR #369, opened 2017-04-17
KFSPTS-8335 setting employee flag on payments

jhulslander:

PR #587, opened 2018-06-08
KFSPTS-12308: Changes needed to keep existing PMW code working due to PMW non-unique tax id number change.

nkimble:


nkimble:
There is a corresponding SQL PR (#369) that needs to go at the same time as this PR.
Please just review this code and do NOT merge until we touch base on Monday 6/11. Thanks

PR #588, opened 2018-06-12
KFSPTS-12220: Fixes to Error handling for ContractsGrantsInvoiceCreateDocument Job

ajd299:


src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuContractsGrantsInvoiceCreateDocumentServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.sql.Date;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+public class CuContractsGrantsInvoiceCreateDocumentServiceImpl extends ContractsGrantsInvoiceCreateDocumentServiceImpl {
+
+    @Override
+    protected void storeValidationErrors(Map<ContractsAndGrantsBillingAward, List<String>> invalidGroup, Collection<ContractsGrantsInvoiceDocumentErrorLog> contractsGrantsInvoiceDocumentErrorLogs, String creationProcessTypeCode) {
```
cah292: I think it would be helpful for this method have a Javadoc comment that explains what was changed, given that this override appears to be almost identical to the original except for a few lines.

src/main/resources/edu/cornell/kfs/module/ar/cu-spring-ar.xml
```diff
@@ -38,4 +38,70 @@
 	<bean id="customerLoadService" parent="customerLoadService-parentBean"
 		class="edu.cornell.kfs.module.ar.batch.service.impl.CuCustomerLoadServiceImpl" />
 
+	<bean id="contractsGrantsInvoiceCreateDocumentService" parent="contractsGrantsInvoiceCreateDocumentService-parentBean"/>
+	<bean id="contractsGrantsInvoiceCreateDocumentService-parentBean" class="edu.cornell.kfs.module.ar.batch.service.impl.CuContractsGrantsInvoiceCreateDocumentServiceImpl" abstract="true">
```
cah292: Based on your other code changes, I don't think you need to override the parent bean. If all you need to change is the "class" attribute, then you can override it on the main "contractsGrantsInvoiceCreateDocumentService" bean so that you don't need to modify the parent bean.

ajd299:
I think I have addressed the comments. Please re-review

PR #590, opened 2018-06-14
KFSPTS-12337 upgrade to 6/7/2018 version of financials

db483:


src/main/java/org/kuali/kfs/module/purap/document/PurchaseOrderDocument.java
```diff
@@ -1797,8 +1780,8 @@ public PersistableBusinessObject getNoteTarget() {
         PurchaseOrderDocument oldest = docDao.findByDocumentHeaderId(PurchaseOrderDocument.class, purchaseOrderDao.getOldestPurchaseOrderDocumentNumber(this.getPurapDocumentIdentifier()));
 
         //KFSMI-9746: added this for null safe checking.
-        if(oldest != null){
-            return oldest;
+        if (oldest != null) {
+            return oldest.getDocumentHeader();
```
cah292: I believe we've purposely customized this to return the reference to "oldest" instead of its document header. Thus, I don't think the document header should be returned here, since that could interfere with notes on our existing POs.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherAccountingLineTotalsValidation.java
```diff
@@ -57,12 +57,15 @@ public boolean validate(AttributedDocumentEvent event) {
...
-                if (persistedDocument.getDocumentHeader().getWorkflowDocument().isSaved() && persistedDocument.getDisbVchrCheckTotalAmount().isZero()) {
+                if (persistedDocument.getDocumentHeader().getWorkflowDocument().isSaved() && persistedDocument
+                        .getDisbVchrCheckTotalAmount().isZero()) {
                     return true;
                 }
 
                 // check total cannot decrease
-                if (!persistedDocument.getDocumentHeader().getWorkflowDocument().isCompletionRequested() && (!persistedDocument.getDisbVchrCheckTotalAmount().equals(dvDocument.getDisbVchrCheckTotalAmount()))) {
+                if (!persistedDocument.getDocumentHeader().getWorkflowDocument()
+                        .isCompletionRequested() && persistedDocument.getDisbVchrCheckTotalAmount().isLessThan(
```
cah292: The old cu-kfs code was using equals() instead of isLessThan() for the comparison of the totals. Can you please check to see if isLessThan() is indeed the correct method or if we purposely customized this to use equals() instead?

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -83,10 +84,16 @@ protected boolean checkWireTransferTabState(DisbursementVoucherDocument dvDocume
...
-        if ((StringUtils.equals(KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_CHECK, dvDocument.getDisbVchrPaymentMethodCode()) || StringUtils.equals(KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_DRAFT, dvDocument.getDisbVchrPaymentMethodCode())) && hasWireTransferValues(dvWireTransfer)) {
-            String questionText = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(CUKFSKeyConstants.QUESTION_CLEAR_UNNEEDED_WIRE_TAB);
-
-            boolean clearTab = super.askOrAnalyzeYesNoQuestion(KFSConstants.DisbursementVoucherDocumentConstants.CLEAR_WIRE_TRANSFER_TAB_QUESTION_ID, questionText);
+        if ((StringUtils.equals(KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_CHECK,
+                dvDocument.getDisbVchrPaymentMethodCode()) || StringUtils.equals(
+                KFSConstants.PaymentSourceConstants.PAYMENT_METHOD_DRAFT,
+                dvDocument.getDisbVchrPaymentMethodCode())) && hasWireTransferValues(dvWireTransfer)) {
+            String questionText = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(
+                    KFSKeyConstants.QUESTION_CLEAR_UNNEEDED_TAB);
```
cah292: Our old cu-kfs code was using QUESTION_CLEAR_UNNEEDED_WIRE_TAB instead of QUESTION_CLEAR_UNNEEDED_TAB. Can you please confirm if our code still needs to use the former?

src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
```diff
@@ -20,9 +17,10 @@ protected void refreshAward(boolean refreshFromLookup) {
         Proposal tempProposal = getAward().getProposal();
         award.setProposal(tempProposal);
         
-        getNewCollectionLine(AWARD_SUBCONTRACTORS).refreshNonUpdateableReferences();
-        getNewCollectionLine(AWARD_PROJECT_DIRECTORS).refreshNonUpdateableReferences();
-        getNewCollectionLine(AWARD_ACCOUNTS).refreshNonUpdateableReferences();
+        getNewCollectionLine(KFSPropertyConstants.AWARD_SUBCONTRACTORS).refreshNonUpdateableReferences();
+        getNewCollectionLine(KFSPropertyConstants.AWARD_PROJECT_DIRECTORS).refreshNonUpdateableReferences();
+        getNewCollectionLine(KFSPropertyConstants.AWARD_FUND_MANAGERS).refreshNonUpdateableReferences();
```
cah292: Can you please check whether we were purposely excluding the refresh of the AWARD_FUND_MANAGERS in our customization? I see that our old cu-kfs code does not refresh the fund managers collection.

src/main/java/org/kuali/kfs/module/purap/document/PurchaseOrderDocument.java
db483: Good point Chad, it looks like this was customized as part of KFSUPGRADE-762 to avoid a conversion during that upgrade.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherAccountingLineTotalsValidation.java
db483: It is not clear which customization changed this code but I will revert to our local customized version.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
db483: This change was done as part of KFSPTS-4366, will revert to locally customized code

src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
db483: This change seems to have gotten lost somehow. I think we need to keep it. It does not look like it is what we have customized locally.

cah292: Ok, I'm fine with keeping this specific change. Thanks for checking.

db483:
Chad, I have made the changes that you suggested. Please review again when you have a chance. Thank you!

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherAccountingLineTotalsValidation.java
```diff
@@ -57,12 +57,15 @@ public boolean validate(AttributedDocumentEvent event) {
...
-                if (persistedDocument.getDocumentHeader().getWorkflowDocument().isSaved() && persistedDocument.getDisbVchrCheckTotalAmount().isZero()) {
+                if (persistedDocument.getDocumentHeader().getWorkflowDocument().isSaved() && persistedDocument
+                        .getDisbVchrCheckTotalAmount().isZero()) {
                     return true;
                 }
 
                 // check total cannot decrease
-                if (!persistedDocument.getDocumentHeader().getWorkflowDocument().isCompletionRequested() && (!persistedDocument.getDisbVchrCheckTotalAmount().equals(dvDocument.getDisbVchrCheckTotalAmount()))) {
+                if (!persistedDocument.getDocumentHeader().getWorkflowDocument()
+                        .isCompletionRequested() && persistedDocument.getDisbVchrCheckTotalAmount().equals(
```
cah292: It looks like the old code was doing a not-equals check instead of an equals check, so this line will need further adjustment or maybe the isLessThan() was correct. Sorry for noticing this earlier.

db483: Thank you for catching that! I have updated the code.

cah292:
After finishing a review of the associated reports, I did find one other thing that may need to be changed. This release updated the spring-security JAR versions, but our POM's override of the spring-security-core dependency doesn't appear to have been updated. Can you please check if our spring-security-core reference in our CU POM needs to have its version updated?

PR #594, opened 2018-06-22
KFSPTS-12304 change ezra service to use document service for routing …

db483:
…documents

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
@@ -456,7 +456,8 @@ public class EzraServiceImpl implements EzraService {
 		agencyDoc.setNewMaintainableObject(agencyMaintainable);
 		try {
 			documentService.saveDocument(agencyDoc);
-			agencyDoc.getDocumentHeader().getWorkflowDocument().route("Automatically created and routed");
+			GlobalVariables.getMessageMap().clearErrorMessages();
+			documentService.routeDocument(agencyDoc, "Automatically created and routed", null);
 		} catch (WorkflowException we) {
 			we.printStackTrace();
 		}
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -456,7 +456,8 @@ private void routeAgencyDocument(Agency agency, Agency oldAgency) {
 		agencyDoc.setNewMaintainableObject(agencyMaintainable);
 		try {
 			documentService.saveDocument(agencyDoc);
-			agencyDoc.getDocumentHeader().getWorkflowDocument().route("Automatically created and routed");
+			GlobalVariables.getMessageMap().clearErrorMessages();
```
jhulslander: Further above, workflow exceptions are caught.  Should those exceptions prevent the document from routing?  If so, clearing the error messages may prevent that.

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -490,8 +490,8 @@ private void routeAwardDocument(Award award, Award oldAward) {
 		awardDoc.getNewMaintainableObject().setBusinessObject(award);;
 		try {
 			documentService.saveDocument(awardDoc);
-			//documentService.routeDocument(awardDoc, "Automatically created and routed", null);
-			awardDoc.getDocumentHeader().getWorkflowDocument().route("Automatically created and routed");
+			GlobalVariables.getMessageMap().clearErrorMessages();
```
jhulslander: same concern as I mentioned in agency

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
```diff
@@ -519,7 +519,8 @@ private void routeProposalDocument(Proposal proposal) {
 		proposalDoc.getNewMaintainableObject().setBusinessObject(proposal);
 		try {
 			documentService.saveDocument(proposalDoc);
-			proposalDoc.getDocumentHeader().getWorkflowDocument().route("Automatically created and routed");
```
jhulslander: same concern as I mentioned in agency

src/main/java/edu/cornell/kfs/module/ezra/service/impl/EzraServiceImpl.java
db483: I think that the Workflow exceptions would still be thrown, they are not cleared with the clearErrorMessages

jhulslander: oK, I shall merge

PR #595, opened 2018-06-22
KFSPTS-12428: Upgrade to 2018-06-14 version of financials

cah292:


cah292:
This is ready for review, but please do NOT merge yet. Also, note that there are no SQL changes for this release, so a SQL PR is not needed for this patch.

db483:
Hi Chad,

Changes look good but I had a question related to https://kualico.atlassian.net/browse/FINP-2887:
Is this change still needed in CuRequisitionDocumentPresentationController after this fix:
editModes.add(RequisitionEditMode.ENABLE_COMMODITY_CODE);

cah292:
Hi Daniela,

You're right, we don't need our custom ENABLE_COMMODITY_CODE workaround anymore, now that the base code is only doing the parameter-based setup for that edit mode. I will update the code accordingly. Thanks for spotting this!

cah292:
I've committed a minor update based on the recent feedback. This is now ready for review again.

PR #596, opened 2018-06-25
KFSPTS-12301 Implement YEDI in the create accounting document

jhulslander:
After this passes code review, please double check the commit message before code merging.


jhulslander:
Thanks I will do that

PR #600, opened 2018-07-04
KFSPTS-12523 upgrade to the 6/21/2018 version of financials

db483:


db483:
retest this please

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -244,8 +244,7 @@
 		<property name="stateService" ref="stateService" />
     </bean>
 
-	<bean id="batchFileDirectoryService" parent="batchFileDirectoryService-parentBean"/>
-	<bean id="batchFileDirectoryService-parentBean" abstract="true" class="edu.cornell.kfs.sys.service.impl.BatchFileDirectoryServiceImpl" />
+	<bean id="batchFileDirectoryService" class="edu.cornell.kfs.sys.service.impl.CuBatchFileDirectoryServiceImpl" />
```
cah292: This should ideally be updated to extend from "batchFileDirectoryService-parentBean". The parent bean doesn't have any special property setup yet, but if that ends up changing, we'll need to make sure the bean inheritance gets straightened out.

cah292:
I've committed a fix for the concern that I had. I will reassign this review to Jay, given that I've now contributed code changes to this PR.

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/sys/service/CuBatchFileDirectoryService.java
```diff
@@ -1,12 +1,13 @@
...
 
+import org.kuali.kfs.sys.batch.service.BatchFileDirectoryService;
 import org.kuali.kfs.sys.businessobject.SystemOptions;
 import org.kuali.rice.core.api.util.KeyValue;
 import org.springframework.cache.annotation.Cacheable;
 
 import java.util.List;
 
-public interface BatchFileDirectoryService {
+public interface CuBatchFileDirectoryService {
```
jhulslander: Why are we changing of name of this class in the KFS upgrade user story?

cah292: With the 06/21 financials patch, KualiCo added their own implementation of BatchFileDirectoryService. I believe our version of the service was renamed to avoid confusion, and to simplify the setup of our custom service's implementation.

src/main/java/edu/cornell/kfs/sys/service/CuBatchFileDirectoryService.java
```diff
@@ -1,12 +1,13 @@
 package edu.cornell.kfs.sys.service;
 
+import org.kuali.kfs.sys.batch.service.BatchFileDirectoryService;
```
jhulslander: this import isn't use, but we may want to to update the interface to extend BatchFileDirectoryService, since the impl class for this interface extends BatchFileDirectoryServiceImpl

cah292: Good point, I'll make that change.

cah292:
I've updated the directory service's interface inheritance as requested. This is ready for review again.

PR #606, opened 2018-07-17
KFSPTS-12304: Fix Ezra Award Documents Routing to Exception

ajd299:
- Use WorkflowDocumentService for routing
- Fix NPE associated with Billing Frequency
- Fix needs confirmation of validity by functionals in DEV & TEST


src/main/java/edu/cornell/kfs/module/cg/document/validation/impl/AwardExtensionRule.java
```diff
@@ -223,7 +227,39 @@ protected boolean checkAccountNotUsedOnOtherAwards(AwardAccount awardAccount) {
...
+                success = false;
+                putFieldError(CGPropertyConstants.AwardFields.BILLING_FREQUENCY_CODE, CGKeyConstants.AwardConstants.ERROR_CG_ACTIVE_BILLS_EXIST, getBillingFrequencyDescription(newAwardCopy));
+            }
+        }
+
+        return success;
+    }
+
+    private String getBillingFrequencyDescription(Award award) {
+        if (award == null || award.getBillingFrequency() == null) {
```
db483: I would suggest using ObjectUtils to check for nulls here

src/main/java/edu/cornell/kfs/module/cg/document/validation/impl/AwardExtensionRule.java
```diff
@@ -223,7 +227,39 @@ protected boolean checkAccountNotUsedOnOtherAwards(AwardAccount awardAccount) {
...
+     * Checks for null on the nullable Billing Frequency Code field. Fixes NullPointerException.
+     */
+    @Override
+    protected boolean checkBillingFrequency() {
+        boolean success = true;
+
+        String newBillingFrequencyCode = newAwardCopy.getBillingFrequencyCode();
+        String oldBillingFrequencyCode = oldAwardCopy.getBillingFrequencyCode();
+
+        if (!org.apache.commons.lang3.StringUtils.equals(newBillingFrequencyCode, oldBillingFrequencyCode)) {
```
db483: I think it may be better to change the StringUtils import to org.apache.commons.lang3.StringUtils and make any needed changes.

src/main/java/edu/cornell/kfs/module/cg/document/validation/impl/AwardExtensionRule.java
db483: I the billingFrequency object null in this case because the new billing frequency code is empty?

ajd299: The NPE was caused by the billing frequency code on the newAward copy being null. I'm not sure what causes the billing frequency code to be null but as it is a nullable field I assumed that we need to handle the null case.

ajd299:
I have address the requested changes to StringUtils. Please re-review.

PR #607, opened 2018-07-20
KFSPTS-12659 set invoicing and fund manager values when creating award doc

db483:


db483:
Hi Jay,

This is now ready for review. Thank you.

PR #612, opened 2018-07-23
KFSPTS-11274 prevent payment works processing errors from stopping subsquent payment works requests

jhulslander:
This needs to be validated locally before merging, but opening it up for code review


src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksKeyConstants.java
```diff
@@ -40,6 +40,7 @@
     public static final String NEW_VENDOR_DETAIL_WAS_NOT_FOUND_ERROR_MESSAGE = "error.paymentworks.new.vendor.detail.not.returned.by.web.service";
     public static final String DUPLICATE_NEW_VENDOR_REQUEST_ERROR_MESSAGE = "error.paymentworks.duplicate.new.vendor.request"; 
     public static final String INITIAL_SAVE_TO_PMW_STAGING_TABLE_FAILED_ERROR_MESSAGE = "error.paymentworks.intial.staging.table.save.failed";
+    public static final String INITIAL_PAYMENT_WORKS_VENOR_SAVE_ERROR_MESSAGE = "error.paymentworks.paymentworksvendor.initial.save";
```
cah292: Minor typo in name of constant.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -99,23 +81,28 @@ private void processEachPaymentWorksNewVendorRequestIntoKFS(List<String> identif
...
             if (canPaymentWorksNewVendorRequestProcessingContinueForVendor(stgNewVendorRequestDetailToProcess, reportData)) {
-                PaymentWorksVendor savedStgNewVendorRequestDetailToProcess = pmwPendingNewVendorRequestInitialSaveToKfsStagingTable(stgNewVendorRequestDetailToProcess);
+                PaymentWorksVendor savedStgNewVendorRequestDetailToProcess = null;
+                try {
+                    savedStgNewVendorRequestDetailToProcess = pmwPendingNewVendorRequestInitialSaveToKfsStagingTable( stgNewVendorRequestDetailToProcess);
+                } catch (Exception e) {
+                    LOG.error("processEachPaymentWorksNewVendorRequestIntoKFS, unable to save payment works vendor for pmwNewVendorRequestId: " + pmwNewVendorRequestId, e);
+                    reportData.getRecordsGeneratingExceptionSummary().incrementRecordCount();
+                    reportData.addPmwVendorThatCouldNotBeProcessed(new PaymentWorksBatchReportRawDataItem(pmwNewVendorRequestId, 
+                            getConfigurationService().getPropertyValueAsString(PaymentWorksKeyConstants.INITIAL_PAYMENT_WORKS_VENOR_SAVE_ERROR_MESSAGE)));
```
cah292: I don't think you need to do the reportData updates within the catch() block, since the relevant updates should already be getting handled by the call to pmwNewVendorSaveToStagingTableWasSuccessful() call below.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -114,8 +115,17 @@ private URI buildPaymentWorksApprovedNewVendorRequestsURI() {
     
     @Override
     public PaymentWorksVendor obtainPmwNewVendorRequestDetailForPmwIdentifier(String pmwNewVendorRequestId, PaymentWorksNewVendorRequestsBatchReportData reportData) {
-        PaymentWorksNewVendorRequestDetailDTO pmwDetailForSpecificVendorDTO = retrieveAllPaymentWorksDetailsForRequestedVendor(buildPaymentWorksPendingNewVendorRequestDetailURI(pmwNewVendorRequestId));
-        PaymentWorksVendor pmwDetailForSpecificVendor = getPaymentWorksDtoToPaymentWorksVendorConversionService().createPaymentWorksVendorFromPaymentWorksNewVendorRequestDetailDTO(pmwDetailForSpecificVendorDTO, reportData);
+        PaymentWorksNewVendorRequestDetailDTO pmwDetailForSpecificVendorDTO = null;
+        PaymentWorksVendor pmwDetailForSpecificVendor = null;
+        try {
+            pmwDetailForSpecificVendorDTO = retrieveAllPaymentWorksDetailsForRequestedVendor(buildPaymentWorksPendingNewVendorRequestDetailURI(pmwNewVendorRequestId));
+        } catch (ResponseProcessingException rpe) {
```
cah292: Is the exception type being caught too specific? Perhaps it should catch a broader exception type instead (like RuntimeException or Exception).

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksKeyConstants.java
jhulslander: thanks, I will change

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
jhulslander: OK, I will remove the report data additions

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
jhulslander: I was using the example exception in the Jira, I can broaden it.

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksKeyConstants.java
jhulslander: with removing the report data elements mentioned below, this can be removed entirely

jhulslander:
I believe I have made the code changes mentioned.  this is ready for another review.  After which, I can try running locally.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -135,14 +119,14 @@ private boolean pmwNewVendorRequestProcessingIntoKfsWasSuccessful(PaymentWorksVe
     
     private boolean canPaymentWorksNewVendorRequestProcessingContinueForVendor(PaymentWorksVendor stgNewVendorRequestDetailToProcess, PaymentWorksNewVendorRequestsBatchReportData reportData) {
         List<String> errorMessages = new ArrayList<String>();
-        if (pmwDtosCouldConvertCustomAttributesToPmwJavaClassAttributes(stgNewVendorRequestDetailToProcess) &&
-            pmwNewVendorAttributesConformToKfsLengthsOrFormats(stgNewVendorRequestDetailToProcess, errorMessages) &&
-            allPmwNewVendorIsoCountriesMapToSingleFipsCountry(stgNewVendorRequestDetailToProcess, errorMessages) &&
-            pmwNewVendorIdentifierDoesNotExistInKfsStagingTable(stgNewVendorRequestDetailToProcess, errorMessages)){
+        if (ObjectUtils.isNotNull(stgNewVendorRequestDetailToProcess)
```
cah292: It looks like that if the vendor object is null, then the report data might not get updated to indicate that the vendor could not be processed. I think the report needs to be updated accordingly when the data for a specific vendor could not be retrieved. If necessary, such reporting could be added to the "obtainPmwNewVendorRequestDetailForPmwIdentifier" method instead.

jhulslander: I think it makes sense to add this report logging here, I will extract the null check into a function, and when null, add to the error List

jhulslander:
I've added back in the null check in the new location, this is ready for review again.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -135,21 +119,30 @@ private boolean pmwNewVendorRequestProcessingIntoKfsWasSuccessful(PaymentWorksVe
...
+                && allPmwNewVendorIsoCountriesMapToSingleFipsCountry(stgNewVendorRequestDetailToProcess, errorMessages)
+                && pmwNewVendorIdentifierDoesNotExistInKfsStagingTable(stgNewVendorRequestDetailToProcess, errorMessages)) {
             return true;
-        }
-        else {
-            if (!errorMessages.isEmpty()){
+        } else {
+            if (!errorMessages.isEmpty()) {
                 reportData.getRecordsThatCouldNotBeProcessedSummary().incrementRecordCount();
                 reportData.addPmwVendorThatCouldNotBeProcessed(new PaymentWorksBatchReportRawDataItem(stgNewVendorRequestDetailToProcess.toString(), errorMessages));
```
cah292: This line assumes that the vendor object is non-null, but the new null-vendor check will allow this line to be called with a null vendor since it updates the error messages list. Some adjustments are needed in order to avoid a NullPointerException.

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -584,6 +584,7 @@ error.paymentworks.new.vendor.custom.field.conversion.exception={0} generated th
 error.paymentworks.new.vendor.detail.not.returned.by.web.service=No detail data was obtained from PaymentWorks web service call for PaymentWorks Vendor ID {0}.
 error.paymentworks.duplicate.new.vendor.request=Duplicate PaymentWorks Vendor ID detected for New Vendor Request.
 error.paymentworks.intial.staging.table.save.failed=Initial attempt to save PaymentWorks New Vendor Request to staging table failed.
+error.paymentworks.paymentworksvendor.initial.save=Unable to save new Payment Works Vendor from Payment Works request.
```
cah292: Is this the correct description to be using for the associated error? This seems to be occurring when the vendor could not be retrieved from PaymentWorks, so perhaps the message should be updated to reflect that.

PR #613, opened 2018-07-24
KFSPTS-12770 convert unit test from easymock to mockito

db483:


src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
```diff
@@ -90,14 +93,7 @@ public void setUp() {
...
 
-        ArrayList<String> methodNames = new ArrayList<>();
-        for (Method method : AdvanceDepositDocument.class.getMethods()) {
-            if (!Modifier.isFinal(method.getModifiers()) && !method.getName().startsWith("set") && !method.getName().startsWith("get")) {
-                methodNames.add(method.getName());
-            }
-        }
-        IMockBuilder<AdvanceDepositDocument> builder = EasyMock.createMockBuilder(AdvanceDepositDocument.class).addMockedMethods(methodNames.toArray(new String[0]));
-        advanceDepositDocument = builder.createNiceMock();
+        advanceDepositDocument = mock(AdvanceDepositDocument.class, CALLS_REAL_METHODS);
```
jhulslander: using Mockito.mock with CALLS_REAL_METHODS, creates an unintiated object, and I think Mockito considers this unsafe.  Take a look at what I did with CuDisbursementVoucherDocumentTest.java.  You can user Powermockito to suppress a parent constructor, and then user power mockito to create a spy.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
```diff
@@ -385,10 +381,10 @@ protected void additionalSetupForFileProcessing() throws IOException, ParseExcep
         fileNames.add(EMPTY_DATA_FILE);
         fileNames.add(GOOD_DATA_FILE_2);
 
-        BatchInputFileService batchInputFileService = EasyMock.createMock(CuBatchInputFileServiceImpl.class);
-        EasyMock.expect(batchInputFileService.listInputFileNamesWithDoneFile(EasyMock.isA(BatchInputFileType.class))).andReturn(fileNames);
-        EasyMock.expect(batchInputFileService.parse(EasyMock.isA(BatchInputFileType.class), EasyMock.isA(byte[].class))).andReturn(setupAchIncomeFilesGood()).andThrow(new RuntimeException()).andReturn(setupAchIncomeFilesGood());
-        EasyMock.replay(batchInputFileService);
+        BatchInputFileService batchInputFileService = mock(CuBatchInputFileServiceImpl.class);
+        when(batchInputFileService.listInputFileNamesWithDoneFile(isA(BatchInputFileType.class))).thenReturn(fileNames);
```
jhulslander: I hadn't seen Mockito.isa before.  We have used Mockito.any in other cases.  Based on https://stackoverflow.com/questions/21430378/mockito-isa-any , we may want to use Mockito.any(BatchInputFileType.class) as the input for the function call.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
```diff
@@ -385,10 +381,10 @@ protected void additionalSetupForFileProcessing() throws IOException, ParseExcep
...
+        when(batchInputFileService.parse(isA(BatchInputFileType.class), isA(byte[].class))).thenReturn(setupAchIncomeFilesGood()).thenThrow(new RuntimeException()).thenReturn(setupAchIncomeFilesGood());
```
jhulslander: similar to above, we may want to switch this to use Mockito.any here.

Also, how does putting the thenReturn/thenThrow calls work here?

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
db483: thank you, I'll take a look at Powermockito and the example you pointed out.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
db483: I'll look into this

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AdvanceDepositServiceImplTest.java
db483: first call will return setupAchIncomeFilesGood(), next call will throw exception, subsequent call will return setupAchIncomeFilesGood()

jhulslander: Interesting, so that makes a call order dependency to be aware of, but cool technique.  Thank you.

PR #616, opened 2018-08-01
KFSPTS-12638: Update PMW Vendor DAO class to use secure parameters

ajd299:
- Refactored to eliminate duplicate code
- Using JDBC PreparedStatement
- Would it be more efficient to use an ORM like OJB or an existing Class/Library?

### DIFF
--- a/src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
+++ b/src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
@@ -20,35 +20,32 @@ public class PaymentWorksVendorDaoJdbc extends PlatformAwareDaoBaseJdbc implemen
     
     @Override
     public void updateExistingPaymentWorksVendorInStagingTable(Integer id, String pmwRequestStatus, String kfsVendorProcessingStatus, Timestamp processingTimeStamp) {
-        String updateSql = buildUpdateExistingPaymentWorksVendorInStagingTableSql(id, pmwRequestStatus, kfsVendorProcessingStatus, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, processingTimeStamp, null, null, null);
-        LOG.info("updateExistingPaymentWorksVendorInStagingTable: updateSQL = " + updateSql);
-        getJdbcTemplate().batchUpdate(updateSql);
+        updateExistingPaymentWorksVendorInStagingTable(id, pmwRequestStatus, kfsVendorProcessingStatus, null, null, null, processingTimeStamp, null, null, null);
     }
     
     @Override
     public void updateExistingPaymentWorksVendorInStagingTable(Integer id, String pmwRequestStatus, String kfsVendorProcessingStatus, String kfsAchProcessingStatus, String kfsVendorDocumentNumber, String supplierUploadStatus, Timestamp  processingTimeStamp) {
-        String updateSql = buildUpdateExistingPaymentWorksVendorInStagingTableSql(id, pmwRequestStatus, kfsVendorProcessingStatus, kfsAchProcessingStatus, kfsVendorDocumentNumber, supplierUploadStatus, processingTimeStamp, null, null, null);
-        LOG.info("updateExistingPaymentWorksVendorInStagingTable: updateSQL = " + updateSql);
-        getJdbcTemplate().batchUpdate(updateSql);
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
```diff
@@ -10,79 +13,132 @@
...
+        return sqlFactory.getPreparedSqlStatement();
     }
 
+    private class ParameterizedSqlFactory {
+        private StringBuilder sqlStringBuilder;
+        private ArrayList parameters;
+
+        public ParameterizedSqlFactory(String initialSql) {
+            sqlStringBuilder = new StringBuilder(initialSql);
+            parameters = new ArrayList();
```
cah292: Initialization of the list should use generics. For this case, you can specify an empty generic-type placeholder and the compiler should be able to auto-detect what type should be used.

src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
```diff
@@ -10,79 +13,132 @@
...
+
+        public String toString() {
+            String effectiveSql = sqlStringBuilder.toString();
+            int parameterIndex = parameters.size();
+
+            while (parameterIndex > 0) {
+                int lastQIndex = effectiveSql.lastIndexOf("?");
+                Object parameter = parameters.get(parameterIndex - 1);
+                String parameterSql = parameter instanceof String ? "'" + parameter.toString() + "'" : parameter.toString();
+                effectiveSql = effectiveSql.substring(0, lastQIndex) + parameterSql + effectiveSql.substring(lastQIndex + 1);
```
cah292: This loop is building a new full SQL string upon each iteration. I would recommend using some other alternative that can operate on the string more efficiently (like using a StringBuilder, or creating an intermediate string that replaces all the question marks with placeholders that can be processed by String.format() or a similar substitution method).

src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
```diff
@@ -2,87 +2,122 @@
...
+        updateExistingPaymentWorksVendorInStagingTable(id, pmwRequestStatus, kfsVendorProcessingStatus, kfsAchProcessingStatus, kfsVendorDocumentNumber, supplierUploadStatus, processingTimeStamp, null, null, null);
     }
     
     @Override
     public void updateExistingPaymentWorksVendorInStagingTable(Integer id, String kfsVendorProcessingStatus, Integer vendorHeaderGeneratedIdentifier, Integer vendorDetailAssignedIdentifier, Timestamp processingTimeStamp) {
-        String updateSql = buildUpdateExistingPaymentWorksVendorInStagingTableSql(id, KFSConstants.EMPTY_STRING, kfsVendorProcessingStatus, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, processingTimeStamp, vendorHeaderGeneratedIdentifier, vendorDetailAssignedIdentifier, null);
-        LOG.info("updateExistingPaymentWorksVendorInStagingTable: updateSQL = " + updateSql);
-        getJdbcTemplate().batchUpdate(updateSql);
+        updateExistingPaymentWorksVendorInStagingTable(id, null, kfsVendorProcessingStatus, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING, processingTimeStamp, vendorHeaderGeneratedIdentifier, vendorDetailAssignedIdentifier, null);
```
cah292: The second arg for this call should also be set to an empty string.

src/main/java/edu/cornell/kfs/pmw/batch/dataaccess/impl/PaymentWorksVendorDaoJdbc.java
```diff
@@ -2,87 +2,122 @@
...
+
+        public void appendSql(String sql, Object parameter) {
+            appendSql(sql);
+            parameters.add(parameter);
+        }
+
+        public String toString() {
+            String effectiveSql = sqlStringBuilder.toString();
+            for (Object parameter : parameters) {
+                effectiveSql = effectiveSql.replaceFirst("\\?", parameter instanceof Integer ? ("%d") : "'%s'");
```
cah292: This could still be reworked to be more efficient. However, given that there does not appear to be any code that is actively using this method, I am okay with leaving it as-is for now, and it can be reworked later if necessary.

ajd299:
I have made the EMPTY_STRING change. The toString method may be useful for debugging but as you observed is not being actively used elsewhere in the codebase.

PR #623, opened 2018-08-09
KFSPTS-12788 updated PMW status when our pmwRequestStatus is set to R…

jhulslander:
…ejected


jhulslander:
This needs to go through local validation, but I thought I would create a PR so we can confirm this is what you thought was needed.

PR #624, opened 2018-08-13
KFSPTS-6361 properly handling acct restriction status code in account global doc

jhulslander:
I believe the function CuAccountGlobal.updateAccountRestrictedStatusCodeForAccountBeingEdited is no longer needed. AccountGlobalPreRules.checkForDefaultSubFundGroupStatus properly sets AccountRestrictedStatusCode to the correct value based on the sub fund group code.  So in AccountGlobal, we just need to set the restriction code on the account from the value on the account global doc.  I noticed the new lines appear to be differently indented than the existing code.  I double checked my eclipse preferences, and they appear to be correct, so I suspect the existing code's format is off.  I didn't want to format the whole file considering there were only a few lines added.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
+++ b/src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
@@ -262,8 +262,8 @@ public class CuAccountGlobal extends AccountGlobal implements GlobalObjectWithIn
 		}
 
 		if (StringUtils.isNotBlank(accountRestrictedStatusCode)) {
-		    if (LOG.isInfoEnabled()) {
-		        LOG.info("updateAccountBasicFields, setting accountRestrictedStatusCode to: " + accountRestrictedStatusCode);
+		    if (LOG.isDebugEnabled()) {
+		        LOG.debug("updateAccountBasicFields, setting accountRestrictedStatusCode to: " + accountRestrictedStatusCode);
 		    }
 		    account.setAccountRestrictedStatusCode(accountRestrictedStatusCode);
 		}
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -17,21 +17,21 @@
...
-import org.kuali.kfs.sys.KFSPropertyConstants;
-import org.kuali.kfs.sys.context.SpringContext;
-import org.kuali.rice.core.api.datetime.DateTimeService;
 import org.kuali.kfs.krad.bo.GlobalBusinessObjectDetailBase;
 import org.kuali.kfs.krad.bo.PersistableBusinessObject;
 import org.kuali.kfs.krad.service.BusinessObjectService;
 import org.kuali.kfs.krad.util.ObjectUtils;
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.sys.context.SpringContext;
+import org.kuali.rice.core.api.datetime.DateTimeService;
```
ajd299: These imports are the same, but they are now correctly alphabetical order, so no change is necessary.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
```diff
@@ -290,7 +262,10 @@ private void updateAccountBasicFields(Account account) {
 		}
 
 		if (StringUtils.isNotBlank(accountRestrictedStatusCode)) {
-			updateAccountRestrictedStatusCodeForAccountBeingEdited(subFundGroup, accountRestrictedStatusCode, account);
+		    if (LOG.isDebugEnabled()) {
```
ajd299: Is there a reason why the code is checking if debug is enabled on this line, but not on line 147? Does the debug function within the Logger class check to see if it is enabled before writing to the log?

jhulslander: Yes, 
The contents of a log statement are processed first, then shown if the applicable log level is achieved.  If a log statement only displays static values, then there is no need to surround it with an if statement.  However, if there is some logic being processed to create the log statement, it is a good idea to surround it with an if statement to avoid the processing except when necessary.

src/main/java/edu/cornell/kfs/coa/businessobject/CuAccountGlobal.java
jhulslander: I ran the organize imports function in eclipse.  I can undo that if you want.

PR #629, opened 2018-08-20
 KFSPTS-6361 properly handling acct restriction status code in account global

jhulslander:
I remove CuAccountGlobal.java.updateAccountRestrictedStatusCodeForAccountBeingEdited as the account global document resets the restriction code based on the sub fund group selected, so this code is no longer needed and causes accounts to not be properly updated with the desired restriction code.
This PR represents a second attempt to solve this problem.  Additional code has been added to check if the restriction code on the account global document is correct for the list of account's existing sub fund group code, when the account global document doesn't try and change the sub fund group code

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
+++ b/src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
@@ -1,5 +1,7 @@
 package edu.cornell.kfs.coa.document.validation.impl;
 
+import java.text.MessageFormat;
+
 import org.apache.commons.lang3.StringUtils;
 import org.kuali.kfs.coa.businessobject.AccountGlobalDetail;
 import org.kuali.kfs.coa.document.validation.impl.AccountGlobalRule;
@@ -7,6 +9,7 @@ import org.kuali.kfs.kns.document.MaintenanceDocument;
 import org.kuali.kfs.krad.util.GlobalVariables;
 import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.KFSKeyConstants;
+import org.kuali.kfs.sys.KFSPropertyConstants;
 
 import edu.cornell.kfs.sys.CUKFSKeyConstants;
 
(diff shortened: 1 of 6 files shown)


src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    protected boolean processCustomRouteDocumentBusinessRules(MaintenanceDocument document) {
+        boolean valid = super.processCustomRouteDocumentBusinessRules(document);
+        if (doesAccountGlobalHaveAccountRestrictionCodeAndNoSubFUndGroupCode()) {
+            valid = isTheNewRestrictionCodeValidForEachAccount() && valid;
+        }
+        return valid;
+    }
+
+    protected boolean doesAccountGlobalHaveAccountRestrictionCodeAndNoSubFUndGroupCode() {
```
ajd299: Extra capital in method name SubFUnd => SubFund

src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
```diff
@@ -0,0 +1,51 @@
...
+@SuppressWarnings("deprecation")
+public class CuAccountGlobalRule extends AccountGlobalRule {
+    
+    @Override
+    protected boolean processCustomRouteDocumentBusinessRules(MaintenanceDocument document) {
+        boolean valid = super.processCustomRouteDocumentBusinessRules(document);
+        if (doesAccountGlobalHaveAccountRestrictionCodeAndNoSubFUndGroupCode()) {
+            valid = isTheNewRestrictionCodeValidForEachAccount() && valid;
+        }
+        return valid;
```
ajd299: This if statement could possibly be refactored to 
if (valid && doesAccountGlobalHaveAccountRestrictionCodeAndNoSubFundGroupCode()) {
valid = isTheNewRestrictionCodeValidForEachAccount()
}

src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
jhulslander: ok

src/main/java/edu/cornell/kfs/coa/document/validation/impl/CuAccountGlobalRule.java
jhulslander: I want to show these error even if there are pre-existing errors from super.processCustomRouteDocumentBusinessRules, so I don't think should be changed

ajd299: Okay I understand now thanks

jhulslander:

PR #632, opened 2018-08-22
KFSPTS-12866 change CINV default document description

jhulslander:


src/main/java/edu/cornell/kfs/module/arl/CuArParameterConstants.java
```diff
@@ -0,0 +1,5 @@
+package edu.cornell.kfs.module.arl;
```
ajd299: Was the arl intentional? Or did you mean to put ar?

jhulslander: good catch,  I shall change

PR #634, opened 2018-08-24
 KFSPTS-12996 hide invoice fields on DV and RCDV docs

db483:


src/main/webapp/WEB-INF/tags/fp/dvPayment.tag
```diff
@@ -332,28 +333,42 @@
...
+	                                              attributeEntry="${dvAttributes.invoiceDate}"
+	                                              readOnly="${!fullEntryMode && !paymentHandlingEntryMode}" /></td>
+	            		</tr>
+	            		 <tr>
+            		</c:when>
+            		<c:otherwise>
+	        			<tr>
+-              			<th scope="row"><div align="right"><kul:htmlAttributeLabel attributeEntry="${dvAttributes.disbVchrCheckStubText}"/></div></th>
+-              			<td><kul:htmlControlAttribute attributeEntry="${dvAttributes.disbVchrCheckStubText}" 
+							property="document.disbVchrCheckStubText" readOnly="${!fullEntryMode && !paymentHandlingEntryMode}"/></td>
```
cah292: Is there a formatting issue that requires the "disbVchrCheckStubText" field to exclude the rowspan setup when the invoice fields are not visible? If so, then are changes also needed to field in any of the other table rows?

cah292:
retest this please

src/main/webapp/WEB-INF/tags/fp/dvPayment.tag
cah292: After running this code locally, things seem to be rendering properly. It looks like the extra rowspan setup allows for keeping the two new invoice fields in the same column.

cah292:
My concerns have been addressed after trying out this code locally, so the changes look good to me. I will trigger one more PR run just to be safe, and will merge this code once the tests pass.

cah292:
retest this please

PR #648, opened 2018-09-25
KFSPTS 12390 Stale Check Job

ajd299:


ajd299:
retest this please

src/main/java/com/rsmart/kuali/kfs/cr/batch/service/impl/StaleCheckExtractServiceImpl.java
```diff
@@ -0,0 +1,329 @@
...
+
+    protected void validateCheckReconciliation(StaleCheckBatchRow staleCheckRow, CheckReconciliation checkReconciliation, List<String> validationErrors) {
+        if (!StringUtils.equalsIgnoreCase(staleCheckRow.getBankCode(), checkReconciliation.getBankCode())) {
+            validationErrors.add("The Bank Code in the file (" + staleCheckRow.getBankCode() + ") row does not match the bank on the Check (" +
+                    checkReconciliation.getBankCode() + ").");
+        }
+
+        try {
+            Double checkTotalAmountParsed = NumberFormat.getNumberInstance(java.util.Locale.US).parse(staleCheckRow.getCheckTotalAmount()).doubleValue();
+            KualiDecimal checkTotalAmount = new KualiDecimal(checkTotalAmountParsed);
```
cah292: KualiDecimal has a constructor that accepts a String as input, so you don't need to parse the String value with NumberFormat explicitly.

ajd299:
retest this please

ajd299:
retest this please

ajd299:
retest this please

ajd299:
retest this please

ajd299:
Ready for review again

src/main/java/com/rsmart/kuali/kfs/cr/dataaccess/CheckReconciliationDao.java
```diff
@@ -73,4 +73,6 @@
      * @return
      */
     public List<PaymentGroup> getAllPaymentGroupForSearchCriteria(KualiInteger disbNbr, Collection<String> bankCodes);
+
+    public CheckReconciliation findByCheckNumber(String checkNumber);
```
cah292: Is the check number enough of a constraint for uniqueness when searching for CR rows? If we don't expect the bank account number to change, then I think we will be okay.

ajd299: I'm not sure, would you recommend adding the bank code?

cah292: I think adding bank code will be enough, unless the CR rows are using several different bank account numbers for the same bank (which I do not think is the case).

ajd299:
Ready for re-review

PR #656, opened 2018-10-03
KFSPTS-13325 update pom to 2018-09-13 version of financials and get t…

db483:
…hings to compile


db483:
Chad, 

Note that I have brought in the AwardRule.java so that I could modify the method access to protected for the validation methods that we have locally customized. KualiCo has updated the access to default and that was causing issues with our locally customized code. I have reached out to Bryan and explained how we customize these methods and he will bring up this issue to KualiCo. He expects they will revert back to methods having protected access level rather than default. Once KualiCo makes this change we should be able to just remove the AwardRule file from the cu-kfs project.

db483:
Please also note that I have created https://jira.cornell.edu/browse/KFSPTS-13489 to update put local module config spring files to use the p syntax.

PR #657, opened 2018-10-03
 KFSPTS-12914 previous error message prevent creation of EIRT document

jhulslander:


src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -1289,6 +1283,16 @@ protected void rejectElectronicInvoiceFile(ElectronicInvoiceLoad eInvoiceLoad, S
         }
     }
 
+    private void clearErrorMessagesToPreventRiceSaveAndRoutErrors() {
```
cah292: Minor typo in method name (should be "...AndRouteErrors" instead of "...AndRoutErrors").

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -1289,6 +1283,16 @@ protected void rejectElectronicInvoiceFile(ElectronicInvoiceLoad eInvoiceLoad, S
...
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("clearErrorMessagesToPreventRiceSaveAndRoutErrors, Logging previously created error messages");
+            for (String key : GlobalVariables.getMessageMap().getErrorMessages().keySet()) {
+                LOG.debug("clearErrorMessagesToPreventRiceSaveAndRoutErrors, key: " + key + "value: " + GlobalVariables.getMessageMap().getErrorMessages().get(key));
```
cah292: In the Map returned by MessageMap.getErrorMessages(), the elements appear to be of type AutoPopulatingList. This particular List implementation doesn't appear to contain the needed toString() override for printing the list contents, so if you indeed want to log the contents here, you may need additional code to implement another way of printing them.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
jhulslander: thanks

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
jhulslander: thanks for noticing, I will see if I can adjust the printout without significant effort, if not, I will remove this.

PR #661, opened 2018-10-11
KFSPTS 12390 Stale Check Job

ajd299:


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
```diff
@@ -203,9 +205,10 @@ else if (paymentGroup.getDisbursementType().getCode().equals(PdpConstants.Disbur
                     glPendingTransaction.setFinancialObjectCode(paymentAccountDetail.getFinObjectCode());
                     glPendingTransaction.setFinancialSubObjectCode(paymentAccountDetail.getFinSubObjectCode());
                 }
+
+                glPendingTransaction.setProjectCd(paymentAccountDetail.getProjectCode());
```
jhulslander: so we are only setting the project code if we are NOT in a stale situation.  are we sure we always want project code to be empty in the stale check situation?

ajd299: Yes I have confirmed with the functionals that this is indeed the case.

jhulslander: Thanks for confirming

PR #663, opened 2018-10-17
 KFSPTS-5546 remove spring context calls from CuDisbursementVoucherPaymentReasonServiceImpl

jhulslander:
CuDisbursementVoucherPaymentReasonServiceImpl already had the service injected via Spring, but the function still called out to SpringConext get an instance.  It seems to be working correctly with this change


src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherPaymentReasonServiceImplTest.java
```diff
@@ -1,85 +1,107 @@
...
+        payeeTypeCds.add("X");
+        payeeTypeCds.add("S");
+        return payeeTypeCds;
+    }
+
+    public class TestableDisbursementVoucherPaymentReasonServiceImpl extends DisbursementVoucherPaymentReasonServiceImpl {
+
+        /*
+         * This is the same function as is in DisbursementVoucherPaymentReasonServiceImpl, except it doesn't use SpringContext
+         */
```
nkimble: Is this comment necessary?

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherPaymentReasonServiceImpl.java
```diff
@@ -4,24 +4,21 @@
...
 
         for (String payeeTypeCode : payeeTypeCodes) {
-            String description = SpringContext.getBean(DisbursementVoucherPayeeService.class).getPayeeTypeDescription(payeeTypeCode);
+            String description = disbursementVoucherPayeeService.getPayeeTypeDescription(payeeTypeCode);
             if (payeeTypeDescriptions.indexOf(description) == -1) {
                 payeeTypeDescriptions.add(description);
             }
         }
 
         return this.convertListToString(payeeTypeDescriptions);
```
nkimble: Is "this" needed here?

jhulslander: I should be able to remove that, thanks

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherPaymentReasonServiceImplTest.java
jhulslander: I thought it would be helpful, but I can remove it

jhulslander:
I've made changes per the code review this is ready for review again.

PR #664, opened 2018-10-17
KFSPTS-12363: Fix handling of CNCR traveler statuses

cah292:
NOTE: There are both code and SQL PRs for this change. Please make sure both PRs are good to go before merging.


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
```diff
@@ -149,6 +151,14 @@ public String getFileContents(String fileName) {
         }
     }
 
+    @Override
+    public boolean isValidTravelerStatus(String status) {
```
jhulslander: would it make more sense to rename this to isVallidTravelerForEmployeePDPStatus?

src/main/java/edu/cornell/kfs/concur/ConcurParameterConstants.java
```diff
@@ -10,6 +10,7 @@
     public static final String CONCUR_CUSTOMER_PROFILE_SUB_UNIT = "CONCUR_CUSTOMER_PROFILE_SUB_UNIT";
     public static final String DEFAULT_TRAVEL_REQUEST_OBJECT_CODE = "DEFAULT_TRAVEL_REQUEST_OBJECT_CODE";
     public static final String CONCUR_PENDING_CLIENT_OBJECT_CODE_OVERRIDE = "CONCUR_PENDING_CLIENT_OBJECT_CODE_OVERRIDE";
+    public static final String CONCUR_VALID_TRAVELER_STATUSES = "CONCUR_VALID_TRAVELER_STATUSES";
```
jhulslander: I think we should name this such that it states these are valid concur status to be used in PDP with the employee payee type

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurBatchUtilityServiceImpl.java
cah292: Ok, I'll see if I can rename this to better reflect what it is checking for.

src/main/java/edu/cornell/kfs/concur/ConcurParameterConstants.java
cah292: Ok, I'll rename the parameter accordingly based on the changes I make to the utility service method.

cah292:
I've updated the code based on the PR feedback. This is ready for review again.

PR #666, opened 2018-10-24
KFSPTS-13696: Disable chunking when uploading vendors to PMW

cah292:
The known KFSPTS-12898 issue is causing our JAX-RS code to use the CXF REST client instead, and that client will auto-chunk large requests by default. This fix updates the relevant code to prevent the supplier-upload process from chunking the requests, since that PaymentWorks endpoint doesn't appear to support chunked requests.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
@@ -415,15 +415,15 @@ public class PaymentWorksWebServiceCallsServiceImpl implements PaymentWorksWebSe
 
     private void disableRequestChunkingIfNecessary(Client client, Invocation.Builder requestBuilder) {
         if (client instanceof org.apache.cxf.jaxrs.client.spec.ClientImpl) {
-            if (LOG.isDebugEnabled()) {
-                LOG.debug("disableRequestChunkingIfNecessary: Explicitly disabling chunking because KFS is using a CXF JAX-RS client");
-            }
+            LOG.info("disableRequestChunkingIfNecessary: Explicitly disabling chunking because KFS is using a JAX-RS client of CXF type "
+                    + client.getClass().getName());
             ClientConfiguration cxfConfig = WebClient.getConfig(requestBuilder);
             HTTPConduit conduit = cxfConfig.getHttpConduit();
             HTTPClientPolicy clientPolicy = conduit.getClient();
             clientPolicy.setAllowChunking(false);
-        } else if (LOG.isDebugEnabled()) {
-            LOG.debug("disableRequestChunkingIfNecessary: The JAX-RS client implementation does not need to explicitly disable chunking");
+        } else {
+            LOG.info("disableRequestChunkingIfNecessary: There is no need to explicitly disable chunking for a JAX-RS client of type "
+                    + client.getClass().getName());
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -397,14 +402,31 @@ private Invocation buildMultiPartRequestForSupplierUpload(Client client, URI uri
...
+        
+        return requestBuilder
                 .accept(MediaType.APPLICATION_JSON_TYPE)
                 .header(PaymentWorksWebServiceConstants.AUTHORIZATION_HEADER_KEY, 
                              PaymentWorksWebServiceConstants.AUTHORIZATION_TOKEN_VALUE_STARTER + getPaymentWorksAuthorizationToken())
                 .buildPost(Entity.entity(multiPart, MediaType.MULTIPART_FORM_DATA_TYPE));
     }
 
+    private void disableRequestChunkingIfNecessary(Client client, Invocation.Builder requestBuilder) {
+        if (client instanceof org.apache.cxf.jaxrs.client.spec.ClientImpl) {
```
jhulslander: I thought this was a problem only when the request was over a certain size, like 30 vendors or so.  Should we disable chunking only when the instance is as coded above and the size limit is reached?

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -397,14 +402,31 @@ private Invocation buildMultiPartRequestForSupplierUpload(Client client, URI uri
...
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("disableRequestChunkingIfNecessary: Explicitly disabling chunking because KFS is using a CXF JAX-RS client");
+            }
+            ClientConfiguration cxfConfig = WebClient.getConfig(requestBuilder);
+            HTTPConduit conduit = cxfConfig.getHttpConduit();
+            HTTPClientPolicy clientPolicy = conduit.getClient();
+            clientPolicy.setAllowChunking(false);
+        } else if (LOG.isDebugEnabled()) {
+            LOG.debug("disableRequestChunkingIfNecessary: The JAX-RS client implementation does not need to explicitly disable chunking");
```
jhulslander: A couple thoughts here.

As coded now, we will only not enable chunking when the client is something other than org.apache.cxf.jaxrs.client.spec.ClientImpl .  If we stay with that, I think want to include the instance type of the client object.  Also, should it be something other than debug, something would have changed that we should probably be made aware of I think

If we want to add further restrictions, we should add the vendor count to the log statement, and surround it with if (LOG.isDebugEnabled())

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
cah292: The point is that we never want the data to be sent in a chunked manner, since the PaymentWorks endpoint cannot handle that. This fix just forces the JAX-RS client to always use non-chunked mode, rather than letting CXF decide whether or not to chunk the data.

Also, at this point in the code, the data is in an InputStream so the number of rows cannot be easily derived. If we wanted to add a row-count-related restriction, we would need to update the method signatures to accept new or different arguments accordingly.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
cah292: Ok, I'll work on updating this to add more log information about the JAX-RS client in use, and to change the log level accordingly.

Also, the calling code in PaymentWorksUploadSuppliersServiceImpl is already logging the number of vendors to upload, so we don't need to log it again here. Doing so here instead would be possible, but would require changing various method signatures to accept new or different arguments.

cah292:
I've adjusted the logging based on the initial PR feedback. This is now ready for review again.

PR #671, opened 2018-10-29
KFSPTS-13687: Force OJB Award references to map to CuAward

cah292:
This fix allows for Award-related OJB queries to map to CuAward instead, and also affects Award-related reference descriptors similarly. I did have to override the Award OJB metadata to implement this fix, but apart from adding a row-reader attribute and an extent-class element, the Award OJB XML should be the same as in base code.

This change also acts as an alternative fix to KFSPTS-13641 (from PR #665), so I have removed its related OJB XML override as well.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
+++ b/src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
@@ -3,6 +3,7 @@ package edu.cornell.kfs.sys.broker.accesslayer;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.ojb.broker.PersistenceBrokerException;
 import org.apache.ojb.broker.accesslayer.RowReaderDefaultImpl;
 import org.apache.ojb.broker.metadata.ClassDescriptor;
@@ -34,7 +35,8 @@ public class RowReaderMappingBaseClassToSingleExtentClass extends RowReaderDefau
         List<?> extentClassNames = baseClassDescriptor.getExtentClassNames();
         if (extentClassNames.size() != 1) {
             throw new PersistenceBrokerException("The class descriptor for " + baseClassDescriptor.getClassNameOfObject()
-                    + " should have had exactly 1 extent class, but instead had " + extentClassNames.size());
+                    + " should have had exactly 1 extent class, but instead had " + extentClassNames.size() + " extents: "
+                    + extentClassNames);
         }
         
         String extentClassName = (String) extentClassNames.get(0);
@@ -42,6 +44,9 @@ public class RowReaderMappingBaseClassToSingleExtentClass extends RowReaderDefau
         if (extentClassDescriptor == null) {
             throw new PersistenceBrokerException("Could not find " + extentClassName
                     + " extent class descriptor for base class " + baseClassDescriptor.getClassNameOfObject());
+        } else if (StringUtils.equals(baseClassDescriptor.getClassNameOfObject(), extentClassDescriptor.getClassNameOfObject())) {
+            throw new PersistenceBrokerException("The base class descriptor for " + baseClassDescriptor.getClassNameOfObject()
+                    + " should not have been selected as the implicit descriptor for extent class " + extentClassName);
         }
         
         return extentClassDescriptor;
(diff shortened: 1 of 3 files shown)


src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
```diff
@@ -14,6 +14,132 @@
...
+        </collection-descriptor>
+        <collection-descriptor name="awardOrganizations" element-class-ref="org.kuali.kfs.module.cg.businessobject.AwardOrganization"
+                               collection-class="org.apache.ojb.broker.util.collections.ManageableArrayList" proxy="true" auto-retrieve="true" auto-update="object" auto-delete="object">
+            <orderby name="chartOfAccountsCode" sort="ASC"/>
+            <orderby name="organizationCode" sort="ASC"/>
+            <inverse-foreignkey field-ref="proposalNumber"/>
+        </collection-descriptor>
+    </class-descriptor>
+    
     <class-descriptor class="edu.cornell.kfs.module.cg.businessobject.CuAward" table="CG_AWD_T">
```
jhulslander: Should CuAward still be mapped here?

src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
```diff
@@ -0,0 +1,50 @@
...
+    public RowReaderMappingBaseClassToSingleExtentClass(ClassDescriptor classDescriptor) {
+        super(classDescriptor);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    protected ClassDescriptor selectClassDescriptor(Map row) throws PersistenceBrokerException {
+        ClassDescriptor baseClassDescriptor = getClassDescriptor();
+        List<?> extentClassNames = baseClassDescriptor.getExtentClassNames();
+        if (extentClassNames.size() != 1) {
```
jhulslander: should this be greater than 1?  I would think no extent classes defined would return super's selectClassDescriptor or perhaps baseClassDescriptor.

src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
```diff
@@ -0,0 +1,50 @@
...
+            throw new PersistenceBrokerException("The class descriptor for " + baseClassDescriptor.getClassNameOfObject()
+                    + " should have had exactly 1 extent class, but instead had " + extentClassNames.size());
```
jhulslander: would it be worth while listing the values in extentClassNames?  If we get here, that mean there is a misconfiguration.  I am just thinking of making it easier to find the error.

src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
cah292: Yes, we still need the CuAward mapping since it includes column mappings that are not in the Award OJB XML, in addition to adding a mapping for the extended attribute.

src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
cah292: The extentClassNames list roughly equates to the "extent-class" elements in the OJB metadata. This RowReader is intended for very specific situations, where we want to replace queries for a base object with those of a single specific subclass (such as replacing Award with CuAward). I could modify this to be more lenient about how many (or if any) extents are defined, but that could increase the risk of overlooking configuration errors.

src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
cah292: Yes, that would be a good addition for troubleshooting purposes. I'll update this accordingly to print the specific extent classes.

src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
jhulslander: sounds good

src/main/java/edu/cornell/kfs/sys/broker/accesslayer/RowReaderMappingBaseClassToSingleExtentClass.java
jhulslander: Thanks for explaining, lets leave as is

cah292:
I've made some more changes based on the initial feedback. In addition, I added some new unit tests for this functionality. The tests ended up discovering one minor issue, which I have also fixed in the latest update.

PR #673, opened 2018-11-01
KFSPTS-13778: More fixes for Sub-Account Global

cah292:
This updates the KFSPTS-12164 feature to perform better handling of closed and expired accounts during sub-account creation (including handling of the continuation-accounts prompt), and to also properly show the off-campus-indicator prompt when any of the new sub-accounts are flagged as off-campus.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
+++ b/src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
@@ -6,7 +6,6 @@ import org.kuali.kfs.coa.document.validation.impl.MaintenancePreRulesBase;
 import org.kuali.kfs.kns.document.MaintenanceDocument;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.sys.KFSConstants;
-import org.kuali.kfs.sys.KFSPropertyConstants;
 import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.rice.core.api.config.property.ConfigurationService;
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
```diff
@@ -1,13 +1,18 @@
 package edu.cornell.kfs.coa.document.validation.impl;
 
-import edu.cornell.kfs.sys.CUKFSConstants;
+import org.apache.commons.lang3.StringUtils;
+import org.kuali.kfs.coa.businessobject.Account;
 import org.kuali.kfs.coa.document.validation.impl.MaintenancePreRulesBase;
+import org.kuali.kfs.kns.document.MaintenanceDocument;
+import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.KFSPropertyConstants;
```
ajd299: Is this import used?

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
```diff
@@ -47,4 +55,34 @@ protected boolean checkOffCampus(MaintenanceDocument maintenanceDocument) {
...
+        } else {
+            return subAccountGlobal.getSubAccountGlobalNewAccountDetails().stream()
+                    .anyMatch(SubAccountGlobalNewAccountDetail::isOffCampusCode);
+        }
+    }
+
+    protected boolean checkContinuationAccountsForNewSubAccounts(MaintenanceDocument maintenanceDocument) {
+        SubAccountGlobal subAccountGlobal = (SubAccountGlobal) maintenanceDocument.getNewMaintainableObject().getBusinessObject();
+        subAccountGlobal.getSubAccountGlobalNewAccountDetails().stream()
+                .forEach(this::updateDetailToUseContinuationAccountIfNecessary);
```
ajd299: Is .stream() needed since the getSubAccountGlobalNewAccountDetails() returns a List?

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
cah292: Good catch, it's not used so I'll remove it.

src/main/java/edu/cornell/kfs/coa/document/validation/impl/SubAccountGlobalPreRules.java
cah292: Good point, the list also has a forEach() method so I don't need to explicitly create the stream first. I'll update this line accordingly.

cah292:
I've made some changes based on the initial feedback. This is ready for review again.

ajd299:
Looks good merging

PR #674, opened 2018-11-01
KFSPTS-13687: Force OJB Award references to map to CuAward

cah292:
This is an updated version of the #671 changes. The original solution was not sufficient to fix situations where the Award reference was proxied, so I've added overrides of the relevant classes so that their Award references will map to CuAward.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
+++ b/src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
@@ -14,6 +14,132 @@
         <field-descriptor name="everify" column="E_VERIFY_IND" jdbc-type="VARCHAR" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbCharBooleanConversion" />
     </class-descriptor>
     
+    <class-descriptor class="org.kuali.kfs.module.cg.businessobject.Award" table="CG_AWD_T"
+            row-reader="edu.cornell.kfs.sys.broker.accesslayer.RowReaderMappingBaseClassToSingleExtentClass">
+        <extent-class class-ref="edu.cornell.kfs.module.cg.businessobject.CuAward"/>
+        <field-descriptor name="proposalNumber" column="CGPRPSL_NBR" jdbc-type="VARCHAR" primarykey="true" index="true"/>
+        <field-descriptor name="objectId" column="OBJ_ID" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="versionNumber" column="VER_NBR" jdbc-type="BIGINT" locking="true"/>
+        <field-descriptor name="awardBeginningDate" column="CGAWD_BEG_DT" jdbc-type="DATE"/>
+        <field-descriptor name="awardEndingDate" column="CGAWD_END_DT" jdbc-type="DATE"/>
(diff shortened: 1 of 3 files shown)


src/main/resources/edu/cornell/kfs/module/cg/cu-ojb-cg.xml
```diff
@@ -125,6 +251,92 @@
...
+        <field-descriptor name="proposalAwardTypeCode" column="CGPRPSL_AWD_TYP_CD" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="agencyNumber" column="CG_AGENCY_NBR" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="proposalStatusCode" column="CGPRPSL_STAT_CD" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="federalPassThroughAgencyNumber" column="CG_FEDPT_AGNCY_NBR" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="cfdaNumber" column="CG_CFDA_NBR" jdbc-type="VARCHAR"/>
+        <field-descriptor name="proposalFellowName" column="CGPRPSL_FELLOW_NM" jdbc-type="VARCHAR"/>
+        <field-descriptor name="proposalPurposeCode" column="CGPRPSL_PURPOSE_CD" jdbc-type="VARCHAR" index="true"/>
+        <field-descriptor name="proposalProjectTitle" column="CGPRPSL_PROJ_TTL" jdbc-type="VARCHAR"/>
+        <field-descriptor name="active" column="ROW_ACTV_IND" jdbc-type="VARCHAR" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbCharBooleanConversion"/>
+        <reference-descriptor name="award" class-ref="edu.cornell.kfs.module.cg.businessobject.CuAward" auto-retrieve="true" auto-update="none"
```
jhulslander: I think you mentioned in the standup this OJB mapping is required due a previously fixed problem.  Should that previous fix be re-engineered to avoid needing this mapping (if so a different user story may be needed for that)?  I am concerned with are going to need update more and more object mapping in OJB just to link to CUAward.  Is there away to avoid this?  Am I overly concerned?

cah292: The only objects that I found that have OJB reference/collection mappings to Award were AwardAccount and Proposal. The other references should already be covered by our existing EBO mapping customizations and the new row-mapping customizations from this PR.

The alternative would be to remove the CuAward BO and transfer its custom column data to the extended attributes, and also move our custom "populateFromProposal" method handling to a separate helper class. The latter point is the biggest concern since that could increase the risk of accidentally overlooking changes to that method when generating upgrade reports, unless we decide to overlay Award to customize that method.

Replacing CuAward with just the base Award would require more work in the short term, but perhaps it would provide more of a benefit in the long term if we can be sure not to overlook future KualiCo changes to the "populateFromProposal" method (and also transfer the relevant data to the extended attribute instead). I'll start looking into this approach a little more to see just how much work it would be.

cah292:
I'm closing this review so that I can explore the alternative approach of replacing CuAward with the regular Award BO. If that approach ends up not working out, then I'll reopen this PR accordingly.

PR #686, opened 2018-11-16
KFSPTS-13806: Truncate PMW business legal name to 45 characters. Truncate PMW legal last-first name combination to 43 characters with comma and space delimiter so total characters are 45.

nkimble:


nkimble:
Legal name is either the business legal name or a concatenation of legal last name and legal first name delimited by a comma and a single space. Both are restricted to a total of 45 characters by the KFS business rules. The edoc deals with creating a single database legal name entity out of these three constructs such they conform to the edocs business rules.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -195,49 +195,15 @@ private boolean validateVendorType(String vendorType, List<String> errorMessages
...
-            allValidationPassed = false;
-            LOG.info("companyLegalNameEnteredConformsToMaxKfsLength: KFS legal name max length is " + VendorConstants.MAX_VENDOR_NAME_LENGTH +
-                     " PMW legal name length received is " + stgNewVendorRequestDetailToProcess.getRequestingCompanyLegalName().length());
+    private boolean legalNameWasEntered(PaymentWorksVendor stgNewVendorRequestDetailToProcess, List<String> errorMessages) {
+        if ( (StringUtils.isNotBlank(stgNewVendorRequestDetailToProcess.getRequestingCompanyLegalLastName()) &
+              StringUtils.isNotBlank(stgNewVendorRequestDetailToProcess.getRequestingCompanyLegalFirstName()))
+              || (StringUtils.isNotBlank(stgNewVendorRequestDetailToProcess.getRequestingCompanyLegalName())) ) {
+            return true;
+        } else {
+            errorMessages.add(getConfigurationService().getPropertyValueAsString(PaymentWorksKeyConstants.ERROR_LEGAL_NAME_NULL_OR_BLANK));
```
jhulslander: This error message is for legal name missing or first & last name missing.  Should it be renamed so it mentions that?

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
```diff
@@ -619,22 +619,59 @@ private String determineKfsOwnershipTypeCodeFromPmwTaxClassificationCode(Payment
...
+            return truncatedLegalLastName;
+        }
+    }
+
+    private static String truncateLegalFirstNameToMaximumAllowedLengthWhenFormattedWithLegalLastName(String legalLastName, String legalFirstName) {
+        int maxAllowedLengthOfFirstName = VendorConstants.MAX_VENDOR_NAME_LENGTH - VendorConstants.NAME_DELIM.length() - legalLastName.length();
+        if (legalFirstName.length() <= maxAllowedLengthOfFirstName) {
+            return legalFirstName;
+        } else {
+            String truncatedLegalFirstName = legalFirstName.substring(0, maxAllowedLengthOfFirstName);
```
jhulslander: If the length of the legal last name is already the max length, then the legal first name would get truncated to an empty string.  I think this code may handle it, but if you put in an explicit check for that, it may make the code intention more easily understood.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
```diff
@@ -619,22 +619,59 @@ private String determineKfsOwnershipTypeCodeFromPmwTaxClassificationCode(Payment
...
+            LOG.info("truncateLegalFirstNameToMaximumAllowedLengthWhenFormattedWithLegalLastName: Received legalFirstName '"
```
jhulslander: Since the length of the last name is part of the reason how the first name is truncated, I think it may be helpful to add the last name to this log message.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
nkimble: Legal name is considered to be one thing by the database or the business rules (depending upon the individual or business tax reporting) and is composed later on in the code. This is a hard error that is done up front that will require the Purchasing Staff to address the problem in PMW and will stop a PVEN edoc from being created (force hand entry into KFS of the data).  Therefore; the Purchasing staff will actually see which type of legal name it is in PMW because one-and-only-one legal name is allowed (EIN only presents business legal name and SSN presents first last name) so IMHO getting that granular here may add more confusion.

jhulslander: OK, fair enough

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
nkimble: Making message more descriptive.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
nkimble: Good catch.  I have reworked this section.
To help clarify: 45 is max length of legal name no matter how it is created. If first and last is used to create it then two of those characters need to be a comma followed by a space to delimit them. 
So: If last is at the max or would be at the max with the delimiter added, I agree we should set first to empty string OR return the actual first name if the whole thing can be concatenated OR  create a sub-string to use for first.

jhulslander: that's a good solution, thanks

nkimble:
retest this please

PR #690, opened 2018-11-23
 KFSPTS-13806: Truncate PMW business legal name to 45 characters. Truncate PMW legal last-first name combination to 43 characters with comma and space delimiter so total characters are 45. Truncate contact name.

db483:


src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
```diff
@@ -619,21 +609,37 @@ private String determineKfsOwnershipTypeCodeFromPmwTaxClassificationCode(Payment
...
         return kfsVendorDataWrapper;
     }
-    
+
     private KfsVendorDataWrapper populateFirstLastLegalName(PaymentWorksVendor pmwVendor, KfsVendorDataWrapper kfsVendorDataWrapper) {
         kfsVendorDataWrapper.getVendorDetail().setVendorFirstLastNameIndicator(true);
-        kfsVendorDataWrapper.getVendorDetail().setVendorFirstName(pmwVendor.getRequestingCompanyLegalFirstName());
-        kfsVendorDataWrapper.getVendorDetail().setVendorLastName(pmwVendor.getRequestingCompanyLegalLastName());
+        kfsVendorDataWrapper.getVendorDetail().setVendorLastName(truncateValueToMaxLength(pmwVendor.getRequestingCompanyLegalLastName(), VendorConstants.MAX_VENDOR_NAME_LENGTH));
+        kfsVendorDataWrapper.getVendorDetail().setVendorFirstName(truncateLegalFirstNameToMaximumAllowedLengthWhenFormattedWithLegalLastName(pmwVendor.getRequestingCompanyLegalLastName(), pmwVendor.getRequestingCompanyLegalFirstName()));
```
jhulslander: Should this function take in kfsVendorDataWrapper.getVendorDetail().getVendorLastName() instead of . pmwVendor.getRequestingCompanyLegalLastName()?  I am asking because the previous line truncated the value as needed.  I am not sure if it matters really.

db483: Good point Jay. In this case I don't think it makes a difference because the same max vendor name length is used in both cases. However, if this was changed to use a different max length for vendor last name then it would not work correctly. I will update the code for clarity.

db483:
Jay, the problems we ran into were not related to the truncating code not working but to another field that needed to be truncated and wasn't included in the initial set of requirements. I do however agree that unit tests would be good. I shall add them.

db483:
This is ready for review again. Thank you.

PR #706, opened 2019-01-03
KFSPTS-14093 upgrade to the 12/13 version of financials and get thing…

db483:
…s to compile


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCreateCollectorFileServiceImpl.java
```diff
@@ -109,18 +112,23 @@ protected int calculateBatchSequenceNumber() {
...
         
-        List<? extends BusinessObject> searchResults = batchFileLookupableHelperService.getSearchResults(criteria);
+        List<? extends BusinessObject> searchResults = batchFileLookupableHelperService.getSearchResults(BatchFile.class, criteria);
         return searchResults.size();
     }
+    
+    private String buildDateRangeStringForCurrentDate() {
+        java.sql.Date midnight = dateTimeService.getCurrentSqlDateMidnight();       
+        Date now = new Date(System.currentTimeMillis());
+        String rangeForCurrentDate = String.format(DATE_RANGE_FORMAT, Long.toString(midnight.getTime()), Long.toString(now.getTime()));
```
cah292: Do these dates have to be specified as long integers instead of date strings? If so, does the upper value need to be changed to the end of the day instead of the current time?

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
```diff
@@ -1,49 +1,56 @@
...
-    public CreateDoneBatchFileAuthorizationService getCreateDoneAuthorizationService() {
+	public List<Map<String, Object>> getActionLinks(BusinessObjectBase businessObject, Person user) {
+		BatchFile batchFile = (BatchFile) businessObject;
+		List<Map<String, Object>> actionLinks = new LinkedList<>();
+
+		if (canCreateDoneFile(batchFile, user)) {
+			Map<String, Object> createDoneLink = new LinkedHashMap<>();
+			createDoneLink.put("label", "Create Done");
+			createDoneLink.put("url", KRAD_URL_PREFIX + getCreateDoneUrl(batchFile));
+			createDoneLink.put("method", "GET");
```
cah292: This class uses a lot of string literals for property names or standard property values. Can some or most of these be replaced with constants instead? (If needed, I'm okay with leaving them as-is if we intend to perform further cleanup on this file in a future sprint.)

db483: good point. I have modeled this on base code that that is not using any constants. I will add some constants to our local constants file.

PR #708, opened 2019-01-09
KFSPTS-12904 Add Total Program Outlays to CGB invoices

cah292:


src/main/java/edu/cornell/kfs/module/ar/CuArPropertyConstants.java
```diff
@@ -7,4 +7,5 @@
         public static final String PARTIAL_BILL = "partialBill";
     }
 
+    public static final String TOTAL_PROGRAM_OUTLAYS_TO_DATE = "totalProgramOutlaysToDate";
```
jhulslander: Since this a CINV template element, shouldn't it go in the ContractsAndGrantsBillingAwardFields sub class?

cah292: I'm not sure if ContractsAndGrantsBillingAwardFields is the right place to put this constant, since the new field is derived from some fields on the invoice details instead of the award. Or is that subclass meant to encompass more than just fields that are closely tied to the award?

jhulslander: I added ContractsAndGrantsBillingAwardFields recently, it was meant to hold CU custom place holders on the CINV PDF template

cah292: Ok, thanks for clarifying. I'll move the constant into that nested class.

PR #709, opened 2019-01-10
KFSPTS-13258 Final Bill and Billing Period editable

ajd299:
- Make Final Bill and Billing Period editable even if the billing period is expired
- Confirm before submit if the final bill indicator is checked


ajd299:
screen shot 2019-01-15 at 4 29 53 pm

cah292:
Do you only experience that error with that specific document, or do you see it on other docs as well? Also, have you tried clearing your browser cache or temporarily using a different browser?

ajd299:
Yes I get the same behavior in Firefox and Firefox in-private browsing (new session/cache)

image

cah292:
It looks like there were some impacting changes to the invoiceGeneral.tag file for the 2018-12-13 financials patch. Please make sure you're using the 2018-12-13 patch's version of that file for your overlay. Also, you should rebase your feature branch so that it uses the 2018-12-13 patch code that was recently moved to the develop branch.

ajd299:
It looks like rebasing caused some side effects to this pull request. I noticed a new pull request would only have 5 files changed. Is it okay if we try to disapprove this one and create a new one?

cah292:
Yes, creating a new PR is fine with me. Thanks!

PR #714, opened 2019-01-16
KFSPTS-13258 CINV make final bill and billing period editable

ajd299:


ajd299:
- Can you take a look at how/where I am doing the validation and error handling and give any feedback?
- Do you know if there is anyway to add a newline in the question text? I have tried "\n" and "<br/>"?
screen shot 2019-01-18 at 10 21 51 am

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +56,115 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+                }
+
+                Pair<Date, Date> awardDateRange = parseDateRange(contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getAwardDateRange());
+                Date awardStartDate = awardDateRange.getKey();
+                if (dateTimeService.dateDiff(billingPeriodStartDate, awardStartDate, false) >= 1) {
+                    warningMessages.add(configurationService.getPropertyValueAsString(CUKFSKeyConstants.WARNING_CINV_BILLING_PERIOD_START_DATE_BEFORE_AWARD_START_DATE));
+                }
+            } catch(ParseException ex) {
+                LOG.error("getContractsGrantsInvoiceDocumentWarningMessage: " + ex.getMessage());
+                warningMessages.add("ParseException occurred while parsing the billing period. Do you want to Proceed?[br]" + ex.getMessage());
```
cah292: Another warning message that could be moved to application resources and use placeholder substitution.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +56,115 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+            }
+        }
+
+        String ret = CollectionUtils.isEmpty(warningMessages) ? null : StringUtils.join(warningMessages, "[br]");
```
cah292: StringUtils.join() should already return a null/empty string for a null/empty list, so you may not need the extra null-check here. Also, it might be good to move the "[br]" separator to its own constant.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +56,115 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+                try {
+                    Pair<Date, Date> billingPeriod = parseDateRange(contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getBillingPeriod());
+                    Date billingEndDate = billingPeriod.getValue();
+                    DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
+                    if (dateTimeService.dateDiff(billingEndDate, contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getLastBilledDate(), false) != 0) {
+                        contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().setLastBilledDate(billingEndDate);
+                    }
+                }
+                catch(Exception ex) {
+                    LOG.error("promptForFinalBillConfirmation setting last billed date to end date: " + ex.getMessage());
```
cah292: Should execution of the action still continue if an error happens here, or should the document be returned to the user with an explanation of the error?

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -55,4 +179,15 @@ private KualiDecimal findAwardBudgetTotal(ContractsGrantsInvoiceDocumentForm con
         return budgetTotalAmount;
     }
 
+    private Pair<Date, Date> parseDateRange(String dateRange) throws ParseException {
+        if (StringUtils.isEmpty(dateRange) || dateRange.length() != 24) {
+            dateRange = dateRange == null ? KFSConstants.EMPTY_STRING : dateRange;
```
cah292: The lines above could use some of the cleanup mentioned earlier (use a constant for expected length, use a StringUtils helper method for null/blank defaults, etc.).

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -55,4 +179,15 @@ private KualiDecimal findAwardBudgetTotal(ContractsGrantsInvoiceDocumentForm con
...
+            throw new ParseException("The Date Range is Invalid [" + dateRange + "]. The expected length is 24 (YYYY-MM-DD to YYYY-MM-DD).", 0);
+        }
+        DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
+        Date endDate = dateTimeService.convertToSqlDate(dateRange.substring(14));
+        Date startDate = dateTimeService.convertToSqlDate(dateRange.substring(0, 10));
```
cah292: You may want constants for some of these literal indexes as well.

src/main/webapp/WEB-INF/tags/module/ar/invoiceGeneral.tag
```diff
@@ -0,0 +1,273 @@
+<%@ include file="/jsp/sys/kfsTldHeader.jsp"%>
+<%@ attribute name="readOnly" required="true" description="If document is in read only mode"%>
+
```
cah292: The KualiCo version of "invoiceGeneral.tag" has a license header at the top of it. You may want to include that header in the overlay as well.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +56,115 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+            ActionForward forward = promptForFinalBillConfirmation(mapping, form, request, response, KFSConstants.ROUTE_METHOD, warningMessage, contractsGrantsInvoiceDocument);
+            if (forward != null) {
+                return forward;
+            }
+        }
+
+        return super.route(mapping, form, request, response);
+    }
+
+    protected String getContractsGrantsInvoiceDocumentWarningMessage(ContractsGrantsInvoiceDocument contractsGrantsInvoiceDocument) {
```
cah292: This method seems to have too many of the validation pieces located directly within it. You may want to create separate helper methods for some of the validation sections if possible.

cah292:
Also, I would recommend double-checking the general formatting of this new code, to make sure it's adhering to our standards (appropriate use of spaces, putting "else" or "catch" on the same line as the closing curly brace of the previous block, etc.).

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
ajd299: I think that it would be a better user experience practice to stop and explain the error. Should I use GlobalVariables.getMessageMap().putError?

cah292: Yes, using GlobalVariables.getMessageMap().putError would be the best option here. Just make sure you use the KFS version of GlobalVariables and not the Rice one.

ajd299:
I've implemented the suggested changes. Please take a look and resolve conversations that are fixed.

For breaking up the validation, it seems like some code duplication has occurred. Could I use a separate class for this? Should it be in a certain file or nested within the CuContractsGrantsInvoiceDocumentAction class?

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +57,138 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+                LOG.error("getContractsGrantsInvoiceDocumentWarningMessage: " + ex.getMessage());
+                String warningMessage = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(CUKFSKeyConstants.WARNING_CINV_DATE_RANGE_PARSE_EXCEPTION);
+                warningMessages.add(warningMessage + CuArConstants.QUESTION_NEWLINE_STRING + ex.toString());
+            }
+        }
+
+        return StringUtils.join(warningMessages, CuArConstants.QUESTION_NEWLINE_STRING);
+    }
+
+    protected void validateBillingPeriodDateRange(ContractsGrantsInvoiceDocument contractsGrantsInvoiceDocument, List<String> warningMessages) throws ParseException {
```
cah292: To keep code duplication to a minimum while still keeping the method length reasonable, I would suggest taking the code from the three date validation methods below and combining them into this "validateBillingPeriodDateRange" method. That way, you don't have to keep repeating the variable initialization for start/end dates or the service references.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
cah292: I agree that this section needs to add an error to GlobalVariables and return to the user. Please update this area accordingly.

src/main/java/edu/cornell/kfs/sys/CUKFSKeyConstants.java
```diff
@@ -223,4 +223,11 @@
...
     public static final String ERROR_DOCUMENT_CONTRACT_GRANT_INVOICE_PRORATE_NO_AWARD_BUDGET_TOTAL = "error.document.contractsGrantsInvoice.prorate.no.award.budget.total";
+    public static final String WARNING_CINV_FINAL_BILL_INDICATOR = "warning.document.contractGrantsInvoice.final.bill.indicator";
+    public static final String WARNING_CINV_BILLING_PERIOD_END_DATE_BEFORE_LAST_BILLED_DATE = "warning.document.contractGrantsInvoice.billingPeriodEndDate.before.lastBilledDate";
+    public static final String WARNING_CINV_BILLING_PERIOD_END_DATE_BEFORE_BILLING_PERIOD_START_DATE = "warning.document.contractGrantsInvoice.billingPeriodEndDate.before.billingPeriodStartDate";
+    public static final String WARNING_CINV_BILLING_PERIOD_START_DATE_BEFORE_LAST_BILLED_DATE = "warning.document.contractGrantsInvoice.billingPeriodStartDate.before.lastBilledDate";
+    public static final String WARNING_CINV_BILLING_PERIOD_START_DATE_BEFORE_AWARD_START_DATE = "warning.document.contractGrantsInvoice.billingPeriodStartDate.before.awardStartDate";
+    public static final String WARNING_CINV_DATE_RANGE_INVALID_FORMAT_LENGTH = "warning.document.contractGrantsInvoice.dateRange.expectedFormat";
```
cah292: This constant's name doesn't quite match what would be expected for the given error key. I would recommend updating the constant name or error key accordingly for naming consistency.

ajd299:
I've implemented the suggested changes. Please take a look and resolve conversations that are fixed and give any additional feedback on the refactored code.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +58,121 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+            } else if (ConfirmationQuestion.YES.equals(buttonClicked) && contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().isFinalBillIndicator()) {
+                try {
+                    Pair<Date, Date> billingPeriod = parseDateRange(contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getBillingPeriod());
+                    Date billingEndDate = billingPeriod.getRight();
+                    DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
+                    if (dateTimeService.dateDiff(billingEndDate, contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getLastBilledDate(), false) != 0) {
+                        contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().setLastBilledDate(billingEndDate);
+                    }
+                } catch(Exception ex) {
+                    GlobalVariables.getMessageMap().putError(KFSConstants.GLOBAL_ERRORS, KFSKeyConstants.ERROR_CUSTOM, "promptForFinalBillConfirmation setting last billed date to end date: " + ex.getMessage());
```
cah292: Now that this message is going to be displayed to the user, please update the wording accordingly to make it more user-friendly.

Also, should this code block return to the MAPPING_BASIC forward after adding the error to GlobalVariables?

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +58,121 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+
+        String billingPeriod = contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getBillingPeriod();
+        if (StringUtils.length(billingPeriod) != CuArConstants.CINV_DATE_RANGE_EXPECTED_FORMAT_LENGTH) {
+            billingPeriod = StringUtils.defaultString(billingPeriod);
+            String warningMessage = SpringContext.getBean(ConfigurationService.class).getPropertyValueAsString(CUKFSKeyConstants.WARNING_CINV_DATE_RANGE_INVALID_FORMAT_LENGTH);
+            warningMessages.add(MessageFormat.format(warningMessage, new String[] {billingPeriod}));
+        } else {
+            try {
+                validateBillingPeriodDateRange(contractsGrantsInvoiceDocument, warningMessages);
+            } catch(ParseException ex) {
```
cah292: Missing whitespace before opening parentheses.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +58,121 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+                forward = mapping.findForward(KFSConstants.MAPPING_BASIC);
+            } else if (ConfirmationQuestion.YES.equals(buttonClicked) && contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().isFinalBillIndicator()) {
+                try {
+                    Pair<Date, Date> billingPeriod = parseDateRange(contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getBillingPeriod());
+                    Date billingEndDate = billingPeriod.getRight();
+                    DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
+                    if (dateTimeService.dateDiff(billingEndDate, contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getLastBilledDate(), false) != 0) {
+                        contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().setLastBilledDate(billingEndDate);
+                    }
+                } catch(Exception ex) {
```
cah292: Missing whitespace before opening parentheses.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
cah292: MessageMap.putError() will just print the referenced error message and parameters; it won't print the stack trace unless you explicitly include it (which would not be user-friendly). I would suggest adding a separate LOG.error() call to log more details about the exception and its stack trace.

src/main/java/edu/cornell/kfs/module/ar/document/web/struts/CuContractsGrantsInvoiceDocumentAction.java
```diff
@@ -41,6 +59,124 @@ public ActionForward prorateBill(ActionMapping mapping, ActionForm form, HttpSer
...
+                    Date billingPeriodEndDate = billingPeriod.getRight();
+                    DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
+                    if (dateTimeService.dateDiff(billingPeriodEndDate, contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().getLastBilledDate(), false) != 0) {
+                        contractsGrantsInvoiceDocument.getInvoiceGeneralDetail().setLastBilledDate(billingPeriodEndDate);
+                    }
+                } catch (Exception ex) {
+                    ConfigurationService configurationService = SpringContext.getBean(ConfigurationService.class);
+                    String message = configurationService.getPropertyValueAsString(CUKFSKeyConstants.ERROR_CINV_SETTING_LAST_BILLED_DATE);
+                    GlobalVariables.getMessageMap().putError(KFSConstants.GLOBAL_ERRORS, KFSKeyConstants.ERROR_CUSTOM, message);
+                    return mapping.findForward(RiceConstants.MAPPING_BASIC);
```
cah292: KFSConstants also has the MAPPING_BASIC constant. I would suggest using that here, for consistency with the rest of this class.

PR #715, opened 2019-01-17
KFSPTS-14298 upgrade to the 12/20/18 and 1/10/19 version of financials

db483:


db483:
Chad,

Do you have some specific examples in mind when referring to the various attachment download operations? Thank you.

cah292:
I think we would need to test downloading attachments on KFS documents, as well as the template-download action(s) on CGB lookups and the CSV-export feature on Letter of Credit Review Documents.

db483:
Thank you Chad, I have updated the jira with additional information.

db483:
Good questions Chad.

I need some help with the changes related to FINP-4572, I will ping you when you get in.

Regarding the Amazon S3 version of the attachment service, I have checked with KualiCo and they said it is optional and it is disabled by default.

PR #716, opened 2019-01-17
KFSPTS-14389 Fix new-sub-account routing setup for GSACC

cah292:
This fix is needed to correct an issue that came up during the KFSPTS-14181 DEV validation.


ajd299:
Code changes look good, I noticed that sometimes the property tag is indented and sometimes it is not. This is however already not consistent in the existing code. I will merge once the tests pass

PR #717, opened 2019-01-18
KFSPTS-14402 Fix CNCR job to do file search by relative path

cah292:
The 2018-12-13 financials patch required us to switch this one file-searching area of Concur to use the updated batch file lookup. However, that lookup is now expecting the search directory criteria to be specified as relative directories. This PR updates the appropriate section to use relative directory criteria instead.


nkimble:

PR #718, opened 2019-01-24
KFSPTS-13473 Allow negative item quantity/amount on IB/SB

cah292:
The functionals have requested that certain IB/SB item fields should allow negative values, so this PR implements those changes. Note that IB/SB items only undergo very simple validation and I believe they are informational only; thus, these changes should not negatively affect any GL entries that get generated.

Unlike other item BOs, the ones for IB/SB docs treat quantity as an integer instead of a decimal number. The existing Rice/KFS validation patterns are either for decimal numbers or nonnegative integers, so I added a new pattern that allows for a larger integer range. I did that to replace the quantity field's existing validation pattern, which only allowed for nonnegative integers before.

### DIFF
--- a/src/main/java/edu/cornell/kfs/kns/datadictionary/validation/fieldlevel/IntegerValidationPattern.java
+++ b/src/main/java/edu/cornell/kfs/kns/datadictionary/validation/fieldlevel/IntegerValidationPattern.java
@@ -1,0 +1,22 @@
+package edu.cornell.kfs.kns.datadictionary.validation.fieldlevel;
+
+import org.kuali.kfs.krad.datadictionary.validation.FieldLevelValidationPattern;
+
+import edu.cornell.kfs.sys.CUKFSConstants;
+
+/**
+ * Validation pattern that allows for integer values only.
+ * Unlike the "NumericValidation" helper bean, this implementation
+ * allows for negative integers.
+ */
+@SuppressWarnings("deprecation")
+public class IntegerValidationPattern extends FieldLevelValidationPattern {
+
+    private static final long serialVersionUID = 8850022886274526658L;
+
+    @Override
+    protected String getPatternTypeName() {
+        return CUKFSConstants.INTEGER_VALIDATION_PATTERN_TYPE;
+    }
+
+}
--- a/src/main/resources/CU-ApplicationResources.properties
+++ b/src/main/resources/CU-ApplicationResources.properties
@@ -693,4 +693,4 @@ pdp.send.ach.notification.error.report.title=PDP Send ACH Advice Notification Er
 pdp.send.ach.notification.error.report.line.format=%25s %25s %25s %55s
 
 validationPatternRegex.integer=-?[0-9]+
-error.format.edu.cornell.kfs.kns.datadictionary.validation.fieldlevel.IntegerValidationPattern=The {0} must be an integer.
+error.format.edu.cornell.kfs.krad.datadictionary.validation.fieldlevel.IntegerValidationPattern={0} must be an integer.
--- a/src/main/resources/CU-ApplicationResources.properties
+++ b/src/main/resources/CU-ApplicationResources.properties
@@ -691,3 +691,6 @@ error.document.contractsGrantsInvoice.prorate.no.award.budget.total=The budget t
 pdp.send.ach.notification.error.report.file.prefix=pdpSendAchAdviceNotificationsErrorReport
 pdp.send.ach.notification.error.report.title=PDP Send ACH Advice Notification Error Report
 pdp.send.ach.notification.error.report.line.format=%25s %25s %25s %55s
+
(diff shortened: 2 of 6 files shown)


src/main/java/edu/cornell/kfs/kns/datadictionary/validation/fieldlevel/IntegerValidationPattern.java
```diff
@@ -0,0 +1,22 @@
+package edu.cornell.kfs.kns.datadictionary.validation.fieldlevel;
```
jhulslander: since this object extends FieldLevelValidationPattern and that is in the KRAD package, should this object also be in KRAD?

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -691,3 +691,6 @@ error.document.contractsGrantsInvoice.prorate.no.award.budget.total=The budget t
 pdp.send.ach.notification.error.report.file.prefix=pdpSendAchAdviceNotificationsErrorReport
 pdp.send.ach.notification.error.report.title=PDP Send ACH Advice Notification Error Report
 pdp.send.ach.notification.error.report.line.format=%25s %25s %25s %55s
+
+validationPatternRegex.integer=-?[0-9]+
+error.format.edu.cornell.kfs.kns.datadictionary.validation.fieldlevel.IntegerValidationPattern=The {0} must be an integer.
```
jhulslander: Will "[0]" be a field name or a value?  If it is a field name, I think "the" can be eliminated.  The error doesn't read all that well.

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/InternalBillingItem.xml
```diff
@@ -0,0 +1,19 @@
...
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+                http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
+
+    <!--
+        To properly allow the Data Dictionary to interpret the "exclusiveMin" property as unset,
+        we must set it to null instead of an empty string.
+     -->
+
+    <bean id="InternalBillingItem-itemUnitAmount" parent="InternalBillingItem-itemUnitAmount-parentBean"
+            p:exclusiveMin="#{null}"/>
```
jhulslander: I haven't seen exclusiveMin in DD before.  What does it do (I'd like to understand)?

src/main/java/edu/cornell/kfs/kns/datadictionary/validation/fieldlevel/IntegerValidationPattern.java
cah292: Even though the FieldLevelValidationPattern superclass is in the KRAD package, the majority of its implementation subclasses are in the KNS package. I put this class in the KNS package for consistency with those other implementations, but I'm fine with relocating it to the KRAD package for consistency with the superclass instead. I'll proceed with that change.

src/main/resources/CU-ApplicationResources.properties
cah292: The placeholder is for a field name. I'll remove the extra "the" to make it easier to read. Thanks.

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/InternalBillingItem.xml
cah292: "exclusiveMin" is a constraint that can be added for DD-based field validation. It validates that a field's value is greater than the given value, though it only works for numeric or date values, or for strings formatted as numeric or date values. (There's also a related "inclusiveMax" constraint for restricting the upper bound.)

jhulslander: Thanks for the explanation.  Why did we need to add it to itemUnitAmount, since it doesn't look like we are applying different validation to it.

cah292: The "exclusiveMin" validation is independent of the validation pattern replacement that also needed to be made on one of the properties. In base code, both of the affected item fields had "exclusiveMin" values of 0, thus preventing users from entering negative values. That was why the "exclusiveMin" validation needed to be cleared out on the two fields.

cah292:
I've committed some changes to address the items that were found in the initial review. I also updated one of the XML comments to help clarify the intent of the null-bean-property-value setup.

src/main/resources/edu/cornell/kfs/fp/businessobject/datadictionary/InternalBillingItem.xml
jhulslander: Thanks for the clarification

PR #724, opened 2019-01-31
KFSPTS-14399 Fix fund manager routing on CINV documents

cah292:
The base service already has the bulk of the necessary logic for finding the appropriate fund managers. We just needed to override the appropriate method so that the logic could be tied into the right place for routing purposes.


ajd299:
Code looks good and is very readable and formatted correctly to our practices.
One suggestion I had was to consider extracting the building the criteria from the qualification to a helper method. Let me know your thoughts.

cah292:
I've updated the code to make the criteria-building section easier to read. This is ready for review again.

ajd299:
Changes look good, merging

PR #726, opened 2019-02-01
KFSPTS-14513 add changes to file search to improve lookup speed on Create Disencumbrance

db483:


db483:
This adds back a customization that was removed with the 12/13/18 financials upgrade.

db483:
Good point Chad, I will add them.

PR #731, opened 2019-02-05
KFSPTS-14460 remove  CULaborObjectService, CULaborObjectDao, and related unit test

jhulslander:


jhulslander:
I've updated the commit message as suggested.  This is ready for review again

PR #738, opened 2019-02-12
KFSPTS-14388: Convert file to p-notation and add missing base code attributes-beans for financialObjectLevel.financialConsolidationObjectCode.

nkimble:


src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/ObjectCode.xml
```diff
@@ -25,304 +25,254 @@
...
-    <property name="title" value="Object Code Lookup"/>
-    
-    <property name="defaultSort">
-      <bean parent="SortDefinition">
-        <property name="attributeNames">
-          <list>
-            <value>financialObjectCode</value>
-          </list>
+    <bean id="ObjectCode-lookupDefinition" parent="ObjectCode-lookupDefinition-parentBean"/>
+    <bean id="ObjectCode-lookupDefinition-parentBean" abstract="true" parent="LookupDefinition"
```
cah292: KualiCo has a third bean in this lookupDefinition hierarchy, which might have been needed for Access Security or some other related override. Does that third bean need to be added here now, or are you planning to address that in the follow-up PR?

cah292: Also, one of these base beans has a multipleValuesResultSetLimit of -1 (unlimited?). Should that be added in here yet?

nkimble: Good catches.  I have added both items noted because the multipleValuesResultSetLimit of -1 is set in a base parent bean that would not be present should the third bean you noted as being missing from this lookupDefinition hierarchy not be present.

PR #739, opened 2019-02-13
KFSPTS-14648 covert search results from BatchFile to CreateDoneBatchF…

db483:
…ile so that the correct getActionLinks method is called

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
@@ -1,12 +1,18 @@
 package edu.cornell.kfs.sys.businessobject.lookup;
 
+import java.io.FileNotFoundException;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import javax.ws.rs.core.MultivaluedMap;
+
 import org.apache.commons.lang.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.kuali.kfs.krad.bo.BusinessObjectBase;
 import org.kuali.kfs.krad.util.KRADConstants;
 import org.kuali.kfs.krad.util.UrlFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
```diff
@@ -1,12 +1,18 @@
...
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import javax.ws.rs.core.MultivaluedMap;
+
 import org.apache.commons.lang.StringUtils;
```
cah292: This import should be switched over to lang3 StringUtils instead.

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
```diff
@@ -16,12 +22,36 @@
...
+	private static final Logger LOG = LogManager.getLogger(CreateDoneBatchFileLookupableHelperServiceImpl.class);
 
 	protected CreateDoneBatchFileAuthorizationService createDoneAuthorizationService;
 
+	@Override
+	public List<BusinessObjectBase> getSearchResults(Class<? extends BusinessObjectBase> businessObjectClass,
+			MultivaluedMap<String, String> fieldValues) {
+
+		List<BusinessObjectBase> results = super.getSearchResults(businessObjectClass, fieldValues);
+		List createDoneBatchFiles = new ArrayList<CreateDoneBatchFile>();
```
cah292: This list is missing the use of generics. Should it just use BusinessObjectBase as the generic type instead?

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
```diff
@@ -16,12 +22,36 @@
...
+		for (Object file : results) {
+			BatchFile batchFile = (BatchFile) file;
+			CreateDoneBatchFile createDoneBatchFile;
+			try {
+				createDoneBatchFile = new CreateDoneBatchFile(batchFile.getId());
+				createDoneBatchFiles.add(createDoneBatchFile);
+			} catch (FileNotFoundException e) {
+				LOG.error(
+						"An error has occured while creating a CreateDoneBatchFile from the BatchFile object in the search results: "
```
cah292: The method name should be added to the error message.

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
db483: That throws an error, not sure what the generics should be.

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
db483: I will make the change

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
db483: I will add the method name.

src/main/java/edu/cornell/kfs/sys/businessobject/lookup/CreateDoneBatchFileLookupableHelperServiceImpl.java
cah292: What kind of error are you getting? It looks as though setting the list type to BusinessObjectBase instead of CreateDoneBatchFile would work. Or is that causing errors on other lines as well?

db483: That will work, I misunderstood your suggestion initially. Thank you.

db483:
Chad, I've made the changes and this is now ready for review.

PR #743, opened 2019-02-18
KFSPTS-14728 handling attachment mime types in create accounting document job better

jhulslander:


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -70,6 +70,25 @@ public Attachment createAttachmentFromBackupLink(Document document, AccountingXm
         }      
     }
 
+    protected String findMimeType(String uploadFileName) {
+        String mimeType = URLConnection.guessContentTypeFromName(uploadFileName);
+        if (StringUtils.isEmpty(mimeType)) {
+            LOG.error("findMimeType, could not determine mime type from file name using URLConnection object. the file name is " + uploadFileName);
+            String[] splitFileName = StringUtils.split(uploadFileName, ".");
+            if (splitFileName != null) {
+                int lastArrayElement = splitFileName.length - 1;
```
cah292: This variable should be renamed, to make it clear that it holds the element index and not the element itself.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    
+    @Test
+    public void testFindMimeTypeEmptyFileName() {
+        String mimeType = attachmentService.findMimeType(StringUtils.EMPTY);
+        assertEquals(StringUtils.EMPTY, mimeType);
+    }
+    
+    @Test
+    public void testFindMimeTypeNullFileName() {
+        String mimeType = attachmentService.findMimeType(StringUtils.EMPTY);
```
cah292: This test case is actually using an empty string instead of a null string. Should it be updated to use a null string, or should this test method just be removed altogether?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
jhulslander: can do

src/test/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImplTest.java
jhulslander: I'll fix this test case, good catch

jhulslander:
Sorry about the Jira number confusion.  In hindsight, I think this should be a seperate user story, I will change the commit message to reference a new jira number

jhulslander:
I've made some changes based on the code review and moved this to a new user story.  I've updated the initial commit message, and the PR description.When you are ready to merge this PR, please double check the commit message that it has the right jira number

jhulslander:
retest this please

PR #744, opened 2019-02-21
KFSPTS-14670 Add first-invoice suspension category for CINV

cah292:


src/main/java/edu/cornell/kfs/module/ar/CuArPropertyConstants.java
```diff
@@ -1,11 +1,23 @@
...
     public static class ContractsAndGrantsBillingAwardFields {
         public static final String FINAL_BILL = "finalBill";
         public static final String PARTIAL_BILL = "partialBill";
         public static final String TOTAL_PROGRAM_OUTLAYS_TO_DATE = "totalProgramOutlaysToDate";
+        public static final String ACCOUNT_DETAILS_CHART_OF_ACCOUNTS_CODE = ArPropertyConstants.ACCOUNT_DETAILS
+                + KFSConstants.DELIMITER + KFSPropertyConstants.CHART_OF_ACCOUNTS_CODE;
+        public static final String ACCOUNT_DETAILS_CONTRACT_CONTROL_CHART_OF_ACCOUNTS_CODE = ArPropertyConstants.ACCOUNT_DETAILS
+                + KFSConstants.DELIMITER + KFSPropertyConstants.ACCOUNT
+                + KFSConstants.DELIMITER + KFSPropertyConstants.CONTRACT_CONTROL_CHART_OF_ACCOUNTS_CODE;
+        public static final String ACCOUNT_DETAILS_CONTRACT_CONTROL_ACCOUNT_NUMBER = ArPropertyConstants.ACCOUNT_DETAILS
```
jhulslander: These variable names don't seem to represent the values being concatenated together.

src/main/java/edu/cornell/kfs/module/ar/document/validation/impl/FirstInvoiceForAwardSuspensionCategory.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private FinancialSystemDocumentService financialSystemDocumentService;
+    private ContractsGrantsInvoiceDocumentDao contractsGrantsInvoiceDocumentDao;
+
+    @Override
+    public boolean shouldSuspend(ContractsGrantsInvoiceDocument contractsGrantsInvoiceDocument) {
+        return priorSuccessfulInvoicesDoNotExist(contractsGrantsInvoiceDocument);
+    }
+
+    protected boolean priorSuccessfulInvoicesDoNotExist(ContractsGrantsInvoiceDocument contractsGrantsInvoiceDocument) {
```
jhulslander: This logic reads a little difficult to me.  Do you think it would be easier to read if this function were "priorSuccessfullInvoicesExist" . And then suspect only if that function is false?  I think it may be better to avoid the negative logic.  What do you think?

src/main/java/edu/cornell/kfs/module/ar/CuArPropertyConstants.java
cah292: After further consideration, I think I can improve the variable names by introducing a new nested class to CuArPropertyConstants and putting the variables in there. I'll work on restructuring these variables/constants accordingly.

src/main/java/edu/cornell/kfs/module/ar/document/validation/impl/FirstInvoiceForAwardSuspensionCategory.java
cah292: Yes, I agree that this would make a little more sense as a positive-named method instead. I was initially trying to avoid the use of "!" in the shouldSuspend() method, but doing so may have made this more confusing. I'll update the name of this method and how it's used by shouldSuspend().

cah292:
I've made some more updates to this PR. I've revised the method that shouldSuspend() depends on so that the logic is now (hopefully) a little clearer. I also reorganized the account detail field constants, and added a new constant for one of the referenced fields so that its setup is more consistent with the related constants.

PR #746, opened 2019-02-28
KFSPTS-14446 persist action notes in document requeuer job

jhulslander:
Please don't merge this yet.  I need to do some more local validation, and I want to see what happens with the prod document requeuer issue.

I used SpringContext and service locator functions to get services instead of using Spring XML configuration as these services are registered in Spring when he applicable spring context is loading.  Also, to keep this transactional, I apparently need to get  a reference to the services inside the for loop.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
+++ b/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
@@ -7,15 +7,17 @@ public class ActionItemNoteDetailDto {
     private String docHeaderId;
     private String actionNote;
     private Timestamp noteTimeStamp;
+    private String orginalActionItemId;
     
     public ActionItemNoteDetailDto() {
         
     }
     
-    public ActionItemNoteDetailDto(String principleId, String docHeaderId, String actionNote, Timestamp noteTimeStamp) {
+    public ActionItemNoteDetailDto(String principleId, String docHeaderId, String actionNote, String orginalActionItemId, Timestamp noteTimeStamp) {
         this.principleId = principleId;
         this.docHeaderId = docHeaderId;
         this.actionNote = actionNote;
+        this.orginalActionItemId = orginalActionItemId;
         this.noteTimeStamp = noteTimeStamp;
     }
 
@@ -51,6 +53,14 @@ public class ActionItemNoteDetailDto {
         this.noteTimeStamp = noteTimeStamp;
     }
 
+    public String getOrginalActionItemId() {
+        return orginalActionItemId;
+    }
+
+    public void setOrginalActionItemId(String orginalActionItemId) {
+        this.orginalActionItemId = orginalActionItemId;
+    }
+
     @Override
     public String toString() {
         StringBuilder sb = new StringBuilder();
@@ -58,6 +68,7 @@ public class ActionItemNoteDetailDto {
         sb.append("' docHeaderId: '").append(docHeaderId);
         sb.append("' noteTimeStamp: '").append(noteTimeStamp).append("'");
         sb.append("' actionNote: '").append(actionNote).append("'");
+        sb.append("' original action item id: '").append(orginalActionItemId).append("'");
         return sb.toString();
     }
 
--- a/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
+++ b/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
@@ -1,8 +1,10 @@
 package edu.cornell.kfs.sys.dataaccess;
 
+import java.io.Serializable;
 import java.sql.Timestamp;
 
-public class ActionItemNoteDetailDto {
+public class ActionItemNoteDetailDto implements Serializable {
+    private static final long serialVersionUID = 9115076028165986273L;
     private String principleId;
     private String docHeaderId;
     private String actionNote;
@@ -66,8 +68,8 @@ public class ActionItemNoteDetailDto {
         StringBuilder sb = new StringBuilder();
         sb.append("principleId: '").append(principleId);
         sb.append("' docHeaderId: '").append(docHeaderId);
-        sb.append("' noteTimeStamp: '").append(noteTimeStamp).append("'");
-        sb.append("' actionNote: '").append(actionNote).append("'");
+        sb.append("' noteTimeStamp: '").append(noteTimeStamp);
+        sb.append("' actionNote: '").append(actionNote);
         sb.append("' original action item id: '").append(orginalActionItemId).append("'");
         return sb.toString();
     }
--- a/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
+++ b/src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
@@ -1,18 +1,22 @@
 package edu.cornell.kfs.sys.dataaccess;
 
+import java.sql.Timestamp;
+
 public class ActionItemNoteDetailDto {
     private String principleId;
     private String docHeaderId;
     private String actionNote;
+    private Timestamp noteTimeStamp;
     
     public ActionItemNoteDetailDto() {
         
     }
     
-    public ActionItemNoteDetailDto(String principleId, String docHeaderId, String actionNote) {
+    public ActionItemNoteDetailDto(String principleId, String docHeaderId, String actionNote, Timestamp noteTimeStamp) {
         this.principleId = principleId;
         this.docHeaderId = docHeaderId;
         this.actionNote = actionNote;
+        this.noteTimeStamp = noteTimeStamp;
     }
 
     public String getPrincipleId() {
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -1,45 +1,100 @@
 package edu.cornell.kfs.sys.service.impl;
 
+import edu.cornell.kfs.sys.CUKFSConstants;
+import edu.cornell.kfs.sys.dataaccess.ActionItemNoteDetailDto;
 import edu.cornell.kfs.sys.dataaccess.DocumentMaintenanceDao;
 import edu.cornell.kfs.sys.service.DocumentMaintenanceService;
+
+import org.apache.commons.collections.CollectionUtils;
```
cah292: This should use the collections4 version of CollectionUtils instead.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -1,45 +1,100 @@
...
             LOG.info("requeueDocuments: Requesting requeue for document: " + docId);
-            DocumentRefreshQueue documentRequeuer = KewApiServiceLocator.getDocumentRequeuerService(CoreConfigHelper.getApplicationId(), docId, WAIT_TIME_NO_WAIT);
+            DocumentRefreshQueue documentRequeuer = (DocumentRefreshQueue) SpringContext.getService(CUKFSConstants.DOCUMENT_REFRESH_QUEUE_SERVICE_SPRING_CONTEXT_NAME);
             documentRequeuer.refreshDocument(docId);
+            
+            List<ActionItemNoteDetailDto> noteDetailsForDocument = findNoteDetailsForDocument(noteDetails, docId);
+            for (ActionItemNoteDetailDto detailDto : noteDetailsForDocument) {
+                List<ActionItem> actionItems = findActionItem(detailDto);
+                if (CollectionUtils.isNotEmpty(actionItems)) {
+                    ActionItem item = actionItems.get(0);
```
cah292: Users can potentially have multiple action items on the same document; not sure how likely that would happen in this requeue case. You may want to force the code to work with the user's highest-priority action item for the document, to ensure that the note doesn't get added to the wrong item. Such prioritization can be handled by Rice's ActionListPriorityComparator class, and by Cynergy's ActionListBuilder.ActionListPriorityAndIdComparator class. (The latter might be preferable because of our Cynergy action list customizations.)

Alternately, if this requeue is only happening in a very limited case where a user would not have multiple action items (not even a "READ NOTES" FYI in addition to APPROVE), then such prioritization-grabbing might not be needed.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -1,45 +1,100 @@
...
+                        KRADServiceLocator.getDataObjectService().save(actionItemExtension);
+                    }
+                }
+            }
         }
-
         return true;
     }
+    
+    private ActionItemExtension findActionItemEtencsion(String actionItemId) {
```
cah292: Minor typo in method name, should be changed from "findActionItemEtencsion" to "findActionItemExtension"

src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
```diff
@@ -0,0 +1,77 @@
+package edu.cornell.kfs.sys.dataaccess;
+
+import java.io.Serializable;
+import java.sql.Timestamp;
+
+public class ActionItemNoteDetailDto implements Serializable {
+    private static final long serialVersionUID = 9115076028165986273L;
+    private String principleId;
```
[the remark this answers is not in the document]
jhulslander: OK, I will change

src/main/java/edu/cornell/kfs/sys/dataaccess/ActionItemNoteDetailDto.java
```diff
@@ -0,0 +1,77 @@
...
+    private String docHeaderId;
+    private String actionNote;
+    private Timestamp noteTimeStamp;
+    private String orginalActionItemId;
```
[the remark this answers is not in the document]
jhulslander: OK, I will change

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
```diff
@@ -167,6 +148,91 @@ private String retrieveTableNameFromAnnotations(Class className) {
...
+                addDocumentIdAndRoleIdsToSelectStatement(selectStatement, docTypeIds, roleIds);
+
+                queryResultSet = selectStatement.executeQuery();
+
+                while (queryResultSet.next()) {
+                    String principleId = queryResultSet.getString(1);
+                    String docHeaderId = queryResultSet.getString(2);
+                    String actionNote = queryResultSet.getString(3);
+                    Timestamp noteTimeStamp = queryResultSet.getTimestamp(4, Calendar.getInstance());
+                    String origianlActionItemId = queryResultSet.getString(5);
```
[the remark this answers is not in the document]
jhulslander: OK, I will change

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
jhulslander: good catch

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
jhulslander: Currently we are only requeueing fiscal officer action items.  I haven't seen multiple action items for a single person/document in my local testing,  Let me see if I can the priority processing you mentioned

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
jhulslander: OK, I will fix

jhulslander:
I've made some changes based off the code review, this is ready for review again. ActionListBuilder.ActionListPriorityAndIdComparator is protected, as is many functions in ActiionListBuilder.  I brought some functionality in ActionListBuilder into DocumentMaintenanceServiceImpl . Let me know if you think an alternative approach would be better.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -1,45 +1,112 @@
...
+    
+    private ActionItem findActionItem(ActionItemNoteDetailDto detailDto) {
+        QueryByCriteria query = QueryByCriteria.Builder.fromPredicates(
+                PredicateFactory.equal(KFSPropertyConstants.PRINCIPAL_ID, detailDto.getPrincipalId()),
+                PredicateFactory.equal(CUKFSConstants.DOCUMENT_ID, detailDto.getDocHeaderId()));
+        List<ActionItem> items = KRADServiceLocator.getDataObjectService().findMatching(ActionItem.class, query).getResults();
+        ActionListBuilder actionListBuilder = new ActionListBuilder();
+        ActionItem selectedActionItem = null;
+        Comparator<ActionItemContract> priorityComparator = new ActionListPriorityComparator();
+        for (ActionItem currentItem : items) {
```
cah292: You could potentially simplify this search by using the Collections.max() method, though you may need to check for an empty collection before invoking it.

jhulslander:
I think the collections.max function makes this more readable code, thanks for the suggestion.  I have made the change, and this is ready for review again

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -1,45 +1,108 @@
...
     }
+    
+    private ActionItem findActionItem(ActionItemNoteDetailDto detailDto) {
+        QueryByCriteria query = QueryByCriteria.Builder.fromPredicates(
+                PredicateFactory.equal(KFSPropertyConstants.PRINCIPAL_ID, detailDto.getPrincipalId()),
+                PredicateFactory.equal(CUKFSConstants.DOCUMENT_ID, detailDto.getDocHeaderId()));
+        List<ActionItem> actionItems = KRADServiceLocator.getDataObjectService().findMatching(ActionItem.class, query).getResults();
+        ActionItem selectedActionItem = null;
+        if (LOG.isDebugEnabled() ) {
+            LOG.info("findActionItem, number of action items for principle " +  detailDto.getPrincipalId() + " and cocument number "
```
cah292: There are a couple typos in this message that should be fixed ("principle" and "cocument"). Also, since the block is checking whether debug logging is enabled, I would suggest logging this message at the debug level, unless you meant it to be displayed at the info level (in which case the "if" condition should be changed instead).

jhulslander:
thanks, I fixed the typos and the log level

PR #747, opened 2019-03-05
KFSPTS-14586 remove customized parse method as the convertParsedObject…

db483:
…ToVO method is now called in base class

### DIFF
--- a/src/main/java/edu/cornell/kfs/vnd/batch/VendorInactivateConvertBatchCsvInputFileType.java
+++ b/src/main/java/edu/cornell/kfs/vnd/batch/VendorInactivateConvertBatchCsvInputFileType.java
@@ -73,22 +73,6 @@ public class VendorInactivateConvertBatchCsvInputFileType  extends CsvBatchInput
         return true;
     }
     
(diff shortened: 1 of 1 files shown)


db483:
Hi Chad,

I have tested this locally. I was able to first reproduce the issue and then validate that after the fix the job no longer throws an exception and completes successfully.

The change may be a bit confusing so let me try to explain the issue.

In the base CsvBatchInputFileTypeBase class the parse method used to just return the parsedContents without calling the convertParsedObjectToVO. With FINP-4262 the parse method was updated to return convertParsedObjectToVO(parsedContents).

The customized parse method in VendorInactivateConvertBatchCsvInputFileType.java was also calling convertParsedObjectToVO. This call was needed before the changes in FINP-4262, but after it is no longer needed as the method is already called in the super. The exception occurred on this second call to the convertParsedObjectToVO method as the value passed in was no longer the expected one - a list of Map but a List of VendorInactivateConvertBatch objects.

Please let me know if you have further questions or you would like me to share the test file.

Thank you.

PR #752, opened 2019-03-11
KFSPTS-9705 clean up custom bank code validation

cah292:
This PR updates our customized bank code validation to be much cleaner, and to also make use of superclass functionality where possible. The customization has also been modified to be more consistent with the superclass changes that were made in base code.

The old customization had a feature for auto-clearing the bank account (and warning the user) in certain cases. We do not appear to need such cases anymore, so that part of the functionality has been removed.

I also added a new static method to properly allow the CuDisbursementVoucherBankCodeValidation class to call our custom code. (No changes were needed to CuDisbursementVoucherBankCodeValidation itself, due to the way it was trying to call the static method before.)

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuAccountsPayableBankCodeValidation.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuAccountsPayableBankCodeValidation.java
@@ -1,55 +1,90 @@
 package edu.cornell.kfs.module.purap.document.validation.impl;
 
 import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.krad.util.GlobalVariables;
 import org.kuali.kfs.module.purap.PurapPropertyConstants;
 import org.kuali.kfs.module.purap.document.AccountsPayableDocumentBase;
 import org.kuali.kfs.module.purap.document.PaymentRequestDocument;
 import org.kuali.kfs.module.purap.document.VendorCreditMemoDocument;
 import org.kuali.kfs.module.purap.document.validation.impl.AccountsPayableBankCodeValidation;
+import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.KFSParameterKeyConstants;
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.sys.businessobject.Bank;
+import org.kuali.kfs.sys.context.SpringContext;
 import org.kuali.kfs.sys.document.validation.event.AttributedDocumentEvent;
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuAccountsPayableBankCodeValidation.java
```diff
@@ -1,55 +1,61 @@
...
+    // This method is private on the superclass, so it has been copied into this class and tweaked accordingly.
+    protected boolean isDocumentTypeUsingBankCode(AccountsPayableDocumentBase apDocument) {
+        String documentTypeName = apDocument.getDocumentHeader().getWorkflowDocument().getDocumentTypeName();
+        ParameterEvaluator evaluator = getParameterEvaluatorService().getParameterEvaluator(
+                Bank.class, KFSParameterKeyConstants.BANK_CODE_DOCUMENT_TYPES, documentTypeName);
+        return evaluator.evaluationSucceeds();
+    }
+
+    protected ParameterEvaluatorService getParameterEvaluatorService() {
+        return SpringContext.getBean(ParameterEvaluatorService.class);
```
jhulslander: should ParameterEvaluatorService be made a member variable, and only call sprincontext here if the member variable isn't set?

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -354,9 +354,9 @@ error.document.paymentmethod.flagrequired=At least one of the {0}, {1}, and {2}
 error.document.paymentmethod.not.pdp.and.interdept=A payment method may not be both interdepartmental and processed by PDP.
 error.document.paymentmethod.no.bank.when.interdept=A bank code may not be entered when the payment method is an interdepartmental charge.
 
-# Additional DV document messages - KITT-592
-error.document.disbursementvoucher.bank.required=A bank code is required for payment method {0}.
-warning.document.disbursementvoucher.bank.not.required=Bank codes are not required on payment method {0}, the entered value has been removed.
+# Additional bank-related document messages - KITT-592
```
jhulslander: Can this comment just be removed, I don't think it adds much value any longer

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuAccountsPayableBankCodeValidation.java
cah292: Yes, I think I can refactor this to use Spring to inject the service. Thanks for the suggestion.

src/main/resources/CU-ApplicationResources.properties
cah292: Ok, I'll remove the comment.

PR #755, opened 2019-03-15
KFSPTS-14929: Changes so CINV edoc description, invoice template, and GLPE properly reference the contract control account.

nkimble:


nkimble:
These changes are based upon KualiCo changes for FINP-4726 which fixed this same issue for the GLCP items.  My question to the reviewer is: Should the method in base code where this customization was modeled from also be locally customized so that all three places needing this data are using the same logic to obtain the contact control account number?

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -549,5 +552,29 @@ public void recalculateObjectCodeByCategory(ContractsGrantsInvoiceDocument contr
         super.recalculateObjectCodeByCategory(contractsGrantsInvoiceDocument, invoiceDetail, total, invoiceDetailAccountObjectCodes);
     }
     
+    @Override
+    protected String getRecipientAccountNumber(List<InvoiceAccountDetail> accountDetails) {
```
db483: I could not find where this method is being used

nkimble: This method is being called on line 93 of this file.

db483: Thank you for explaining, that looks good.

nkimble:
Noting Daniela and I discussed off-line the fact that FINP-4726 changed base code file ContractsGrantsInvoiceDocumentServiceImpl lines 320 to 347 to address the GLCP issue for contract control number.  

That is where I obtained the bulk of the logic in this customization of that same data for these two additional places.  To ensure the same logic is used in ALL three CGB locations where contract control account is needed, the base code method from FINP-4726 will also be customized locally to use method "determineContractControlAccount". Please do not merge until this change is added.

nkimble:
All three locations were this change is needed to reference the contract control account have been made. Ready for review again.

PR #756, opened 2019-03-22
KFSPTS-14884 Fix scroll position bug in some browsers

cah292:
This is nearly identical to the "page.tag" file from the base financials code. The only change is to the generation of the "body" tag within the "c:otherwise" block, where instead of invoking the restoreScrollPosition() function directly, the function is now being passed into a setTimeout() function call so that its execution will be delayed slightly. This fixes an issue where some browsers (most notably IE) are not properly restoring the previous scroll position immediately following a change to the field focus.


ajd299:
Code looks good, merging.

PR #761, opened 2019-04-01
KFSPTS-14894 Add Agency Extension and Agency Origin

cah292:


src/main/java/edu/cornell/kfs/module/cg/businessobject/options/AgencyOriginValuesFinder.java
```diff
@@ -0,0 +1,35 @@
...
+public class AgencyOriginValuesFinder extends KeyValuesBase {
+
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public List<KeyValue> getKeyValues() {
+        Collection<AgencyOrigin> agencyOrigins = getKeyValuesService().findAll(AgencyOrigin.class);
+        List<KeyValue> keyValues = new ArrayList<>();
+        keyValues.add(new ConcreteKeyValue(KFSConstants.EMPTY_STRING, KFSConstants.EMPTY_STRING));
+        agencyOrigins.stream()
```
jhulslander: should we add a sort here?

src/main/resources/edu/cornell/kfs/module/cg/businessobject/datadictionary/AgencyOrigin.xml
```diff
@@ -0,0 +1,128 @@
...
+            p:validationPattern-ref="AnyCharacterWithWhitespaceValidation">
+        <property name="control">
+            <bean parent="TextControlDefinition" p:size="55"/>
+        </property>
+    </bean>
+
+    <bean id="AgencyOrigin-active" parent="AgencyOrigin-active-parentBean"/>
+    <bean id="AgencyOrigin-active-parentBean" abstract="true" parent="GenericAttributes-activeIndicator" p:name="active"/>
+
+    <!-- Business Object Inquiry Definition -->
```
jhulslander: I think this comment can be removed

src/main/java/edu/cornell/kfs/module/cg/document/CuAgencyMaintainableImpl.java
```diff
@@ -0,0 +1,26 @@
...
+     * some custom processing that will prematurely save the agency before performing any auto-setup
+     * of keys on the referenced objects. Thus, the agency extension needs to have its agency number
+     * manually set here, to avoid errors when persisting it.
+     */
+    @Override
+    public void prepareForSave() {
+        super.prepareForSave();
+        Agency agency = (Agency) getDataObject();
+        AgencyExtendedAttribute agencyExtension = (AgencyExtendedAttribute) agency.getExtension();
+        agencyExtension.setAgencyNumber(agency.getAgencyNumber());
```
jhulslander: should there be error handling if the agency doesn't have an agency number yet?  Such as in the initial save scenario?

src/main/java/edu/cornell/kfs/module/cg/businessobject/options/AgencyOriginValuesFinder.java
cah292: It looks like KeyValuesService has a search method that allows for sorting, so I will use that to perform the sort. (For now I'll put the sort in descending order so that the "USA" drop-down option will be the first non-blank item.) Thanks for the suggestion.

src/main/resources/edu/cornell/kfs/module/cg/businessobject/datadictionary/AgencyOrigin.xml
cah292: Ok, I'll remove this comment.

src/main/java/edu/cornell/kfs/module/cg/document/CuAgencyMaintainableImpl.java
cah292: The Agency Number is a required field, so it should have a value once the document is successfully submitted for routing. Prior to routing, the agency extension will only be stored in the maintenance XML and not in the agency extension table, so I don't think a null agency number would break anything here.

jhulslander: OK, thanks for clarifying

cah292:
I've implemented some more changes based on the initial feedback. This is ready for review again.

PR #764, opened 2019-04-08
KFSPTS-14984: Gracefully handle PaymentWorks sending values too long for KFS staging table columns.

nkimble:


src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
```diff
@@ -121,6 +121,12 @@
         public static final String NEW_VENDOR_REQUESTS_SUPPLIER_UPLOAD_REPORT_NAME = "New Vendor Requests Upload Suppliers to PaymentWorks";
     }
     
+    public static final class PaymentWorksVendorStagingTableParseExceptionConstants {
+        public static final String DEFAULT_EXCEPTION_MESSAGE = "Could not determine exception failure reason for initial KFS staging table save of PaymentWorks vendor.";
+        public static final String EXCEPTION_REASON_START_STRING = "value too large for column";
+        public static final char EXCEPTION_REASON_END_CHAR = ')';
```
jhulslander: I am wondering if these constants should be renamed.  Something to indicate the default is a message added to the a user report, and the other two are used in an oracle error message search.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -351,17 +354,37 @@ private PaymentWorksVendor setStatusValuesForPaymentWorksProcessedNewVendorRejec
...
     }
     
+    private String parseSqlExceptionStringForOracleError(String exceptionString) {
+        int startPosition = exceptionString.indexOf(PaymentWorksConstants.PaymentWorksVendorStagingTableParseExceptionConstants.EXCEPTION_REASON_START_STRING);
+        int endPosition = exceptionString.indexOf(PaymentWorksConstants.PaymentWorksVendorStagingTableParseExceptionConstants.EXCEPTION_REASON_END_CHAR, startPosition);
+        String reasonForSqlException = PaymentWorksConstants.PaymentWorksVendorStagingTableParseExceptionConstants.DEFAULT_EXCEPTION_MESSAGE;
+        
+        if (requestedIndexWasFound(startPosition) && requestedIndexWasFound(endPosition)) {
+            reasonForSqlException = exceptionString.substring(startPosition, (endPosition + 1));
+        }
```
jhulslander: do you think it would be helpful to add an else clause logging that the error message couldn't be parsed?

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
nkimble: I struggled with this since the actual exception being generated is not an true java.sql.SQLException where it is being caught. Rice is wrapping it weirdly and what is being seen by the parse routine is the following before it gets to the actual stack trace with that toString call; therefore, I decided to get the first occurrence of why the column size is too small and keep the naming more generic to match Rice.  

2019-04-08 13:22:51,430 [scheduler_Worker-1] u:/d: FATAL org.kuali.rice.core.framework.persistence.jta.KualiTransactionInterceptor :: Exception caught by Transaction Interceptor, this will cause a rollback at the end of the transaction.
org.springframework.jdbc.UncategorizedSQLException: OJB operation; uncategorized SQLException for SQL []; SQL state [72000]; error code [12899]; ORA-12899: value too large for column "KFS"."CU_PMW_VENDOR_T"."PO_LN1_ADDR" (actual: 57, maximum: 50)
; nested exception is java.sql.SQLException: ORA-12899: value too large for column "KFS"."CU_PMW_VENDOR_T"."PO_LN1_ADDR" (actual: 57, maximum: 50)


IMHO, if we make these constants more specific, then the parse routine probably should be adjusted to get the second occurrence of "value too large for column"..... and would we run the risk of not having that second value present in some cases?

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
nkimble: Getting the default error message in the report implies that the error message could not be parsed. 
Do we need the additional noise in the logs since the stacktrace will be in the log from the Exception catch block?

src/main/java/edu/cornell/kfs/pmw/batch/PaymentWorksConstants.java
jhulslander: Thanks for clarification, lets go with as is

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
jhulslander: fair enough, lets go with as is

PR #765, opened 2019-04-10
KFSPTS-15027 consolidate concur collector transactions

jhulslander:


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
```diff
@@ -253,6 +259,70 @@ protected void updateCollectorBatchWithOriginEntries() {
         collectorBatch.setTotalRecords(totalRecords);
         collectorBatch.setTotalAmount(totalDebitAmount);
     }
+    
+  private void consolidateOriginEntries() {
```
cah292: This new code is a little harder to read when it doesn't use as many spaces. I would suggest modifying this code so that the spacing is in groups of 4 instead of groups of 2, for consistency with the other code in this file.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
```diff
@@ -253,6 +259,70 @@ protected void updateCollectorBatchWithOriginEntries() {
...
+        nextSequenceNumber.increment();
+        entry.setTransactionLedgerEntrySequenceNumber(nextSequenceNumber.toInteger());
+      }
+      collectorBatch.setOriginEntries(consolidatedOriginEntries);
+    }
+    LOG.info("consolidateOriginEntries, consolidated number of origin entries: " + consolidatedOriginEntries.size());
+  }
+
+  private OriginEntryFull findMatchingOriginEntryFull(List<OriginEntryFull> entries, OriginEntryFull searchEntry) {
+    for (OriginEntryFull entry : entries) {
```
cah292: This full-iteration matching could become very inefficient if the SAE file has a large number of rows in it. If there's a way to limit this check to a certain subset (such as groupings by document number) while still preserving the original entry ordering as much as possible, that would help with keeping this code efficient for larger SAE files.

cah292: If this ends up being infeasible to refactor, or if the Production files are not actually large enough to cause this code to slow down significantly, then we might be okay with moving this forward as-is.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
```diff
@@ -1,11 +1,15 @@
...
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.mutable.MutableInt;
```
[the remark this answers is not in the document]
jhulslander: will change

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
jhulslander: ah, when I was working on checkstyle I was changing these settings, I will set it back and reformat my changes, thanks for noticing

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
jhulslander: I will get some large SAE file and test performance locally.  I'd like to only do some optimization for performance if we really need it.

jhulslander:
I've made some changes based on the code review.  
To check on the performance question, I got some of the larger SAE files we have in our staging directory.   The largest number of collector lines from an SAE file was 320 records, and the the consolidation routine merged them into 266 records in less than a second locally.  I think for the scale of the SAE files we have received, this performs well enough.  Unless you see a bug or some changes needed for clarity, I think leaving as is should be fine for now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurStandardAccountingExtractCollectorBatchBuilder.java
cah292: Given the results of your SAE file testing, I'm fine with leaving this consolidation algorithm as-is. Thanks for checking!

PR #766, opened 2019-04-11
KFSPTS-15268 add fix for issue with opening docs in Capital Asset sp…

db483:
…ace; changes brought in from FINP-5105 fix


db483:
there were some changes to: performForceUppercaseCycleSafe

db483:
Chad, you are right, the changes to performForceUppercaseCycleSafe are not in.

db483:
I have removed the BusinessObjectDictionaryService as we discussed in slack. It has no relevant changes.

cah292:
As per the discussion in Slack, this change will be merged in a new PR instead, so that it can be merged to develop without commit conflicts.

PR #769, opened 2019-04-15
 KFSPTS-15277 switch to extended attribute for custom field

db483:


db483:
I have updated this customization to be an extended attribute as the lookup no longer worked after base code was updated to use the new lookup API.

PR #774, opened 2019-04-18
KFSPTS-15091 enable YEBA in create accounting documents job

jhulslander:


jhulslander:
Since YEBA is very much like BA, I didn't think it was required.  If you think more tests would be preferable, let me know and I can add some

cah292:
I think a few more simple test cases would be a good addition, just to make sure the YEBA processing is properly handling the month amounts and such.

jhulslander:
OK, I shall work on that tomorrow

jhulslander:
I've added additional unit tests for multiple YEBA docs in one file, and a YEBA doc in a multi doc type file.  This is ready for review again

PR #775, opened 2019-04-19
KFSPTS-15192 remove final status date customization from CINV doc

jhulslander:
This user story was to remove the customization for setFinalStatusDate and use base code . That left only one difference from base code, and that was setting TOTAL_PROGRAM_OUTLAYS_TO_DATE.  I think it makes sense to use the parent's getTemplateParameterList function, and just add in the new map values for TOTAL_PROGRAM_OUTLAYS_TO_DATE.  When reviewing these changes, please confirm there were no other CU customizations I may have missed.


jhulslander:
retest this please

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -10,481 +10,42 @@
...
-                    totalCostInvoiceDetail.getAmountRemainingToBill());
-            parameterMap.put(
-                    CuArPropertyConstants.ContractsAndGrantsBillingAwardFields.TOTAL_PROGRAM_OUTLAYS_TO_DATE,
+            Map<String, Object> parameterMap =  new HashMap<String, Object>();
+            
+            parameterMap.put(CuArPropertyConstants.ContractsAndGrantsBillingAwardFields.TOTAL_PROGRAM_OUTLAYS_TO_DATE,
                     totalCostInvoiceDetail.getTotalAmountBilledToDate().add(document.getInvoiceGeneralDetail().getCostShareAmount()));
+            
+            Map<String, String> localParameterMap = new PdfFormattingMap(parameterMap);
+            templateParameters.putAll(new PdfFormattingMap(localParameterMap));
```
cah292: The sub-map is getting wrapped inside PdfFormattingMap twice by the code above. I would suggest wrapping it inside only one PdfFormattingMap before passing it to the main map.

Alternatively, you could convert the total-program-outlays-to-date value to a String directly and then pass it to the main map. This service should already have a variable for the ContractsGrantsBillingUtilityService, which is needed for performing the proper KualiDecimal-to-String conversion that PdfFormattingMap does.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -10,481 +10,42 @@
...
-        parameterMap.put(KFSPropertyConstants.PAYEE + "." + KFSPropertyConstants.ZIPCODE, document.getBillingZipCode());
-        parameterMap.put(ArPropertyConstants.ADVANCE_FLAG,
-                ArConstants.BillingFrequencyValues.isPredeterminedBilling(document.getInvoiceGeneralDetail()));
-        parameterMap.put(ArPropertyConstants.REIMBURSEMENT_FLAG,
-                !ArConstants.BillingFrequencyValues.isPredeterminedBilling(document.getInvoiceGeneralDetail()));
-        parameterMap.put(
-                ArPropertyConstants.ACCOUNT_DETAILS + "." + KFSPropertyConstants.CONTRACT_CONTROL_ACCOUNT_NUMBER,
-                getRecipientAccountNumber(document.getAccountDetails()));
-        
-        if (document.getInvoiceGeneralDetail().isFinalBillIndicator()) {
```
cah292: This "if-else" block needs to be preserved, since it contains our customization for adding the FINAL_BILL and PARTIAL_BILL indicators to the PDF.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
jhulslander: ah, I will fix.  miss step from refactoring.  I'd like to continue to use PdfFormattingMap so the string conversions are consistent.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
jhulslander: good catch, thank you

jhulslander:
I've made some changes based on the PR, this is ready for review again.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
cah292: Sounds good. Thanks!

PR #772, opened 2019-04-16
KFSPTS-15174 upgrade to then 3/21/19, 3/28/19 version of financials

db483:

PR #778, opened 2019-04-23
KFSPTS-15174 Upgrade to 03/28/2019 version of financials

cah292:
This is an updated version of the #772 PR for upgrading to the 03/28/2019 financials version. It takes Daniela's changes from the other PR, and adds some additional cleanup ones on top of that. My changes primarily involve updating the overlays and tweaking a few unit tests.

There are some extension classes that don't have all of the code cleanup that was done in this release. However, the relevant areas only had formatting changes, so it may not make sense to update all of them at this time. I would suggest using a different user story for further extension cleanup if it's needed yet.

Note that once the #777 changes for KFSPTS-14683 get merged, I may have to rebase this PR because of the updated made to CUKFSConstants. Or, if this PR gets merged first, then I will rebase the other PR.


src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
```diff
@@ -22,6 +22,71 @@
     public static final String DATE_FORMAT_mm_dd_yyyy_hh_mm_ss_am = "MM/dd/yyyy hh:mm:ss a";
     public static final String DATE_FORMAT_yyyy_MM_dd_T_HH_mm_ss_SSS = "yyyy-MM-dd'T'HH:mm:ss.SSS";
     
+
+    public static final String NON_CONFIDENTIAL_ATTACHMENT_TYPE_LABEL = "Default";
+    
+    // I Want document constants
+    public static final String I_WANT_DOC_ITEM_TAB_ERRORS = "document.item*,newIWantItemLine*";
```
jhulslander: This block looks like it just reformatting CU custom constants. Should that be part of this PR?

cah292: This reformatting block was carried over from the original PR for this upgrade's changes. I don't have a strong preference either way on whether the changes to this file should go forward. (If they do, though, then we would need to rebase this PR if #777 gets merged first.) We can discuss with Daniela tomorrow if needed.

cah292: @db483 Do you think that the CUKFSConstants code reordering should still be included with this patch? I'm not sure if it's in scope for the patch, given that it's not extending from the base KFSConstants class and our other constants classes haven't been reordered likewise.

db483: The reformatting is not needed for the upgrade, it was just to keep the KFSConstants and our local class in similar pattern. I am fine with not including this change if you think that would be better. thank you.

cah292: Ok, I'll remove this change for now, and we can add it back in for a future user story when needed. Thanks!

cah292:
I've undone the CUKFSConstants changes, so that file should now be the same as before. This is ready for review again.

jhulslander:
Thanks Chad, I think this is ready to merge when we are ready

PR #781, opened 2019-04-24
KFSPTS-14887: Customize the CGB Invoice Transmission Subject and Message.

nkimble:


src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    
+    @Override
+    protected String getSubject(ContractsGrantsInvoiceDocument invoice) {
+        String grantNumber = invoice.getInvoiceGeneralDetail().getAward().getProposal().getGrantNumber();
+        String subject;
+        String message;
+        if (StringUtils.isBlank(grantNumber)) {
+            subject = kualiConfigurationService.getPropertyValueAsString(ArKeyConstants.CGINVOICE_EMAIL_SUBJECT_NO_GRANT_NUMBER);
+            message = MessageFormat.format(subject, invoice.getDocumentNumber(), 
+                    invoice.getInvoiceGeneralDetail().getAward().getProposal().getProposalNumber());
```
cah292: You can get the proposal number from the InvoiceGeneralDetail directly, so you don't have to walk the object graph this far to get it.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        } else {
+            subject = kualiConfigurationService.getPropertyValueAsString(ArKeyConstants.CGINVOICE_EMAIL_SUBJECT);
+            message = MessageFormat.format(subject, invoice.getDocumentNumber(), 
+                    invoice.getInvoiceGeneralDetail().getAward().getProposal().getGrantNumber(), 
+                    invoice.getInvoiceGeneralDetail().getAward().getProposal().getProposalNumber());
```
cah292: Another spot where the proposal number can be read from the InvoiceGeneralDetail instead.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
nkimble: This code was pull directly from base with the minimal amount of changes performed. 
Do we really want to make this change and deviate even further with this mod?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
nkimble: Same issue here. Do we really want to deviate from base code even further?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
cah292: Base code isn't using proposal number here. Nevertheless, after further discussion with Nancy, I'm fine with leaving it like this so that it's consistent with the grant number derivation.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuAREmailServiceImpl.java
cah292: Similarly to the other spot, deriving the value like this is fine with me, and will leave it as-is.

PR #782, opened 2019-04-25
 KFSPTS-15359 moving check address values to parameters

jhulslander:
This code change moves check address values to parameters.


src/main/java/edu/cornell/kfs/pdp/CUPdpParameterConstants.java
```diff
@@ -13,6 +13,16 @@
     public static final String UPDATED_PAYEE_ACH_ACCOUNT_EMAIL_SUBJECT = "UPDATED_PAYEE_ACH_ACCOUNT_EMAIL_SUBJECT";
     public static final String UPDATED_PAYEE_ACH_ACCOUNT_EMAIL_BODY = "UPDATED_PAYEE_ACH_ACCOUNT_EMAIL_BODY";
     public static final String PDP_ACH_INVALID_EMAIL_ERROR_REPORT_TO_ADDRESSES = "PDP_ACH_INVALID_EMAIL_ERROR_REPORT_TO_ADDRESSES";
+    
+    public static final class CuPayeeAddressService {
+        public static final String COMPONENT_CODE = "CuPayeeAddressService";
```
cah292: It's a little confusing having the parameter component in this spot and under this name. Could this constant be moved to a different constants class (like CUPdpConstants), or could it be changed to a different name (like CU_PAYEE_ADDRESS_SERVICE_COMPONENT)?

jhulslander: I'll rename the variable to CU_PAYEE_ADDRESS_SERVICE_COMPONENT  .... I'd like to keep it in this location so it is easy to find with the parameters that it is used with.

jhulslander:
I updated the component code name as suggested and also the parameter names as well to include 
_parameter

PR #784, opened 2019-05-01
KFSPTS-5700 make CuCommodityCodeUpdateServiceImplTest a micro test

jhulslander:


src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
```diff
@@ -1,45 +1,56 @@
...
-import org.kuali.kfs.sys.context.KualiTestBase;
-import org.kuali.kfs.sys.context.SpringContext;
+import java.util.ArrayList;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
 import org.kuali.kfs.vnd.businessobject.CommodityCode;
-import org.kuali.rice.core.api.config.property.ConfigurationService;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
```
cah292: Do you need to use PowerMockito for this test? Since the mocked service implementation appears to be in the same package as this test, and since the "updateCommodityCodes" method has "protected" visibility, you should be able to just use plain Mockito to create the spy and mock the protected method.

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
```diff
@@ -1,45 +1,56 @@
...
-		assertEquals(noCommodityList.size(), noList.size());
 		assertNotSame(content.length, 0);
-		assertNotSame(someCommodityCodes.size(), 0);
-		
-		assert(commodityCodeUpdateService.loadCommodityCodeFile(DATA_FILE_PATH));
+		assertEquals(NUMBER_OF_RECORDS_IN_DATA_FILE, someCommodityCodes.size());
+	}
+	
+	@Test
+	public void testloadCommodityCodeFile() {
```
cah292: This method has almost the same name as another test method in this class, where the only difference is that the first "L" has been lowercased. I would suggest renaming this method and/or the other method accordingly to avoid confusion.

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
```diff
@@ -1,45 +1,56 @@
...
-		byte[] zeroLength = {};
-		ArrayList<CommodityCode> noList = new ArrayList();
-		byte[] noContent = commodityCodeUpdateService.getFileContent("no such file");
-		ArrayList<CommodityCode> noCommodityList = (ArrayList<CommodityCode>) commodityCodeUpdateService.parseCommodityCodeList(zeroLength);
 		byte[] content = commodityCodeUpdateService.getFileContent(DATA_FILE_PATH);
 		ArrayList<CommodityCode> someCommodityCodes = (ArrayList<CommodityCode>) commodityCodeUpdateService.parseCommodityCodeList(content);
 		
-		assertEquals(noContent.length, zeroLength.length);
-		assertEquals(noCommodityList.size(), noList.size());
 		assertNotSame(content.length, 0);
```
cah292: Should this test be using "assertNotSame"? I think that does a "object1 != object2" check instead of a "!object1.equals(object2)" check. (Note that "assertNotSame" only has Object-comparing variants, so primitives might be getting "boxed" prior to comparsion.) It may be better to use "assertNotEquals" instead.

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
jhulslander: Yes, I removed powermockito and was able to use Mockito with slightly different mocking syntax

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
jhulslander: Thanks, updated function names

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
jhulslander: thanks for the info, I switched

jhulslander:
I've made some changes based on the notes, this is ready for review again

src/test/java/edu/cornell/kfs/vnd/batch/service/impl/CuCommodityCodeUpdateServiceImplTest.java
```diff
@@ -1,45 +1,55 @@
...
+
+    @After
+    public void tearDown() {
+        commodityCodeUpdateService = null;
+    }
+
+    @Test
+    public void testParseCommodityCodeList() {
+        byte[] content = commodityCodeUpdateService.getFileContent(DATA_FILE_PATH);
+        ArrayList<CommodityCode> someCommodityCodes = (ArrayList<CommodityCode>) commodityCodeUpdateService.parseCommodityCodeList(content);
```
cah292: I don't think it's necessary to cast the returned collection to an ArrayList, since the assertions below are only checking the collection's total size.

jhulslander: OK, changed

jhulslander:

PR #785, opened 2019-05-01
KFSPTS-15342 Improve due date handling for CINV docs

cah292:
NOTE: There is also a related nonprod-sql PR that needs to go out together with this.

Some of the functionality is using a less-than-ideal hook so as to avoid overlaying the CINV document class, and creating a CINV subclass involves too many other changes. In addition, I customized things so that a preliminary due date is set on document creation, and the due date gets its last revision when the document is processed/finalized. Please let me know if you think an overlay would be a better hook for some of this functionality, and if you think that just setting the due date once at document finalization is enough.


src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -178,4 +187,66 @@ protected CustomerInvoiceDetail createSourceAccountingLinesByContractControlAcco
...
     }
 
+    /**
+     * Overridden to also update the invoice due date.
+     * 
+     * NOTE: There isn't an easy way to hook in the invoice-due-date updates independently of this
+     * without overlaying the CINV document class. (This method only appears to be called when CINV docs
+     * enter PROCESSED status, so hooking in here is okay for now.) If ContractsGrantsInvoiceDocument
+     * gets updated to make it easier to hook in custom PROCESSED-status handling, then we should rework
+     * this customization.
```
jhulslander: Should we create a kualico ticket with suggested changes to ContractsGrantsInvoiceDocument?  If so, we should also create a follow user story for us to rework this

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -178,4 +187,66 @@ protected CustomerInvoiceDetail createSourceAccountingLinesByContractControlAcco
...
+        if (customerNetTerms.isPresent()) {
+            netTermsInDays = customerNetTerms.get();
+        } else {
+            String defaultNetTerms = parameterService.getParameterValueAsString(
+                    ArConstants.AR_NAMESPACE_CODE, ArConstants.CUSTOMER_COMPONENT, CuArParameterKeyConstants.CG_INVOICE_TERMS_DUE_DATE);
+            try {
+                netTermsInDays = Integer.parseInt(defaultNetTerms);
+            } catch (NumberFormatException e) {
+                LOG.error("calculateInvoiceDueDate, Default net terms parameter is malformed; will use zero as the default instead", e);
+                netTermsInDays = 0;
```
jhulslander: It might be a good idea to check with the functionals what should happen if the parameter is improperly formatted

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
cah292: Ok, I'll create a PTS ticket for us to follow up with KualiCo to make this part easier to customize.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
cah292: Good point, I'll check to see what the functionals want the behavior to be here.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
cah292: I've created KFSPTS-15550 to keep track of the CINV override limitations.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
cah292: The functionals have confirmed that an error should be thrown in this spot if the net terms parameter value is invalid. I'll update the code accordingly. Thanks for pointing this out!

cah292:
I've made some more changes so that a blank or invalid net-terms parameter value will cause an error. I believe the concerns from the initial review have been addressed, and this is now ready for review again.

PR #789, opened 2019-05-08
KFSPTS-14965 Add AVs to Create Accounting Document Job

cah292:
This PR adds the ability to create AV documents via the Create Accounting Document Job. Some new document and account line XML elements had to be added to support this change.

Note that only the "Adjustment" and "Accrual" AV types are allowed for AVs that are generated by this batch job. This PR explicitly prevents the "Recode" AV type from being used, as per the functional requirements.

There were also a few areas where I had to duplicate and tweak some needed logic from base code, due to where the original code had placed some of the logic and to also allow for making that logic micro-testable. Hopefully we can rework that logic in the future if KualiCo makes the appropriate changes.


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
```diff
@@ -0,0 +1,234 @@
...
+        A accountingLine = super.buildAccountingLine(accountingLineClass, documentNumber, xmlLine);
+        validateAndPopulateAccountingLineWithDebitCreditAmounts(accountingLine, xmlLine);
+        return accountingLine;
+    }
+
+    protected <A extends AccountingLine> void validateAndPopulateAccountingLineWithDebitCreditAmounts(
+            A accountingLine, AccountingXmlDocumentAccountingLine xmlLine) {
+        if (hasAmountBeenSpecified(xmlLine.getAmount())) {
+            throw new ValidationException(
+                    "AV accounting line cannot use the 'amount' element to specify an amount; use the relevant debit/credit amount element instead");
```
jhulslander: I'd suggesting moving this and other validation exception messages to properties

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
```diff
@@ -0,0 +1,234 @@
...
+        WorkflowDocument workflowDocument = document.getDocumentHeader().getWorkflowDocument();
+        return workflowDocument.getDateCreated();
+    }
+
+    /*
+     * NOTE: Much of this period-searching logic has been duplicated and modified from AuxiliaryVoucherForm.
+     * If KualiCo updates the related base code to be more micro-test-friendly, then we should update
+     * the code below accordingly.
+     */
+    protected AccountingPeriod findEligibleAccountingPeriod(
```
jhulslander: Can this be simplified if the XML document takes in a numberic accounting period and let the document validate if the accounting period is valid?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
cah292: Ok, I'll get these messages moved to properties accordingly.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
cah292: After further analysis, it looks like the AV document validation rules do contain sufficient checks for ensuring that the accounting period is eligible for use. I will simplify this piece further to just check that the referenced accounting period actually exists.

As for searching by year-and-period-code instead of by period name, I will check with the functionals to see if they are fine with such a setup.

cah292:
I've committed some changes to move the error messages out to properties, and to remove some redundant accounting period validation. This is ready for review again.

Also, as noted in the user story for this change, the functionals still want to specify the accounting period by name rather than by year-and-period-code combo.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+        Optional<Date> reversalDate = validateAndGetReversalDateIfApplicable(
+                documentEntry, period, avTypeCode, documentCreateDateWithoutTime);
+        if (reversalDate.isPresent()) {
+            document.setReversalDate(reversalDate.get());
+        }
+    }
+
+    protected DateTime getDocumentCreateDate(AuxiliaryVoucherDocument document) {
+        WorkflowDocument workflowDocument = document.getDocumentHeader().getWorkflowDocument();
+        return workflowDocument.getDateCreated();
```
jhulslander: Why do we need to get the date created from the workflow document?  Wouldn't it always be the current date?

src/main/resources/edu/cornell/kfs/fp/cu-fp-resources.properties
```diff
@@ -0,0 +1,9 @@
+error.create.accounting.document.av.amount.required=AV accounting line did not specify a debit or credit amount
```
jhulslander: I think this is great to create a new FP resources properties file. Do we have a user story to move appropriate properties from application resources into this file?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AuxiliaryVoucherDocumentGenerator.java
cah292: Good point, we don't run this job at around midnight so it should be safe to assume that the current date is also the create date. I'll adjust this accordingly.

src/main/resources/edu/cornell/kfs/fp/cu-fp-resources.properties
cah292: I know we had discussed moving the properties into module-specific files at some prior retro meetings, but I'm having trouble finding any follow-up user stories. We can create some new user stories for this effort if they haven't been created yet.

cah292:
I've committed another set of changes based on the latest PR feedback. This is now ready for review again.

PR #792, opened 2019-05-16
KFSPTS-14985 Import KIM feed to KFS with minimal changes

cah292:
This is the initial import of the Cynergy KIM feed process into KFS, to convert it into a KFS batch job. Note that the CynergyKimFeed and SimpleCommandLineParser classes have been copied over mostly as-is from the "cynergy-kim-feed" project. As a result, those two classes still need significant cleanup to bring them up-to-date with our latest coding standards; the KFSPTS-14994 user story will handle that refactoring in the near future.

A couple minor changes were made to CynergyKimFeed to make it run and behave better for KFS processing:

[1] Updated its log4j setup to bring it more in line with our other KFS classes.
[2] Updated the class to allow for injecting the config properties by passing in a Properties object, rather than reading the settings from an external file or from system properties.
[3] Renamed the "main" method to "runKimFeed" instead, and also adjusted its parameter list.

Please let me know if you find anything else that still needs minimal refactoring at this stage.


cah292:
NOTE: The following config PR needs to be merged as well, so that the KFS unit tests can launch successfully:

https://github.com/CU-CommunityApps/config/pull/17

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -166,7 +167,30 @@
...
+	<bean id="runKimFeedStep" class="edu.cornell.kfs.sys.batch.RunKimFeedStep" parent="step">
+		<property name="kimFeedProperties">
+			<props>
+				<prop key="db-url0">${edw.datasource.url}</prop>
+				<prop key="db-username0">${edw.datasource.username}</prop>
+				<prop key="db-password0">${edw.datasource.password}</prop>
+				<prop key="db-url1">${rice.datasource.url}</prop>
+				<prop key="db-username1">${rice.datasource.username}</prop>
+				<prop key="db-password1">${rice.datasource.password}</prop>
+				<prop key="skip-delta-flag-updates">${kim.feed.skip.delta.flag.updates}</prop>
```
jhulslander: I am wondering if we should move kim.feed.skip.delta.flag.updates, kim.feed.load.latest.delta.only and kim.feed.load.delta.with.date to KFS parameters.  It makes total sense for the database connection information to be controlled from a properties file.  I know there is a seperate user story to refactor the job into our standards, but it may make sense to move these now since they affect properties files that will need to be added to each developer's system, and to the KFS application servers.

cah292: I was hoping to keep these changes to a minimum, but I agree that moving the non-datasource-credential properties to parameters would be good to do now, which should also make it easier to perform testing (and make it easier for the DevOps config changes as well). I'll update this accordingly. Thanks!

cah292:
I've made some more updates based on the initial review. The job now uses new parameters to customize the non-datasource-related parts of the configuration, and a related nonprod-sql PR has been created as a result. I've also created a follow-up user story (KFSPTS-15599) to have one of the new parameters auto-changed accordingly by our post-refresh scripts for KFS DEV.

The new parameters have been set up to consolidate a few of the settings, resulting in two parameters instead of three. (The new code will translate these into the appropriate legacy properties accordingly.) Please let me know if you think this alternate setup should be postponed until later.

This is ready for review again, but please don't merge until we're ready for DevOps to add the new EDW datasource config properties to the KFS DEV properties file.

jhulslander:
retest this please

jhulslander:
I will review again tomorrow, but I think this code is ready to merge assuming the unit tests run and we coordinate with devops and the other developers to update the appropriate properties files.

src/main/java/cynergy/CynergyKimFeed.java
```diff
@@ -0,0 +1,1085 @@
...
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.support.rowset.SqlRowSet;
+
+import util.SimpleCommandLineParser;
+
+public class CynergyKimFeed {
+
+    private static final Logger LOG = LogManager.getLogger(CynergyKimFeed.class);
+
+	// Helper constant for inserting blank address rows.
```
jhulslander: would it be worth while to do a source format of this class on this commit?  It should make future reviews easier.

src/main/java/util/SimpleCommandLineParser.java
```diff
@@ -0,0 +1,88 @@
+package util;
```
jhulslander: if you agree with doing a source format on the class above, this could use the same

nkimble:
@jhulslander My suggestion would be to just lift the code with as few changes as possible for this user story and get it to work as a batch job. And, then in the next user story tackle all the package/class structuring to get the code into our normal formats. My reasoning: Should we start to see issues with the KIM feed it will be *just* because of the switch to a batch job and *not* because of any other refactoring coding changes introduced.

jhulslander:
I agree with as small changes as possible.  My hope is the changes I suggested make it easier to refactor the code moving forward and not impact the way the current code works.

src/main/java/cynergy/CynergyKimFeed.java
cah292: By "source format" are you referring to just the whitespace usage, or are you referring to other formatting like comments, line breaks, use of curly braces, etc.?

I would prefer to keep the changes in this PR as minimal as possible, since the CynergyKimFeed class will likely get completely replaced by DAO classes during the future refactoring efforts.

jhulslander: I was referring to the eclipse function selecting the source of a java class, and then right clicking and choosing source, then format .... but I am fine with leaving as if the future refactor will be mostly a a replacement by DAOs

cah292: It looks like the "source: format" feature would add way more changes than what I would have intended for an almost-drop-in setup of the existing class, even if they're only minor changes. If it's okay with you, I would prefer to wait on the formatting cleanup until the follow-up user story, which may totally replace the CynergyKimFeed class with some better DAOs anyway.

jhulslander: OK, I am fine with proceeding

PR #793, opened 2019-05-17
 KFSPTS-15572 Increase max length on customer address fields 

cah292:


cah292:
The related SQL PR has been updated accordingly with additional field-length-change SQL. No new code changes should be needed, since the DD beans for the new affected fields appear to be inheriting from the fields that this PR is already modifying.

This PR and the SQL PR are ready for review again.

PR #797, opened 2019-05-22
KFSPTS-14985 Import KIM feed to KFS with minimal changes

cah292:
This is an updated version of the #792 PR changes. This revision of the feature fixes an issue where CynergyKimFeed was using an older outdated version of the relevant JDBC driver.

Note that this PR copies the KIM feed code into KFS with only minimal changes, so the code does not adhere to our current coding standards at the moment. We have a follow-up PR that will focus on refactoring the KIM feed code accordingly.


cah292:
retest this please

PR #801, opened 2019-05-31
KFSPTS-15575 Add FINP-4678 milestone/PDBS fix from KualiCo

cah292:
This PR takes the FINP-4678 milestone/PDBS fix from the 2019-05-02 KualiCo patch, and integrates it into our current cu-kfs code. This fix has been included by using overlays of the relevant files, so that we can easily remove the overlays once we upgrade cu-kfs to financials version 2019-05-02 or newer.

These overlays should be the same as the 2019-04-25 versions of the files, except that the FINP-4678 fixes have been added and a few explanation comments have also been added.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/ar/businessobject/MilestoneSchedule.java
+++ b/src/main/java/org/kuali/kfs/module/ar/businessobject/MilestoneSchedule.java
@@ -1,0 +1,362 @@
+/**
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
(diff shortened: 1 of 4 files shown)


cah292:
KualiCo typically references the sub-tasks instead of the parent tickets in their commit messages. I used one of the helper scripts from our cu-kfs-upgrades project to retrieve the information about FINP-4678, and from that I saw that the user story had sub-tasks FINP-5324 and FINP-5325. I then found two commits that referenced sub-task FINP-5324, and brought in the changes from those.

PR #809, opened 2019-06-21
KFSPTS-15755 Fix exception on Rass batch job when reportsToAgency is null

jhulslander:


src/main/java/edu/cornell/kfs/rass/RassConstants.java
```diff
@@ -38,5 +38,8 @@ public static boolean isSuccessfulResult(RassObjectUpdateResultCode resultCode)
     public static final String RASS_DEFAULT_PROPOSAL_AWARD_TYPE_PARAMETER = "RASS_DEFAULT_PROPOSAL_AWARD_TYPE";
     
     public static final String PROPOSAL_ORG_CHART = "IT";
+    
+    public static final String REPORTING_NAME = "reportingName";
+    public static final String REPORTS_TO_AGENCY_NUMBER = "reportsToAgencyNumber";
```
cah292: Both of these constants already appear to be available in the KFSPropertyConstants class, so you don't need to add them here.

jhulslander:
i've made some changes based on the comments.  This is ready for review again

PR #817, opened 2019-07-09
KFSPTS-15327 eInvoice KFS API

ajd299:


src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @GET
+    @Path("vendors")
+    public Response getVendor(@Context HttpHeaders headers) {
+        try {
+            List<String> vendorNumbers = getVendorNumbersFromRequest();
+            List<VendorDetail> vendors = getCuEinvoiceDao().getVendors(vendorNumbers);
+            List<Properties> vendorsSerialized = vendors.stream().map(vendor -> getVendorProperties(vendor)).collect(Collectors.toList());
+            return Response.ok(gson.toJson(vendorsSerialized)).build();
+        } catch (BadRequestException ex) {
```
jhulslander: I think . we could eliminate the catch for BadRequestException, and just used the generic exception catch.  Unless there is a reason to NOT log the BadRequestException case.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,248 @@
...
+
+            HashMap<String, String> vendorCombinedPk = parseVendorNumber(vendorNumber);
+            VendorDetail vendorDetail = getLookupDao().findObjectByMap(VendorDetail.class.newInstance(), vendorCombinedPk);
+            if (ObjectUtils.isNull(vendorDetail)) {
+                return respondNotFound();
+            }
+            String responseBody = gson.toJson(getVendorProperties(vendorDetail));
+            return Response.ok(responseBody).build();
+        }
+        catch (BadRequestException ex) {
```
jhulslander: similar thought here about merging the catches

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
ajd299: I have added logging statements to the catch block of the BadRequestException. For the generic Exception the server is responding with status code 500 internal server error, but with BadRequestException it should be responding with status code 400 Bad Request

jhulslander: Thanks for the clarification

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,248 @@
...
+    @Path("vendors")
+    public Response getVendor(@Context HttpHeaders headers) {
+        try {
+            List<String> vendorNumbers = getVendorNumbersFromRequest();
+            List<VendorDetail> vendors = getCuEinvoiceDao().getVendors(vendorNumbers);
+            List<Properties> vendorsSerialized = vendors.stream().map(vendor -> getVendorProperties(vendor)).collect(Collectors.toList());
+            return Response.ok(gson.toJson(vendorsSerialized)).build();
+        } catch (BadRequestException ex) {
+            LOG.error(ex);
+            return respondGetVendorsBadRequest();
```
jhulslander: Thanks for the clarification,  It looks like many of these functions are doing the same error handling.  To consolidate, we could just catch exception in these functions, and in the catch call an error handling function that could include an instanceof check so it could do different things if the exception is BadRequestException or whatever else.  Also, the log statements should include a function name

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private LookupDao getLookupDao() {
+        if (ObjectUtils.isNull(lookupDao)) {
+            lookupDao = SpringContext.getBean(LookupDao.class);
+        }
+        return lookupDao;
+    }
+
+    private CuEinvoiceDao getCuEinvoiceDao() {
+        if (ObjectUtils.isNull(null)) {
```
jhulslander: This will always return true, did you want cuEinvoiceDao instead of null here

ajd299:
Please have a look at my latest changes and let me know if they addressed your comments.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,242 @@
...
+            HashMap<String, String> map = new HashMap<>();
+            map.put(PurapPropertyConstants.PURAP_DOC_ID, purapDocumentIdentifier);
+            PurchaseOrderDocument poDoc = getLookupDao().findObjectByMap(PurchaseOrderDocument.class.newInstance(), map);
+            if (ObjectUtils.isNull(poDoc)) {
+                return respondNotFound();
+            }
+            String responseBody = serializePoDocumentToJson(poDoc);
+            return Response.ok(responseBody).build();
+        } catch (Exception ex) {
+            LOG.error(ex);
```
jhulslander: one more log statement that needs the function all

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
ajd299: Thanks for catching this bug

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -0,0 +1,242 @@
...
+            return respondInternalServerError(ex);
```
jhulslander: should this error handling also check for BadRequestException?

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
ajd299: I have consolidated into 1 catch block and using instance of in a ternary statement.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
ajd299: No, the functions called from getPurchaseOrder will not produce a BadRequestException. getPurchaseOrder is simpler than getVendor and getVendors because there is no parsing of VendorNumber(s) in getPurchaseOrder.

ajd299:
Please see my latest commit and comment.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
jhulslander: OK, thanks

PR #818, opened 2019-07-18
KFSPTS-15813 eInvoice KFS Purchase Order Integration

ajd299:


src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -170,22 +170,47 @@ private VendorAddress getVendorAddress(List<VendorAddress> vendorAddresses) {
...
         }
         poProperties.put(CUPurapConstants.Einvoice.ITEMS, poLines.toArray());
     }
 
+    private void safelyAddProperty(Properties properties, String key, String value) {
+        properties.put(key, StringUtils.defaultIfBlank(value, KFSConstants.EMPTY_STRING));
+    }
+
+    private void safelyAddProperty(Properties properties, String key, KualiDecimal value) {
+        properties.put(key, ObjectUtils.isNull(value) ? KFSConstants.EMPTY_STRING : value.floatValue());
```
cah292: Should the KualiDecimal's value just be converted to a String instead of a float, to help avoid the risk of losing data? If you need to convert it to a primitive, I would suggest converting to a "double" instead if you can. The same goes for the equivalent BigDecimal method below.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -170,22 +170,47 @@ private VendorAddress getVendorAddress(List<VendorAddress> vendorAddresses) {
...
+    }
+
+    private void safelyAddProperty(Properties properties, String key, BigDecimal value) {
+        properties.put(key, ObjectUtils.isNull(value) ? KFSConstants.EMPTY_STRING : value.floatValue());
+    }
+
+    private void safelyAddProperty(Properties properties, String key, Integer value) {
+        properties.put(key, ObjectUtils.isNull(value) ? KFSConstants.EMPTY_STRING : value.intValue());
```
cah292: Do you need to call intValue() here? This particular line of code is just going to implicitly convert the primitive int back into an Integer.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -170,22 +170,47 @@ private VendorAddress getVendorAddress(List<VendorAddress> vendorAddresses) {
...
+                PurchaseOrderItem poItem = (PurchaseOrderItem)obj;
                 Properties lineProps = new Properties();
-                safelyAddProperty(lineProps, CUPurapConstants.AMOUNT.toLowerCase(), purApItem.getTotalAmount().toString());
-                safelyAddProperty(lineProps, CUPurapConstants.Einvoice.QUANTITY, purApItem.getItemQuantity().toString());
-                safelyAddProperty(lineProps, CUPurapConstants.Einvoice.UNIT_OF_MEASURE, purApItem.getItemUnitOfMeasureCode());
-                safelyAddProperty(lineProps, CUPurapConstants.Einvoice.DESCRIPTION, purApItem.getItemDescription());
-                lineProps.put(CUPurapConstants.Einvoice.UNIT_PRICE, purApItem.getItemUnitPrice());
-                lineProps.put(CUPurapConstants.Einvoice.LINE_NUMBER, purApItem.getItemLineNumber());
-                lineProps.put(CUPurapConstants.Einvoice.PART_NUMBER, purApItem.getItemAuxiliaryPartIdentifier());
+                safelyAddProperty(lineProps, CUPurapConstants.AMOUNT.toLowerCase(), poItem.getTotalAmount());
```
cah292: I would suggest adding a new CUPurapConstants.Einvoice.AMOUNT constant that has a lowercase value, so that you can use a more consistent convention for tracking the Properties keys.

src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -170,22 +170,47 @@ private VendorAddress getVendorAddress(List<VendorAddress> vendorAddresses) {
...
+                safelyAddProperty(lineProps, CUPurapConstants.Einvoice.INVOICED_AMOUNT.toLowerCase(), poItem.getItemInvoicedTotalAmount());
```
cah292: Do you need the toLowerCase() call here? The referenced constant already appears to have a lowercase value.

PR #820, opened 2019-07-19
KFSPTS-15981 Upgrade to 2019-05-23 version of financials

cah292:
NOTE: This release introduces a new General Ledger Transfer (GLT) document. I've added its new doc type XML to this PR to make it more convenient to get and ingest the XML. (The base financials code has the GLT doc type embedded in a larger doc type XML file, so that larger file is not feasible to ingest because it would affect several other doc types as well.) If you think the doc type should be placed elsewhere, please let me know.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/gl/document/workflow/GeneralLedgerTransferDocument.xml
+++ b/src/main/resources/edu/cornell/kfs/gl/document/workflow/GeneralLedgerTransferDocument.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    This is the KualiCo-delivered General Ledger Transfer Document workflow XML,
+    which has been separated out into its own separate file for ingestion convenience.
+ -->
+<data xmlns="ns:workflow" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="ns:workflow resource:WorkflowData">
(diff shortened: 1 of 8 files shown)


cah292:
NOTE: The related SQL has not been applied to UTEST yet, but the unit tests still seem to be able to pass without those changes.

src/main/resources/edu/cornell/kfs/gl/document/workflow/GeneralLedgerTransferDocument.xml
```diff
@@ -0,0 +1,90 @@
...
+                ${kfs.url}/financialGeneralLedgerTransfer.do?methodToCall=docHandler
+            </docHandler>
+            <helpDefinitionURL>
+                default.htm?turl=WordDocuments%2Fgeneralledgertransfer.htm
+            </helpDefinitionURL>
+            <active>
+                true
+            </active>
+            <routingVersion>
+                2
```
jhulslander: since this is a new document type, shouldn't the version be 1?

cah292: The "routingVersion" doesn't refer to a BO version number; it refers to the KEW doc type routing setup. Version 1 specifies an archaic "route level" routing scheme, whereas version 2 represents the route node routing scheme that is used today.

jhulslander: ah, right

jhulslander:
This can be merged when we are ready to move forward with this patch.

jhulslander:
retest this please

PR #821, opened 2019-07-24
KFSPTS-15671 Fix vendor contact phone number validation

cah292:
This fixes a bug where base code was not always validating vendor contact phone numbers at the required times, which was allowing our PMW batch jobs to route (but not finalize) vendors whose contact phone extension numbers were too long. The KualiCo-delivered business rules for such areas should now be getting enforced properly by this fix.

Note that the fix has been implemented via overlays of files from base financials code, so that we can get this contributed back to KualiCo in the near future. I've added some comments to these files to help point out which areas have changed; please let me know if you think the comments need to be updated or removed.

### DIFF
--- a/src/main/java/org/kuali/kfs/vnd/document/validation/impl/VendorRule.java
+++ b/src/main/java/org/kuali/kfs/vnd/document/validation/impl/VendorRule.java
@@ -1,0 +1,1741 @@
+/**
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2019 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/main/java/org/kuali/kfs/vnd/document/validation/impl/VendorRule.java
```diff
@@ -0,0 +1,1741 @@
...
+                    }
+                }
+            }
+        }
+        // refresh contacts
+        if (vendor.getVendorContacts() != null) {
+            for (VendorContact contact : vendor.getVendorContacts()) {
+                contact.refreshNonUpdateableReferences();
+                // ==== CU Customization ====
+                if (contact.getVendorContactPhoneNumbers() != null) {
```
jhulslander: should this use StringUtils.isNotEmpty instead?

cah292: contact.getVendorContactPhoneNumbers() returns a collection, not a string. I used a non-null check here for consistency with the "vendor.getVendorContacts() != null" check from a few lines farther up.

jhulslander: oh, thanks for the clarification

PR #826, opened 2019-07-30
KFSPTS-15510 Fix sub-fund program description max length

cah292:
This is an updated version of the #823 fix, which additionally converts the description text box into a textarea.


cah292:
retest this please

PR #831, opened 2019-08-07
KFSPTS-16217 update RASS batch job report to make each line atomic

db483:


db483:
retest this please

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassReportServiceImpl.java
```diff
@@ -284,6 +284,19 @@ private void writeHeaderForResultsDetailsByResultCode(Class businessObjectClass,
 
         reportWriterService.writeFormattedMessageLine(reportMessage);
     }
+    
+    private String buildErrorLines(String prefix, String errorMessage) {
+        StringBuffer errorLines = new StringBuffer();
```
cah292: Unless you need multi-thread appending or need to work with APIs that only accept StringBuffer, I would suggest using a StringBuilder instead.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassReportServiceImpl.java
```diff
@@ -284,6 +284,19 @@ private void writeHeaderForResultsDetailsByResultCode(Class businessObjectClass,
...
+        String[] lines = errorMessage.split(KFSConstants.NEWLINE);
+        for (String line : lines) {
+            if (StringUtils.isNoneBlank(line)) {
```
cah292: Since you're only checking a single String with this call, I think it would be better to use an isNotBlank() check instead of an isNoneBlank() check.

db483: Thank you Chad. I have changed it.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassReportServiceImpl.java
db483: updated

PR #842, opened 2019-08-27
KFSPTS-16330: Customized Proposal such that federal pass through indicator business rule matches Award business rule.

nkimble:


nkimble:
Requested changes have been made and code is ready to be reviewed again.

PR #844, opened 2019-08-28
KFSPTS-16427 remove CFDA number from RASS integration

jhulslander:


jhulslander:
I ended up removing CFDA from RassXmlAwardEntry since we aren't going to be using for the integration.  That fixed the unit test that failed.  This is ready for review again.

PR #641, opened 2018-09-10
KFSPTS-7507: Add exception handling to allow processing to continue when IWNT cannot send email on finalize.

nkimble:

PR #846, opened 2019-08-30
KFSPTS-16391: RASS change to add default value of N to KFS system parameter DEFAULT_AWARD_ACCOUNT for account attribute active indicator.

nkimble:


nkimble:
This PR must be performed with nonprod-sql PR #641.
NOTE: The corresponding SQL is **required** for the code to perform correctly. I would suggest merging these changes into the repository with the shortest amount of time between when the build occurs and the SQL is applied so other developers working on RASS are **not* impacted:

src/main/java/edu/cornell/kfs/rass/batch/AwardTranslationDefinition.java
```diff
@@ -84,34 +87,41 @@ public void processCustomTranslationForBusinessObjectCreate(
...
-        awardAccount.setChartOfAccountsCode(defaultChartAndAccount.getLeft());
-        awardAccount.setAccountNumber(defaultChartAndAccount.getRight());
+        awardAccount.setChartOfAccountsCode(defaultAwardAccountAttributes.get(KFSPropertyConstants.CHART));
+        awardAccount.setAccountNumber(defaultAwardAccountAttributes.get(KFSPropertyConstants.ACCOUNT_NUMBER));
+        awardAccount.setActive(defaultAccountIndicatorIsActive(KFSPropertyConstants.ACCOUNT_ACTIVE_INDICATOR));
         return awardAccount;
     }
+    
+    protected boolean defaultAccountIndicatorIsActive(String rowIndicatorValue) {
+        return StringUtils.equals(rowIndicatorValue, KFSConstants.ACTIVE_INDICATOR) ? true : false;
```
cah292: You can just return the StringUtils.equals() result directly; making it a ternary of "? true : false" is redundant.

src/main/java/edu/cornell/kfs/rass/batch/AwardTranslationDefinition.java
```diff
@@ -84,34 +87,41 @@ public void processCustomTranslationForBusinessObjectCreate(
...
         }
-        
         AwardAccount awardAccount = new AwardAccount();
         awardAccount.setProposalNumber(xmlAward.getProposalNumber());
         awardAccount.setPrincipalId(directorPrincipalId);
-        awardAccount.setChartOfAccountsCode(defaultChartAndAccount.getLeft());
-        awardAccount.setAccountNumber(defaultChartAndAccount.getRight());
+        awardAccount.setChartOfAccountsCode(defaultAwardAccountAttributes.get(KFSPropertyConstants.CHART));
+        awardAccount.setAccountNumber(defaultAwardAccountAttributes.get(KFSPropertyConstants.ACCOUNT_NUMBER));
+        awardAccount.setActive(defaultAccountIndicatorIsActive(KFSPropertyConstants.ACCOUNT_ACTIVE_INDICATOR));
```
cah292: This call to the indicator-is-active method is passing in the property key itself. I think it should instead be passing in the value that is associated with the key, given that the method treats its argument as being the value and not the key.

src/main/java/edu/cornell/kfs/rass/batch/AwardTranslationDefinition.java
```diff
@@ -84,34 +87,41 @@ public void processCustomTranslationForBusinessObjectCreate(
...
         }
-        
-        String defaultChart = StringUtils.substringBefore(defaultChartAndAccount, CUKFSConstants.COLON);
-        String defaultAccount = StringUtils.substringAfter(defaultChartAndAccount, CUKFSConstants.COLON);
-        if (StringUtils.isBlank(defaultChart) || StringUtils.isBlank(defaultAccount)) {
-            throw new RuntimeException("Default chart-and-account parameter cannot have a blank chart or account");
+        String defaultChart = StringUtils.substringBefore(defaultAttributes, CUKFSConstants.COLON);
+        defaultAttributes = StringUtils.substringAfter(defaultAttributes, CUKFSConstants.COLON);
+        String defaultAccountNumber = StringUtils.substringBefore(defaultAttributes, CUKFSConstants.COLON);
+        String defaulAccountActiveIndicator = StringUtils.substringAfter(defaultAttributes, CUKFSConstants.COLON);
```
cah292: Minor typo in variable name; should be changed from "defaulAccountActiveIndicator" to "defaultAccountActiveIndicator"

PR #858, opened 2019-09-25
KFSPTS-16513 Create RASS reports for each XML file

cah292:
This PR updates the RASS batch job so that each processable XML file will have a unique report file generated for it, rather than lumping all of the report data into one file. For any files that have a complete parsing failure (such as from malformed XML), those files will be lumped into one error report file.

I had to perform a fair amount of refactoring for the RASS service (and related unit test(s)) so that the agency/proposal/award updates would be organized on a per-file basis. I've also added a new unit test to perform some basic testing of the report file generation process.

Please let me know if you notice any obvious problems with these refactoring changes.

### DIFF
--- a/src/main/java/edu/cornell/kfs/rass/RassConstants.java
+++ b/src/main/java/edu/cornell/kfs/rass/RassConstants.java
@@ -8,7 +8,6 @@ public final class RassConstants {
 
     public static final String RASS_MAINTENANCE_NEW_ACTION_DESCRIPTION = "Creation";
     public static final String RASS_PARSE_ERRORS_BASE_FILENAME = "errors";
-    public static final String XML_FILE_EXTENSION = ".xml";
 
     public enum RassParseResultCode {
         SUCCESS,
--- a/src/main/java/edu/cornell/kfs/rass/RassConstants.java
+++ b/src/main/java/edu/cornell/kfs/rass/RassConstants.java
@@ -7,6 +7,8 @@ public final class RassConstants {
     }
 
     public static final String RASS_MAINTENANCE_NEW_ACTION_DESCRIPTION = "Creation";
+    public static final String RASS_PARSE_ERRORS_BASE_FILENAME = "errors";
+    public static final String XML_FILE_EXTENSION = ".xml";
 
     public enum RassParseResultCode {
         SUCCESS,
--- a/src/main/java/edu/cornell/kfs/rass/batch/RassStep.java
+++ b/src/main/java/edu/cornell/kfs/rass/batch/RassStep.java
@@ -33,7 +33,6 @@ public class RassStep extends AbstractStep {
         Map<String, RassXmlFileProcessingResult> processingResults = rassService.updateKFS(successfulResults);
         RassBatchJobReport report = new RassBatchJobReport(parseResults, processingResults);
         rassReportService.writeBatchJobReports(report);
-        //sendReport(report);
         return true;
     }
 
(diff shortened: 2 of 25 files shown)


src/main/java/edu/cornell/kfs/rass/batch/RassStep.java
```diff
@@ -29,10 +30,9 @@ public boolean execute(String jobName, Date jobRunDate) throws InterruptedExcept
             return true;
         }
         List<RassXmlFileParseResult> successfulResults = findSuccessfullyParsedFileResults(parseResults);
-        RassXmlProcessingResults processingResults = rassService.updateKFS(successfulResults);
+        Map<String, RassXmlFileProcessingResult> processingResults = rassService.updateKFS(successfulResults);
         RassBatchJobReport report = new RassBatchJobReport(parseResults, processingResults);
-        rassReportService.writeBatchJobReport(report);
-        sendReport(report);
```
jhulslander: it doesn't look like this function is used in RassStep any longer, should it be removed?

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassReportServiceImpl.java
```diff
@@ -81,56 +86,111 @@ public String getFileContents(String fileName) {
...
+    private void writeErrorReport(List<RassXmlFileParseResult> errorResults) {
+        initializeNewReport(RassConstants.RASS_PARSE_ERRORS_BASE_FILENAME);
+        reportWriterService.writeFormattedMessageLine("RASS Error Report");
+        reportWriterService.writeNewLines(1);
+        reportWriterService.writeFormattedMessageLine(
+                "Unexpected errors were encountered when attempting to read the following RASS XML files.");
+        reportWriterService.writeFormattedMessageLine(
+                "The failures were likely the result of incorrect XML formatting.");
+        reportWriterService.writeFormattedMessageLine(
+                "Specific details are available in the RASS batch job logs.");
```
jhulslander: Should these three lines be moved to a property or a parameter?

src/main/java/edu/cornell/kfs/rass/batch/service/RassReportService.java
```diff
@@ -4,7 +4,7 @@
 
 public interface RassReportService {
     
-    void writeBatchJobReport(RassBatchJobReport rassBatchJobReport);
+    void writeBatchJobReports(RassBatchJobReport rassBatchJobReport);
     
     void sendReportEmail();
```
jhulslander: since it looks like sendReportEmail() is just used internally to RassReportServiceImpl, should this function no longer be exposed on the interface?

src/main/java/edu/cornell/kfs/rass/RassConstants.java
```diff
@@ -7,6 +7,8 @@ private RassConstants() {
     }
 
     public static final String RASS_MAINTENANCE_NEW_ACTION_DESCRIPTION = "Creation";
+    public static final String RASS_PARSE_ERRORS_BASE_FILENAME = "errors";
+    public static final String XML_FILE_EXTENSION = ".xml";
```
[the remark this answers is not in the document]
cah292: I had meant to create a follow-up user story to standardize the ".xml" constant usage. I agree that adding this constant to CUKFSConstants is fine for this user story, and the other usage of the constant can be cleaned up in a follow-up story. I'll change this accordingly.

src/main/java/edu/cornell/kfs/rass/batch/RassStep.java
cah292: Good catch, I'll remove that method.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassReportServiceImpl.java
cah292: Ok, I'll get these three strings externalized accordingly.

src/main/java/edu/cornell/kfs/rass/batch/service/RassReportService.java
cah292: Good point, I'll remove it from the interface.

cah292:
I've committed some changes based on the initial PR feedback. This is ready for review again.

PR #863, opened 2019-10-02
KFSPTS-16702 prevent errors from halting processing

jhulslander:


src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassServiceImpl.java
```diff
@@ -174,15 +174,22 @@ protected void removeDoneFileQuietly(String rassInputFileName) {
...
+                                RassObjectUpdateResultCode.ERROR, e.getMessage());
+                    }
                     if (RassObjectUpdateResultCode.isSuccessfulResult(result.getResultCode())) {
                         documentTracker.addDocumentIdToTrack(result);
                     } else if (RassObjectUpdateResultCode.ERROR.equals(result.getResultCode())) {
                         documentTracker.addObjectUpdateFailureToTrack(result);
                         groupingResultCode = RassObjectGroupingUpdateResultCode.ERROR;
                     }
                     objectResults.add(result);
+
```
nkimble: empty line that is not needed

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassServiceImpl.java
```diff
@@ -174,15 +174,22 @@ protected void removeDoneFileQuietly(String rassInputFileName) {
...
-                    RassBusinessObjectUpdateResult<R> result = rassUpdateService.processObject(
-                            typedXmlObject, objectDefinition, documentTracker);
+                    RassBusinessObjectUpdateResult<R> result;
+                    try {
+                        result = rassUpdateService.processObject(typedXmlObject, objectDefinition, documentTracker);
+                    } catch (Throwable e) {
+                        LOG.error("updateBOs, caught an error while processing the objects", e);
+                        result = new RassBusinessObjectUpdateResult<>(objectDefinition.getBusinessObjectClass(), objectDefinition.printObjectLabelAndKeys(typedXmlObject), 
+                                RassObjectUpdateResultCode.ERROR, e.getMessage());
+                    }
```
nkimble: I realize that everything "should" be caught to set the error in the catch clause but in the event it does not, should "result" have a default value set outside the try-catch block?

src/test/java/edu/cornell/kfs/rass/RassTestConstants.java
```diff
@@ -14,6 +14,7 @@ private RassTestConstants() {
     public static final String DEFAULT_PROPOSAL_AWARD_TYPE = "S";
     public static final Integer DEFAULT_DD_FIELD_MAX_LENGTH = Integer.valueOf(50);
     public static final String ERROR_OBJECT_KEY = "ERR";
+    public static final String ERROR_OBJECT_KEY_FOR_EXPECTED_RESULTS = "Agency " + ERROR_OBJECT_KEY;
```
nkimble: Should this be named differently to reflect this is for Agency expected results? 
Meaning would we possible need this type of constant for Award, Proposal, etc. in the future where the name of the constant is too generic for the specific error that is currently being trapped?

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassServiceImpl.java
jhulslander: OK, I will remove

src/test/java/edu/cornell/kfs/rass/RassTestConstants.java
jhulslander: Yes, I will rename to signify it is Agency specific

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassServiceImpl.java
jhulslander: hmmm, good point, I will add an open else to the if block below that should handle the odd condition

jhulslander: Upon looking further, I don't think there is a case where "result" wouldn't be set,  However, I did some refactoring that will hopefully make this cleaner.

jhulslander:
I've made some changes based on the code review, this is ready for review again

PR #866, opened 2019-10-07
KFSPTS-16733 Fix AWRD-edit rendering when RASS adds fund mgr

cah292:
Our RASS code generally auto-handles inserting filler objects into old-maint-BO lists when needed. However, the fund managers list is not affected by the current auto-handling, due to its custom setup for populating that list. This PR updates the RASS Award handling to auto-insert into the old-BO fund managers list when needed, which is necessary to prevent document-rendering problems when RASS edits an existing Award to add a fund manager.


db483:
Brilliant!

PR #868, opened 2019-10-08
KFSPTS-16684 Improve RASS handling of null values

cah292:
This PR fixes some areas of the RASS batch job where null values are not being handled properly, especially with regards to primitive boolean BO fields.

I wasn't able to find a practical way to include the handling in RassValueConverterBase without reworking several method signatures. (I was hoping to use ObjectPropertyUtils to fix this, but some of the nested path handling doesn't work unless a BO instance is used, as opposed to the BO class that the converter uses.) Thus, I fixed the issue by adding a new primitive-value converter instead, to ensure that an appropriate wrapped "false" boolean will be used in the place of null Boolean values.


cah292:
Also, note that the changes from KFSPTS-16703 fixed the parameter-mapping issues related to null values, so a fix for that is not needed in this PR.

db483:
Chad,

Changes look good but I am not sure if assuming false for empty cost share and zero for total amount is ok. I think we need to check with Janet what the behavior should be in that case in terms of determining if the award/proposal should be created.

cah292:
As per the functional feedback that Daniela forwarded to me, I have committed some more changes to this PR. The is-creation-allowed checks will now return false if the total amount or cost-share-required flag are null. In addition, the following fields have now been marked as "required" in the XML mappings (if they weren't marked as such already on Proposals or Awards):

* Direct Cost Amount
* Indirect Cost Amount
* Total Amount
* Budget Total
* Cost Share Required

The configuration has been updated so that the cost-share-required flag will no longer auto-default to false. (The required-field checking should handle the flag appropriately to prevent an attempt to insert a null primitive Boolean into the BO.) I also added a total-amount property mapping to the Award translation metadata, so that a required-field check will be enforced there. Note that the existing Award total-amount setter is a no-op method (the getter derives the total at runtime instead), and that this mapping is only intended for validation purposes. If needed, I can create a follow-up user story to add a validation-only property mapping type for convenience.

PR #872, opened 2019-10-23
KFSPTS-16448: Have FastTrack mod output invoice number and invoice date for DVs when that data exists.

nkimble:


nkimble:
Note: No code changes were made for the InvoiceDate as that value was already being set regardless of the type of customer payment when there was data present.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -412,6 +412,20 @@ protected void writeExtractCheckFile(PaymentStatus extractedStatus, PaymentProce
         }
     }
 
+    private String createFormattedRemittanceIdTextBasedOnPresenceOfInvoiceDataForDv(String invoiceNumber, Date invoiceDate, String customerPaymentDocumentNumber) {
+        StringBuffer formattedRemittanceIdText = new StringBuffer("");
+
+        if (StringUtils.isNotBlank(invoiceNumber) && ObjectUtils.isNotNull(invoiceDate)) {
+            formattedRemittanceIdText.append(invoiceNumber);
+            formattedRemittanceIdText.append(" Doc No:");
```
cah292: For the combined string that has both invoice number and document number, I believe Marcia wanted the doc number label to be changed to "Doc:" instead. If needed, you can confirm with Marcia whether this label should be "Doc No:" or just "Doc:". (I think the remittance ID field has a large enough max length to accommodate either option.)

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -412,6 +412,20 @@ protected void writeExtractCheckFile(PaymentStatus extractedStatus, PaymentProce
         }
     }
 
+    private String createFormattedRemittanceIdTextBasedOnPresenceOfInvoiceDataForDv(String invoiceNumber, Date invoiceDate, String customerPaymentDocumentNumber) {
+        StringBuffer formattedRemittanceIdText = new StringBuffer("");
```
cah292: I would suggest using a StringBuilder instead for better efficiency. The builder also defaults to being empty, so you don't need to explicitly specify an empty string as the initial value.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Marcia has confirmed we can use "Doc:" as the label for both this change and the existing code paths when there is no invoice number or date. Fixing.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
nkimble: Changing to StringBuilder as suggested. 
Since the existing code explicitly set the value to an empty string when neither condition was encountered, I would like to retain that for the initialization if you are ok with that.

nkimble:
Ready to be reviewed again.

PR #874, opened 2019-10-24
KFSPTS-16943: Set default country code in business object ACHBank to US in constructor utilitzed for Federal Bank Data file loading by the batch job pdpLoadFederalReserveBankDataStep.

nkimble:


nkimble:
The only changes from base code are the addition of setter and getter for bankCountryCode and the addition of line 102.

src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
```diff
@@ -0,0 +1,254 @@
...
+        // the ojb insert explicitly specifies every column it knows about. this will work for now.
+        bankCountryCode = KFSConstants.COUNTRY_CODE_UNITED_STATES;
+    }
+
+    /**
+     * This constructor takes a line of data from FedACH Directory file and populates the ACHBank object
+     *
+     * @param fileData line of data from the FedACH Directory file
+     */
+    public ACHBank(String fileData) {
```
cah292: The original file appears to have a call to "setBankDataViewCode" in this method, but I don't see that call in this overlay. Can you please add that back in if necessary?

nkimble: Good catch. Thanks.

nkimble:
Ready for review again

PR #875, opened 2019-10-25
KFSPTS-16329 handle poorly formatted attachment links

jhulslander:


src/test/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImplTest.java
```diff
@@ -1122,6 +1129,22 @@ public void setClient(Client client) {
...
+        protected Invocation buildClientRequest(String url, Collection<WebServiceCredential> creds) throws URISyntaxException {
+            if (StringUtils.equals(url, 
+                    "https://www.cornell.edu/v1_0/invoice?year=2019&month=09Sep&filename=J%20Aron%20&%20Company%20B34367.pdf")) {
+                return allowClientToFailWithPoorlyFormatedUrl(url);
+            } else {
+                return super.buildClientRequest(url, creds);
+            }
+        }
+
+        private Invocation allowClientToFailWithPoorlyFormatedUrl(String url) throws URISyntaxException {
```
db483: Could you help me understand why this method is needed?

src/test/resources/edu/cornell/kfs/fp/batch/xml/single-di-bad-attach-document-test.xml
```diff
@@ -0,0 +1,49 @@
...
+                    <sub_object_cd></sub_object_cd>
+                    <project></project>
+                    <org_ref_id></org_ref_id>
+                    <line_description></line_description>
+                    <amount>100.00</amount>
+                </Accounting>
+            </TargetAccountingLineList>
+            <BackupDocumentLinks>
+                <BackupLink>
+                	<Link>https://www.cornell.edu/v1_0/invoice?year=2019&amp;month=09Sep&amp;filename=J%20Aron%20&amp;%20Company%20B34367.pdf</Link>
```
db483: there's an extra space or tab at the beginning of this link

src/test/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImplTest.java
```diff
@@ -427,6 +429,12 @@ public void testServiceError() throws Exception {
         boolean results = createAccountingDocumentService.createAccountingDocumentsFromXml();
         assertFalse("When there is a problem calling services, the job should fail", results);
     }
+    
+    @Test
+    public void testLoadSingleFileWithSingleDIDocumentWithBadAttachment() throws Exception {
```
db483: Is the attachment bad or the URI? would it be better to make this method name more specific to what the problem is?

jhulslander: how about changing it to testLoadSingleFileWithSingleDIDocumentWithBadAttachmentUrl() ?

src/test/resources/edu/cornell/kfs/fp/batch/xml/single-di-bad-attach-document-test.xml
jhulslander: I'll fix the formatting

src/test/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImplTest.java
db483: sounds good, thank you

src/test/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImplTest.java
jhulslander: Whem the bad attachment URL example is passed to the function call "super.getClient().target(uri).request()", a UriBuilderException is thrown.  I wanted to create this scenario like we see in the batch job.  I am guessing a better function name would help here.  Do you have a recommendation?

db483: When this error happens in production is it being thrown by the same code in buildClientRequest? I still don't understand why we had to override the buildClientRequest implementation

jhulslander: The client is set to a mock client that is useful for the other tests.  THe client was needed for this test so that it would throw the real exception.
And Yes, this where it happening in production I believe

db483: I understand now. I think changing this method's name would make sense. Something that would explain it is not using the mock client.

src/test/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImplTest.java
```diff
@@ -1122,6 +1129,22 @@ public void setClient(Client client) {
             this.mockClient = client;
         }
         
+        @Override
+        protected Invocation buildClientRequest(String url, Collection<WebServiceCredential> creds) throws URISyntaxException {
+            if (StringUtils.equals(url,
```
db483: Jay and I chatted about changing the override to use a boolean instead of checking on the specific URL

jhulslander:
I've made some adjustments, this is ready for review again.

PR #877, opened 2019-10-29
KFSPTS-16884: Renumber Cornell mod specific suspension codes due to base code adding a new suspension code.

nkimble:


jhulslander:
I spoke with Nancy and reviewed Analysis on https://jira.cornell.edu/browse/KFSPTS-16960 ... it looks like my concerns have been considered.  Kualico patch SQL should handle these collisions for us.

PR #878, opened 2019-10-30
KFSPTS-16992 Fix error email handling in PDP Load Payments

cah292:
There are cases where a bad PDP file will throw an exception other than ParseException, which interferes with sending error emails since such code normally only does so for ParseException. This fix changes the PDP Load Payments code so that other types of exceptions can also trigger PDP error emails.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
@@ -164,16 +164,13 @@ public class CuPaymentFileServiceImpl extends PaymentFileServiceImpl {
         }
     }
     
-    /**
-     * This method sends error emails in a manner similar to that from PaymentInputFileType.parse(),
-     * except that our code will also attempt to send an email to the customer if possible.
-     */
     private void sendErrorEmailForGenericException(
             Exception genericException, String incomingFileName, PaymentFileLoad paymentFile) {
         PaymentFileLoad paymentFileForEmail;
         try {
             paymentFileForEmail = getCustomerProfileFromUnparsableFile(incomingFileName, paymentFile);
         } catch (RuntimeException e) {
+            LOG.error("sendErrorEmailForGenericException: Could not read file contents as plain text", e);
             paymentFileForEmail = paymentFile;
         }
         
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -159,6 +164,25 @@ private void assignDisbursementTypeCode(PaymentGroup pg) {
...
+     * This method sends error emails in a manner similar to that from PaymentInputFileType.parse(),
+     * except that our code will also attempt to send an email to the customer if possible.
+     */
+    private void sendErrorEmailForGenericException(
+            Exception genericException, String incomingFileName, PaymentFileLoad paymentFile) {
+        PaymentFileLoad paymentFileForEmail;
+        try {
+            paymentFileForEmail = getCustomerProfileFromUnparsableFile(incomingFileName, paymentFile);
+        } catch (RuntimeException e) {
+            paymentFileForEmail = paymentFile;
```
jhulslander: should we log this exception?

jhulslander: should paymentFileForEmail be set to something if the parse fails?

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
```diff
@@ -159,6 +164,25 @@ private void assignDisbursementTypeCode(PaymentGroup pg) {
         }
     }
     
+    /**
+     * This method sends error emails in a manner similar to that from PaymentInputFileType.parse(),
```
jhulslander: I am not sure if we need these comments

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
cah292: Ok, I'll add some logging here just in case.

However, I don't understand what you mean about "paymentFileForEmail be set to something if the parse fails". If the plain-text parse fails, it'll get set to the "paymentFile" value that was passed into the method, though "paymentFile" could be null in such a case. When the email-sending code gets called, it will automatically ignore the passed-in "paymentFileForEmail" value if it's null, so I don't think we need to worry about a NullPointerException here.

If there's some other potential problem here that you meant to point out, please let me know.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
cah292: Removing comments.

cah292:
I've committed some more changes in response to the initial feedback. This is ready for review again.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileServiceImpl.java
jhulslander: Thanks for the further explanation, I think that answers the concern I was trying to mention.

PR #879, opened 2019-10-30
KFSPTS-16975 adding bank state and postal code setters to ACHBank, also added logging to help with future debugging

jhulslander:
We were not able to reproduce this problem locally, however adding these setter should get us past these problems.  I added some log statements in hopes to help future debugging needs

### DIFF
--- a/src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
+++ b/src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
@@ -18,6 +18,9 @@
  */
 package org.kuali.kfs.pdp.businessobject;
 
+import org.apache.commons.lang.builder.ToStringBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
 import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.businessobject.PostalCode;
@@ -25,6 +28,7 @@ import org.kuali.kfs.sys.businessobject.State;
 import org.kuali.rice.core.api.mo.common.active.MutableInactivatable;
 
 public class ACHBank extends PersistableBusinessObjectBase implements MutableInactivatable {
+    private static final Logger LOG = LogManager.getLogger(ACHBank.class);
 
     protected String bankRoutingNumber;
     protected String bankOfficeCode;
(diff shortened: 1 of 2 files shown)


src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
```diff
@@ -18,13 +18,17 @@
  */
 package org.kuali.kfs.pdp.businessobject;
 
+import org.apache.commons.lang.builder.ToStringBuilder;
```
cah292: Please use the lang3 version of ToStringBuilder instead.

src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
```diff
@@ -252,4 +256,14 @@ public String getBankCountryCode() {
     public void setBankCountryCode(String bankCountryCode) {
         this.bankCountryCode = bankCountryCode;
     }
+
+    public void setBankState(State bankState) {
+        LOG.info("setBankState, setting bank state to: " + ToStringBuilder.reflectionToString(bankState));
```
cah292: Should these logging statements be set to debug level instead of info level?

Also, if you switch over to using lang3 ToStringBuilder, I think that version of it does not allow null values for "reflectionToString".

src/test/java/org/kuali/kfs/pdp/businessobject/ACHBankTest.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Test
+    public void testSetBank() {
+        bank.setBankState(state);
+        assertEquals(state, bank.getBankState());
+    }
+    
+    @Test
+    public void testSetBankMull() {
```
cah292: Minor typo in method name; should be changed from "...BankMull" to "...BankNull"

src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
jhulslander: OK

src/main/java/org/kuali/kfs/pdp/businessobject/ACHBank.java
jhulslander: Good point about making these debug log statements.  I shall change.  And yes, you are correct ToStringBuilder errors on null with the lang3 version

src/test/java/org/kuali/kfs/pdp/businessobject/ACHBankTest.java
jhulslander: will fix

jhulslander:
I've made some changes, this is ready for review again

PR #886, opened 2019-11-04
KFSPTS-16683 upgrade to the 2019-07-25 version of financials

db483:


cah292:
Also, I see that this KFS release updates a lot of its values finders so that they use Spring injection to populate the KeyValuesService where needed. I'll create a follow-up user story to update our cu-kfs values finders to do the same.

PR #888, opened 2019-11-05
KFSPTS-16448: Have FastTrack mod output invoice number and invoice date to FastTrack file and advices for DVs when that data exists.

nkimble:


nkimble:
Marcia's preliminary validation in DEV indicated the original changes looked good but advices also needed to be updated with that same information.  This PR is adding those advice changes to that original commit.

PR #892, opened 2019-11-19
KFSPTS-15559 Update create-acct-doc to handle JAXB errors

cah292:
Certain JAXB errors (such as XML adapter errors) don't cause the Create Accounting Documents parsing to fail by default, but configuring the parsing to auto-fail on such errors is not desirable either since it will fail the whole file and not an individual XML-based doc. Instead, this PR sets up JAXB listeners and related code so that these types of JAXB errors will get recorded, and then the ones associated with a particular XML-based document will cause just that document to fail validation. (Any XML-based documents within the file that don't have such errors will still get processed normally, as long as there wasn't some other fatal error.)

This new code also attempts to make the error messages a little more user-friendly by printing the error line number and removing the exception classname from the message. Please let me know if you think the exception classname should still remain in the final error message.

Also, the "CreateDate" in the "DocumentOverview" element is not being validated by these updates since it's at the file level instead of the document level, and I don't think the existing Create Accounting Documents code is validating it either. I was thinking of creating a separate user story to add in validation of that element, but if you think it should be included in this PR's changes, please let me know.


jhulslander:
Chad, I don't think the error messages displayed on the create accounting document report should contain a class name.  That would only add confusion to our customers reading the report.  If we could include the class name detail in the log, that would be helpful for a developer though.

As far as validating the create date for the file, I am inclined to think that isn't necessary.  It is just value displayed on the report.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
```diff
@@ -22,6 +28,11 @@
 public class CreateAccountingDocumentValidationServiceImpl implements CreateAccountingDocumentValidationService {
     private static final Logger LOG = LogManager.getLogger(CreateAccountingDocumentValidationServiceImpl.class);
     
+    private static final Pattern EXCEPTION_MESSAGE_PATTERN = Pattern.compile(
+            "^(?<exceptionClassname>([\\w$]+\\.)+[\\w$]+(Exception|Error|Throwable))((: ?)(?<detailMessage>.+))?$");
```
jhulslander: I'm wondering if this pattern string would be better to keep in a property file.  I am not sure where it is best kept.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
```diff
@@ -126,6 +138,78 @@ private boolean isExplanationValid(String explanation, CreateAccountingDocumentR
...
+        if (lineNumber != -1) {
+            return formatMessageFromProperty(CuFPKeyConstants.ERROR_CREATE_ACCOUNTING_DOCUMENT_XML_ADAPTER_ERROR,
+                    lineNumber, subMessage);
+        } else {
+            return subMessage;
+        }
+    }
+    
+    private String getErrorSubMessage(ValidationEvent validationError) {
+        String eventMessage = getErrorEventMessage(validationError);
```
jhulslander: This function is getting rather long, can it be refactored into some smaller functions?

cah292:
Thanks for the feedback. With regards to adding the classname information in the logs, the "getErrorSubMessage" method should be doing so.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
```diff
@@ -126,6 +138,78 @@ private boolean isExplanationValid(String explanation, CreateAccountingDocumentR
...
+                + eventMessage);
+        Matcher matcher = EXCEPTION_MESSAGE_PATTERN.matcher(eventMessage);
+        if (!matcher.matches()) {
+            return eventMessage;
+        }
+        
+        String exceptionClassname = matcher.group(EXCEPTION_CLASSNAME_GROUP);
+        if (StringUtils.isBlank(exceptionClassname)) {
+            throw new IllegalStateException(
+                    "Pattern matched message but it does not contain exception classname; this should NEVER happen!");
```
jhulslander: Since we don't want to display the exception class on the report, do we need to throw this exception? perhaps just log the anomaly,

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
```diff
@@ -126,6 +138,78 @@ private boolean isExplanationValid(String explanation, CreateAccountingDocumentR
...
+            if (StringUtils.isBlank(errorMessage)) {
+                errorMessage = getStringProperty(CuFPKeyConstants.ERROR_CREATE_ACCOUNTING_DOCUMENT_GENERIC_ERROR);
+            }
+        }
+        return errorMessage;
+    }
+    
+    private int getErrorLineNumber(ValidationEvent event) {
+        ValidationEventLocator locator = event.getLocator();
+        return locator != null ? locator.getLineNumber() : -1;
```
jhulslander: is there a situation were ValidationEventLocator would not have the line number?  If so should the conditional also check that?

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentListener.java
```diff
@@ -0,0 +1,50 @@
+package edu.cornell.kfs.fp.batch.xml;
+
+import javax.xml.bind.Unmarshaller;
+import javax.xml.bind.ValidationEvent;
+import javax.xml.bind.ValidationEventHandler;
+
+public class AccountingXmlDocumentListener extends Unmarshaller.Listener implements ValidationEventHandler {
```
jhulslander: Should the class name be AccountingXmlDocumentUnmarshalListener since this for unmarshaling?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
cah292: Ok, I'll see if I can externalize this string to one of our config property files.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
cah292: I'll get this refactored into a few more smaller functions.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
cah292: Good point, I'll update this to log the situation instead of triggering a failure.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
cah292: According to the ValidationEventLocator Javadocs, the getLineNumber() method is supposed to return -1 if the line number is unavailable. Thus, I don't think any changes are needed here.

src/main/java/edu/cornell/kfs/fp/batch/xml/AccountingXmlDocumentListener.java
cah292: Good idea, I'll rename this class accordingly.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentValidationServiceImpl.java
jhulslander: Sounds good

cah292:
I've committed some changes based on the initial feedback. This is ready for review again.

PR #896, opened 2019-11-27
KFSPTS-17347 add ObjectUtils null checks to prevent errors

db483:


jhulslander:
These changes look fine, and can be merged once the unit tests pass

PR #902, opened 2019-12-11
KFSPTS-17197 removing spring context from values finder

jhulslander:
This PR is ready to merge after it passes code review.


src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -330,5 +330,13 @@
...
     <bean id="batchFileLookupable-parentBean" class="org.kuali.kfs.kns.lookup.KualiLookupableImpl" abstract="true"
           p:lookupableHelperService-ref="batchFileLookupableHelperService"/>
+          
+	<bean id="simpleChartValuesFinder" class="edu.cornell.kfs.coa.businessobject.options.CuSimpleChartValuesFinder"
+		p:parameterService-ref="parameterService"
+		p:keyValuesService-ref="keyValuesService"
+		p:financialSystemUserService-ref="financialSystemUserService"/>
+	
+	<!-- Temporary alias to allow legacy Struts pages to work properly, due to class change in bean override above. -->
+	<alias name="simpleChartValuesFinder" alias="cuSimpleChartValuesFinder"/>
```
cah292: Was it because of the "financialSystemUserService" reference that you had to move these beans into "cu-spring-sys" instead? It would be preferable to keep these beans in the "KfsBaseAttributeDefinitions.xml" file instead.

Not sure if you tried this yet, but you could try using a service-importing factory bean to import the "financialSystemUserService" bean into the DD bean area. That way, you can reference the service and still keep the values finders within the DD bean files. (You can also try this for the IWantDocumentService injection.)

Many of KFS's "lookup-beans" files import the "spring-kfs-imported-rice-beans.xml" file, which contains an abstract "grlBeanImporter" bean that can be used as the basis for importing KFS services into the DD bean area. You may be able to create child beans of "grlBeanImporter" that can import the FinancialSystemUserService and IWantDocumentService. (You'll have to experiment to see if "spring-kfs-imported-rice-beans.xml" has to be explicitly imported beforehand.)

jhulslander: Yes, I moved simpleChartValuesFinder to get access to financialSystemUserService . I will give the import idea a try

jhulslander:
Thanks Chad, the "grlBeanImporter" idea worked great.  This is ready for review again.

src/main/java/edu/cornell/kfs/fp/businessobject/options/PaymentMethodValuesFinderNonInterdepartmental.java
```diff
@@ -1,10 +0,0 @@
-package edu.cornell.kfs.fp.businessobject.options;
-
-import edu.cornell.kfs.fp.CuFPConstants;
-
-
-public class PaymentMethodValuesFinderNonInterdepartmental extends PaymentMethodValuesFinder {
-    static {
-        filterCriteria.put(CuFPConstants.INTERDEPARTMENTAL_PAYMENT, CuFPConstants.NO);
```
cah292: This subclass was modifying the static filterCriteria Map on the superclass once loaded by the classloader. If there are documents which use PaymentMethodValuesFinder specifically, then we need to make sure that this class removal does not mess up their drop-down values.

src/main/java/edu/cornell/kfs/module/purap/businessobject/options/CollegeLevelOrganizationValuesFinder.java
```diff
@@ -34,4 +30,8 @@
         return keyValues;
     }
 
+    public void setiWantDocumentService(IWantDocumentService iWantDocumentService) {
```
cah292: Can you please confirm that this setter is named correctly for being invoked by Spring, even though the first "i" is lowercased? Does it need to be renamed to "setIWantDocumentService" instead?

jhulslander: I tested the I want document locally, and the college/division select list was populated.  I think that confirms this is working as expected.  Please let me know if you think something else should be checked.

cah292: Yes, you're correct that this is the right naming for this setter, after I did some searching online and experimented with Eclipse's getter/setter generation feature. Sorry for the confusion.

src/main/java/edu/cornell/kfs/fp/businessobject/options/PaymentMethodValuesFinderNonInterdepartmental.java
cah292: I'm okay with this class removal moving forward, but the functional users will need to verify that the drop-down values are still correct for the places that depend on the base PaymentMethodValuesFinder (such as Payment Request and Credit Memo documents).

PR #903, opened 2019-12-13
KFSPTS-16809 Move Sub-Fund Program to new lookup framework

cah292:
Note that there are two PRs for this change: One in cu-kfs and one in nonprod-sql.

The Sub-Fund Program objects were still on the old lookup framework, which is capable of truncating certain field values that are too long. This PR migrates the Sub-Fund Program lookup to the new framework so that it doesn't experience the truncation issue.


cah292:
The stored-value truncation issue (which also impacts the inquiry screen) appears to have already been fixed by KFSPTS-15510, where updates were made so that the database and the Data Dictionary would have the same max length for this field.

I'm not sure what you mean about adding the lookup. The Sub-Fund Program already had a lookup, but it was using the old framework (which can truncate the displayed values of certain fields). For this PR, I added the configuration needed to have the Sub-Fund Program lookup use the new framework instead, which doesn't have the displayed-value truncation issue.

jhulslander:
I was referring to adding lookupAttributeDefinitions and lookupResultAttributeDefinitions tags but I think you answered my question.  I shall merge these PRs

PR #904, opened 2019-12-18
KFSPTS-17118 Fix detail validation for PDP Load Payments

cah292:
The PDP Load Payments Job was not properly validating certain Payment Detail field lengths on the input files, which could lead to unexpected errors when saving to the database. This PR adds in some Payment Detail field length checks that are similar to those for the job's Payment Group validation. Additionally, some of the Payment Detail data dictionary configuration was overridden, so that the max lengths would be consistent with the column lengths in the database.

There are still some account-related fields that would be good to add to this validation, but that work will be deferred to a future sprint, as per the input from the functionals.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
@@ -1,10 +1,14 @@
 package edu.cornell.kfs.pdp.service.impl;
 
+import java.util.List;
 import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kuali.kfs.krad.util.ErrorMessage;
+import org.kuali.kfs.krad.util.MessageMap;
+import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.pdp.PdpKeyConstants;
 import org.kuali.kfs.pdp.businessobject.PayeeType;
 import org.kuali.kfs.pdp.businessobject.PaymentDetail;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
```diff
@@ -216,11 +221,34 @@ protected void checkPaymentGroupPropertyMaxLength(PaymentGroup paymentGroup, Mes
         }
     }
 
+    // This is an altered copy of checkPaymentGroupPropertyMaxLength() that operates on Payment Details instead.
+    protected void checkPaymentDetailPropertyMaxLength(PaymentDetail paymentDetail, MessageMap errorMap) {
+        for (String propertyName : paymentDetailPropertiesToCheckMaxLength) {
+            String propertyValue = (String) ObjectUtils.getPropertyValue(paymentDetail, propertyName);
+            if (StringUtils.isNotEmpty(propertyValue)) {
+                Integer maxLength = dataDictionaryService.getAttributeMaxLength(PaymentDetail.class, propertyName);
+                if ((maxLength != null) && (maxLength < propertyValue.length())) {
```
jhulslander: if max length isn't set, should we log it as an error condition?

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
```diff
@@ -216,11 +221,34 @@ protected void checkPaymentGroupPropertyMaxLength(PaymentGroup paymentGroup, Mes
         }
     }
 
+    // This is an altered copy of checkPaymentGroupPropertyMaxLength() that operates on Payment Details instead.
```
jhulslander: do we need this comment?

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
cah292: The not-null setup here was how the related base code was set up, but I agree that some extra error handling would be good to add here if the length is null. I'll update this to throw an error if the max length is not specified, since such a situation would indicate a more significant problem with the system.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
cah292: I don't this comment is really adding much value, so I'll remove it. Thanks.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentFileValidationServiceImpl.java
cah292: After further discussion with Jay, I will update this to just use the Logger to log the no-max-length situation, rather than throwing an error.

cah292:
I've committed some changes based on the initial PR feedback. This is ready for review again.

PR #907, opened 2020-01-03
KFSPTS-16511: New integration to create csv file of closed accounts for WorkDay. Code has ability to create initial seed file as well as incremental file based on KFS system parameter.

nkimble:

PR #909, opened 2020-01-10
KFSPTS-16511: New integration to create csv file of closed accounts for Workday. Code has ability to create initial seed file as well as incremental file based on KFS system parameters.

nkimble:


nkimble:
This PR relies upon nonprod SQL PR #706.
The code in this PR is the same as what was submitted in PR #907 with the following changes:
* The term “labor” was removed from this mod.
* Method CreateClosedAccountsCsvServiceImpl.ensureOutbundDirectoriesExist that was removed during the previous code review per the reviewer’s request was put back.  Further local testing with this set of changes proved that Spring does not automatically create directories and you need to manually ensure the file system has the directory you intend to access; otherwise, the code will generate a “java.io.FileNotFoundException:” with a “(No such file or directory)” pointing to the full directory path and file the code is attempting to access.
* Code was added to have the outbound CSV file created in one directory and moved to another directory once all data had been written to the file and the file was closed.  This change was needed to prevent data loss while the outbound file is being created due to the use of the AWS inotify docker containers monitoring feature that is being used to move files from the cloud to onprem servers.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/ClosedAccountsByDateRangeDaoJdbc.java
```diff
@@ -0,0 +1,75 @@
...
+    public List<ClosedAccount> obtainClosedAccountsDataFor(Map<String, Date> dateRange) {
+        List<ClosedAccount> closedAccounts = findAllClosedAccountsFor(dateRange);
+        return closedAccounts;
+    }
+    
+    private List<ClosedAccount> findAllClosedAccountsFor(Map<String, Date> dateRange) {
+        try {
+            RowMapper<ClosedAccount> mapRow = new RowMapper<ClosedAccount>() {
+                public ClosedAccount mapRow(ResultSet resultSet, int rowNumber) throws SQLException {
+                    ClosedAccount laborClosedAccountDataRow = new ClosedAccount();
```
jhulslander: a few cases of labor used here

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CreateClosedAccountsCsvServiceImpl.java
```diff
@@ -0,0 +1,306 @@
...
+        if (ObjectUtils.isNotNull(closedAccountsData) && closedAccountsData.size() > 0) {
+            LOG.info("writeClosedAccountsToCsvFormattedFile: Have the data. Will now write out " + closedAccountsData.size() + " closed accounts to CSV file.");
+            writeClosedAccountsToCsvFile(closedAccountsData);
+        } else {
+            LOG.info("writeClosedAccountsToCsvFormattedFile: NO CSV file being created. " 
+                    + (ObjectUtils.isNotNull(closedAccountsData) ? (closedAccountsData.size() + " accounts being written to file.") : ("null returned for requested account data.") ));
+        }
+    }
+    
+    private void ensureOutbundDirectoriesExist() {
```
jhulslander: I spoke with Nancy about this, and spring should create the directories at boot time if the directories are defined in <property name="directoryPath"> similar to below.  

<property name="batchFileDirectories">
            <list merge="true">
                <value>${staging.directory}/coa/whatever</value>
            </list>
        </property>

nkimble: Jay and I discussed this code review suggestion again when it would now work as initially described.  Based on that second information sharing, it was concluded that the existing code is the solution to implement.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/ClosedAccountsByDateRangeDaoJdbc.java
nkimble: removed..

PR #924, opened 2020-02-19
KFSPTS-11929 leave the role id null so that the role id sequence is used to set it when saving to the database

db483:


src/main/java/edu/cornell/kfs/sec/document/CuSecurityModelMaintainableImpl.java
```diff
@@ -0,0 +1,12 @@
+package edu.cornell.kfs.sec.document;
+
+import org.kuali.kfs.sec.businessobject.SecurityModel;
+import org.kuali.kfs.sec.document.SecurityModelMaintainableImpl;
+
+public class CuSecurityModelMaintainableImpl extends SecurityModelMaintainableImpl {
+
+    protected String buildModelRoleId(SecurityModel securityModel) {
```
cah292: Please add the "Override" annotation if this overriding a superclass method.

db483: Good point! Adding

PR #926, opened 2020-02-25
KFSPTS-17565 update pom to the 9/5/2019 version of financials and get…

db483:
… things to compile

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
         <skipTests>true</skipTests>
         <test.excludes>**/*Test.java.bogus</test.excludes>
 
-        <kfs.version>2019-08-15</kfs.version>
+        <kfs.version>2019-09-05</kfs.version>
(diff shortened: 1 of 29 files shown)


db483:
[1] updated
[2] not sure what needs to change for cuVendorService-parentBean
[3] updated
[4] updated

cah292:
After further analysis, I don't think "cuVendorService" or "cuVendorService-parentBean" need to be updated at this time. That bean doesn't replace the base "vendorService" financials bean, and for the cases that actually use "cuVendorService", they don't seem to be impacted by the changes from this financials upgrade.

I'm fine with leaving those vendor service beans as-is, though I will create a follow-up user story to determine whether we still need the "cuVendorService" bean at all. Sorry for the confusion.

db483:
Thank you Chad. I have added all other requested changes.

PR #928, opened 2020-03-04
KFSPTS-16119 Remove mod that increased error columns sizes

db483:


ajd299:
The 2 files seem safe to be remove. Is it feasible to also remove CuContractsGrantsInvoiceCreateDocumentServiceImpl.storeValiationErrors?

db483:
Good question. We still have code in place that checks for null amount before setting so we cannot yet remove ir:
 KualiDecimal awardTotalAmount = award.getAwardTotalAmount() == null ? KualiDecimal.ZERO : award.getAwardTotalAmount();

ajd299:
Okay thank you for pointing that out. The code looks good.

PR #932, opened 2020-03-04
KFSPTS-18056 fixing concur event notification

jhulslander:


src/test/resources/edu/cornell/kfs/concur/rest/xmlObjects/fixture/travel-request-detail-dto.xml
```diff
@@ -0,0 +1,93 @@
+<TravelRequestDetails xmlns="http://www.concursolutions.com/api/travelrequest/2012/06" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
```
cah292: This file contains several "key" fields ("CashAdvanceKey", "RequestKey", etc.). Are any of those fields considered sensitive.

jhulslander: I don't think so, it is just a hashed value, but it might be a good idea to clear it out

cah292: Thanks, the test file looks good now.

jhulslander:
I've cleaned up some the test data you found.  Also, this was a from a "Test user" submission in concur, so is also a second piece of info that the values aren't sensitive, I did clean up some of those values

PR #933, opened 2020-03-09
KFSPTS-17693 update pay reqs when related checks are canceled

jhulslander:
@cah292  This makes updates as requested in the user story.  I wasn't able to verify this fixes the bug locally.  I think this scenario is going to need to be tested in dev unless you know of steps to verify locally

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
@@ -93,6 +93,7 @@ public class CuPendingTransactionServiceImpl extends PendingTransactionServiceIm
 	private DocumentService documentService;
     private NoteService noteService;
     private CheckReconciliationDao checkReconciliationDao;
+    protected PurapAccountRevisionService purapAccountRevisionService;
 
     /**
      * @see org.kuali.kfs.pdp.service.PendingTransactionService#generateCRCancellationGeneralLedgerPendingEntry(org.kuali.kfs.pdp.businessobject.PaymentGroup)
@@ -510,7 +511,7 @@ public class CuPendingTransactionServiceImpl extends PendingTransactionServiceIm
 
             // manually save cm account change tables (CAMS needs this)
 
-                SpringContext.getBean(PurapAccountRevisionService.class).cancelCreditMemoAccountRevisions(cm.getItems(), cm.getPostingYearFromPendingGLEntries(), cm.getPostingPeriodCodeFromPendingGLEntries());
+            purapAccountRevisionService.cancelCreditMemoAccountRevisions(cm.getItems(), cm.getPostingYearFromPendingGLEntries(), cm.getPostingPeriodCodeFromPendingGLEntries());
       
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
```diff
@@ -962,7 +957,14 @@ protected boolean generateEntriesPaymentRequest(PaymentRequestDocument preq, Lis
 
           
         }
-
+        
+        if (CollectionUtils.isNotEmpty(preq.getItems())) {
```
cah292: Based on the description for this user story, I think the cancel-revisions logic should be executed when "summaryAccounts" is non-empty. Can this "if" condition be updated accordingly, or be removed and have its internal code moved into the block above that already tests the "summaryAccounts" list?

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
```diff
@@ -962,7 +957,14 @@ protected boolean generateEntriesPaymentRequest(PaymentRequestDocument preq, Lis
...
+            LOG.info("generateEntriesPaymentRequest, cancel previous revisions of preq document: " + preq.getDocumentNumber());
+            purapAccountRevisionService.cancelPaymentRequestAccountRevisions(preq.getItems(), preq.getPostingYearFromPendingGLEntries(), 
+                    preq.getPostingPeriodCodeFromPendingGLEntries());
+        } else {
+            LOG.info("generateEntriesPaymentRequest, unable to cancel previous revisions becuase they items are empty for document: " + preq.getDocumentNumber());
```
cah292: Minor typo in label; should be changed from "...becuase they items..." to "...because the items..." instead.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
jhulslander: I will move this code further up so it is in if statement that checks if summaryAccounts is populated.  I do think the if statement checking preq.getItems() is needed because when looking at PurapAccountRevisionService.cancelPaymentRequestAccountRevisions, it will error of getItems is null, and won't do anything if it is empty

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
jhulslander: ok, will fix

jhulslander:
I've made some changes based on the code review, this is ready again

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
```diff
@@ -959,11 +954,16 @@ protected boolean generateEntriesPaymentRequest(PaymentRequestDocument preq, Lis
...
             }
-
-          
+            
+            if (CollectionUtils.isNotEmpty(preq.getItems())) {
+                LOG.info("generateEntriesPaymentRequest, cancel previous revisions of preq document: " + preq.getDocumentNumber());
+                purapAccountRevisionService.cancelPaymentRequestAccountRevisions(preq.getItems(), preq.getPostingYearFromPendingGLEntries(), 
+                        preq.getPostingPeriodCodeFromPendingGLEntries());
+            } else {
+                LOG.info("generateEntriesPaymentRequest, unable to cancel previous revisions becuase the items are empty for document: " + preq.getDocumentNumber());
```
cah292: Minor typo in label; should be changed from "...becuase..." to "...because..." instead.

jhulslander:
the typo has been fixed, this is ready for review again

PR #937, opened 2020-03-12
KFSPTS-18168 update pom checkstyle version to 8.29 on com.puppycrawl.…

db483:
…tools


cah292:
Shouldn't this be upgraded to 8.29 instead of 8.28?

db483:
Correct, updating

PR #940, opened 2020-03-16
KFSPTS-18181 adding vendor address ID to create accounting document Job DV processing

jhulslander:
Please see https://jira.cornell.edu/browse/KFSPTS-18181 for the business rules we are attempting to implement.

Some error handling and edge case handling is deferred to future user stories.  I think we also want to defer unit testing till later

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
@@ -181,7 +181,7 @@ public class CuDisbursementVoucherDocumentGenerator extends AccountingDocumentGe
             fieldValues.put(KFSPropertyConstants.VENDOR_ADDRESS_GENERATED_ID, paymentInfo.getPayeeAddressId());
             fieldValues.put(KFSPropertyConstants.ACCOUNT_ACTIVE_INDICATOR,  KFSConstants.ACTIVE_INDICATOR);
             vendorAddresses = businessObjectService.findMatching(VendorAddress.class, fieldValues);
-        } catch (Throwable e) {
+        } catch (Exception e) {
             LOG.error("findVendorAddress, there was an error attempting to find the vendor address", e);
             throw new ValidationException("Unable to find the vendor address: " + e.getMessage());
         }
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
```diff
@@ -147,6 +150,51 @@ protected void validateAndPopulatePayeeTypeCodeAndPayeeNumber(DisbursementVouche
...
+                        " vendorDetailAssignedIdentifier: " + payeeArray[1] + 
+                        ", payee address id: " + paymentInfo.getPayeeAddressId());
+            }
+            Map<String, String> fieldValues = new HashMap<String, String>();
+            fieldValues.put(KFSPropertyConstants.VENDOR_HEADER_GENERATED_ID, payeeArray[0]);
+            fieldValues.put(KFSPropertyConstants.VENDOR_DETAIL_ASSIGNED_ID, payeeArray[1]);
+            fieldValues.put(KFSPropertyConstants.VENDOR_ADDRESS_GENERATED_ID, paymentInfo.getPayeeAddressId());
+            fieldValues.put(KFSPropertyConstants.ACCOUNT_ACTIVE_INDICATOR,  KFSConstants.ACTIVE_INDICATOR);
+            vendorAddresses = businessObjectService.findMatching(VendorAddress.class, fieldValues);
+        } catch (Throwable e) {
```
cah292: Is catching a Throwable going to be too broad? Should this just capture Exception instead?

jhulslander: I wanted to make sure every odd thing got caught here, and got handled ..... I was concerned with making sure any string split errors were handled

cah292: I would think the string-splitting errors would extend from Exception. If you're catching at the level of Throwable, the code could end up catching out-of-memory errors and things like that.

jhulslander: OK, I will switch to exception

jhulslander:

PR #941, opened 2020-03-19
KFSPTS-17530 Update cu-kfs POM for Java 11 use

cah292:
There are PRs for this in cornell_customizations, cynergy-server-customizations, cynergy-standalone, and cu-kfs.

This PR adds in some cu-kfs POM changes that are similar to those that Jay temporarily added when making cu-kfs work with Java 11. It also tweaks the Cynergy version number slightly to account for the changes proposed in the related Cynergy PRs. If you think some of these POM changes are unneeded, please let me know.

I'm not certain if the PR unit tests are running on Java 11 yet. Please check with DevOps on whether that has been set up.

Also, depending on what artifacts have already been deployed to nexus as a result of the IntDev testing, this PR's unit tests might not pass until the cornell_customizations PR has been merged and the build jobs have executed appropriately.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>edu.cornell.kfs</groupId>
     <artifactId>kfs-war</artifactId>
@@ -19,12 +20,13 @@
         <buildNumber>6.0.1-SNAPSHOT</buildNumber>
         <cynergy.development>true</cynergy.development>
         <job.name>${project.artifactId}</job.name>
+
         <skipTests>true</skipTests>
         <test.excludes>**/*Test.java.bogus</test.excludes>
 
         <kfs.version>2019-12-19</kfs.version>
         <rice.version>2.7.0</rice.version>
-        <cynergy.version>2.7.0</cynergy.version>
+        <cynergy.version>2.7.0.j11</cynergy.version>
         <cynergy.version.suffix>current</cynergy.version.suffix>
 
         <aws-java-sdk-dynamodb.version>1.11.18</aws-java-sdk-dynamodb.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -68,53 +78,48 @@
...
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-compiler-plugin</artifactId>
+                    <version>${maven.compiler.version}</version>
+                    <configuration>
+                        <source>${java.version}</source>
+                        <target>${java.version}</target>
+                        <forceJavacCompilerUse>true</forceJavacCompilerUse>
```
db483: why is this needed?

pom.xml
```diff
@@ -68,53 +78,48 @@
...
-                    <instrumentation>
-                        <excludes>
-                            <exclude>org/kuali/**/*.class</exclude>
-                            <exclude>com/rsmart/**/*.class</exclude>
-                        </excludes>
-                    </instrumentation>
-                </configuration>
-            </plugin>
-            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
db483: why is this needed with java 8 version?

pom.xml
cah292: After consulting with Jay, he indicated that he had added this while attempting to get the code working with Java 11, but he believes that this setting is no longer needed. I will go ahead and remove this part of the setup, and verify that I can still build the WAR locally.

pom.xml
cah292: The compiler plugin's addition of "groupId" may have been added so that the related configuration in the "pluginManagement" section above could properly map to the plugin setup here. I would suggest keeping this change in place, so that the POM will properly refer to the "maven-compiler-plugin" using a group ID and artifact ID pair.

cah292:
I've committed a change to remove that unneeded pluginManagement configuration from the POM, and I confirmed that I can successfully build the WAR locally with the change in place. (I also had a local version of the related updated Cynergy JAR so that I could get things to compile.)

cah292:
I will also create a KDO to have DevOps update the Cynergy build jobs to start building the new Java 11 artifacts. Hopefully that will allow for the PR unit tests to start passing, provided that the unit tests are also set up to work with Java 11.

cah292:
Just so you're aware, I've rebased this branch to pull in the new 09/26 financials patch changes.

cah292:
I've rebased this branch to include the latest 2019-10-31 financials patch changes from the develop branch.

cah292:
As per KFSPTS-18060, I have removed our override of the glassfish JAXB dependency, and have updated the affected unit tests so that they will be ignored for now. We will fix and re-enable the two affected unit test classes in a future user story.

I see that Chad Bower has also made several changes to this feature branch. Those updates will need to be reviewed as well.

cah292:
@db483 @chadbower I've committed a change that should fix the unit test problems we had discussed earlier today. When I had initially updated the POM for Java 11 based on one of our other user stories, the POM's surefire configuration was updated to allow for injecting inclusions through a new property. However, the POM was not updated to include a default setting for this inclusion, so nothing was being included by default. I've removed the inclusion injection setup so that the plugin will now use the default inclusion settings. Sorry about that.

The unit tests should now be runnable again for this branch.

db483:
Fantastic! Thank you very much Chad for quickly finding and fixing the issue!

db483:
retest this please

cah292:
@chadbower Are there POM changes here that you need to revert before we merge these Java 11 changes to DEV? The removal of the "test.excludes" property should be safe to undo, since that was not what caused the recent unit test problems. I'm also wondering if we should remove the jacoco plugin for now, and whether we need that "jvnet-nexus-staging" repository entry that you added.

cah292:
As per further discussion with Chad B and Jay, I have removed Chad B's unneeded config changes. After that, I squashed all of the commits together, and then performed a rebase to bring in the latest develop changes and resolve merge conflicts. I believe the needed updates have been made now.

PR #942, opened 2020-03-25
KFSPTS-18277 fix scrubber null pointer exception

db483:


src/main/resources/edu/cornell/kfs/gl/cu-spring-gl.xml
```diff
@@ -189,9 +189,20 @@
         	</map>
         </property>
     </bean>
-    
-    <bean id="scrubberProcess" class="edu.cornell.kfs.gl.batch.service.impl.CuScrubberProcessImpl" parent="scrubberProcess-parentBean"/>
```
cah292: Instead of removing this bean and overriding the parentBean, what if this "scrubberProcess" bean override could be modified to something like the following? That would bring it closer to how it's currently configured in base code. Sorry for not noticing this in the original review of the 09/26 patch changes.

```
<bean id="scrubberProcess" class="edu.cornell.kfs.gl.batch.service.impl.CuScrubberProcessImpl"
    parent="scrubberProcess-parentBean" scope="prototype"
    p:scrubberBadBalanceListingReportWriterService-ref="scrubberBadBalanceListingReportWriterService"/>
```

db483: I am not sure I understand why it would only need the scrubberBadBalanceListingReportWriterService defined here? it is not something we have locally added

cah292: It actually needs two things:

[1] It needs the "scrubberBadBalanceListingReportWriterService" reference, since the base code sets it up on the "scrubberProcess" bean instead of the parentBean.

[2] It needs to have the "scrubberProcess" bean declared with the "prototype" scope.

db483: Got it, thank you!

db483:
I've updated the code. This is ready for review again. Thank you.

jhulslander:
retest this please

PR #943, opened 2020-03-27
KFSPTS-17852 upgrade to the 10/31/2019 version of financials

db483:


src/main/resources/edu/cornell/kfs/module/purap/document/validation/configuration/PurapValidatorDefinitions.xml
```diff
@@ -46,20 +31,6 @@
 			class="edu.cornell.kfs.sys.document.validation.impl.AccountingLineRequiredOverridesValidation" abstract="true">
 		<property name="accountingLineRuleHelperService" ref="purchaseOrderAmendmentAccountingLineRuleHelperService"/>
 	</bean>
-
-	<bean id="Requisition-newIndividualItemValidation" class="edu.cornell.kfs.module.purap.document.validation.impl.RequisitionNewIndividualItemValidation" abstract="true">
```
cah292, line 50: The custom cu-kfs version of RequisitionNewIndividualItemValidation does not extend from the one in base code, and it does not have the commodityCodeIsRequired() method override like in the base code's class. Can you please confirm if we should be using the base code's variant instead? If so, then can we remove the cu-kfs version of RequisitionNewIndividualItemValidation?

db483: Restoring entry for RequisitionNewIndividualItemValidation

src/main/java/org/kuali/kfs/module/purap/service/impl/ElectronicInvoiceHelperServiceImpl.java
```diff
@@ -0,0 +1,1886 @@
...
+    protected DateTimeService dateTimeService;
+    protected ParameterService parameterService;
+    protected DocumentService documentService;
+    protected AttachmentService attachmentService;
+    protected NoteService noteService;
+    protected BankService bankService;
+    protected KualiRuleService kualiRuleService;
+    private BusinessObjectService businessObjectService;
+    private DataDictionaryService dataDictionaryService;
+    //CU customization change from priate to protected
```
cah292, line 157: Minor typo in comment.

db483: updated

db483:
I've added the requested changes, please review again. Thank you.

PR #946, opened 2020-04-06
KFSPTS-17450: Fix to make lastBilledDate today for first predetermined or milestone invoice created via batch or manually.

nkimble:


src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuVerifyBillingFrequencyServiceImpl.java
```diff
@@ -140,6 +142,22 @@ public BillingPeriod getStartDateAndEndDateOfPreviousBillingPeriod(ContractsAndG
     public BillingPeriod getStartDateAndEndDateOfPreviousBillingPeriod(ContractsAndGrantsBillingAward award, Date calculatedLastBilledDate, AccountingPeriod currPeriod, String creationProcessTypeCode) {
         return BillingPeriod.determineBillingPeriodPriorTo(award.getAwardBeginningDate(), this.dateTimeService.getCurrentSqlDate(), calculatedLastBilledDate, ArConstants.BillingFrequencyValues.fromCode(award.getBillingFrequencyCode()), this.accountingPeriodService, creationProcessTypeCode);
     }
+    
+    /**
+     * CU Customization: The lastBilledDate should be set to the invoice creation date on the first invoice when
+     *                   Invoicing Option is Invoice By Schedule (Milestone or Predetermined) and Last Billed Date (Award) is null.
+     *                   Invoice is considered to be first invoice based on above listed criteria.
+     *                   This adjustment needs to be performed when invoices are created by BOTH Batch and Manual processing.
+     */
```
jhulslander, line 151: I am not sure if we need this comment, and if we do need it, do you think it would be more helpful where this function is used in CuContractsGrantsInvoiceCreateDocumentServiceImpl

src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuVerifyBillingFrequencyServiceImpl.java
```diff
@@ -140,6 +142,22 @@ public BillingPeriod getStartDateAndEndDateOfPreviousBillingPeriod(ContractsAndG
...
+    public boolean isFirstMilestoneOrPredeterminedInvoiceBySchedule(String billingFrequencyCode, String invoicingOptionCode, Date lastBilledDate) {
+        if (StringUtils.equalsIgnoreCase(ArIntegrationConstants.AwardInvoicingOptions.INV_SCHEDULE, invoicingOptionCode)
+                && (ArIntegrationConstants.BillingFrequencyValues.MILESTONE.equals(billingFrequencyCode) 
+                        || ArIntegrationConstants.BillingFrequencyValues.PREDETERMINED_BILLING.equals(billingFrequencyCode))
```
jhulslander, line 155: For consistency I would suggest using StringUtils for this String comparison, like it is above.  The indentation of this if statement makes it hard to read.  Maybe break it into two private methods isFirstMilestoneBill and isFirstPredeterminedInvoiceByScheduleBill.
Also I would suggest just returning the boolean expression, rather than "if (true) return true else return false"

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuVerifyBillingFrequencyServiceImpl.java
jhulslander: OK, lets leave it as is

PR #948, opened 2020-04-09
KFSPTS-18189 fixing rass batch job

jhulslander:


src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
```diff
@@ -545,4 +551,15 @@ public void setMaxStatusCheckAttempts(int maxStatusCheckAttempts) {
...
+    public void setDocumentProcessingQueue(DocumentProcessingQueue documentProcessingQueue) {
+        this.documentProcessingQueue = documentProcessingQueue;
+    }
+
+    public DocumentProcessingQueue getDocumentProcessingQueue() {
+        if (documentProcessingQueue == null) {
+            documentProcessingQueue = (DocumentProcessingQueue) SpringContext.getService(RassConstants.RICE_DOCUMENT_PROCESSING_QUEUE_SERVICE_NAME);
+        }
+        return documentProcessingQueue;
+    }
```
cah292, line 563: Can RassUpdateServiceImpl be configured to have its service reference be set only by injection, and not by falling back to SpringContext calls?

If you're having trouble referencing the DocumentProcessingQueue bean by name in "cu-spring-rass.xml", you could try adding a bean-importing entry near the top of our "cu-spring-kfs.xml" file.

src/test/java/edu/cornell/kfs/rass/batch/service/impl/RassServiceImplTest.java
```diff
@@ -96,6 +97,7 @@ public void setUp() throws Exception {
         
         rassUpdateService = springContext.getBean(RassTestConstants.RASS_UPDATE_SERVICE_BEAN_NAME, TestRassUpdateServiceImpl.class);
         rassUpdateService.setDocumentTracker(this::processMaintenanceDocument);
+        rassUpdateService.setDocumentProcessingQueue(Mockito.mock(DocumentProcessingQueue.class));
```
cah292, line 100: If you end up making the service-injecting change noted earlier, you may have to change this unit test setup as well. You will likely have to update "cu-spring-rass-service-test.xml" and the RassMockServiceFactory class so that you can inject a mock DocumentProcessingQueue instance.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
jhulslander: I was having spring injection issues, that is why I switched to spring contect .... I will try doing the bein import in cu-spring-kfs

jhulslander:
I've made changes based on the notes made, this is ready for review again,.

PR #957, opened 2020-04-20
KFSPTS-18534 Add config-loading workaround for Tomcat 8

cah292:
We've been having difficulties launching KFS DEV under Java 11 and Tomcat 8+ because of a missing property placeholder. After further analysis, we believe that KFS DEV is mistakenly loading the "META-INF/common-config-defaults.xml" configuration file from Rice, rather than the one from KFS. Some quick online searches suggest that Tomcat 8+ no longer has a consistent ordering of JAR dependencies from the WAR's expanded "WEB-INF/lib" directory, so that could be why we're experiencing this issue. We are not certain why this issue didn't affect our INTDEV testing, though.

This PR introduces a potential workaround for the problem. The changes here will add a duplicate of "common-config-defaults.xml" under a different directory and a slightly different name, and will also overlay PropertyLoadingFactoryBean to load this different config file instead. This should fix the missing-property-placeholder problem, but further DEV-deployment testing is needed to confirm that no other issues need fixing.

### DIFF
--- a/src/main/java/org/kuali/kfs/sys/context/PropertyLoadingFactoryBean.java
+++ b/src/main/java/org/kuali/kfs/sys/context/PropertyLoadingFactoryBean.java
@@ -1,0 +1,184 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2019 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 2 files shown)


db483:
retest this please

jhulslander:
retest this please

cah292:
I'm closing this PR since its workaround is insufficient, due to the Tomcat 8 issue affecting more than just the "common-config-defaults.xml" file. We will be implementing a different solution instead, either within the Tomcat configuration or within the POM's build setup. We will open some new PRs once a replacement fix is ready.

PR #959, opened 2020-04-23
KFSPTS-15693: Modification uses Concur SAE data file for both cash advance creation and trip reimbursement. Table that tracks duplicate cash advance requests requires change to secondary key structure due to data differences.

nkimble:


nkimble:
This code PR needs to be merged with SQL PR#765

PR #965, opened 2020-04-30
KFSPTS-18060 Fix namespace handling on Concur DTOs

cah292:
This PR updates the various Concur DTOs so that they can be processed correctly by the upgraded JAXB version. Some package-info.java files have been introduced to set up the appropriate default namespacing on a per-package basis, and some of the DTOs have been moved to different packages accordingly.

The ConcurEventNotificationDTO has to be available in both a namespaced and non-namespaced state, depending on the context. After some searching online for finding out how to set this up, I settled on a subclassing solution to address the issue. The base DTO has been updated to be an abstract class without explicit namespace setup, and then I created DTO subclasses for each context that the DTO is used in. The package-info.java (or lack thereof) in each DTO subclass's package controls what XML namespace to use.

For one of the DTOs that was updated, I only added some minimal testing of it in the corresponding unit test class, since the related unit tests didn't have extensive testing either. Please let me know if you think those unit tests should be made more extensive as part of this effort, or if that can wait for a future user story.


db483:
Chad, is the package-info.java needed to control the XML namespace or can that be done at the DTO level?

cah292:
The XML namespacing can be controlled at both the package level and the DTO level. If it's at the DTO level, though, then it might not be possible or practical for a subclass to override the namespaces on a superclass's fields. (It may be possible to do so by overriding the getters and placing the XmlElement annotations there, but that is a much more verbose setup compared to the package-level overriding.)

db483:
Thank you for explaining this.

PR #968, opened 2020-05-05
KFSPTS-18575 Fix unit tests to work with Java 11

cah292:
One of the failing tests was trying to create (and copy a file to) a directory that could not be auto-created properly on the new Jenkins instance. I was also unable to get the test to run locally in its old state. I fixed it by having the test derive the directory based on a property of the ReceiptProcessingService, so that the path would involve a more file-friendly location (such as the staging directory).

Two other tests were failing in the new Jenkins instance, but were succeeding locally. It turns out that the new Jenkins instance didn't have its Java locale settings set up as expected, which caused the tests to be unable to format currency values properly. Now that DevOps has fixed the Jenkins issue, these two tests can succeed as-is (apart from removing the "Ignore" annotations).


cah292:
retest this please

cah292:
retest this please

cah292:
retest this please

PR #969, opened 2020-05-05
KFSPTS-18207 prevent foreign vendors on recurring DVs

jhulslander:


src/main/resources/edu/cornell/kfs/fp/document/datadictionary/RecurringDisbursementVoucherDocument.xml
```diff
@@ -130,6 +131,31 @@
 		</property>
 	</bean>
 
+	<bean id="RecurringDisbursementVoucherDocument-validations" parent="RecurringDisbursementVoucherDocument-validations-parentBean" />
+	<bean id="RecurringDisbursementVoucherDocument-validations-parentBean" abstract="true" 
+		parent="DisbursementVoucherDocument-validations-parentBean" class="org.springframework.beans.factory.config.MapFactoryBean">
```
cah292, line 136: I see that this bean is a child of a base financials parent bean instead of a base financials non-parent bean. Was that necessary to avoid factory-bean-related issues? Or are there customizations for regular DVs that you needed to exclude from the recurring DVs?

Also, do you need to specify the class again if the parent bean already defines it?

src/main/resources/edu/cornell/kfs/fp/document/datadictionary/RecurringDisbursementVoucherDocument.xml
```diff
@@ -130,6 +131,31 @@
...
+		<property name="sourceMap">
+			<map key-type="java.lang.Class">
```
cah292, line 138: Is it possible to use the merge="true" option here, so that you don't have to copy all of the map entries from the parent bean?

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/CU-DisbursementVoucherValidation.xml
```diff
@@ -98,4 +98,18 @@
 	</property>
 	</bean>
 	
+	<bean id="RecurringDisbursementVoucher-saveDocumentValidation" parent="DisbursementVoucher-saveDocumentValidation-parentBean">
```
cah292, line 101: Should the foreign vendor check be happening at the save-document phase or the route-document phase? (Or both?)

Also, should the parent be the bean without the "-parentBean" suffix? Or is that needed for excluding any custom validation behavior on regular DVs that shouldn't be carried over to recurring DVs?

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/CU-DisbursementVoucherValidation.xml
```diff
@@ -98,4 +98,18 @@
...
+		<property name="validations">
+			<list merge="true">
+				<bean parent="RecurringDisbursementVoucherForeignVendorValidation" scope="prototype">
+                    <property name="parameterProperties">
+                        <list>
+                            <bean parent="accountingDocumentFieldConversion"/>
+                        </list>
+                    </property>
+                </bean>
+			</list>
```
cah292, line 111: There's a mix of tabs and spaces in this section. It would be helpful to keep the formatting consistent.

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/FinancialProcessingValidators.xml
```diff
@@ -35,6 +35,9 @@
           p:parameterEvaluatorService-ref="parameterEvaluatorService" p:parameterService-ref="parameterService"/>
 	<bean id="DisbursementVoucher-bankCodeValidation" class="edu.cornell.kfs.fp.document.validation.impl.CuDisbursementVoucherBankCodeValidation" abstract="true" />
 	<bean id="DisbursementVoucher-payeeStateAndCountryValidation" class="edu.cornell.kfs.fp.document.validation.impl.CuDisbursementVoucherPayeeStateAndCountryValidation" abstract="true">
+		<property name="recurringDisbursementVoucherForeignVendorValidation" ref="RecurringDisbursementVoucherForeignVendorValidation" />
```
cah292, line 38: "RecurringDisbursementVoucherForeignVendorValidation" is an abstract bean. If this reference is not getting materialized as expected, then you may have to change this to use the \<bean parent="..."/\> setup. (In cases like the latter, you might need to specify it as a prototype bean as well, for consistency with similar spots in KFS's validator beans.)

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherTaxServiceImplTest.java
```diff
@@ -85,6 +85,19 @@ public void testIsForeignVendorAndTaxReviewNotRequired_false_PaymentReasonRequir
...
+    @Test
+    public void testIsForeignVendor_true() {
+        boolean isForeign = disbursementVoucherTaxService.isForeignVendor(KFSConstants.PaymentPayeeTypes.VENDOR,
+                FOREIGN_VENDOR_HEADER_ID);
+        assertTrue("isForeign method should have returned true when input is foreign vendor", isForeign);
+    }
+
+    @Test
+    public void testIsForeignVendor_false() {
+        boolean isForeign = disbursementVoucherTaxService.isForeignVendor(KFSConstants.PaymentPayeeTypes.EMPLOYEE, null);
```
cah292, line 97: You should also add a test case method for when the payee is a domestic vendor.

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/FinancialProcessingValidators.xml
cah292: Based on the Spring bean fix that you recently put in for the requisition issue, a possible alternative way to fix this would be to redefine "RecurringDisbursementVoucherForeignVendorValidation" as a non-abstract bean with prototype scope.

src/main/resources/edu/cornell/kfs/fp/document/datadictionary/RecurringDisbursementVoucherDocument.xml
jhulslander: I should be able to switch to non parent, and I will remove the class definition

src/main/resources/edu/cornell/kfs/fp/document/datadictionary/RecurringDisbursementVoucherDocument.xml
jhulslander: yes, I should be able to use the merge option, good idea

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/CU-DisbursementVoucherValidation.xml
jhulslander: it was originally implemented on the route phase, the user story I am working on would like this to happen on the save .... I added code so the validation isn't attempted if the vendor hasn't been selected yet. .... I will also try to not use the non-parent bean

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/FinancialProcessingValidators.xml
jhulslander: Yes, I switched to non abstract and scoped it

jhulslander:

src/main/resources/edu/cornell/kfs/fp/document/validation/configuration/CU-DisbursementVoucherValidation.xml
cah292: Ok, thanks for clarifying. I'm fine with this happening on the save accordingly.

PR #975, opened 2020-05-12
KFSPTS-18611 Allow RASS to update Proposal Grant Number

cah292:
Our legacy Ezra integration allowed for updating just the Grant Number on existing Proposals. The functionals have discovered that we need this feature in place for the new RASS integration as well, so this PR adds such functionality to our RASS code. The PR also updates the report file generation to handle this change.

In addition, it was discovered that our Award documents were somehow causing certain sub-object lists on the Proposals to get cleared out unexpectedly. This PR introduces a partial workaround for that, so that when the Proposal needs to have its Grant Number updated, its related sub-object lists will also get repopulated from the XML file. A full fix for this issue will be implemented in a follow-up user story.


cah292:
retest this please

cah292:
retest this please

src/main/resources/edu/cornell/kfs/rass/cu-spring-rass.xml
```diff
@@ -185,23 +185,23 @@
...
-				<bean parent="rassProperty" p:xmlPropertyName="federalPassThroughAgencyNumber" p:boPropertyName="federalPassThroughAgencyNumber" p:required="false"/>
+						p:valueConverter-ref="rassPrimitiveBooleanConverter" p:skipForObjectEdit="true"/>
+				<bean parent="rassProperty" p:xmlPropertyName="federalPassThroughAgencyNumber" p:boPropertyName="federalPassThroughAgencyNumber" p:required="false" p:skipForObjectEdit="true"/>
 				<bean parent="rassListPropertyWithProposalNumberMapping" p:xmlPropertyName="organizationCode" p:boPropertyName="proposalOrganizations" p:required="false"
 						p:valueConverter-ref="rassProposalOrganizationConverter" p:subObjectDefinition-ref="proposalOrganizationDefinition"/>
-				<bean parent="rassProperty" p:xmlPropertyName="finalReportDueDate" p:boPropertyName="proposalDueDate" p:required="false"/>
+				<bean parent="rassProperty" p:xmlPropertyName="finalReportDueDate" p:boPropertyName="proposalDueDate" p:required="false" p:skipForObjectEdit="true"/>
 				<bean parent="rassListPropertyWithProposalNumberMapping" p:xmlPropertyName="principalAndCoPrincipalInvestigators" p:boPropertyName="proposalProjectDirectors"
```
jhulslander, line 205: this configuration looks like organizationCode and principalAndCoPrincipalInvestigators are editable fields, in addition to grantNumber.  I thought we just wanted grantNumber to be editable.  Would it make sense to have skipForObjectEdit defaulted to false, so only the fields the need to allow edits were need the new XML configuration

src/main/java/edu/cornell/kfs/rass/batch/ProposalTranslationDefinition.java
```diff
@@ -11,16 +11,29 @@
...
+public class ProposalTranslationDefinition extends RassObjectTranslationDefinition<RassXmlAwardEntry, Proposal>
+            implements InitializingBean {
 
 	protected BusinessObjectService businessObjectService;
 	protected DateTimeService dateTimeService;
 	protected ParameterService parameterService;
+	protected RassPropertyDefinition grantNumberMapping;
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
```
jhulslander, line 30: Why is this needed, isn't the mapping setup in cu-spring-rass.xml ?

src/main/java/edu/cornell/kfs/rass/batch/ProposalTranslationDefinition.java
```diff
@@ -11,16 +11,29 @@
...
+        grantNumberMapping = getPropertyMappings().stream()
+                .filter(propertyMapping -> StringUtils.equals(
+                        CuCGPropertyConstants.GRANT_NUMBER, propertyMapping.getBoPropertyName()))
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("No property mapping was found for Grant Number"));
```
jhulslander, line 35: I think this error only happens if our XML configuration isn't setup correctly.  If so, this error message might be more helpful if it states where the mapping should be setup.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
```diff
@@ -277,7 +277,13 @@ protected boolean shouldKeepWaitingForDocument(String documentId) {
         Class<R> businessObjectClass = objectDefinition.getBusinessObjectClass();
         R businessObject = businessObjectSupplier.get();
         List<String> missingRequiredFields = new ArrayList<>();
+        boolean isNewObject = StringUtils.equalsIgnoreCase(maintenanceAction, KRADConstants.MAINTENANCE_NEW_ACTION);
+        boolean isObjectEdit = StringUtils.equalsIgnoreCase(maintenanceAction, KRADConstants.MAINTENANCE_EDIT_ACTION);
+        
         for (RassPropertyDefinition propertyMapping : objectDefinition.getPropertyMappings()) {
+            if (isObjectEdit && propertyMapping.isSkipForObjectEdit()) {
+                continue;
+            }
```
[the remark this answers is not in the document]
cah292: I had set up the condition in this manner so that I wouldn't have to use an "else" block to indent the rest of the loop logic. I could change this to put the remaining logic in an indented block, but If I do that then it would probably make more sense to just make this an "if" block without an "else" (and invert the condition accordingly). Please let me know if you think such a change would make this more readable.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
```diff
@@ -296,7 +302,7 @@ protected boolean shouldKeepWaitingForDocument(String documentId) {
             throw new RuntimeException(objectDefinition.printObjectLabelAndKeys(xmlObject) + " is missing values for the following required fields: "
                     + missingRequiredFields.toString());
         }
-        if (StringUtils.equalsIgnoreCase(maintenanceAction, KRADConstants.MAINTENANCE_NEW_ACTION)) {
+        if (isNewObject) {
```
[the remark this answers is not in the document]
cah292: No, I believe the objects are supposed to have their active status remain the same if they are only being edited. In the past we put in the KFSPTS-16794 fix, to address an issue where the RASS Job was reactivating certain objects by mistake.

src/main/resources/edu/cornell/kfs/rass/cu-spring-rass.xml
cah292: "skipForObjectEdit" is already being defaulted to false, and it's only the Proposal setup that needs to set the value to true at this time. I'll see if I can create some separate parent beans that default "skipForObjectEdit" to true.

Also, as noted in the PR description and the JIRA comments, I've added a temporary fix to repopulate certain properties if the Award document somehow mistakenly cleared them out. (The org code and PI/director properties are the ones that need repopulating, and even then I've set up ProposalTranslationDefinition so that the updates won't happen unless the Grant Number has changed.) Without this fix, most Grant Number updates to Proposals will run into business rule failures. (Our legacy Ezra integration updated the Grant Number via the BO service to get around this.)

src/main/java/edu/cornell/kfs/rass/batch/ProposalTranslationDefinition.java
cah292: The Grant Number mapping is included in the "propertyMappings" list, but I didn't want to loop through the mappings list every time that just this individual mapping was needed. (This one mapping is needed for the modified can-edit check.) I added this afterPropertiesSet() method to set up a new "grantNumberMapping" field automatically once the bean is all configured, without requiring the bean configuration to add the Grant Number mapping twice.

src/main/java/edu/cornell/kfs/rass/batch/ProposalTranslationDefinition.java
cah292: Ok, I'll update this with a better error message.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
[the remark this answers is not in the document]
cah292: After thinking through this further, I'll work on restructuring the code so that the edit-mode-based filtering will occur elsewhere, and so that it doesn't need to rely on the "continue" command.

cah292:
I've committed some changes based on the initial feedback. This is ready for review again.

src/main/java/edu/cornell/kfs/rass/batch/service/impl/RassUpdateServiceImpl.java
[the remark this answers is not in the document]
jhulslander: Thanks for clarifying

PR #980, opened 2020-05-20
KFSPTS-18563 implement PMW foreign vendors step 1, add form mode service

jhulslander:


jhulslander:
I was originally thinking about it just being a boolean, but I think the "mode" function naming convention makes the code clearer to understand.  Also to your point about more modes,  it may be necessary when we get to ACH changes, since that is an unknown set of changes.

src/main/java/edu/cornell/kfs/pmw/batch/businessobject/PaymentWorksVendor.java
```diff
@@ -1105,7 +1110,36 @@ public void setServicesProvidedWithoutInsurance(boolean servicesProvidedWithoutI
...
+    }
+
+    public void setPaymentWorksFormModeService(PaymentWorksFormModeService paymentWorksFormModeService) {
+        this.paymentWorksFormModeService = paymentWorksFormModeService;
+    }
+    
+    @Override
     public String toString() {
+        if (getPaymentWorksFormModeService().shouldUseForeignFormProcessingMode()) {
+            ReflectionToStringBuilder builder = new ReflectionToStringBuilder(this, ToStringStyle.MULTI_LINE_STYLE);
```
db483, line 1127: I like the ReflectionToStringBuilder use, this saves a lot of coding. One thing to think of is whether this is used in the reports and if it will impact what users are used to see in the reports.

jhulslander: It does get printed on the report, so it will be slightly different, but with the style I selected, it should be very similar, but when we get to functional validation, I will let them know and we can get feedback.

db483: If it prints PaymentWorksVendor@xxxx they may not like the object address part. I am not sure what would be an easy way to remove that and keep the multi-line style.

jhulslander: There is a style that supports no class name, which also prevents the Object ID ..... worse case we can create a custom style  that does both and the multi line

PR #984, opened 2020-05-21
KFSPTS-18609 Add PO shipping address to eInvoice CXML

cah292:
There are PRs for this in cu-kfs and einvoice-api-server.

Also, I realized that this PR's new getPurchaseOrderShippingAddress() method shares a lot of the same code as getPurchaseOrder(). It would be good to consolidate some of the common logic, but that might make more sense to do in a subsequent user story. Please let me know your thoughts on this.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
@@ -254,13 +254,15 @@ public class CuEinvoiceApiResource {
     }
 
     private String serializePoShippingAddressToJson(PurchaseOrderDocument poDoc) {
-        // TODO: If the PO is configured to use the receiving address, should that be returned instead?
         Properties addressProps = new Properties();
         safelyAddProperty(addressProps, CUPurapConstants.Einvoice.DOCUMENT_NUMBER, poDoc.getDocumentNumber());
         safelyAddProperty(addressProps, CUPurapConstants.Einvoice.SHIPPING_NAME, poDoc.getDeliveryToName());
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/module/purap/rest/resource/CuEinvoiceApiResource.java
```diff
@@ -233,6 +253,29 @@ private Properties convertUnitOfMeasureToProperties(UnitOfMeasure unitOfMeasure)
...
+        safelyAddProperty(addressProps, CUPurapConstants.Einvoice.ZIPCODE, poDoc.getDeliveryPostalCode());
+        safelyAddProperty(addressProps, CUPurapConstants.Einvoice.COUNTRY_CODE, poDoc.getDeliveryCountryCode());
+        safelyAddProperty(addressProps, CUPurapConstants.Einvoice.COUNTRY_NAME, poDoc.getDeliveryCountryName());
+        safelyAddProperty(addressProps, CUPurapConstants.Einvoice.EMAIL, poDoc.getDeliveryToEmailAddress());
+        return gson.toJson(addressProps);
+    }
+
+    private String buildRoomNumberLine(PurchaseOrderDocument poDoc) {
+        return CUPurapConstants.Einvoice.ROOM_NUMBER_PREFIX + poDoc.getDeliveryBuildingRoomNumber();
+    }
```
ajd299, line 277: This function adds 3-4 lines of code, but makes the code that calls it a little bit cleaner. I am fine with it either way.

cah292: I would prefer to keep it this way for now, since it simplifies the calling code a bit.

ajd299:
Changes look good, merging

PR #987, opened 2020-06-10
KFSPTS-18769: Seven files from 2020-02-13 KualiCo patch moved into cu-kfs base code path with FINP-4769 then being applied to them. That required changes to VerifyBillingFrequencyServiceImpl.java method call BillingPeriod.determineBillingPeriodPriorTo in order to successfully compile. All CGB CU mods previously in place that required basecode methods to be copied into the Cornell package with overrides were updated to reflect the FINP-4769 change and all CGB CU mods were marked with their corresponding JIRA numbers.

nkimble:


src/main/java/edu/cornell/kfs/module/ar/batch/service/CuVerifyBillingFrequencyService.java
```diff
@@ -8,8 +8,10 @@
 import org.kuali.kfs.module.ar.batch.service.VerifyBillingFrequencyService;
 import org.kuali.kfs.module.ar.businessobject.BillingPeriod;
 
+//CUMod: KFSPTS-15655 required this interface exist with no content.
```
jhulslander, line 11: since the interface has content, should this comment be updated, if we want/need to keep it.

src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuVerifyBillingFrequencyServiceImpl.java
```diff
@@ -69,31 +72,45 @@ private Date calculateLastBilledDateBasedOnAwardInvoiceOption(ContractsAndGrants
...
         LOG.info("validateBillingFrequency(private): For Award/Proposal# = " + award.getProposalNumber()
-                + " with calculatedLastBilledDate = " + lastBilledDate
+                + " with calculatedLastBilledDate = " + lastBilledDate + ", checkBillingPeriodEnd = " + checkBillingPeriodEnd
                 + " creationProcessTypeCode = " + ArConstants.ContractsAndGrantsInvoiceDocumentCreationProcessType.getName(award.getCgInvoiceDocumentCreationProcessTypeCode()));
         final Date today = getDateTimeService().getCurrentSqlDate();
         AccountingPeriod currPeriod = getAccountingPeriodService().getByDate(today);
 
-        BillingPeriod billingPeriod = getStartDateAndEndDateOfPreviousBillingPeriod(award, lastBilledDate, currPeriod, award.getCgInvoiceDocumentCreationProcessTypeCode());
+        //basecode: BillingPeriod billingPeriod = getStartDateAndEndDateOfPreviousBillingPeriod(award, currPeriod);
+        BillingPeriod billingPeriod = getStartDateAndEndDateOfPreviousBillingPeriod(award, lastBilledDate, currPeriod, award.getCgInvoiceDocumentCreationProcessTypeCode()); //CUmod
```
jhulslander, line 113: if the cumod comment is needed, I would move that to it's own line above

src/main/java/org/kuali/kfs/module/ar/batch/service/impl/VerifyBillingFrequencyServiceImpl.java
```diff
@@ -0,0 +1,167 @@
+//KualiCo Patch Release 2020-02-13
+//FINP-4769 FK-117 changes applied.
```
jhulslander, line 2: I like these comment, but if you could add what kualiCo patch will have these changes, so we know when the overlay file can be removed

This thought applies to ther

src/main/java/edu/cornell/kfs/module/ar/batch/service/CuVerifyBillingFrequencyService.java
nkimble: Should the mod that gave this interface content was ever removed or needed to be provided to a partner school, then the interface would still need to exist with no content for that original change. That was the intent of the past tense in this comment.

src/main/java/edu/cornell/kfs/module/ar/batch/service/impl/CuVerifyBillingFrequencyServiceImpl.java
nkimble: Moved.

src/main/java/org/kuali/kfs/module/ar/batch/service/impl/VerifyBillingFrequencyServiceImpl.java
nkimble: Added patch FINP is in to seven files in question.

src/main/java/edu/cornell/kfs/module/ar/CuArConstants.java
```diff
@@ -10,6 +10,9 @@
     public static final int CINV_DATE_RANGE_END_DATE_START_INDEX = 14;
     public static final String QUESTION_NEWLINE_STRING = "[br]";
 
+    /*
+     * CUMod: KFSPTS-14970
```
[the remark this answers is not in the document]
jhulslander: thanks for explaining, that makes sense

src/main/java/edu/cornell/kfs/module/ar/batch/service/CuVerifyBillingFrequencyService.java
jhulslander: OK, I understand now ..... I am not sure how to make that clearer, so lets keep it as is

PR #1000, opened 2020-07-31
KFSPTS-16227 save PVEN doc when there are business rules errors

db483:


db483:
Chad, this is ready for review again. A code change was needed so that vendorFaxNumber would be set to null when there is a formatting error. In that case the PVEN would be saved with a business rule error when the method of PO transmission is FAX because fax data is required.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
```diff
@@ -281,14 +282,25 @@ private VendorAddress assignMethodOfPoTransmission(VendorAddress poAddress, Paym
...
     }
+    
+    private void setVendorFaxNumberValue(VendorAddress vendorAddress, String vendorFaxNumber) {
+        try {
+            Class type = ObjectUtils.easyGetPropertyType(vendorAddress, VendorPropertyConstants.VENDOR_FAX_NUMBER);
+            ObjectUtils.setObjectProperty(vendorAddress, VendorPropertyConstants.VENDOR_FAX_NUMBER, type, vendorFaxNumber);
+        } catch (FormatException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+            vendorAddress.setVendorFaxNumber(null);
+            LOG.error("setVendorFaxNumberValue: Vendor Fax Number cannot be set due to exception: " + e);
+            e.printStackTrace();
```
cah292, line 301: It would be better to use Log4j to print the stack trace, rather than using the exception's printStackTrace() method.

db483: Thank you, I am updating this.

db483:

PR #1004, opened 2020-08-06
KFSPTS-19289 fix Organization Global document after 3/20/20 upgrade

db483:


db483:
Note: I also had to remove the existence check on organizationManagerUniversal as this was causing and exception. The same fix was added to the Organization document by KualiCo.

PR #1006, opened 2020-08-11
KFSPTS-19317 Fix group member button display on Person doc

cah292:
The Person document has a bug where the group membership "Inactivate" buttons are visible on uneditable documents. This PR fixes that by adding new group-action-checking logic. The new logic is similar to what the Person doc already uses for showing the role membership actions.

Implementing the changes via overlays seemed like the simplest and least invasive way to do so, given the types of code and config areas that needed adjustment.

### DIFF
--- a/src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
+++ b/src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
@@ -1,0 +1,620 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2020 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
```diff
@@ -0,0 +1,620 @@
...
+        if (StringUtils.isNotEmpty(group.getNamespaceCode())) {
+            Map<String, String> additionalPermissionDetails = new HashMap<>();
+            additionalPermissionDetails.put(KimConstants.AttributeConstants.NAMESPACE_CODE, group.getNamespaceCode());
+            additionalPermissionDetails.put(KimConstants.AttributeConstants.GROUP_NAME, group.getGroupName());
+            if (!getDocumentHelperService().getDocumentAuthorizer(this).isAuthorizedByTemplate(this,
+                    KimConstants.NAMESPACE_CODE, KimConstants.PermissionTemplateNames.POPULATE_GROUP,
+                    GlobalVariables.getUserSession().getPrincipalId(), additionalPermissionDetails, null)) {
+                rulePassed = false;
+            }
+        }
```
jhulslander, line 603: if the person group doesn't have a namespace code, then the group is editable on the person document?  Is that desirable?

cah292: I was trying to set up this assign-group check to be consistent with the assign-role check in the method above it. I'm not certain why an empty-namespace check is being done here, but there should be other business rules on the Person document that prevent an empty-namespace group entry from being added (and there should not be any pre-existing empty-namespace group entries). If needed, I can update this block to return false if an empty namespace is encountered.

cah292:
As requested, I have added some comments to help point out the customized areas of the overlayed files.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
jhulslander: sounds good to me, thanks for explaining

PR #1007, opened 2020-08-11
KFSPTS-19339 Auto-insert empty entries in lookup drop-downs

cah292:
This PR updates the refactored lookups so that, like with the old lookups, they will auto-insert an empty drop-down entry if the values finder does not return one already. Because the change involved updating a private method (buildLookupControlValuesMap), the fix has been implemented as an overlay.


cah292:
As requested, I've added appropriate comments to make it clearer which areas were customized.

PR #1010, opened 2020-08-13
KFSPTS-18900 implement PMW foreign vendors step 3

jhulslander:
This is very similar to the previous PR for this, but I am using the correct Jira number and updates to findPoCountryToUse() function to fix a bug with domestic PO country handling


src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksVendorToKfsVendorDetailConversionServiceImpl.java
```diff
@@ -882,6 +901,27 @@ private KfsVendorDataWrapper createInsuranceCertificateNote(PaymentWorksVendor p
...
+        String poCountryToUse;
+        if (paymentWorksFormModeService.shouldUseForeignFormProcessingMode()) {
+            if (StringUtils.equalsIgnoreCase(poCountryUsCanadaAustraliaOther, PaymentWorksConstants.PO_ADDRESS_COUNTRY_OTHER) ||
+                    StringUtils.isBlank(poCountryUsCanadaAustraliaOther)) {
+                poCountryToUse = poCountry;
+            } else {
+                poCountryToUse = poCountryUsCanadaAustraliaOther;
+            }
+        } else {
+            if (StringUtils.isNotBlank(poCountry)) {
```
db483, line 917: Is this fixing a bug with domestic?

jhulslander: Yes, if poCountry has a value, use that, otherwise use poCountryUsCanadaAustraliaOther.  With the mapping fixed, I expect poCountry to be empty and have the value in  poCountryUsCanadaAustraliaOther.  However, there might be some iterations of data and form details were poCountry has a value, and that value should take precedence

db483: Thank you for explaining this.

PR #1011, opened 2020-08-13
KFSPTS-19308 add support for inquiry by principalName

db483:


src/main/java/org/kuali/kfs/kim/inquiry/PersonInquirableImpl.java
```diff
@@ -0,0 +1,192 @@
...
+        return person;
+
+    }
+
+    @Override
+    public HtmlData getInquiryUrl(BusinessObject businessObject, String attributeName, boolean forceInquiry) {
+        List<String> primaryKeys = new ArrayList<>();
+        primaryKeys.add("principalId");
+        //KFSPTS-19308 CU customization to allow inquiry by prinicpalName
+        primaryKeys.add("principalName");
```
cah292, line 86: Were you specifically encountering inquiry problems if the principalName was not being included in this manner? I'm reluctant about treating non-PK fields as primary keys, given that principalId is the actual primary key (despite principalName also being unique to each user).

Also, there's a small typo in the explanatory comment.

src/main/resources/edu/cornell/kfs/kim/impl/identity/PersonImpl.xml
```diff
@@ -4,6 +4,14 @@
        xsi:schemaLocation="http://www.springframework.org/schema/beans
             http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
 
+    <bean id="PersonImpl" parent="PersonImpl-parentBean">
+        <property name="primaryKeys">
+            <list merge="true">
+                <value>principalName</value>
```
cah292, line 10: If you don't need the getInquiryUrl() method change in the customized inquirable class, then you may not need to make this primary key change to the bean either.

cah292:
After some more investigation, I found one potential alternative that could allow for properly handling the principalName without modifying the primary key list. If you can override the KimModuleService, then you could override the listAlternatePrimaryKeyFieldNames() method so that it returns the principalName field as an alternate key. The InquiryForm appears to allow for using such alternate keys to search for a particular object.

I would suggest using this alternate-key setup, so that you don't have to modify the list of actual PersonImpl primary keys.

db483:
Great find on the listAlternatePrimaryKeyFieldNames. I have updated the fix. This is ready for review again.

PR #1014, opened 2020-08-24
KFSPTS-19381 Fix Person doc phone number validation

cah292:
We have an existing customization in standalone Cynergy that allows for a larger number of valid phone number formats. This PR adds that customization to the KFS side, but restricts it to the Person document and its related pages (which might be the only spot that needs the extra format options).


cah292:
I've added some unit tests to validate the new phone number pattern handling. This is ready for review again.

PR #1017, opened 2020-09-07
KFSPTS-19595 fix KFS Person inquiry issue related to field masking

db483:


db483:
Chad, I have used the code you have put together on https://github.com/CU-CommunityApps/cu-kfs/tree/cah292-post-kim-pers-fix for this fix and added some changes to make it work. The original version was not working as expected. I have tested this locally and the Person Inquiry works as expected both in an unmasked and masked env.

cah292 (approved):
Changes look good, and unit tests pass. The one variable probably could have been named to something other than "workflowDocument" to make it more accurate about what it represents, but I'm okay with deferring such cleanup to a future story, given the need for moving this fix forward promptly.

Merging.

PR #1019, opened 2020-09-11
KFSPTS-19556: Prevent creator on submit and approver on approve from routing I Want edoc to themselves by stopping their ability of entering their own netid in the Approver text field.

nkimble:

PR #1020, opened 2020-09-15
Revert "KFSPTS-19556: Prevent creator on submit and approver on appro…

nkimble:
…ve from routing I Want edoc to themselves by stopping their ability of entering their own netid in the Approver text field. (#1019)"

This reverts commit 4607da114b4fc743498732921775b33de60255cc.


db483 (approved):
Revert looks good.

PR #1021, opened 2020-09-15
KFSPTS-19556: Prevent creator on submit and approver on approve from routing I Want edoc to themselves by restricting them from entering their own netid in the Approver text field.

nkimble:


cah292 (approved):

PR #1022, opened 2020-09-15
KFSPTS-17857 Upgrade to 04/30/2020 version of financials

cah292:
There are PRs for this in cu-kfs, nonprod-sql and cu-kfs-upgrades. Please make sure all of them are good to go before merging.

Please make sure the unit tests pass before merging. Some unit test code changes were needed due to KualiCo's refactoring efforts, and I am not certain whether the tests will pass if the associated upgrade SQL has not been applied to UTEST yet.

Also note the following details about this upgrade:

* Several unneeded overlays and overrides have been removed. If you notice something that should or should not have been removed, please let me know.
* Due to KualiCo's Maven-related refactoring, I had to add "updateDatabaseOnStartup=false" to our institution config so that KFS won't try to run Liquibase updates on startup. (I believe this property was already false by default in prior releases, but now defaults to true in the 04/30 patch.) We should notify KualiCo of this as part of the KFSPTS-19537 sub-task work.
* KualiCo's removal of the Asset Location Type BO also removed one of the Data Dictionary attributes that our Asset lookup customizations rely on. I've updated the customization accordingly to avoid referencing the non-existent bean, and to also display the field as a drop-down with relevant labeling (now that those location types aren't lookupable anymore).
* Our Vendor inquiry customizations were relying on the Inquiry Presentation Controller feature, but KualiCo has removed that feature with this release. I have rewritten the customization accordingly so that it now overrides certain BO Authorization Service methods to handle the custom field-visibility handling.


cah292:
Based on the failed tests from the first run, it appears that they failed because they could not retrieve the OBJ_ID column value from the View for the ObjectCodeCurrent BO. This release recreated that View to add the OBJ_ID column, but this patch's database changes have not been applied to UTEST yet, which would explain why the tests failed.

When we're closer to being ready to merge the patch to DEV, we will need to apply the upgrade SQL to UTEST to see if that fixes the tests.

db483 (approved):
Changes look good.

db483:
Chad, 

Code changes look good. Please go ahead an apply the SQL in UTEST so that the unit tests can run.

Does the updateDatabaseOnStartup=false property need to be added to the utest config file?

cah292:
retest this please

cah292:
Daniela,

I've applied the SQL to UTEST and triggered another run of the unit tests.

As for the "updateDatabaseOnStartup=false" setting, this PR adds it to the institutional config properties file within the cu-kfs project, so I don't think we need to add it to the unit test config property file.

db483:
Thank you for clarifying this.

PR #1026, opened 2020-09-23
KFSPTS-19758 Add Last Scanned netid & Date Fields

ajd299:


db483:
retest this please

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetServiceImpl.java
```diff
@@ -53,7 +59,17 @@ public Asset updateAssetInventory(String capitalAssetNumber, String conditionCod
         Timestamp currentTimestamp = dateTimeService.getCurrentTimestamp();
         asset.setLastInventoryDate(currentTimestamp);
         asset.setLastUpdatedTimestamp(currentTimestamp);
-        return businessObjectService.save(asset);
+        asset = businessObjectService.save(asset);
+
+        AssetExtension assetExtension = businessObjectService.findBySinglePrimaryKey(AssetExtension.class, capitalAssetNumber);
```
jhulslander, line 64: you should be able to do 
AssetExtension assetExtension - (AssetExtension) asset.getExtension()

you'll want to check for null, and create a new instance of AssetExtension as needed.  If you need to create a new AssetExtension object, you'll want to call asset.setExtension().

doing a business object save of asset should save both the asset and asset extension.

jhulslander (changes requested):
Changes look good, I had one suggestion for simplifying getting and saving the extension object.

Also, should the Asset maintenance document inside KFS updated to include the new extension fields?  It might make sense to defer that till later, but I don't think it should be forgotten.

src/main/java/edu/cornell/kfs/module/cam/document/service/impl/CuAssetServiceImpl.java
```diff
@@ -53,7 +62,18 @@ public Asset updateAssetInventory(String capitalAssetNumber, String conditionCod
...
-        return businessObjectService.save(asset);
+
+        AssetExtension assetExtension = (AssetExtension) asset.getExtension();
+        if (ObjectUtils.isNull(assetExtension)) {
+            assetExtension = new AssetExtension();
+            asset.setExtension(assetExtension);
+        }
+        Person person = personService.getPersonByPrincipalName(netid);
+        String lastScannedBy = person.getFirstName() + " " + person.getLastName() + " (" + netid + ")";
+        assetExtension.setLastScannedNetid(lastScannedBy);
```
jhulslander, line 73: The database field and object name imply the data is a netid,  the value being populated is the name following by netid in parenthesis.  I suggest renaming the new values, or just storing the netid.

jhulslander: also, the database field only supports 32, we would need more space if we want to store the name and net id

jhulslander (changes requested):

jhulslander:
Also, what about updating the maintenance document?  I think it is fine to defer that change till a later date, we just don't want to lose it.

ajd299:
Should we add the fields to the maintenance document if they are not editable?

ajd299:
I have made the following improvements:

- Added fields to Maintenance Document with unconditonallyReadonly=true
- Changed field name lastUpdatedNetid to lastUpdatedBy
- Increased field length to 256
- Changed Label to Last Scanned On to Last Scanned Date

jhulslander (approved):

PR #1030, opened 2020-09-28
KFSPTS-19760 Cap Asset Generate Error Document

ajd299:


src/main/java/edu/cornell/kfs/module/cam/document/validation/impl/CuBarcodeInventoryErrorDocumentRule.java
```diff
@@ -0,0 +1,13 @@
+package edu.cornell.kfs.module.cam.document.validation.impl;
+
+import org.kuali.kfs.module.cam.document.BarcodeInventoryErrorDocument;
+import org.kuali.kfs.module.cam.document.validation.impl.BarcodeInventoryErrorDocumentRule;
+
+public class CuBarcodeInventoryErrorDocumentRule extends BarcodeInventoryErrorDocumentRule {
+
+    @Override
+    public boolean validateBarcodeInventoryErrorDetail(BarcodeInventoryErrorDocument document, boolean updateStatus) {
```
db483, line 9: Was it a functional requirement that this validation should be removed?

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -131,6 +146,45 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.CAPITAL_ASSET
         }
     }
 
+    private void createCapitalAssetErrorDocument(String netid, String assetTag, String condition, String buildingCode, String roomNumber) {
+        try {
+            GlobalVariables.clear();
+
+            GlobalVariables.setUserSession(new UserSession("kp378"));
```
db483, line 153: Why is this using a hard coded net id for creating a user session? Are these documents supposed to be initiated by this user? Could you just set the initiator net id? Can this user created these documents manually as well? If so, can they differentiate between the ones automatically created by this integration and the ones they manually created?

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -131,6 +146,45 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.CAPITAL_ASSET
...
+            BarcodeInventoryErrorDocument document = (BarcodeInventoryErrorDocument) getDocumentService().getNewDocument(BarcodeInventoryErrorDocument.class);
+
+            document.getFinancialSystemDocumentHeader().setFinancialDocumentTotalAmount(KualiDecimal.ZERO);
+            String errorDescription = "Error Scanning Asset Tag #" + assetTag;
```
db483, line 157: This error description should be added to a properties or a constants file.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -131,6 +146,45 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.CAPITAL_ASSET
...
+            List<AdHocRouteRecipient> adHocRecipients = new ArrayList<>();
+            getDocumentService().routeDocument(document, "Capital Asset Inventory Error Asset Not Found", adHocRecipients);
+        } catch (Exception ex) {
+            LOG.error("createCapitalAssetErrorDocument", ex);
+        }
+    }
+
+    private BarcodeInventoryErrorDetail getErrorDetail(String netid, String assetTag, String condition, String buildingCode, String roomNumber) {
+        BarcodeInventoryErrorDetail barcodeInventoryErrorDetail = new BarcodeInventoryErrorDetail();
+        barcodeInventoryErrorDetail.setUploadRowNumber(1L);
```
db483, line 175: Please use existing constants or define new ones  for these hard coded values.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -131,6 +146,45 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.CAPITAL_ASSET
         }
     }
 
+    private void createCapitalAssetErrorDocument(String netid, String assetTag, String condition, String buildingCode, String roomNumber) {
+        try {
+            GlobalVariables.clear();
```
db483, line 151: Why is this GlobalVariable.clear() needed here?

db483 (changes requested):
Please see my feedback. Thank you.

src/main/java/edu/cornell/kfs/module/cam/document/validation/impl/CuBarcodeInventoryErrorDocumentRule.java
ajd299: No, however, to reach this point the Asset will not be found, if this wasn't overridden an error would occur and the document would route to error. See the base code version of the method in BarcodeInventoryErrorDocumentRule.java for more details.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
ajd299: Setting the user session to kp378 allows the document initiator to be kp378 and it routes the document to kp378. I'm not sure how to set the document initiator to kp378 or a group?
I have found some base code that may be a clue as to how to do it in DocumentBase.java: wrapDocumentWithMetadataForXmlSerialization

ajd299:
Please see my comments and commits, I could use some guidance on how to best set the document initiator and route it to the Capital Asset Group.

Is routing to group id 34242 ga.distribution.uco-capasset feasible/correct. The requirement is to route it to the Capital Asset Group?
https://cynergy-dev.kuali.cornell.edu/cynergy/kim/identityManagementGroupInquiry.do?methodToCall=inquiry&id=34242#%3Cesapi:encodeForJavaScript%3EtopOfForm%3C/esapi:encodeForJavaScript%3E

src/main/java/edu/cornell/kfs/module/cam/document/validation/impl/CuBarcodeInventoryErrorDocumentRule.java
db483: If we override this then the validation will not happen when the document is created manually either. Will the validation happen at a different point? I a not sure we just want to remove the validation. I will try to run this locally to better understand the issue you are running into. It may be good to have a zoom when you get in. Thank you!

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
db483: Using the UserSession will work, sorry for adding confusion on this one. I see that you have changed the initiator to be the kfs user. Will the routing work as expected in that case?  If it doesn't work then either the kfs user will need some role/permission added or if it is ok for kp378 to initiate these automatically created documents then perhaps we could use a system parameter to store this value. If one day this user no longer works for the university or she changes her job and another person should be responsible for these documents then a code change will be needed if we leave it hard coded. If we use a system parameter then the functional users can update the value in the application and this will keep working without needing a code change.

db483:
retest this please

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
ajd299: There is a related SQL PR to add permission to kfs system user to create/initiate BCIE documents. I am currently working on routing to the capasset distribution group

db483: Thank you for pointing the SQL PR.
For the routing, will you need to change the workflow xml for the BCIE doc type?

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -131,6 +144,42 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.CAPITAL_ASSET
         }
     }
 
+    private void createCapitalAssetErrorDocument(String netid, String assetTag, String condition, String buildingCode, String roomNumber) {
+        try {
+            GlobalVariables.setUserSession(new UserSession(CuCamsConstants.CapAssetApi.KFS_SYSTEM_USER));
```
cah292, line 149: Instead of explicitly changing the current UserSession in this manner, I would suggest using the GlobalVariables.doInNewGlobalVariables() method instead. This helper method will handle the GlobalVariables management automatically; you just have to pass in the new UserSession and a lambda expression (or method reference) for the code you want to run.

db483:
retest this please

ajd299:
Please see my latest commit to remove unused code as we previously discussed

db483 (approved):
Changes look good.

PR #1031, opened 2020-09-29
KFSPTS-19579 added cognito_access_token to the request in hopes of solving session time out issue

jhulslander:
As requested, I added cognito_access_token to the request.  I did run into a problem with the netid field not being populated in the servlet request.  I am not sure if that is related to this change, or a bug with a previous change.


ajd299 (approved):
Changes look good, merging

PR #1033, opened 2020-10-01
KFSPTS-19555 backporting KualiCo fix on FINP-7046

db483:


cah292 (commented):
It would be helpful to have some comments indicating what is being customized in these overlays. Other than that, the changes look good.

db483:
Chad, thank you for your feedback. Nothing is being customized in  these files, they are just base code from a future release being brought in. I will add a comment in the header that this is just backporting FINP-7046 and can be removed when we upgrade to the 9/10/20 version of financials.

cah292 (approved):

PR #1034, opened 2020-10-01
KFSPTS-19069: Remove OJB property changes put in place by KITI-2541 while retaining KITI-2700 mod to remove sales tax references for both source and target accounting lines. At the time of this commit, both the source and target accounting line OJB class descriptors matched base code with just the salesTax reference object being removed.

nkimble:


jhulslander (approved):

PR #1035, opened 2020-10-05
KFSPTS-19929 Fix misplaced class overlay

cah292:
One of our recent overlays of a base financials class was accidentally placed in the wrong package. This PR just moves the file to the correct path.

I've created the KFSPTS-19935 follow-up ticket to investigate why we didn't get alerted to this problem sooner, given that it was causing compilation errors.


nkimble (approved):
File has been moved to the correct location ".../document/se**r**vice/impl".

PR #1037, opened 2020-10-07
KFSPTS-19941 adding xml-api exclusions to fix eclipse build problems

jhulslander:


cah292 (commented):
I'm trying to test out the POM change against a local branch, where I copied in the changes and ran a "Maven -- Update Project" command. Eclipse is no longer reporting any errors in the Project Explorer pane or in the Problems view as a result. However, when I open up certain Java files in cu-kfs, I sometimes see errors about unresolvable types from java.lang or certain List types not being visible. Here are some example classes where I am seeing such problems:

* CuCapAssetInventoryApiResource
* CuElectronicInvoiceHelperServiceImpl
* CuMaintainableXMLConversionServiceImplTest

Can you please confirm whether you needed to make any additional changes to the Eclipse metadata files, apart from the auto-updates made by the "Maven -- Update Project" operation?

Also, note that "Maven -- Update Project" seems to be updating the .project file as well. Should changes to that file be included in this PR, given that we're still tracking it in Git?

cah292 (commented):
I was finally able to fix the problem; it was an odd issue with my local Eclipse IDE. I closed and reopened both the cu-kfs project and the Eclipse IDE, and cleaned cu-kfs as well. This finally resolved the odd errors that I was seeing. (Note that if you try to run a unit test or do a similar operation right afterwards, you may have to wait a minute or two before the test will actually start, because Eclipse may need to reindex a bunch of files.)

As for the .project file changes from the Maven-project-update operation, those might be needed to fix an unknown cu-kfs POM error that I sometimes see in my local Eclipse. I would suggest including those .project changes in this PR as well, or adjusting the Git tracking to not pay attention to the .project file anymore. If you'd rather deal with the .project file in a follow-up user story, please let me know.

jhulslander:
Yes, my Eclipse tends to take a few minutes to update after maven/pom changes.  

Often I have had problems that stem from my eclipse making changes to the .project file, and it being part of the GIT project.  Personally I would be in favor of removing .project file from our repo.  Many of us different versions of Eclipse and different local configurations.   @db483 do you have an opinion on removing the .project file from the repo?  I think it makes sense removing the file with this PR.

db483:
@jhulslander It makes sense to me to remove the .project file. Thank you.

jhulslander:
I checked the three files you mentioned, I am able to open them in eclipse and no errors are reporting.  I am also able to run CuMaintainableXMLConversionServiceImplTest successfully without errors.

The .project file has been removed, and I added .project to .gitignore

cah292 (approved):

PR #1039, opened 2020-10-12
KFSPTS-19961 set maven compiler encoding to UTF-8

db483:


pom.xml
```diff
@@ -155,6 +155,7 @@
                 <version>${maven.compiler.version}</version>
                 <configuration>
                     <release>11</release>
+                    <encoding>UTF-8</encoding>
```
cah292, line 158: Even though this fixes the errors that Jenkins was reporting, Jenkins is still showing warnings about the project using a platform-specific encoding for copying resources. Would it be better to put this configuration in the "project.build.sourceEncoding" property instead, so that it affects the project as a whole? The various Cynergy POMs are currently using that property to specify the UTF-8 encoding.

cah292 (commented):
I had a question about whether the UTF-8 encoding configuration is in the correct place.

cah292 (approved):
Changes look good, unit tests pass, and the unit-test-job logs no longer show encoding-related warnings or errors. Merging.

PR #1040, opened 2020-10-12
KFSPTS-19212 Upgrade to 06/25/2020 financials patch

cah292:
There are PRs for this upgrade in cu-kfs, nonprod-sql and cu-kfs-upgrades.

Note that this upgrade introduced more explicit use of UTF-8 charset and US locale, but only the more closely upgrade-related cu-kfs files have been updated with those changes. I will create separate user stories to get those changes applied to the rest of the cu-kfs files, given that they might be too extensive to be entirely included in this PR.

Also, there are several DB schema changes with this release. I am not certain if the unit tests will pass without applying the related SQL to UTEST.


db483 (approved):
Changes look good.

PR #1041, opened 2020-10-15
KFSPTS-19435 Improve API for CapAsset update operation

cah292:
There are PRs for this in both cu-kfs and CapAsset-React-Website.


ajd299 (approved):
Code looks good, in addition I have tested locally and it seems to still work as expected. Ready to merge

PR #1042, opened 2020-10-16
KFSPTS-19213 upgrade to the 8/27/20 version of financials

db483:


cah292:
I found a few minor items in the pom.xml and PurapConstants.java files that would be good to fix, but otherwise the upgrade changes look good. I've updated those two files accordingly to fix the issues that I had found. Details are in the PR's commit history.

This is ready for review by another developer, and can be merged at the end of the sprint once all three PRs have been approved.

jhulslander (commented):
Code changes look good and can be merged when we are ready

jhulslander (approved):
this is ready to move now

PR #1043, opened 2020-10-16
KFSPTS-19631: Add PS_POSITION_DATA table to batch purgeJob using table specific KFS parameter PS_POSITION_DATA_NUMBER_OF_DAYS_OLD.

nkimble:


nkimble:
This PR depends on the SQL in nonprod-sql PR#850.
Both PR's should be merged at the same time.

db483:
@nkimble just wanted to confirm that this will delete the entries from both the extension table and the base position data table? Thank you.

nkimble:
@db483 I have confirmed that entries in both the base table and extension table are deleted with the code presently under review AND the current setting/configuration/wiring of [auto-delete="object"] in file [cu-obj-ld.xml].

The only thing I did not like with these changes was that business object PositionData had to be modified from base code to include log file output to ensure we had a recording of the records that were deleted in both tables by the purge job.

db483:
@nkimble Thank you for confirming. The code changes look good. I think for the purpose of this user story this implementation is good. 

An alternative would be to add an extension instead of an overlay but that would not buy us much and would still add custom code. 
Another option to consider is to update the TablePurgeRecordsDaoOjb.identifyAndRequestRecordsDeletion so that instead of using the object's toString method it would use it's own implementation to create a string representation of the object being purged. The advantage would be that it can use a consistent string representation without replying on the objects toString implementation. This way there is no need to alter the PositionData implementation. The new method could use something like ToStringBuilder.reflectionToString to build the string representation. If this is something to consider, I would suggest creating a technical debt user story for this change. Please let me know what you think.

Thank you.

nkimble:
Meeting with Boomi team clarified that PS_POSITION_DATA and PS_POSITION_DATA_TX are built from PS_POSITION_DATA_WD so PS_POSITION_DATA_WD is the table with data of record that would need to be purged. Closing this code review to make the appropriate adjustments.

PR #1044, opened 2020-10-20
KFSPTS-20019 Redirect to refactored lookup after PDP format

cah292:
We have a customization that redirects the user to the Format Summary (aka ProcessSummary) lookup after going through the formatting actions. This PR updates that customization so that the user will be redirected to the refactored ProcessSummary lookup, now that KualiCo has removed support for the old version of that lookup.


jhulslander (approved):

PR #1045, opened 2020-10-20
KFSPTS-20089 CapAsset KFS API Building & Room Active

ajd299:


cah292 (approved):
Changes look good, and unit tests pass. I'm also okay with the removal of the getAssetConditions() method, given that none of our code appears to be actively using it.

Merging.

PR #1046, opened 2020-10-20
KFSPTS-19631: Add PS_POSITION_DATA_WD table to batch purgeJob using table specific KFS parameter PS_POSITION_DATA_WD_NUMBER_OF_DAYS_OLD.

nkimble:


nkimble:
This PR depends on the SQL in nonprod-sql PR#856.
Both PR's should be merged at the same time.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -365,6 +366,17 @@
...
+          
+    <bean id="positionDataWorkdayLookupCriteriaPurgeService"
+          class="edu.cornell.kfs.sys.batch.service.impl.PositionDataWorkdayLookupCriteriaPurgeServiceImpl"/>
+          
+    <bean id="positionDataWorkdayTableDetailsForPurge" class="edu.cornell.kfs.sys.businessobject.TableDetailsForPurge"
+          p:businessObjectForRecordsTablePurge="edu.cornell.kfs.module.ld.businessobject.PositionDataWorkday"
+          p:useDefaultDaysBeforePurgeParameter="false"
+          p:nameSpaceCode="KFS-SYS"
+          p:component="PurgeTablesStep"
+          p:parameterName="PS_POSITION_DATA_WD_PARAMETER_NAME"
```
db483, line 378: Should this value be PS_POSITION_DATA_WD_NUMBER_OF_DAYS_OLD?

src/main/java/edu/cornell/kfs/sys/CUKFSParameterKeyConstants.java
```diff
@@ -52,5 +52,6 @@
         public static final String DEFAULT_NAME_SPACE_CODE = "KFS-SYS";
         public static final String DEFAULT_COMPONENT = "PurgeTablesStep";
         public static final String DEFAULT_PARAMETER_NAME = "DEFAULT_NUMBER_OF_DAYS_OLD";
+        public static final String PS_POSITION_DATA_WD_PARAMETER_NAME = "PS_POSITION_DATA_WD_NUMBER_OF_DAYS_OLD";
```
db483, line 55: Is this constant needed?

src/main/java/edu/cornell/kfs/module/ld/businessobject/PositionDataWorkday.java
```diff
@@ -0,0 +1,62 @@
...
+    private BigDecimal positionFullTimeEquivalency;
+    private String positionSalaryPlanDefault;
+    private String positionGradeDefault;
+    private Date inactivationDate;
+
+    public PositionDataWorkday() {
+    }
+
+    public String toString() {
+        SimpleDateFormat sdf = new SimpleDateFormat(KFSConstants.MONTH_DAY_YEAR_DATE_FORMAT);
```
db483, line 38: I am wondering if you could use reflection to output the String value of this object.

db483 (changes requested):
Changes look good but I had one concern regarding the system parameter name and one suggestion for improvement. Thank you.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
nkimble: Changing to match updated parameter name.

src/main/java/edu/cornell/kfs/sys/CUKFSParameterKeyConstants.java
nkimble: Isn't it our normal practice to have a constant defined so that we do not need to hard code a parameter name as a string where we use it in the code?
Based on your other noted concerns, I will change the constant to PS_POSITION_DATA_WD_NUMBER_OF_DAYS_OLD. so it matches the parameter name.

src/main/java/edu/cornell/kfs/module/ld/businessobject/PositionDataWorkday.java
nkimble: I will investigate.

src/main/java/edu/cornell/kfs/sys/CUKFSParameterKeyConstants.java
db483: Hi Nancy, yes, normal practice is to have a constant for parameter names. But this particular one is not used anywhere in the code so I am not sure it is needed.

db483:
Update code looks good. I was wondering whether we should keep all the constants for PositionDataWorkday or just keep inactivationDate as that is the only one being used. Thank you.

db483 (approved):
Changes look good.

PR #1047, opened 2020-10-21
KFSPTS-19988 adding 1099 form type to labor object code to 1099 box mapping

jhulslander:
Please note there is a related non-prod-sql PR that should be merged with this.

### DIFF
--- a/src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
+++ b/src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
@@ -1,0 +1,32 @@
+package edu.cornell.kfs.tax;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+public enum FormTypes1099 {
+	MISC("MISC", "1099-Misc"),
+	NEC("NEC", "1099-NEC");
+	
+	
+	public final String formCode;
+	public final String formDescription;
+	
+	private FormTypes1099 (String formCode, String formDescription) {
+		this.formCode = formCode;
+		this.formDescription = formDescription;
+	}
+	
+	public static FormTypes1099 findFormTypes1099FromFormCode(String formCode) {
+		for (FormTypes1099 type : FormTypes1099.values()) {
+			if (StringUtils.equalsIgnoreCase(type.formCode, formCode)) {
+				return type;
+			}
+		}
+		throw new IllegalArgumentException("Unable to find a 1099 form type for code " + formCode);
+	}
+	
+	public String toString() {
+        return ToStringBuilder.reflectionToString(this, ToStringStyle.NO_CLASS_NAME_STYLE);
+    }
+}
--- a/src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
+++ b/src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
@@ -5,26 +5,26 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
 public enum FormTypes1099 {
-  MISC("MISC", "1099-Misc"), NEC("NEC", "1099-NEC");
+    MISC("MISC", "1099-Misc"), NEC("NEC", "1099-NEC");
 
-  public final String formCode;
-  public final String formDescription;
+    public final String formCode;
+    public final String formDescription;
 
-  private FormTypes1099(String formCode, String formDescription) {
-    this.formCode = formCode;
-    this.formDescription = formDescription;
-  }
-
-  public static FormTypes1099 findFormTypes1099FromFormCode(String formCode) {
-    for (FormTypes1099 type : FormTypes1099.values()) {
-      if (StringUtils.equalsIgnoreCase(type.formCode, formCode)) {
-        return type;
-      }
+    private FormTypes1099(String formCode, String formDescription) {
+        this.formCode = formCode;
+        this.formDescription = formDescription;
     }
-    throw new IllegalArgumentException("Unable to find a 1099 form type for code " + formCode);
-  }
 
-  public String toString() {
-    return ToStringBuilder.reflectionToString(this, ToStringStyle.NO_CLASS_NAME_STYLE);
-  }
+    public static FormTypes1099 findFormTypes1099FromFormCode(String formCode) {
+        for (FormTypes1099 type : FormTypes1099.values()) {
+            if (StringUtils.equalsIgnoreCase(type.formCode, formCode)) {
+                return type;
+            }
+        }
+        throw new IllegalArgumentException("Unable to find a 1099 form type for code " + formCode);
+    }
+
+    public String toString() {
+        return ToStringBuilder.reflectionToString(this, ToStringStyle.NO_CLASS_NAME_STYLE);
+    }
 }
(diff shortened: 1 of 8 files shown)


src/main/resources/edu/cornell/kfs/tax/cu-ojb-tax.xml
```diff
@@ -8,6 +8,7 @@
         <field-descriptor name="versionNumber" column="VER_NBR" jdbc-type="BIGINT" locking="true" />
         <field-descriptor name="active" column="ACTV_IND" jdbc-type="VARCHAR" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbCharBooleanConversion" />
         <field-descriptor name="lastUpdatedTimestamp" column="LAST_UPDT_TS" jdbc-type="TIMESTAMP" index="true"/>
+        <field-descriptor name="formType" column="FORM_TYPE" jdbc-type="VARCHAR" />
```
cah292, line 11: Please see my note on the related SQL PR. If the "FORM_TYPE" does become part of the primary key, then either this user story or the follow-up user story should update the OJB metadata to configure it as such.

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
```diff
@@ -0,0 +1,55 @@
+package edu.cornell.kfs.tax;
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import com.mchange.util.AssertException;
```
cah292, line 9: Can this import be removed?

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
```diff
@@ -0,0 +1,55 @@
...
+
+	@AfterEach
+	void tearDown() throws Exception {
+	}
+
+	@Test
+	void findFormTypes1099FromFormCodeMisc() {
+		FormTypes1099 expectedType = FormTypes1099.MISC;
+		FormTypes1099 actualType = FormTypes1099.findFormTypes1099FromFormCode(FormTypes1099.MISC.formCode);
+		assertEquals(expectedType.formDescription, actualType.formDescription);
```
cah292, line 25: In this test method and the one below it, I would suggest including an assertion that also checks the formCode field.

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
```diff
@@ -0,0 +1,55 @@
...
+		try {
+			FormTypes1099.findFormTypes1099FromFormCode(null);
+			fail("Should have an illegal argument exception");
+		} catch (Exception e) {
+			assertEquals(IllegalArgumentException.class, e.getClass());
+		}
+	}
+	
+	@Test
+	void findFormTypes1099FromFormCodeEmptyString() {
```
cah292, line 46: You may also want a test case for a search value that is non-blank but invalid.

src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
```diff
@@ -0,0 +1,32 @@
...
+	MISC("MISC", "1099-Misc"),
+	NEC("NEC", "1099-NEC");
+	
+	
+	public final String formCode;
+	public final String formDescription;
+	
+	private FormTypes1099 (String formCode, String formDescription) {
+		this.formCode = formCode;
+		this.formDescription = formDescription;
```
cah292, line 17: I'm noticing a lot of tab characters in the new and modified code sections of this PR's files. Can these tabs be converted into spaces instead?

cah292 (changes requested):
I had a few suggestions for improvement, and one note about a change that may need to be added in the follow-up user story. Other than that, the changes look good.

src/main/resources/edu/cornell/kfs/tax/cu-ojb-tax.xml
jhulslander: I have an open question to the functionals on this

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
jhulslander: yes it should

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
jhulslander: OK

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
jhulslander: ok

src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
jhulslander: ok

jhulslander:
I've made the changes suggested, there is an open question about object code / payment reason code mapping to more than one form type.  We should get an answer to that before merging.

src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
```diff
@@ -0,0 +1,30 @@
...
+  public final String formDescription;
+
+  private FormTypes1099(String formCode, String formDescription) {
+    this.formCode = formCode;
+    this.formDescription = formDescription;
+  }
+
+  public static FormTypes1099 findFormTypes1099FromFormCode(String formCode) {
+    for (FormTypes1099 type : FormTypes1099.values()) {
+      if (StringUtils.equalsIgnoreCase(type.formCode, formCode)) {
```
cah292, line 20: Our cu-kfs Java code typically uses 4 spaces per indentation level. Can you please adjust the indentation of this PR's code for consistency with that? Thanks!

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      FormTypes1099.findFormTypes1099FromFormCode(StringUtils.EMPTY);
+      fail(SHOULD_HAVE_AN_ILLEGAL_ARGUMENT_EXCEPTION);
+    } catch (Exception e) {
+      assertEquals(IllegalArgumentException.class, e.getClass());
+    }
+  }
+  
+  @Test
+  void findFormTypes1099FromFormCodeEmptyDummyText() {
```
cah292, line 50: I would suggest removing the word "Empty" from the method name, to more accurately describe the search input that is being tested.

cah292 (changes requested):
I found a couple minor naming and indentation issues, but otherwise the changes look good. I will continue to wait for the functional feedback on whether any primary key adjustments are needed.

src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
jhulslander: ok

src/test/java/edu/cornell/kfs/tax/FormTypes1099Test.java
jhulslander: ok

jhulslander:
OK, I spoke with Cathy Salino for right now, we only want an object code and payment reason code mapped to only one box on one form
So I don't think the primary key needs to be adjusted.  This is ready for review and can be merged when you think it is ready

src/main/resources/edu/cornell/kfs/tax/cu-ojb-tax.xml
cah292: Thanks for confirming this with the functionals. Based on the feedback you provided, I'm fine with the form type as-is.

cah292 (approved):
Changes look good, unit tests pass, and the related SQL PR is also ready. Merging.

PR #1048, opened 2020-10-22
KFSPTS-20113: Remove toString method added for purgeJob from ConcurEventNotification business object as newer technologies are now being utilized.

nkimble:


jhulslander (changes requested):
These changes look fine, but I think there are a couple of imports that can now be removed:
import java.text.SimpleDateFormat;
import java.util.Locale;

nkimble:
Good catch. There are actually 5 imports that are no longer used so I will remove them all.

jhulslander (approved):

PR #1052, opened 2020-11-10
KFSPTS-20103 Add 1099-NEC tab file to tax processing job

cah292:
There are two PRs for this enhancement: One in cu-kfs and one in nonprod-sql.

For the 2020 tax year, we now need to generate tab records for both 1099-MISC and 1099-NEC, instead of just for 1099-MISC like in previous years. This PR implements the needed changes to have our tax batch job generate a 1099-NEC tab file.

A previous user story added a 1099 Form Type column to the Object Code Bucket Mapping BO, to help distinguish between 1099-MISC and 1099-NEC transactions. This PR had to expand these changes further by adding similar columns to the Transaction Override BO and the Transaction Detail BO.

To simplify these changes and to make certain future changes easier, a few areas have been refactored. For example, some of the KFS tax form properties now reference the type of payment (Rents, Royalties, etc.) instead of referencing the box by number. (A new parameter has been added to assist with mapping between the box numbers and these tax properties.) Also, one of the log-message-related enums has been updated to reference application resource properties instead, and a new tax-specific resource properties file has been added to collect the various tax-related messages.

Note that this tax job's code doesn't have very many areas that are suitable for writing micro-tests. Some future refactoring of the tax job code will be needed before writing better and more in-depth tests.


src/main/java/edu/cornell/kfs/tax/dataaccess/impl/Transaction1099Summary.java
```diff
@@ -198,9 +225,117 @@
         
         this.taxBoxIncludedDvCheckStubTextMaps = Collections.unmodifiableMap(tempCheckStubTextMaps);
         this.taxBoxDvCheckStubTextsAreWhitelists = Collections.unmodifiableMap(tempWhitelistFlagsMap);
+        this.taxBoxMinimumReportingAmounts = parseReportingThresholdsForTaxBoxes(parameterService, boxNumberMappings);
+        this.defaultTaxBoxMinimumReportingAmount = computeDefaultTaxBoxMinimumReportingAmount();
+        this.filerAddressFields = parseFilerAddressFields(parameterService);
+    }
+
+    private Map<Tax1099FilerAddressField, String> parseFilerAddressFields(ParameterService parameterService) {
```
db483, line 233: I wonder if a better name would be buildFilerAddressFieldsMapFromParameter?

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/Transaction1099Summary.java
```diff
@@ -198,9 +225,117 @@
...
+        }
+        String[] addressFieldValues = filerAddressFieldsFromParameter.toArray(String[]::new);
+        Map<Tax1099FilerAddressField, String> addressFields = new EnumMap<>(Tax1099FilerAddressField.class);
+        for (int i = 0; i < addressFieldKeys.length; i++) {
+            addressFields.put(addressFieldKeys[i], addressFieldValues[i]);
+        }
+        return Collections.unmodifiableMap(addressFields);
+    }
+
+    private Map<String, TaxTableField> parseBoxNumberMappings(ParameterService parameterService) {
```
db483, line 249: Would a better name be buildBoxNumberMappingsFromParameter?

db483 (changes requested):
The code changes look good but the tax reporting implementation remains very complex which makes it a bit more difficult to maintain and update. I wonder if this is something that would make sense to review as part of the Developer's bookclub.

The refactoring looks good, I am wondering if it was needed for adding the new functionality or was just a good opportunity to improve the code? 

I've had some minor suggestions and one concern.

Thank you.

src/main/java/edu/cornell/kfs/tax/batch/TransactionOverrideCsvBatchInputFileType.java
```diff
@@ -254,12 +263,22 @@ protected Object convertParsedLineToVO(Map<String,String> parsedLine, int lineNu
...
             LOG.error("Found a line that specifies both a 1099 and 1042S tax box override. Line number: " + Integer.toString(lineNumber));
             valid = false;
+        } else if (StringUtils.isBlank(formType1099)) {
+            LOG.error("Found a line with a 1099 box number that is missing a 1099 form type. Line number: "
+                    + Integer.toString(lineNumber));
+            valid = false;
+        } else if (FormTypes1099.findPotentialFormTypes1099FromFormCode(formType1099).isEmpty()
+                && !StringUtils.equalsIgnoreCase(formType1099, NULL_STRING)) {
+            LOG.error("Found a line with a 1099 box number that has an invalid 1099 form type. Line number: "
+                    + Integer.toString(lineNumber));
```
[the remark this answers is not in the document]
cah292: Good catch, yes, this block should also set the "valid" flag to false. I'll update this code accordingly.

src/main/java/edu/cornell/kfs/tax/FormTypes1099.java
```diff
@@ -15,6 +19,14 @@ private FormTypes1099(String formCode, String formDescription) {
         this.formDescription = formDescription;
     }
 
+    public static Optional<FormTypes1099> findPotentialFormTypes1099FromFormCode(String formCode) {
+        try {
+            return Optional.of(findFormTypes1099FromFormCode(formCode));
```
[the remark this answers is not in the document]
cah292: I was trying to set up the ability to search for a form code that might not have an existing mapping. However, given that this method might only be getting used in one other spot in the code, I'll pull this method out of here and place it in the class that relies on it. Further refactoring can be performed at a future date if needed.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/Transaction1099Summary.java
cah292: Ok, I'll rework this method to give it a better name, like the one you suggested.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/Transaction1099Summary.java
cah292: Ok, I'll rework this method to give it a better name, like the one you suggested.

cah292:
I've committed some changes to address the concerns that had been brought up.

This PR's extra refactoring was not a requirement for adding the 1099-NEC functionality. However, since the changes related to the 1099-NEC addition affected those particular areas, and since it was feasible to refactor those areas without an extreme amount of added work, it seemed like this PR would be a good opportunity to perform that work. Those changes should make things a little easier when performing future changes or making more significant refactoring efforts.

I agree that this batch job would make a good bookclub topic. I've also created a follow-up user story (KFSPTS-20262) to devise a more detailed plan for refactoring this batch job in the future. Not sure whether this tax processing should be discussed at the bookclub before or after the KFSPTS-20262 analysis has been performed.

This PR is ready for review again.

db483 (approved):
Changes look good.

db483:
Thank you for creating KFSPTS-20262. I would suggest that we discuss as a team in the next bookclub what would make most sense: review tax code before or after refactoring as part of KFSPTS-20262. My concern is that it may take a while until we are approved to work on improving/refactoring KFSPTS-20262.

PR #1056, opened 2020-11-18
KFSPTS-20162 save PVEN with iso to fips conversion issues

jhulslander:


jhulslander:
retest this please

chadbower:
retest this please

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
```diff
@@ -130,7 +129,6 @@ private boolean canPaymentWorksNewVendorRequestProcessingContinueForVendor(Payme
         if (pmwNewVendorNotNull(stgNewVendorRequestDetailToProcess, errorMessages)
                 && pmwDtosCouldConvertCustomAttributesToPmwJavaClassAttributes(stgNewVendorRequestDetailToProcess)
                 && pmwNewVendorAttributesConformToKfsLengthsOrFormats(stgNewVendorRequestDetailToProcess, errorMessages)
-                && allPmwNewVendorIsoCountriesMapToSingleFipsCountry(stgNewVendorRequestDetailToProcess, errorMessages)
```
cah292, line 133: This method was performing two different types of checks: Those which verify that the required country codes are present, and those which verify that the country codes map to a single FIPS country. Should a failure still happen at this stage if the country data is absent? Or should the PVEN still be created if the country codes are missing?

cah292 (commented):
Changes look good overall, though I had one question about what country-code-related circumstances should allow for generating the PVEN.

I also noticed that legacy-form mode may still prevent the PVEN from being created if certain country data is absent, but the new foreign-form mode appears to allow the PVEN to be created with missing country data. If we no longer need to support the legacy-form mode, then I'm okay with that discrepancy.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksNewVendorRequestsServiceImpl.java
jhulslander: Yes, we want to create a saved PVEN in both situations (can't convert an ISO country to a FIPS country and lack of country provided).  The PMW form requires various countries as specified by Purchasing staff, so the scenario is not likely to present itself.

Interesting find about legacy form processing mode.  Since we are turning on foreign form processing mode this evening, and won't be going back to legacy mode, it should be fine to fix the scenario you mention.  Also, in the near term, we will be removing legacy form only code.

cah292: Ok, thanks for clarifying. I'm fine with this change as-is, then.

jhulslander:
Unless you have further things for me to look at, I think this can be merged once you agree.

cah292 (approved):

PR #1058, opened 2020-11-19
KFSPTS-20361 Remove extra unneeded closing tag from PO

cah292:


jhulslander (approved):
changes look fine, unit tests pass, I shall merge

PR #1059, opened 2020-11-23
KFSPTS-20379 Fix DV income tax bug in 10/29/2020 patch

cah292:
There are some income-tax-related areas where the financials 10/29/2020 patch accidentally compares KualiDecimals against BigDecimals, which can result in unexpected errors. This PR adds some fixes for those by overlaying the DisbursementVoucherTaxServiceImpl class.

Note that some of the problematic areas were fixed by our initial commit of the 10/29/2020 patch to the develop branch. This PR expands the fix to another class that was having issues.


jhulslander (approved):
since the big decimal / Kuali Decimal fix is spread out through this class, using an overlay seems to make the most sense.  Unit tests pass, I shall merge

PR #1060, opened 2020-11-24
KFSPTS-19396 add fix to use isNull criteria instead of equal to empty…

db483:
… String as this does not work in Oracle DB


src/main/java/org/kuali/kfs/fp/document/dataaccess/impl/GeneralLedgerTransferEntryDaoOjb.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        return criteria;
+    }
+
+    public boolean hasTransferredEntries(String documentNumber) {
+        Criteria criteria = new Criteria();
+        criteria.addEqualTo(KFSPropertyConstants.DOCUMENT_NUMBER, documentNumber);
+        criteria.addEqualTo(KFSPropertyConstants.FINANCIAL_SYSTEM_ORIGINATION_CODE, KFSConstants.ORIGIN_CODE_KUALI);
+        criteria.addNotNull(KFSPropertyConstants.GENERAL_LEDGER_TRANSFER_DOCUMENT_NUMBER);
```
jhulslander, line 159: is this also a CU customization?  If so, should we add a note here as well?

jhulslander (commented):
code changes look good, I had one question though

src/main/java/org/kuali/kfs/fp/document/dataaccess/impl/GeneralLedgerTransferEntryDaoOjb.java
db483: This is not CU customization, the only change to this file is the line I have marked.

jhulslander (approved):
Thanks for confirming, this unit tests pass, I shall merge

PR #1061, opened 2020-11-24
KFSPTS-20244 upgrade to the 11-12 version of financials

jhulslander:


db483:
Please set the upgrade version to 11/12/20.

db483:
As KualiCo is adding bulkReindexDocsService, can we remove our local version of this batch job?

PR #1065, opened 2020-12-09
KFSPTS-20549 pull GeneralLedgerPendingEntryServiceImpl.java from 10-29 to fix poster, merged to master

jhulslander:


jhulslander:
Dave gave the sign off to move to test, but I didn't resolve his sub task, I think this can be merged

src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
```diff
@@ -0,0 +1,1081 @@
...
+                        cashOffsetObject.getFinancialObjectCode(), bank.getCashOffsetSubObjectCode());
+                return false;
+            }
+
+            bankOffsetEntry.setFinancialSubObjectCode(bank.getCashOffsetSubObjectCode());
+        }
+
+        bankOffsetEntry.setTransactionEntryOffsetIndicator(true);
+        bankOffsetEntry.setTransactionLedgerEntryAmount(depositAmount.abs());
+        bankOffsetEntry.setUniversityFiscalPeriodCode(financialDocument.getPostingPeriodCode());
```
cah292, line 754: Does this line also need to be updated to set the fiscal period to null? I believe this line had the null-value setup prior to our 10/29/2020 financials upgrade.

cah292 (commented):
I had a question about whether an additional line needs to be updated. Other than that, the changes look good.

src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
jhulslander: This line wasn't changed in recent history, only the line were I commented the CU customization was changed.  Also, the kfs-dev validation only validated that one change.  I don't think we should add changes.

jhulslander: Yes, I double checked the kfs-dev version, this line exists there, and testing didn't find a problem

cah292: Ok, thanks for checking. I'm fine with moving this change forward as-is, but we'll need the functionals to verify that the bank offset entries are being posted to the correct fiscal period. We may need a follow-up PR if this line does end up needing a change.

jhulslander: agreed

cah292 (approved):
Changes look good, unit tests pass, the user story has sub-tasks for preventing merge conflicts when the develop version of the fix moves to master, and my concerns have been addressed for the moment. Merging.

PR #1066, opened 2020-12-09
KFSPTS-20411 Fix wildcard-as-literal handling in tax module

cah292:
The Object Code Bucket Mapping BO sometimes has "*" (an asterisk) in one of its primary key fields. However, one of the default duplicate-object checks on maintenance documents will invoke a search method that can treat asterisks as wildcards. This PR adjusts the bucket mapping maintenance code, so that it will build search criteria that have been adjusted to treat a potential asterisk as a literal.

In addition, there are some lookup fields for bucket mappings and transaction overrides that need wildcard-as-literal handling but were lacking it. Thus, this PR will also adjust the DD setup for those two BOs so that the relevant fields will treat wildcards as literals on their lookup screens.


db483:
Chad,

The changes look good but I agree with Cathy that this should only be a short term solution and we should implement a more robust solution for long term. I believe that should be part of the Tax Process Improvements project that we will work on in the new year. I created an epic for Tax Process Improvements. I would suggest adding a user story there for removing this fix and incorporating a more robust solution in the rewrite.

I see that these changes are fixing more than just the Object Code Bucket Mapping issue. I would suggest updating the user story to reflect that and adding details for the needed testing steps.

Thank you very much Chad!

cah292:
Daniela,

As requested, I've created a new KFSPTS-20566 ticket to implement a future long-term fix for this wildcard-handling issue. I've also updated this PR's user story accordingly to mention the extra changes and to include the appropriate testing steps.

I believe this is ready to be merged if you are satisfied with the changes.

db483 (approved):
Changes look good and my concerns have been addressed.

PR #1069, opened 2020-12-11
KFSPTS-20565 Backport FINP-7275 GLPE fiscal period fix

cah292:
This PR backports the GLPE fiscal period fix from KualiCo's 2020-12-10 patch.

I verified locally that KFS can still boot up with the changes in place, but I ran out of time today to validate the effects of the fix in my local environment. If the team is okay with deferring the testing until the changes move into DEV, then this PR can be merged after another developer approves it.

Note that CuGeneralLedgerPendingEntryServiceImpl currently does not have code changes in this PR; it has a method that sets up the GLPE fiscal period, but it uses the old behavior of setting it to null like we want it to. If you think that service method should be modified to do the parameter-based check on whether to set to null, please let me know.

### DIFF
--- a/src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
+++ b/src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
@@ -1,0 +1,1032 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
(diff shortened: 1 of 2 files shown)


db483:
Chad,

Changes look good and I will merge them for validation in KFS-DEV.
Thank you for pointing out the changes needed in CuGeneralLedgerPendingEntryServiceImpl. I think it would make sense to have a separate user story for that.

Thank you!

db483 (approved):
Changes look good.

PR #1073, opened 2020-12-16
Kfspts 20150

db483:


cah292 (approved):
Changes look good, and unit tests pass. I'm okay with the log message's setup since it follows the format that the superclass is using, though that may need to be revisited in the future if the isPRNCDocument() method gets used for something other than can-auto-approve checks.

Merging.

PR #1075, opened 2020-12-18
KFSPTS-19976 Receipt Processing CSV Parsing Fix Potential Resource Leak

ajd299:


src/main/java/edu/cornell/kfs/module/receiptProcessing/CsvBatchInputFileTypeBase.java
```diff
@@ -76,15 +79,14 @@ public Object parse(byte[] fileByteContent) throws ParseException {
...
-            //use csv reader to parse the csv content
-            CSVReader csvReader = new CSVReader(
-                    new InputStreamReader(new ByteArrayInputStream(fileByteContent), StandardCharsets.UTF_8),
-                    ',', '"', '|');
+        validateCSVHeader(fileByteContent, headerList);
+
+        final CSVParser csvParser = new CSVParserBuilder().withSeparator(',').withQuoteChar('"').withEscapeChar('|').build();
+        try (
+                final InputStreamReader inputStreamReader = new InputStreamReader(new ByteArrayInputStream(fileByteContent), StandardCharsets.UTF_8);
+                final CSVReader csvReader = new CSVReaderBuilder(inputStreamReader).withSkipLines(1).withCSVParser(csvParser).build();
```
cah292, line 88: This reader is being configured to skip a line, but a few lines down from here the existing code will automatically remove the first line in the list. Does that other line-deleting code need to be removed?

src/main/java/edu/cornell/kfs/module/receiptProcessing/CsvBatchInputFileTypeBase.java
```diff
@@ -113,6 +115,15 @@ public Object parse(byte[] fileByteContent) throws ParseException {
         return convertParsedObjectToVO(parsedContents);
     }
 
+    protected void validateCSVHeader(byte[] fileByteContent, List<String> headerList) throws ParseException {
+        try (ByteArrayInputStream validateFileContents = new ByteArrayInputStream(fileByteContent)) {
+            validateCSVFileInput(headerList, validateFileContents);
```
cah292, line 120: There are some potential resource leaks in the validateCSVFileInput() method that also need to be fixed. That method currently creates its own InputStreamReader and CSVReader but never closes them.

(Note that, due to the setup of the data that this class is being used for, the validation can succeed even though its reader doesn't have the same CSV parser setup as the main data-reading code.)

cah292 (changes requested):
I noticed a couple items that still need to be addressed, but the other changes so far look good.

ajd299:
Please review latest changes

src/main/java/edu/cornell/kfs/module/receiptProcessing/CsvBatchInputFileTypeBase.java
```diff
@@ -125,23 +133,24 @@ public Object parse(byte[] fileByteContent) throws ParseException {
...
-
-        if (!CollectionUtils.isEqualCollection(expectedHeaderList, inputHeaderList)) {
-            errorMessage = "CSV Batch Input File contains incorrect number of headers";
-            //collection has same elements, now check the exact content orders by looking at the toString comparisons
-        } else if (!expectedHeaderList.equals(inputHeaderList)) {
-            errorMessage = "CSV Batch Input File headers are different";
-        } else {
-        		errorMessage = validateDetailRowsContainExpectedNumberOfFields(expectedHeaderList, csvReader);
-        }
+        try (CSVReader csvReader = new CSVReader(new InputStreamReader(fileContents, StandardCharsets.UTF_8))) {
```
cah292, line 136: Just to confirm, will the enclosed InputStreamReader be auto-closed when the CSVReader closes? Or does the reader need to be defined on a line prior to the CSVReader so that it will auto-close accordingly by the try-with-resources block?

cah292 (commented):
I had one more question about the revisions, but otherwise the changes look good.

src/main/java/edu/cornell/kfs/module/receiptProcessing/CsvBatchInputFileTypeBase.java
ajd299: Please see the latest commit. 

Question: are the final keywords advisable?

cah292: In the case of local variables, I'm fine either way with including or excluding the "final" keyword. If your code needs to have a certain local variable remain unchanged, make sure that the code at least treats it as "effectively" final if it's not explicitly declared final.

cah292 (approved):
Revised changes look good, and unit tests pass. Merging.

PR #1076, opened 2020-12-22
KFSPTS-20409 Fix vendor payee type setup from DV XML docs

cah292:
The Create Accounting Documents Job had a bug in its DV generation process, where DV vendors were always given the "V" payee type, regardless of the vendor type defined on the vendor header record. This can cause the DV to erroneously allow the payment-reason-related business rules to pass, since some types of vendors are not allowed to be specified for particular DV payment reasons. This PR fixes the issue, where it will set up DV vendor payees with the correct payee type code, based on the map returned by the DisbursementVoucherPayeeServiceImpl.getVendorPayeeTypeCodeMapping() method. (The existing DV business rules will stop the document from being submitted if the given vendor payee is the wrong type for the given payment reason.)

A simple unit test and supporting classes were added, to verify that the type codes are being mapped appropriately for DV vendor payees. It would be good to perform some additional refactoring of the create-acct-doc unit test setup to validate the DV generation process in more detail, but that is likely beyond the scope of this user story.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
@@ -24,6 +24,7 @@ import org.kuali.kfs.sys.KFSPropertyConstants;
 import org.kuali.kfs.sys.businessobject.AccountingLine;
 import org.kuali.kfs.sys.service.UniversityDateService;
 import org.kuali.kfs.vnd.businessobject.VendorAddress;
+import org.kuali.kfs.vnd.businessobject.VendorDetail;
 import org.kuali.kfs.vnd.document.service.VendorService;
 import org.kuali.rice.core.api.util.type.KualiDecimal;
 
@@ -39,6 +40,7 @@ import edu.cornell.kfs.fp.batch.xml.DisbursementVoucherPrePaidTravelOverviewXml;
 import edu.cornell.kfs.fp.businessobject.CuDisbursementVoucherPayeeDetail;
 import edu.cornell.kfs.fp.document.CuDisbursementVoucherDocument;
 import edu.cornell.kfs.fp.document.service.CuDisbursementVoucherDefaultDueDateService;
+import edu.cornell.kfs.fp.document.service.CuDisbursementVoucherPayeeService;
 import edu.cornell.kfs.pdp.CUPdpConstants;
 
 public class CuDisbursementVoucherDocumentGenerator extends AccountingDocumentGeneratorBase<CuDisbursementVoucherDocument> {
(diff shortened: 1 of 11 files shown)


src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherPayeeServiceImpl.java
```diff
@@ -182,4 +182,12 @@ protected String getPayeeEntityId(String payeeIdNumber) {
 
         return entityId;
     }
+
+    @Override
+    public String getPayeeTypeCodeForVendorType(String vendorTypeCode) {
+        if (StringUtils.isBlank(vendorTypeCode)) {
+            return null;
```
db483, line 189: Should this return null or should this be considered an exception? Should there be a unit test for this case when vendorTypeCode is blank?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
```diff
@@ -136,29 +139,35 @@ protected void validateAndPopulatePayeeTypeCodeAndPayeeNumber(DisbursementVouche
         String payeeTypeCode = paymentInfo.getPayeeTypeCode();
         String payeeId = paymentInfo.getPayeeId();
         boolean isEmployee = false;
+        String convertedPayeeTypeCode;
```
db483, line 142: Would it make sense to default this to payeeTypeCode?

db483 (changes requested):
Changes look good but I had a couple of questions. Thank you.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CuDisbursementVoucherDocumentGenerator.java
cah292: Ok, I'll adjust this to default the variable to payeeTypeCode for clarity.

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherPayeeServiceImpl.java
cah292: This blank-value check was just meant as a safeguard against passing a null search value to a Map instance controlled by base code, in case KualiCo ends up changing the Map to a type that doesn't allow null inputs. I'll work on adding some unit test methods to test out various inputs to this method.

cah292:
I've committed a few changes to address the initial feedback. Also note that I created a separate unit test class for testing the new vendor-type-to-payee-type method, since our existing unit test class for our CU payee service is not a micro-test.

db483 (approved):
Changes look good.

PR #1077, opened 2021-01-04
KFSPTS-20273 removing CuDisbursementVoucherNonresidentInformationVali…

jhulslander:
Per discussion with Cathy, we are going to try and remove our customization, and check to see if it works for us.  Upon review of CuDisbursementVoucherNonresidentInformationValidation and DisbursementVoucherNonresidentInformationValidation and it looks like base code includes the functionality we need.  Cathy will need to review the parameter INCOME_CLASS_CODES_REQUIRING_STATE_TAX that is used to control if zero tax will pass validation


db483:
If we are removing this customization now I think we will have to backport https://kualico.atlassian.net/browse/FINP-7271. Otherwise we will want to wait until we upgrade to 12-10-2020 version of financials: https://kuali-financials.zendesk.com/hc/en-us/articles/360060594913-Financials-Production-Release-2020-12-10/

Do we also need to review validateFederalStateTaxPercentsForIncomeClass and see if it is ok for our local validation to be replaced by validateFederalStateTaxPercentsForIncomeClass?

jhulslander:
I spoke with Daniela and confirmed the functions Daniela mentioned seem to be what we want to run, but testing will confirm that.

Per Cathy in today's standup, we will wait to do this till after we the 12/10 patch brought in

PR #1079, opened 2021-01-06
KFSPTS-20170 add webservice API call so cappasset can determine if the test banner should be displayed

jhulslander:


ajd299:
Code looks good, will merge when the corresponding capasset PR is merged.

ajd299 (approved):

PR #1082, opened 2021-01-12
KFSPTS-18645: Obtained most current version of file from base code for our overlay and reapplied the Cornell modification to clean up unused items.

nkimble:


nkimble:
Note that the base code version of the file was pulled locally with our mod reapplied to make future upgrade comparisons easier.

jhulslander (changes requested):
This code changes looks good, I think we may also need to update bean id Purchasing-importItemValidation in purapvalidatordefinitions.xml ... dictionaryValidationService no longer seems to be used.

nkimble:
Base code still has that service defined in that bean. That is why I left it in our overlay. Please confirm my findings.
I don't think we should mod that file away from base code if your review confirms.

jhulslander (approved):
OK, things look good, PurchasingImportItemValidation extends PurchasingAccountsPayableImportItemValidation and PurchasingAccountsPayableImportItemValidation contains DictionaryValidationService.  So the sprint configuration is correctly.

unit tests pass, code changes look good, I shall merge

PR #1083, opened 2021-01-15
KFSPTS-20744 Add fiscal period feature to custom GLPE code

cah292:
Our prior sprints preserved a GLPE-related helper method that KualiCo had removed, but it currently does not contain the advanced fiscal-period-deriving functionality from the FINP-7275 fix in base code. This PR updates the custom method to take advantage of the FINP-7275 changes.

This PR does not update the documents that rely on our custom method for GLPE generation. (I think our PREQ documents are the only ones actively using the custom CU GLPE method.) If we decide to allow for more use of the fiscal-period-choosing feature via a drop-down or other means, we would need to alter the affected documents appropriately, and I would suggest doing that in a separate user story.

### DIFF
--- a/src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
+++ b/src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
@@ -79,8 +79,9 @@ import java.util.List;
 import java.util.Map;
 
 /*
- * CU Customization: Backported the FINP-7275 fix from the 2020-12-10 financials patch.
- * This overlay can be removed once we upgrade to the 2020-12-10 patch or later.
+ * CU Customization: Backported the FINP-7275 fix from the 2020-12-10 financials patch,
+ * and updated one of the new methods from that release to protected visibility.
+ * This overlay can be removed or simplified once we upgrade to the 2020-12-10 patch or later.
  */
 /**
  * This class is the service implementation for the GeneralLedgerPendingEntry structure. This is the default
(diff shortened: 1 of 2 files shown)


src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
```diff
@@ -79,8 +79,9 @@
 import java.util.Map;
 
 /*
- * CU Customization: Backported the FINP-7275 fix from the 2020-12-10 financials patch.
- * This overlay can be removed once we upgrade to the 2020-12-10 patch or later.
+ * CU Customization: Backported the FINP-7275 fix from the 2020-12-10 financials patch,
+ * and updated one of the new methods from that release to protected visibility.
+ * This overlay can be removed or simplified once we upgrade to the 2020-12-10 patch or later.
```
db483, line 84: Should this comment be more clear about the changes needed when we upgrade to 12/10?
I will ask KualiCo to change the method's visibility to protected.

db483:
Chad, what kind of changes would be needed to the PREQ document code to  allow for the use of the fiscal-period-choosing feature via a drop-down or other means? Isn't this something supported by base code and controlled by system parameters? Thank you.

db483 (changes requested):
Changes look good but I had one suggestion for improvement. Thank you/

src/main/java/org/kuali/kfs/sys/service/impl/GeneralLedgerPendingEntryServiceImpl.java
cah292: I will update the comment for clarity.

cah292:
With regards to enabling the fiscal period drop-down on PREQs, I'm not sure if such documents are fully set up to allow for it. I was having difficulty finding any areas in the code where the PREQs were using such drop-downs, so we may have to add them manually. Also, the description on the "FISCAL_PERIOD_SELECTION_DOCUMENT_TYPES" parameter notes that only certain documents (most notably FP documents) have been validated with this feature, so PREQs may not be fully compatible with the feature even if they were added to the parameter.

The code changes from this PR will at least give us the option of using the parameter to enable the new fiscal period feature on the PREQs, but more investigation is needed to verify whether enabling that feature would work as-is. (Getting this working for PREQs may require changing their getPostingPeriodCode() method, among other things.)

cah292:
As requested, I've updated the comment on the overlayed file for clarity.

db483:
Chad, Thank you for explaining this.

db483 (approved):
Changes look good.

db483:
I submitted a request to KualiCo to have the determineFiscalPeriodCode method updated from private to protected.

PR #1086, opened 2021-01-22
KFSPTS-20854: Fix null object checks on accounting line validation introduced by 12-03-2020 FINP-7183. To be fixed by FINP-7325.

nkimble:


nkimble:
This is ready for review. Unit tests are running.

Question: Would you recommend that lines 430, 431, and 432 also have ObjectUtils.isNull/isNotNull checks added for the "accountingLine" input parameter where those respective local objects are set to "null" when "accountingLine" is detected as null?

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
```diff
@@ -0,0 +1,589 @@
...
+    public List<AccountingLineValidationError> getAccountingLineValidationErrors(AccountingLine accountingLine, boolean useShortMessages) {
+        List<AccountingLineValidationError> validationErrors = new ArrayList<>();
+
+        //get the accounting line sequence string to identify which line has error.
+        String accountIdentifyingPropertyName = getAccountIdentifyingPropertyName(accountingLine);
+
+        // retrieve accounting line objects to validate
+        Chart chart = accountingLine.getChart();
+        Account account = accountingLine.getAccount();
+        ObjectCode objectCode = accountingLine.getObjectCode();
```
cah292, line 432: Nancy mentioned the possibility of adding some non-null checks for this section of the code. However, since this particular method relies heavily on validation code that requires a non-null accounting line, it might be better to just have this method return an empty validation errors list if the accounting line is null.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    public boolean hasAccountRequiredOverrides(AccountingLine line, String overrideCode) {
+        boolean retVal = true;
+        AccountingLineOverride override = AccountingLineOverride.valueOf(overrideCode);
+        Account account = line.getAccount();
+        if (AccountingLineOverride.needsExpiredAccountOverride(account)
+                && !override.hasComponent(AccountingLineOverride.COMPONENT.EXPIRED_ACCOUNT)) {
+            Account continuation = accountService.getUnexpiredContinuationAccountOrNull(account);
+            if (continuation == null) {
```
cah292, line 139: This spot wasn't impacted by the FINP-7183 changes, but it may still be a good idea to perform the ObjectUtils null check here.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
```diff
@@ -0,0 +1,589 @@
...
+
+    /**
+     * Validates an accounting line. Side effects: refreshes referenced BOs and sets global error messages
+     *
+     * @param accountingLine
+     * @return boolean whether the accounting line is valid
+     */
+    @Override
+    public boolean validateAccountingLine(AccountingLine accountingLine) {
+        if (accountingLine == null) {
```
cah292, line 390: Another spot that may need the ObjectUtils update, despite not being changed by FINP-7183.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
```diff
@@ -0,0 +1,589 @@
...
+    public void setAccountService(AccountService accountService) {
+        this.accountService = accountService;
+    }
+
+    public void setConfigurationService(ConfigurationService configurationService) {
+        this.configurationService = configurationService;
+    }
+
+    // This exists so tests can avoid static mocking
+    String getDashSubAccountNumber() {
```
cah292, line 586: It looks like this overlay is for the 2020-12-17 version of the class. However, that release only seems to have made some minor updates for unit testing convenience, such as the addition of this particular method. I'm okay with keeping those changes in here if necessary.

nkimble:
Unit tests have completed successfully.

cah292 (commented):
I noticed a few more spots in this file that may need adjustment, but other than that the overlay looks good.

Also, based on KualiCo's FINP-7183 commit, this appears to be the only file that we need to overlay with a fix.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
nkimble: Removing these lines and adjusted two other area to keep file in sync with KualiCo 12-03-2020 released version.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
nkimble: Adjusting as recommended.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
nkimble: Added null check.

src/main/java/org/kuali/kfs/sys/document/service/impl/AccountingLineRuleHelperServiceImpl.java
nkimble: Added null check.

nkimble:
This is ready to be reviewed again.

cah292 (approved):

PR #1095, opened 2021-02-17
KFSPTS-20958: Upgrade to the 01-21-2021 version of financials.

nkimble:


nkimble:
Chad: FYI for this review, I did do some formatting/style adjustments in the HigherEducationFunction.xml and the SubFundGroup.xml to ensure that I did not miss the changes KualiCo was making in base code.  This was done with the intent of getting those files closer to base code to make future compares easier.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/SubFundGroup.xml
```diff
@@ -37,277 +39,272 @@
...
+                <ref bean="SubFundGroup-active"/>
+                <ref bean="SubFundGroup-fundGroupBudgetAdjustmentRestrictionLevelCode"/>
+                <ref bean="SubFundGroup-accountRestrictedStatusCode"/>
+            </list>
+        </property>
+    </bean>
 
-	<!-- Business Object Lookup Definition -->
+    <bean id="SubFundGroup-inquirySectionDefinition" parent="SubFundGroup-inquirySectionDefinition-parentBean"/>
+    <bean id="SubFundGroup-inquirySectionDefinition-parentBean" abstract="true" parent="InquirySectionDefinition"
```
cah292, line 215: Our old version of this bean was setting "numberOfColumns" to 1. However, the newer type of "section" bean doesn't seem to have this property, so I am not certain whether it is worth keeping it on this legacy bean still.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/SubFundGroup.xml
```diff
@@ -37,277 +39,272 @@
...
+          p:title="">
+        <property name="inquiryFields">
+            <list>
+                <bean parent="FieldDefinition" p:attributeName="subFundGroupCode"/>
+                <bean parent="FieldDefinition" p:attributeName="subFundGroupDescription"/>
+                <bean parent="FieldDefinition" p:attributeName="subFundGroupTypeCode"/>
```
cah292, line 221: A FieldDefinition for our custom extension field should be reinserted between these two, like in the old version of the bean.

cah292 (changes requested):
I found a few spots in the Spring bean files that could use further cleanup, but otherwise the upgrade changes look good.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/HigherEducationFunction.xml
```diff
@@ -1,51 +1,50 @@
...
-			</list>
-		</property>
-	</bean>
+    <bean id="HigherEducationFunction-inquiryDefinition-override"
+        parent="DataDictionaryBeanOverride">
+        <property name="beanName" value="HigherEducationFunction-inquiryDefinition" />
+        <property name="fieldOverrides">
+            <list>
+                <bean parent="FieldOverrideForListElementInsert">
+                    <property name="propertyName" value="inquirySections[0].inquiryFields" />
```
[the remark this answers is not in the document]
nkimble: Could you identify an existing example of what you are referring to?
Changes I am trying to make locally to address your concern are not working as expected.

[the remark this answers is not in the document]
cah292: Given where the override needs to be, you could add a new "FieldOverrideForListElementInsert" bean into the "fieldOverrides" list property on the "HigherEducationFunction-inquiryDefinition-override" bean. KualiCo's new inquiry setup is very recent, and I don't think we have an existing example of a field-override bean for that. However, it looks like such a bean should have a setup like the following, to perform the equivalent insert from our existing bean above:

```xml
<bean parent="FieldOverrideForListElementInsert"
      p:propertyName="sections[0].fields"
      p:propertyNameForElementCompare="name"
      p:element-ref="HigherEducationFunction-financialHigherEdFunctionNm"
      p:insertAfter-ref="HigherEducationFunction-extension-financialHigherEdFunctionDescription"/>
```

Note that the "element" and "insertAfter" properties might need to be replaced with child beans instead of bean refs, if the inquiry or other related screens run into unexpected issues.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/SubFundGroup.xml
cah292: It looks like KualiCo removed the explicit numberOfColumns=1 setting over 5 years ago, so our bean setup must not have been updated to reflect that until now. (I also confirmed that the 01/21/2021 version of the bean does not have a setting for that property.) I'm fine with excluding "numberOfColumns" from this bean.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/HigherEducationFunction.xml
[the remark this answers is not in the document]
cah292: As another alternative, you could add a DataDictionaryBeanOverride bean that operates on the "HigherEducationFunction-sectionDefinition" bean instead, and then configure the field override to insert the element into the "fields" list. Perhaps this way would be a little clearer, so that you can simplify the list property path in the field-override bean.

nkimble:
This is ready to be reviewed again.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/HigherEducationFunction.xml
```diff
@@ -1,51 +1,42 @@
...
-			</list>
-		</property>
-	</bean>
+    <bean id="HigherEducationFunction-inquiryDefinition-override"
+        parent="DataDictionaryBeanOverride">
+        <property name="beanName" value="HigherEducationFunction-inquiryDefinition" />
+        <property name="fieldOverrides">
+            <list>
+                <bean parent="FieldOverrideForListElementInsert"
+                      p:propertyName="sections[0].fields"
```
cah292, line 34: It seems that this field override has been updated to only affect the new section field list. I don't see any new beans that perform a field override on the legacy section field list. Can you please confirm whether the older-style HigherEducationFunction inquiry is still openable? If it is, then we'll still need a field override that affects the legacy field list.

cah292 (commented):
I had a question about one of the bean overrides that got updated. Other than that, the revised changes look good.

nkimble:

cah292 (approved):
Changes look good, unit tests pass, my concerns have been addressed, and the related cu-kfs-upgrades PR is also ready. This can be merged once we get closer to the end of the sprint.

cah292:
Changes have already been approved to move forward, and we are now at the end of the sprint. Merging.

PR #1098, opened 2021-02-26
KFSPTS-20397 add cache to AWS Secrets manager and create testing batch job

jhulslander:
I noticed that we no longer required the dynamo DB dependency so I removed that.  It turns our the dynamo library also provided org.apache.httpcomponents to be available at compile time and run time.  I had to add a new org.apache.httpcomponents dependency that wasn't scoped for testing.  Please let me know if you are aware of a better maven configuration.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -795,15 +795,10 @@
             </exclusions>
         </dependency>
         <dependency>
-	          <groupId>com.amazonaws</groupId>
-	          <artifactId>aws-java-sdk-secretsmanager</artifactId>
-	          <version>${aws-java-sdk.version}</version>
+          <groupId>com.amazonaws</groupId>
+          <artifactId>aws-java-sdk-secretsmanager</artifactId>
+          <version>${aws-java-sdk.version}</version>
         </dependency>
-        <dependency>
-					  <groupId>com.amazonaws.secretsmanager</groupId>
-					  <artifactId>aws-secretsmanager-caching-java</artifactId>
-					  <version>1.0.1</version>
-				</dependency>
         <dependency>
             <groupId>backport-util-concurrent</groupId>
             <artifactId>backport-util-concurrent</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
         <cynergy.version>2.7.2</cynergy.version>
         <cynergy.version.suffix>current</cynergy.version.suffix>
 
-        <aws-java-sdk.version>2.14.7</aws-java-sdk.version>
+        <aws-java-sdk.version>1.11.339</aws-java-sdk.version>
         <backport-util-concurrent.version>2.2</backport-util-concurrent.version>
         <checkstyle.version>8.37</checkstyle.version>
         <gson.version>2.8.2</gson.version>
@@ -784,10 +784,21 @@
         </dependency>
 
         <dependency>
-            <groupId>software.amazon.awssdk</groupId>
-            <artifactId>secretsmanager</artifactId>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-dynamodb</artifactId>
             <version>${aws-java-sdk.version}</version>
-        </dependency> 
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+          <groupId>com.amazonaws</groupId>
+          <artifactId>aws-java-sdk-secretsmanager</artifactId>
+          <version>${aws-java-sdk.version}</version>
+        </dependency>
         <dependency>
             <groupId>backport-util-concurrent</groupId>
             <artifactId>backport-util-concurrent</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -942,6 +942,17 @@
             <classifier>tests</classifier>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>${httpclient.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>javax.validation</groupId>
             <artifactId>validation-api</artifactId>
             <version>${validation-api.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
         <cynergy.version>2.7.2</cynergy.version>
         <cynergy.version.suffix>current</cynergy.version.suffix>
 
-        <aws-java-sdk.version>1.11.339</aws-java-sdk.version>
+        <aws-java-sdk.version>2.14.7</aws-java-sdk.version>
         <backport-util-concurrent.version>2.2</backport-util-concurrent.version>
         <checkstyle.version>8.37</checkstyle.version>
         <gson.version>2.8.2</gson.version>
@@ -784,21 +784,10 @@
         </dependency>
 
         <dependency>
-            <groupId>com.amazonaws</groupId>
-            <artifactId>aws-java-sdk-dynamodb</artifactId>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>secretsmanager</artifactId>
             <version>${aws-java-sdk.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-logging</groupId>
-                    <artifactId>commons-logging</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-          <groupId>com.amazonaws</groupId>
-          <artifactId>aws-java-sdk-secretsmanager</artifactId>
-          <version>${aws-java-sdk.version}</version>
-        </dependency>
+        </dependency> 
         <dependency>
             <groupId>backport-util-concurrent</groupId>
             <artifactId>backport-util-concurrent</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -795,10 +795,15 @@
             </exclusions>
         </dependency>
         <dependency>
-          <groupId>com.amazonaws</groupId>
-          <artifactId>aws-java-sdk-secretsmanager</artifactId>
-          <version>${aws-java-sdk.version}</version>
+	          <groupId>com.amazonaws</groupId>
+	          <artifactId>aws-java-sdk-secretsmanager</artifactId>
+	          <version>${aws-java-sdk.version}</version>
         </dependency>
+        <dependency>
+					  <groupId>com.amazonaws.secretsmanager</groupId>
+					  <artifactId>aws-secretsmanager-caching-java</artifactId>
+					  <version>1.0.1</version>
+				</dependency>
         <dependency>
             <groupId>backport-util-concurrent</groupId>
             <artifactId>backport-util-concurrent</artifactId>
--- a/src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
+++ b/src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
@@ -61,7 +61,7 @@ public class AwsSecretValidationStep extends AbstractStep {
             awsSecretService.logCacheStatus();
             awsSecretService.clearCache();
             
-            waitFiveSecondsAfterUpdatingSecretAndClearingCacheBeforeRetrieveSecretFromAws();
+            waitForAwsSecretUpdateToFullyPost();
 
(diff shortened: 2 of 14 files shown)


src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
```diff
@@ -0,0 +1,167 @@
...
+                    false, AmazonSecretValidationShared.class);
+            confirmAmazonSecretValidationShared(sharedSpaceSecrets);
+            service.logCacheStatus();
+        } catch (JsonProcessingException e) {
+            LOG.error("validateUsingNewAWSServiceFromSpring, had an error calling AWS Secret Service ", e);
+            throw new RuntimeException(e);
+        }
+    }
+    
+    protected void waitAfterUpdatingSecretBeforeGettingSecret() {
```
cah292, line 153: Is there caching in AWS itself or in the AWS Java libraries that makes it necessary to wait after performing an update? If such a wait is necessary, that could make it more difficult to get Concur to use the AWS Secrets Manager properly.

I see that AWS also has an official Java library to help with AWS secrets caching. Would that library address any of the wait-after-update concerns, or would it be simpler to just go with our custom caching solution?

If we expect this kind of waiting to be more commonplace with the move to AWS secrets, it might be helpful to make this method more accessible to other areas of the code.

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
```diff
@@ -39,6 +45,17 @@
 
     @BeforeEach
     void setUp() throws Exception {
+        org.apache.log4j.LogManager.getLogger(AwsSecretServiceImpl.class).setLevel(org.apache.log4j.Level.DEBUG);
+        
+        /* 
+         * This logging was helpful in debugging things.  we don't want this logging long term, but it would be helpful to 
+         * keep them to be easily recovered
```
cah292, line 52: Is there a better way to structure or explain the helper logging, so that we don't have to keep commented-out code around in our project?

cah292 (commented):
Changes look good overall, though I had some questions about the newest updates in this revised PR.

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
jhulslander: The cache library I found related to secrets manager API is https://github.com/aws/aws-secretsmanager-caching-java ..... but that only works with the 1.x version of the API.  Per recommendation from AWS, we are now using the latest 2.x version of the API.  Were you able to find a cache service with the new API?

In a ticket we opened with Amazon, I asked if there was latency that I should be aware of.   Amazon support responded with "The latency shouldn't be more than a couple of seconds if the secret was properly updated."

In our normal course of business we won't need to do this wait as we will be keeping the value in cache.  I found where if I updated a secret, cleared the cache and then immediately checked the secret value, I wasn't retrieving the the updated value.  Adding a few second wait seems to have resolved the problem.  In normal course of business, we will use cache, so I don't think this wait will be needed.  Should I change this code to clarify the purpose?  Do you think a wait will be needed in a normal business process?

pom.xml
```diff
@@ -952,6 +941,17 @@
             <scope>test</scope>
             <classifier>tests</classifier>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
```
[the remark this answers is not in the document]
jhulslander: I agree, I am surprised this was needed, but I tested again this morning, and I get compile time errors without this change.  I am not sure if there an alternative maven configuration needed here.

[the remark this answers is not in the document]
cah292: It looks like the "kfs-ar" dependency should have had httpclient in compile scope, but since we're referencing "provided" scope for the "kfs-ar" JAR (along with many of the other KFS JARs), perhaps that's why it's not getting pulled in from KFS as expected.

I'm okay with this POM change for now, until we can find a better way to pull this dependency in transitively.

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
cah292: Thanks for the clarification on the caching library. I see now that it only supports the 1.x API like you had said. (It does have some pending PRs to add support for the 2.x API, but I don't know when they will finally move forward.) I'm fine with using our own custom caching in the meantime.

I can't think of any batch jobs yet where we would need to perform a manual wait. With the Concur Event Notification process, I think the read/write of the tokens is handled through the ConcurAccessTokenService, so an extra wait may not be needed as long as our processes all go through this service to read/write the credential data (and where this service would have its own unique AWS secret service instance).

As for the web page that handles manual access token read/write, I'm not sure if it would be worth adding a wait in that spot. The page will need to manually clear the cache prior to at least its initial load; however, as long as there aren't multiple users changing the token at once and as long as we're not using this page around the time of the Concur job, we may not need to add a wait for this page.

If you still plan to keep this wait in place for at least the sample AWS validation job, I would suggest renaming the method to make the intent clearer. The same goes for the related method in the integration test.

jhulslander: I think we need the wait for these test cases, but am having a hard time coming up with a good name, how about waitFiveSecondsAfterUpdatingSecretAndClearingCacheBeforeRetrieveSecretFromAws?

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
jhulslander: I think I found a better way

jhulslander:
This is ready for review again.  I am not sure if I like the new name for the "wait 5 seconds" function, but I can't think of anything better.  If you can think of a better alternative, please let me know.

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
```diff
@@ -0,0 +1,167 @@
...
+                    false, AmazonSecretValidationShared.class);
+            confirmAmazonSecretValidationShared(sharedSpaceSecrets);
+            service.logCacheStatus();
+        } catch (JsonProcessingException e) {
+            LOG.error("validateUsingNewAWSServiceFromSpring, had an error calling AWS Secret Service ", e);
+            throw new RuntimeException(e);
+        }
+    }
+    
+    protected void waitFiveSecondsAfterUpdatingSecretAndClearingCacheBeforeRetrieveSecretFromAws() {
```
cah292, line 153: How about "waitForAwsSecretUpdateToFullyPost" or something along those lines?

jhulslander: that works

jhulslander:
I've made some additional changes, this is ready for review again

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
```diff
@@ -0,0 +1,168 @@
...
+            awsSecretService.logCacheStatus();
+            awsSecretService.clearCache();
+            
+            waitForAwsSecretUpdateToFullyPost();
+
+            AmazonSecretValidationInstance amazonInstanceSpaceSecretsAfterProcessing = awsSecretService.getPojoFromAwsSecret(
+                    AWS_SECRET_NAME_VALIDATION_INSTANCE, true, AmazonSecretValidationInstance.class);
+            confirmInstanceSecretValues(newSecretValues, amazonInstanceSpaceSecretsAfterProcessing);
+
+            LOG.info("excute, The values were retrieved from Amazon Secret Manager as expected");
```
cah292, line 70: Minor typo in log message.

src/main/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImpl.java
```diff
@@ -4,95 +4,169 @@
...
         try {
-            getSecretValueResult = client.getSecretValue(getSecretValueRequest);
-        } catch (SdkClientException e) {
+            GetSecretValueRequest valueRequest = GetSecretValueRequest.builder()
+                    .secretId(fullAwsKey)
+                    .build();
+            GetSecretValueResponse valueResponse = client.getSecretValue(valueRequest);
+            return valueResponse.secretString();
+        } catch (SecretsManagerException e) {
             LOG.error("getSingleStringValueFromAwsSecret, had an error getting value for secret " + fullAwsKey, e);
```
cah292, line 117: I think this log message should be updated to reference the "retrieveSecretFromAws" method instead.

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
```diff
@@ -77,25 +97,79 @@ void testBooleanSetAndGet() {
...
+            LOG.error("waitForAwsSecretUpdateToFullyPost. had an error waiting", e);
+            throw new RuntimeException(e);
+        }
+    }
+    
     @Test
     void testPojo() throws JsonMappingException, JsonProcessingException {
+        testPojoWithCache();
+        awsSecretServiceImpl.clearCache();
+        testPojoWithOutCache();
```
cah292, line 128: The previous versions of this PR had separate test cases for the POJO-with-cache and POJO-without-cache tests. Was there a particular reason why you decided to combine them into one test?

Such a setup also got me to think about another issue. This class's wait-for-posting delay is only being called in the middle of the test-without-cache methods, after making an explicit secret update. Since those methods affect the same secrets as their related test-with-cache methods, and since the test-with-cache ones are not waiting for the secret update to post in AWS, then is another wait-for-posting delay needed at the start of the test-without-cache methods (or end of the test-with-cache methods)? That would help avoid using stale secret values at the start of some of the test-without-secret methods, given that they retrieve the existing secrets first.

cah292 (commented):
I noticed a few more items that may need to be remediated, but other than that the changes look good.

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
jhulslander: @dp65 one of the goals is for the code to "speak for itself" as in comments are needed to clarify  the code ..... the function names will tell the next developer why we are adding a wait

src/main/java/edu/cornell/kfs/sys/aws/batch/AwsSecretValidationStep.java
jhulslander: excute to execute

src/main/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImpl.java
jhulslander: good catch

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
jhulslander: adding the extra wait seems to have avoided the issue that caused me to make this change ..... thanks, good idea

jhulslander:

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
```diff
@@ -78,12 +98,36 @@ void testBooleanSetAndGet() {
     }
     
     @Test
-    void testPojo() throws JsonMappingException, JsonProcessingException {
+    void testBooleanSetAndGetWithOutCache() {
```
cah292, line 101: Since you added the extra wait to the related POJO-without-cache method, should this method also have the extra wait?

cah292 (commented):
There's one quick follow-up item I wanted to check on. Apart from that, the changes should be ready to merge after the unit tests pass.

src/test/java/edu/cornell/kfs/sys/service/impl/AwsSecretServiceImplIntegrationTest.java
jhulslander: can't hurt

cah292: So, if the delay is needed to make the POJO tests work because they use the same secret, then should the delay be added to this secret boolean test as well? The two boolean tests also use the same secret.

jhulslander: I hadn't seen issues with that test in terms of latentency issues, but it would be good to add it for consistency, thanks for noticing.

jhulslander:

cah292 (approved):

PR #1099, opened 2021-03-05
KFSPTS-20884 add Shipping Title to PO cxml

db483:
Note that this was based on the existing ShipPayTerms.


jhulslander (approved):

PR #1112, opened 2021-05-11
KFSPTS-21564: Modify code to use EGAs instead of netids for From email addresses when email sent from KFS. Required for Cornell AppSMTP service change.

nkimble:
Please review this code but DO NOT merge yet.
Item 1, Item 4, and Item 6 as discussed in Description for sub-task KFSPTS-21577 have been addressed by these code changes.

### DIFF
--- a/src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
@@ -1,0 +1,28 @@
+package edu.cornell.kfs.krad.service.impl;
+
+import org.kuali.kfs.krad.service.impl.KualiExceptionIncidentServiceImpl;
+
+public class CuKualiExceptionIncidentServiceImpl extends KualiExceptionIncidentServiceImpl {
+    
+    //Cornell Customization: Base code implemenation override required due to AppSMTP 
+    //                       restriction where From email address must be a registered EGA.
+    @Override
+    protected String getFromAddress() {
+    /* Base code at time of cutomization creation
+     * 
+        Person actualUser = GlobalVariables.getUserSession().getActualPerson();
+
+        String fromEmail = actualUser.getEmailAddress();
+        if (StringUtils.isNotBlank(fromEmail)) {
+            return fromEmail;
+        } else {
+            return this.getMessageTemplate().getFromAddress();
+        }
+     *
+     */
+        
+        // "kr.incident.mailing.list" is set to @incident_email in file kfs-config.properties.erb
+        // @incident_email is then defined in each environment's puppet configuration file
+        return this.getMessageTemplate().getFromAddress();
+    }
+}
--- a/src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
@@ -4,25 +4,14 @@ import org.kuali.kfs.krad.service.impl.KualiExceptionIncidentServiceImpl;
 
 public class CuKualiExceptionIncidentServiceImpl extends KualiExceptionIncidentServiceImpl {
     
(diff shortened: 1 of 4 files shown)


src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.kuali.kfs.krad.service.impl.KualiExceptionIncidentServiceImpl;
+
+public class CuKualiExceptionIncidentServiceImpl extends KualiExceptionIncidentServiceImpl {
+    
+    //Cornell Customization: Base code implemenation override required due to AppSMTP 
+    //                       restriction where From email address must be a registered EGA.
+    @Override
+    protected String getFromAddress() {
+    /* Base code at time of cutomization creation
```
cah292, line 11: Some of these comments should be cleaned up. The original code in commented form can be removed, since we can go to the superclass to view it if needed. It's also not clear about how "kr.incident.mailing.list" relates to this unless you look at the bean file.

In addition, you need to override the appropriate service bean so that it will use this new subclass. (I think "cf.kradExceptionIncidentService" is the bean name.)

src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
```diff
@@ -0,0 +1,28 @@
+package edu.cornell.kfs.krad.service.impl;
+
+import org.kuali.kfs.krad.service.impl.KualiExceptionIncidentServiceImpl;
+
+public class CuKualiExceptionIncidentServiceImpl extends KualiExceptionIncidentServiceImpl {
+    
+    //Cornell Customization: Base code implemenation override required due to AppSMTP
```
cah292, line 7: Minor typo in comment.

cah292 (changes requested):
I found a few things that should be updated, but otherwise the changes look good.

src/main/java/edu/cornell/kfs/krad/service/impl/CuKualiExceptionIncidentServiceImpl.java
nkimble: Comments clean up. Thanks for catching the bean definition.

src/main/resources/edu/cornell/kfs/krad/config/CUKRADSpringBeans.xml
```diff
@@ -72,5 +72,7 @@
 
     <bean id="cf.businessObjectAuthorizationService"
           class="edu.cornell.kfs.kns.service.impl.CuBusinessObjectAuthorizationServiceImpl"/>
+          
+   <bean id="cf.kradExceptionIncidentService" class="edu.cornell.kfs.krad.service.impl.CuKualiExceptionIncidentServiceImpl"/>
```
cah292, line 76: It looks like the base financials version of the "cf.kradExceptionIncidentService" bean does not follow the "parentBean" convention. You may have to include the full XML of the base "cf.kradExceptionIncidentService" bean when you customize it here.

cah292 (changes requested):
I found one more item that needs fixing, but otherwise the changes look good.

src/main/resources/edu/cornell/kfs/krad/config/CUKRADSpringBeans.xml
nkimble: fixed

cah292 (changes requested):
The service bean looks good, but the unit tests are failing because the bean's file doesn't have a namespace entry for the "p" prefix. Please update "CUKRADSpringBeans.xml" to include the appropriate "xmlns:p" entry at the top of the file.

cah292 (commented):
Changes look good, and unit tests pass. I will wait for confirmation that we're ready to move forward with the fixes before merging.

cah292 (approved):
I've received confirmation from Nancy that we are ready to move these changes forward. Merging.

PR #1116, opened 2021-05-20
KFSPTS-21519 make payee ach job transactional so one error does not prevent subsequent documents

jhulslander:
Please note, we were unable to recreate the situation locally.  Adding a transactional annotation is the solution I think will fix this.  If there is something else you think would be worth adding, please let me know.


cah292 (changes requested):
In this case, you'll need more changes than just adding the "Transactional" annotation. There's an example in the RASS module on how to create a separate transaction for routing each new document.

jhulslander:
closing this PR, I will create a new one with the changes needed

PR #1118, opened 2021-06-01
KFSPTS-21744 Remove unneeded feedbackKey attribute

cah292:
KualiCo removed the "feedbackKey" attribute from various tags for the KEW-to-KFS patch. This PR fixes one of our custom tags to also remove that attribute.


db483 (approved):
Changes look good.

PR #1119, opened 2021-06-02
KFSPTS-21705 Fix sequence handling for KEW conversion

cah292:
The base KEW-to-KFS financials patch has some OJB repository files that hard-code MySQL as the database platform to use. This PR overlays the relevant files so that they will use the Oracle platform instead.

Note that our pre-upgrade SQL was using OJB's "Oracle9i" platform, which has a few differences from OJB's regular Oracle platform type. This PR reinstates the use of this "Oracle9i" type, and also adjusts the Kuali-specific "dbPlatform" bean override's setup in an attempt to avoid confusion with the OJB "Oracle9i" type. (Rice/KFS used to have a matching Oracle9iDatabasePlatform class, but it seems to have been removed for KEW-to-KFS since it was merely a subclass of OracleDatabasePlatform with no extra changes.)

### DIFF
--- a/src/main/resources/OJB-repository.xml
+++ b/src/main/resources/OJB-repository.xml
@@ -1,0 +1,38 @@
+<!--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
+    Copyright 2005-2021 Kuali, Inc.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


src/main/resources/OJB-repository.xml
```diff
@@ -0,0 +1,38 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<!--
+    CU Customization: Explicitly change the database platform attribute from "MySQL" to "Oracle9i" instead.
+ -->
+<descriptor-repository version="1.0">
+    <jdbc-connection-descriptor
+            jcd-alias="dataSource"
+            platform="Oracle9i"
```
db483, line 27: Is the Oracle9i referring to the database version?

db483 (changes requested):
Changes look good. I had a question about the platform=Oracle9i. Thank you.

src/main/resources/OJB-repository.xml
db483: Thank you for clarifying this.

db483 (approved):
Changes look good and my question was answered.

PR #1124, opened 2021-06-10
KFSPTS-21770 update httpclient.version to 4.5.13

jhulslander:


cah292 (commented):
Just to confirm, are any dependency exclusions or WAR filtering needed to prevent the old JAR from being added into our built WAR? The base financials code also has the httpclient dependency.

jhulslander:
There are a number of versions of httpclient (4.0.2, 4.3.6, 4.5.5, 4.5.7, 4.5.8, 4.5.9, and now 4.5.13) used by KFS and its dependencies.  I am not sure what exclusions we want to add.  What are your thoughts?

cah292:
When you build the KFS WAR locally from this branch, what version(s) of the httpclient JARs are being included in the WAR? If only the 4.5.13 version is included, then no further changes are needed. If one of the earlier versions is included, then you may need to add exclusions to whatever is pulling in the older version.

jhulslander:
When I go into target/kfs-war-current/WEB-INF/lib I find both httpclient-4.5.13.jar and httpclient-4.5.5.jar.  In Eclipse's dependency hierarchy, it looks like 4.5.5 is omitted by 4.5.13.  I am not sure how to find out where the older 4.5.5 version is coming from.

cah292:
The httpclient 4.5.5 JAR is likely being pulled in from the base financials WAR's "WEB-INF/lib" directory. You could try updating the cu-kfs POM's "maven-war-plugin" configuration to explicitly exclude the JAR, to see if that successfully omits it. There are already some similar JAR exclusions in our setup there, under the section that handles the "kfs-web" overlay.

jhulslander:
Thanks for the details.  I have added the exclusion.  After a rebuild, I only see httpclient-4.5.13.jar in the lib directory. This is ready for review again

cah292 (commented):
The changes look good. I'll merge this once the unit tests pass.

cah292 (approved):

PR #1121, opened 2021-06-07
KFSPTS-21650 do NOT merge this PR, creating to test kew upgrade unit …

jhulslander:
…testing

do NOT merge this PR, just testing

PR #1125, opened 2021-06-11
KFSPTS-21650 Add missing 2021-01-28 changes impacting unit tests

cah292:
There were a few minor non-KEW-related changes from the 2021-01-28 patch, but one of our overlays was not updated to account for that and it started causing our unit tests to fail. This PR fixes the problematic file to correctly account for the 2021-01-28 financials patch changes.

There's still one config-related problem that DevOps needs to deal with for the upgrade PRs, but once that's taken care of, all of the unit tests on this PR should pass. Make sure you open the details on the checks/tests in Jenkins, to verify that the tests indeed passed.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/purap/RequisitionStatuses.java
+++ b/src/main/java/org/kuali/kfs/module/purap/RequisitionStatuses.java
@@ -19,6 +19,7 @@
 package org.kuali.kfs.module.purap;
 
 import java.util.HashMap;
+import java.util.Map;
 
 public final class RequisitionStatuses {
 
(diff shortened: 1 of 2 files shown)


cah292:
The first run was erroneously marked as successful for some reason. There were still several unit tests that failed, but they may have been due to the config issue I noted in the PR #1121 comments. Once that config problem is fixed, we should re-run the unit tests and see if they all pass.

cah292:
After further consultation with Jay, and after further follow-up analysis, the remaining test failures don't appear to be related to a fault in our unit test config property files. Instead, it appears to be the result of a conflict between some of our bean cleanup for the upgrade and some of KualiCo's testing Spring bean files. Some of my initial KEW-to-KFS changes also involved cleaning up our module configuration Spring beans, to make them override the base beans instead of specifying our own independently-named ones. However, with the test-only Spring files that are loaded for unit testing by default, a couple of them have their own overrides of the PURAP and AR module configuration beans, thus wiping out our overrides and causing downstream problems with our unit tests.

I've committed a change to the cu-kfs "institutional-config.properties" file that will override the relevant property and have it exclude the interfering test Spring files. (We don't appear to be actively using the features introduced by those excluded files.) Hopefully this will be enough to get the rest of our unit tests to pass.

jhulslander:
retest this please

jhulslander:
@cah292 the b2b config changes along with the changes in this PR seem to be our answer.  If you look at the details of this run, the unit tests are passing now.  We have an open issue with GitHub not recognizing when jobs fail, but that will be handled on my test PRs

cah292:
@jhulslander Thanks for fixing the config!

db483 (approved):
Changes look good and all unit tests are now passing.

PR #1126, opened 2021-06-15
KFSPTS-21719 Update Location BO conversion rules

cah292:
This PR makes the additional updates needed for converting Location-related maintenance documents for KEW-to-KFS compatibility, and for converting old Rice-side ones into KFS-compatible ones. Note that some of the Location conversion rules were already added by prior user stories.

We used to have custom Cynergy variants of the various Location-related BOs. However, I have not found any previously-created maintenance documents that actually used one of these custom Cynergy Location BOs, so I did not add conversion rules for them in this PR. If you think they should still be added, please let me know.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
+++ b/src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
@@ -239,11 +239,6 @@
                 <match>restricted</match>
                 <replacement/>
             </pattern>
-            <!-- CU-specific pattern for removing a legacy CynergyCountryBo field. -->
-            <pattern>
-                <match>isoCountryCode</match>
-                <replacement/>
-            </pattern>
         </pattern>
         <pattern>
             <class>org.kuali.kfs.kim.impl.permission.GenericPermission</class>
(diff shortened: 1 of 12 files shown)


db483 (approved):
Changes look good. I had one question about the rule on Country class. Thank you.

cah292:
@db483 What was the question you had concerning the changes to the Country conversion rules? I don't see a question posted on this PR.

If your question is related to the handling of the country "restricted" indicators, KualiCo's rules didn't correctly update those and I had to make some adjustments. The old Rice Country BOs had a flag for indicating whether a country was restricted or not, and this flag initially had a different property name in older Rice versions. The KFS version of the Country BO no longer has this flag, so I updated the conversion rules to remove both property variants of the flag. I will add the details on this to KFSPTS-21841 for further follow-up with KualiCo.

If you had other questions about the Country conversion rules or about a different conversion rule, please let me know.

db483:
@cah292 yes, that was my question, if there is a bug in the way KualiCo handled the conversion. Thank you for creating the user story KFSPTS-21841 for reporting this to KualiCo. 
My question was in the review but maybe because I approved the PR it did not keep it? I am not sure why that was lost.

PR #1127, opened 2021-06-15
KFSPTS-21534 Fix email reminder process

cah292:
This PR backports certain parts of KualiCo's FINP-7413 updates. Most of KualiCo's changes revolved around removing Quartz-related code and configuration, which complicates backporting the changes completely. As an alternative to forcibly excluding that code in our builds, this PR instead sets the "email.reminder.lifecycle.enabled" property to false, which is a much simpler way to disable the problematic Quartz-related parts for the 2021-01-28 release.

The batch-step-related parts of FINP-7413 have been backported in this PR, though they only introduced some minor setup improvements. If you think that part of the fix is not worth backporting at this time, please let me know.

Note that performing the Tidal and OpDoc setup for running the daily/weekly email reminder jobs is currently beyond the scope of this PR. We will handle that work in a future user story.

### DIFF
--- a/src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
+++ b/src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
@@ -23,7 +23,8 @@ import org.kuali.kfs.kew.mail.service.ActionListEmailService;
 import java.util.Date;
 
 /**
- * CU Customization: Backported the version of this class from the 2021-04-01 financials patch.
+ * CU Customization: Backported the version of this class from the 2021-04-01 financials patch,
+ * to bring in part of the FINP-7413 changes.
  * 
  * Batch step implementation for the Daily Email
  */
--- a/src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
+++ b/src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
@@ -1,0 +1,42 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2021 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.sys.batch;
+
+import org.kuali.kfs.kew.mail.service.ActionListEmailService;
+
+import java.util.Date;
+
+/**
+ * CU Customization: Backported the version of this class from the 2021-04-01 financials patch.
+ * 
+ * Batch step implementation for the Daily Email
+ */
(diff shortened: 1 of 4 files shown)


src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.sys.batch;
+
+import org.kuali.kfs.kew.mail.service.ActionListEmailService;
+
+import java.util.Date;
+
+/**
+ * CU Customization: Backported the version of this class from the 2021-04-01 financials patch.
```
db483, line 26: Should the FINP number be added as well? Thank you.

src/main/java/org/kuali/kfs/sys/batch/WeeklyEmailStep.java
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.sys.batch;
+
+import org.kuali.kfs.kew.mail.service.ActionListEmailService;
+
+import java.util.Date;
+
+/**
+ * CU Customization: Backported the version of this class from the 2021-04-01 financials patch.
```
db483, line 26: Should the FINP number be added as well? Thank you.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
```diff
@@ -410,4 +410,19 @@
 
 	<bean id="cuApiJsonWebRequestReader" class="edu.cornell.kfs.sys.web.service.impl.CuApiJsonWebRequestReaderImpl" />
 
+    <!--
+        Override of daily and weekly email reminder batch steps to backport changes from the 2021-04-01 patch.
+        These two bean overrides should be removed when we upgrade to the 2021-04-01 patch or later.
```
db483, line 415: Should the FINP number be added as well? Thank you.

src/main/resources/institutional-config.properties
```diff
@@ -15,6 +15,7 @@ dailyEmail.cronExpression=0 0 1 * * ?
 weeklyEmail.cronExpression=0 0 2 ? * 2
 periodic.thread.dump=false
 xml.pipeline.lifecycle.enabled=false
+email.reminder.lifecycle.enabled=false
```
db483, line 18: It would be good to add a comment about why this was set to false and that it should be removed when we upgrade to 4/1/21 version of financials.

db483 (changes requested):
Changes look good but I had some small suggestions for improvement.
Please create a user story in the KEW backlog for setting up the new batch jobs in Tidal. Thank you.

src/main/java/org/kuali/kfs/sys/batch/DailyEmailStep.java
cah292: Will update comment.

src/main/java/org/kuali/kfs/sys/batch/WeeklyEmailStep.java
cah292: Will update comment.

src/main/resources/edu/cornell/kfs/sys/cu-spring-sys.xml
cah292: Will update comment.

src/main/resources/institutional-config.properties
cah292: Ok, I'll add a comment accordingly.

cah292:
I've updated the comments as suggested. Also, yesterday I created the KFSPTS-21875 follow-up ticket to get the email reminder jobs runnable through Tidal.

db483 (approved):

PR #1129, opened 2021-06-21
KFSPTS-21428 back-port FINP 7310

jhulslander:
I opted to back-port this change as an overlay as there was enough private elements in PaymentApplicationAdjustmentDocument such that it was starting to get confusing to add this change in CuPaymentApplicationAdjustmentDocument.

When reviewing this, the only changes are in the functions getTotalDollarAmount() and anAppaIsBeingAdjusted().  The rest of the file is base code


cah292 (approved):
Changes look good, and unit tests pass. (The unit tests ran successfully prior to today's DevOps maintenance of the PR unit test process.) Merging.

PR #1133, opened 2021-06-22
KFSPTS-21430 back-port FINP-7396

jhulslander:
I brought in ContractsGrantsPaymentHistoryReportLookupableHelperServiceImpl as a new overlay item, then back-ported FINP-7396


cah292 (commented):
FINP-7396 also included a change that replaced the use of ArPropertyConstants.RANGE_LOWER_BOUND_KEY_PREFIX with the use of KFSPropertyConstants.RANGE_LOWER_BOUND_KEY_PREFIX instead. (Both constants appear to have the same value, so it shouldn't matter which one we go with.) Should that change be included as well, or should that wait until we move to the actual 2021-03-11 version of the file?

jhulslander:
I originally opted to not make that switch because they are the same value, but I think it would be good to switch from ArPropertyConstants to KFSPropertyConstants .... I will update this PR accordingly.  Thanks for noting

jhulslander:

cah292 (approved):

PR #1139, opened 2021-06-23
KFSPTS-21723: Add XML conversion rules for rSmart related class names in Security Model and Security Definition maintenance documents.

nkimble:


cah292 (approved):
Changes look good, and unit tests pass. Also, given Nancy's comments on the user story's review sub-task, I agree that these conversion changes are minor enough that we don't need new test cases for them. We can add unit test cases later in a future PR if it ends up being bigger than we had originally thought.

Merging.

PR #1135, opened 2021-06-22
KFSPTS-21720 Fix doc type and parameter XML conversion

cah292:
This PR updates the maintenance XML conversion rules to properly handle older Parameter and Document Type maintenance documents. Most of the needed changes were already added by previous commits, but this PR had to add a few more changes to convert some classnames and to remove some unexpected JPA-related elements.

PR #1142, opened 2021-06-30
KFSPTS-21724 Fix vendor XML conversion

cah292:
This PR fixes the maintenance XML conversion rules for Vendor documents.

Our prior conversion rules were moving most of the vendor header extension fields to the parent, and also removed the Locale field. However, several years ago we ended up adding this extension object back in, and kept its Locale field. To prevent the conversion rules from removing the Locale field while still allowing them to properly move the other fields, this PR adds some new conversion rule setup that allows for picking and choosing which fields to move to the parent element.

Some corresponding test cases have been added as well, which have been altered accordingly to avoid exposing existing vendor data in the new unit test cases, and which have altered any confidential fields to contain dummy data instead.


cah292:
NOTE: These changes may or may not conflict with the #1135 changes. Be sure to check for merge conflicts and re-trigger the unit tests when one gets merged before the other.

cah292:
retest this please

cah292:
retest this please

nkimble (approved):
Changes are appropriate. Unit tests have passed. Merging to kew upgrade branch.

PR #1146, opened 2021-07-12
KFSPTS-21967 Update 2021-07-12 rebased code for KEW-to-KFS compatibility

cah292:
Some follow-up code changes were needed as a result of today's rebase of the kfs-kew-upgrade branch. I will post details on the changes and the resolved merge conflicts on the KFSPTS-21967 user story.

NOTE: This PR merges into the branch used for the rebasing efforts, rather than into kfs-kew-upgrade directly.

The unit tests should run with the KEW upgrade setup, but if that ends up not being the case then we'll need to make adjustments.


nkimble (changes requested):
Changes in the PR appear appropriate for the documented reasons in the JIRA.

There may be an additional change needed.
File “cu-spring-ar.xml” still references the removed class “edu.cornell.kfs.module.ar.batch.service.impl.CuVerifyBillingFrequencyServiceImpl”
Should that spring file be updated to remove that reference as well?

cah292:
@nkimble I'm not seeing the CuVerifyBillingFrequencyServiceImpl in the "cu-spring-ar.xml" file. Can you please check again?

Remember that this PR is requesting a merge into the "kfs-kew-rebase-2021-07-12" branch, so please look at that branch's version of the file.

nkimble (approved):
I had the correct branch but for some reason that particular XML file had not sync'ed correctly within eclipse after checkout. You are correct. That reference is not present.
My concerns have been addressed. Merging.

PR #1147, opened 2021-07-14
KFSPTS-21559 Add KFS version of ClamAV scanning

cah292:
This PR adds an implementation of the AntiVirusService that integrates with ClamAV to perform the file scans. It takes the code from Cynergy's ClamScanResult and ClamAVService classes, and updates it to conform to our current coding standards and to make various other improvements. (These two original classes can be found in the cornell_customizations project.)

In addition, I've reworked the package structure of the existing antivirus-related classes in the KEW upgrade code, to put them under the KNS/KRAD area instead.

The Cynergy version included the ability to invoke the ClamAV "ping" and "stats" commands, though they don't seem to have been actively used by our code. I've included them in this PR for now, but if you think we don't need to call them, please let me know so that I can remove them. (I also adjusted the "stats" invocation to use the same command start/end setup as the other commands for simplicity and consistency.)

I was able to add some unit tests to validate the new code, which create a mock ClamAV process to handle the needed Socket and to read/write data accordingly. The tests also rely on some existing test files from the CU Attachment Service unit tests, as well as some new test files to validate other cases.


src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
```diff
@@ -0,0 +1,217 @@
...
+                        ClamAVDelimiters.NULL_SUFFIX, fullResponse.length() - chunkLength);
+            }
+        } while (responseTerminatorIndex == -1 && chunkLength > 0);
+        
+        if (responseTerminatorIndex == -1) {
+            LOG.error("readClamAVResponse, End of stream was reached before reading full ClamAV response!  " +
+                    "No content will be returned.  " + fullResponse.length() +
+                    " characters were received prior to abrupt end.");
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("readClamAVResponse, Response text received prior to stream end: " + fullResponse);
```
jhulslander, line 114: should this be a debug log.  Would this be helpful in all cases?

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
```diff
@@ -0,0 +1,217 @@
...
+            }
+            return KFSConstants.EMPTY_STRING;
+        } else if (responseTerminatorIndex < fullResponse.length() - 1) {
+            int extraContentLength = fullResponse.length() - responseTerminatorIndex - 1;
+            LOG.warn("readClamAVResponse, ClamAV sent at least an extra " + extraContentLength +
+                    " characters of response data beyond the delimiter!  The extra characters will be ignored.");
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("readClamAVResponse, Response text received beyond end-of-content delimiter: " +
```
jhulslander, line 122: similar concern here, would this be helpful in all cases?

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
```diff
@@ -0,0 +1,217 @@
...
+        return responsePrecedingTerminator;
+    }
+
+    @Override
+    public ClamAVScanResult scan(byte[] fileContents) throws IOException {
+        try (
+            InputStream preLoadedFileStream = new ByteArrayInputStream(fileContents);
+        ) {
+            return scan(preLoadedFileStream);
+        }
```
jhulslander, line 137: no catch here?

jhulslander (changes requested):
This change looks good for the most part.  I found a couple of things to consider.  Why did you change the class path from antivirus to krad?  I think it makes more sense to be under an antivirus path.  
I remember adding change to the antivirus code to log file names (among other details) when an infected file was found.  I just want to make sure that hasn't been lost.

jhulslander:
The log message I was referring to
ERROR edu.cornell.kfs.krad.service.impl.CuAttachmentServiceImpl :: createAttachment, virus protection failure!  uploadedFileName: eicar.com.txt, mimeType: text/plain, attachmentTypeCode: null, scan result: FAILED, parent object: org.kuali.kfs.sys.businessobject.FinancialSystemDocumentHeader@670d7b55[financialDocumentTotalAmount=0.00,correctedByDocumentId=<null>,financialDocumentInErrorNumber=<null>,financialDocumentStatusCode=?,workflowDocumentStatusCode=S,applicationDocumentStatus=<null>,initiatorPrincipalId=1012585,workflowDocumentTypeName=DV,workflowCreateDate=2021-07-16 13:59:58.645,processedDate=<null>,documentNumber=37738464,documentDescription=Jay test dv,organizationDocumentNumber=<null>,documentTemplateNumber=<null>,explanation=<null>,workflowDocument=org.kuali.rice.kew.impl.document.WorkflowDocumentImpl@12590ff6,versionNumber=2,objectId=6088b795-0026-47b0-9027-554996712d64,lastUpdatedTimestamp=2021-07-16 14:00:49.323,newCollectionRecord=false]
Jul 16 14:01:20 kfs-dev-i-0158094eb3d8b4a94 tc =2021-07-16 14:01:20,104 [http-nio-8080-exec-72] u:ccs1/d: FATAL org.kuali.rice.core.framework.persistence.jta.KualiTransactionInterceptor :: Exception caught by Transaction Interceptor, this will cause a rollback at the end of the transaction.

src/main/java/edu/cornell/kfs/krad/CUKRADConstants.java
```diff
@@ -0,0 +1,27 @@
...
+        public static final byte Z_PREFIX_BYTE = 0x7A;
+        public static final byte NULL_SUFFIX_BYTE = 0;
+    }
+
+    public static final class ClamAVResponses {
+        public static final String STREAM_PREFIX = "stream:";
+        public static final String FOUND_SUFFIX = "FOUND";
+        public static final String ERROR_SUFFIX = "ERROR";
+        public static final String RESPONSE_OK = STREAM_PREFIX + " OK";
+        public static final String PING_RESPONSE_OK = "PONG";
```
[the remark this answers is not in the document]
cah292: This is the response that ClamAV is expected to return from a successful run of its "ping" command. If you think the constant name is still not clear enough, I'll see if I can find a better name for it.

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
cah292: Under further consideration, I agree that it makes sense to have this content visible in all cases and just debug-logging ones. I will update this logging accordingly.

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
cah292: I will update this to use warn or error logging instead of debug logging.

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
cah292: The scan() method that accepts a byte array already had "throws IOException" in its method signature, so I thought that it would not be necessary to perform a catch here. Or, if it's just unclear as to why this section doesn't have a "catch" or "finally" block at all, it's because I had set it up as a try-with-resources block, which can allow for omitting an explicit catch/finally block afterwards.

cah292:
The log message that you referred to should still work properly. That logging section of CuAttachmentServiceImpl was not altered by this PR.

In regards to the package name changes, The antivirus scanning only seems to be used by our customized version of the KNS/KRAD AttachmentService, so I thought it would make sense to move it into the CU KRAD packages. However, I agree that having "antivirus" in the package path would be good for keeping this functionality grouped separately from the rest of KNS/KRAD, so I will look into using a "krad.antivirus" sub-package for the appropriate classes.

jhulslander:
krad.antivirus works

src/main/java/edu/cornell/kfs/krad/CUKRADConstants.java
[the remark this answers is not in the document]
jhulslander: I think given the situation, the variable name is OK ..... I had just assumed a typo.  I am not sure what is better.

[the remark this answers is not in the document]
cah292: I'm thinking of renaming the variable to "RESPONSE_PING_SUCCESS" instead. Does that sound okay?

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
jhulslander: It was unclear why there was not catch or finally block I don't often see a standalone try block

src/main/java/edu/cornell/kfs/krad/CUKRADConstants.java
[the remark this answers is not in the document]
jhulslander: yeah, that would be OK

src/main/java/edu/cornell/kfs/krad/service/impl/ClamAVAntiVirusServiceImpl.java
cah292: I'll add a "catch" block here, to avoid confusion and to also provide a spot for logging the exception before re-throwing it.

cah292:
I've committed some changes based on the PR feedback. This is ready for review again.

src/main/java/edu/cornell/kfs/krad/antivirus/ScanResult.java
```diff
@@ -1,4 +1,4 @@
-package edu.cornell.kfs.krad.service;
+package edu.cornell.kfs.krad.antivirus;
```
jhulslander, line 1: should this be edu.cornell.kfs.krad.antivirus.service?

src/main/java/edu/cornell/kfs/krad/antivirus/service/impl/ClamAVAntiVirusServiceImpl.java
```diff
@@ -110,18 +110,14 @@ private String readClamAVResponse(Reader socketReader) throws IOException {
             LOG.error("readClamAVResponse, End of stream was reached before reading full ClamAV response!  " +
```
jhulslander, line 110: since the full response is logged below, is this log statement needed?  Could it be combined?

jhulslander (changes requested):
changes look good, I found a few small things that could be cleaned up

src/main/java/edu/cornell/kfs/krad/antivirus/ScanResult.java
cah292: I wasn't sure whether this enum should be in the "service" package or not, but I can move it there to reduce confusion.

src/main/java/edu/cornell/kfs/krad/antivirus/service/impl/ClamAVAntiVirusServiceImpl.java
cah292: Ok, I'll combine the messages, and will do the same for the "else if" block below.

cah292:
I've committed some more changes based on the latest PR feedback. This is ready for review again.

jhulslander (approved):
Code changes look good, my concerns have been addressed, I shall merge

PR #1149, opened 2021-07-20
KFSPTS-21725 Fix base COA XML conversion

cah292:
This PR updates the maintenance XML conversion process (plus some data dictionary beans) so that older COA module maintenance documents can be successfully opened. The main user story has a comment with a summary of which documents were broken and what fixes were implemented.

Note that this PR only works with base financials COA documents (or CU-customized versions of base COA documents). The conversion of CU-specific COA documents will be handled in a separate user story.


nkimble:
retest this please

cah292:
I've committed some changes to fix the GDLG document issue that had been mentioned in the user story. This is now ready for review again.

nkimble (approved):
Concerns raised in corresponding JIRA user story have been addressed. Merging.

PR #1150, opened 2021-07-21
KFSPTS-22019 CU custom fix for FINP-7572

jhulslander:
Please start to review this change, but hold off on merging this code till we get some feedback.

There are a number of rules that determine if the "Error Correct" action is available on the CINV document or not.  The current KualiCo rule prevents CINV from being error corrected if the CINV has APP or APPA that are marked as not adjusted.  Janet believes this is incorrect.  A CINV document  should be able to be error corrected, if no payments have been applied, or if all the payments have been adjusted to zero.
Janet showed me a report in base code called "Outstanding Invoice Report."  I looked at how "Payment Applied" is calculated, and made use of the same service call in this code change.


jhulslander:
Janet thinks my work around may work for now.  I am going to close this PR for now

PR #1151, opened 2021-07-23
KFSPTS-21726 Fix acct reversion XML conversion

cah292:
This PR fixes the XML conversion rules pertaining to Account Reversion Global documents. These appear to be the only Cornell-specific COA documents requiring further XML conversion changes that are not already covered by the existing rules.


nkimble (approved):
Changes look appropriate. Merging.

PR #1152, opened 2021-07-26
KFSPTS-22142 Add local customization to fix Role lookup from a child/collection object on a maintenance doc

db483:


cah292 (approved):
Changes look good, and unit tests pass. Also, given that the superclass already treats the fieldValues map as mutable, I'm okay with this custom subclass treating it as mutable, and we can make follow-up changes in another user story if it ends up becoming a problem.

Merging.

PR #1157, opened 2021-08-04
KFSPTS-21729 Fix CG XML conversion

cah292:
This PR fixes the maintenance XML conversion rules for the relevant CG module documents. The related user story has more details on which documents needed fixes and why.


src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
```diff
@@ -462,6 +462,34 @@
                 <replacement>nonresidentReview</replacement>
             </pattern>
         </pattern>
+        <pattern>
+            <class>org.kuali.kfs.module.cg.businessobject.AwardProjectDirector</class>
+            <pattern>
+                <match>projectDirector</match>
```
db483, line 468: Why did these person reference objects need a rule to be removed?

db483 (changes requested):
I had a question about the updates. Thank you.

src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
cah292: For the document types that needed fixing, the more recent documents lacked "class" attributes on these elements/fields that would have spelled out one of the Person or PersonImpl classnames. As a result, the usual conversion rules that match on the Person or PersonImpl classnames did not identify these elements as needing conversion. That's why it was necessary to add new conversion rules like this one.

KualiCo's KIM-to-KFS conversion appears to have been what caused the more recent documents to stop displaying the Person(Impl) "class" attributes for the more recent documents. As part of that effort, KualiCo modified many of the Person references on the business objects so that they would use the PersonImpl class directly, instead of using the Person interface. This likely caused XStream's object-to-XML conversion to omit the "class" attribute, since it could derive the exact PersonImpl type directly from the BO property's type.

db483: Thank you for explaining this Chad. Do we have a note about this in the KEW upgrade confluence documentation? I looked for that this morning but could not find it.

cah292: I don't believe this has been noted in the KEW upgrade Confluence documentation yet. And just to clarify, it was the KIM-to-KFS upgrade (not the KEW-to-KFS upgrade) that made the Person-to-PersonImpl changes that I had described above. Those changes don't seem to be causing problems in DEV/TEST/PROD; they only seemed to start causing problems with the KEW upgrade, since the Person-related elements on the pre-KEW docs have sub-elements that reference Rice classnames.

db483: Thank you Chad. Should we note this on the KEW upgrade conversion documentation?

cah292: I've updated the following Confluence page with the details (down at the bottom of the page): https://confluence.cornell.edu/x/6_XQFw

db483: Thank you very much Chad, the updated documentation looks good.

db483 (approved):
Changes look good.

PR #1158, opened 2021-08-05
KFSPTS-21731 Fix LD XML conversion

cah292:
This PR adds some minor XML conversion changes to fix attempts to open old LD maintenance docs. In the case of the LD module, the only extra thing needed was a minor property name change on the LaborBenefitRateCategory objects.


nkimble (changes requested):
The changes included with this commit look good. 
Waiting for additional fixes for issues found with LD BCAL edocs before moving forward.

cah292:
I've committed some changes to fix the conversion of legacy LD BCAL documents. This is ready for review again.

nkimble (approved):
Additional changes for BCAL documents look good and fix the reported issue. Thank you for addressing that problem. 
All changes for this PR are appropriate. Merging.

PR #1159, opened 2021-08-10
KFSPTS-21968 Fix reference attribute conversion

cah292:
This PR updates the maintenance XML conversion process so that the XML element paths within "reference" attributes will also be converted. That way, if an XML element is renamed or has certain special conversions applied (like the move-to-parent conversion), the updated "reference" path should still lead to the intended value. Without converting the reference paths that need updates, XStream will fail when converting the XML into Java objects.

Some additional conversion rules have been added to address the "reference" issue and a few other issues with at least two docs that are known to be affected: Security Model and Asset Global.


db483 (approved):
Changes look good.

PR #1161, opened 2021-08-13
KFSPTS-22337: Remove DAO OJB references to cynergy schema and switch references from KRCL_ cynergy tables to SH_ KFS tables.

nkimble:


db483 (approved):
Changes look good.

PR #1174, opened 2021-09-03
KFSPTS-21560 Update CU RouteHeaderService for KEW upgrade

cah292:
This PR migrates the needed portions of our old Cynergy RouteHeaderService customization to KFS, and updates them for KEW-to-KFS compatibility. Only the custom "getFinalizedDatesForDocumentType" service method needed to be kept and remediated, since the 1099/1042S tax batch job relies upon it.

The old version of the method in CynergyRouteHeaderServiceImpl used JPA and the Kuali Criteria API to retrieve the desired results. Since KFS still uses OJB and has removed the Rice JPA-related code for the KEW-to-KFS upgrade, this PR reimplements the method to use an OJB DAO and OJB report queries instead.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/util/CuOjbUtils.java
+++ b/src/main/java/edu/cornell/kfs/sys/util/CuOjbUtils.java
@@ -3,7 +3,7 @@ package edu.cornell.kfs.sys.util;
 import java.util.Iterator;
 import java.util.Objects;
 import java.util.Spliterator;
-import java.util.function.Consumer;
+import java.util.Spliterators;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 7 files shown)


src/main/java/edu/cornell/kfs/sys/util/CuOjbUtils.java
```diff
@@ -0,0 +1,64 @@
...
+        return buildStreamForQueryResults(Object[].class, reportQueryIterator);
+    }
+
+    public static <T> Stream<T> buildStreamForQueryResults(Class<T> queryResultType, Iterator<?> queryIterator) {
+        Objects.requireNonNull(queryResultType, "queryResultType cannot be null");
+        Objects.requireNonNull(queryIterator, "queryIterator cannot be null");
+        return StreamSupport.stream(() -> new QuerySpliterator<>(queryResultType, queryIterator), 0, false);
+    }
+
+    private static class QuerySpliterator<T> implements Spliterator<T> {
```
db483, line 22: What is the main reason for using a Spliterator here? Are you trying to improve a performance issue?

db483 (changes requested):
Changes look good but I am trying to understand the reasoning behind the decision to use a Spliterator for this implementation. Thank you.

src/main/java/edu/cornell/kfs/sys/util/CuOjbUtils.java
cah292: In the DAO class, I was trying to use a Stream to convert the query results into a Map. The results are returned as an Iterator, though, and I haven't found any Java methods to directly convert an Iterator to a Stream. That's why I wrote this Spliterator implementation.

As an alternative, I found some Java examples online that use a lambda expression to create an Iterable object from an Iterator, and that lambda-created Iterable can then be used to create the needed Spliterator for integrating with Streams. I could update this code to switch over to that, so that we don't need to write the Spliterator ourselves.

db483: Thank you for clarifying this Chad. It sounds like the new example you have found may simplify the implementation.

cah292: Ok, I'll rewrite this to use that simplification. Thanks for the input.

cah292: After further analysis, I noticed that Java has some helper methods for wrapping an Iterator as a Spliterator (which can then be converted into a Stream). I'll update the code to use this strategy.

cah292:
I've committed some changes to simplify the new CuOjbUtils class, so that it now makes use of Spliterator helper methods instead of creating a custom Spliterator implementation.

db483 (approved):
Changes look good.

PR #1175, opened 2021-09-03
KFSPTS-21006 Add KEW email changes to KFS

cah292:
NOTE: There are two PRs for this change: One in cu-kfs and one in nonprod-sql. Please make sure both are good to go before merging.

NOTE: Please do NOT merge this until I comment otherwise. I still need to get final confirmation on some of the email wording from the functionals.

This PR moves the relevant Cynergy action list email customizations into KFS:

* Overlayed the email XML stylesheet so that the email layout and wording is similar to what we have for the current Cynergy Production action list emails. (Note that KEW-upgraded KFS only allows for a single hard-coded email stylesheet, and no longer uses the DB-stored stylesheets that were available in Rice/Cynergy.)
* Customized the ActionListEmailService so that Production-like email behavior can be enabled in non-Production environments.
  * Note that this PROD-like setup is independent of the email mode setting in the regular EmailService. Because KEW-upgraded KFS uses the regular EmailService to send action list emails, it's still possible to force such emails to be written to the logs or sent to a test address, even if the action list email service is in forced-Production mode.
  * Also note that the property for controlling this action list email service behavior will be a little different than the equivalent Cynergy one. The old Cynergy property is named "cynergy.sendProductionEmail" and is meant to be a true/false value. The new cu-kfs property is named "cu.actionlist.email.environment" and is meant to be an environment code. (In the cu-kfs default properties file, this property will be set to the current environment code by default.)


src/main/java/edu/cornell/kfs/kew/mail/service/impl/CuActionListEmailServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+import org.kuali.kfs.kew.mail.service.impl.CustomizableActionListEmailServiceImpl;
+
+public class CuActionListEmailServiceImpl extends CustomizableActionListEmailServiceImpl {
+
+    @Override
+    protected boolean isProduction() {
+        return super.isProduction() || shouldForceProductionBehavior();
+    }
+
+    protected boolean shouldForceProductionBehavior() {
```
db483, line 14: Are we going to need to set the cu.actionlist.email.environment to prod in a non-prod env to get prod like behaviour? Why do we need prod like behavior in non-prod if this service is using the regular email service? Thank you.

db483 (changes requested):
I have a question about this implementation. Thank you.

src/main/java/edu/cornell/kfs/kew/mail/service/impl/CuActionListEmailServiceImpl.java
cah292: The two services behave a little differently for their Production modes:

* With the regular email service, it uses a KFS parameter to control whether to log the email, send to a test address, or send to the actual address. (Note that this parameter is ignored when running in Production.)
* With the action list email service and its related email content service, they rely on config properties to control whether the action list emails should be sent to the actual recipient or a test address, as well as whether to include extra debug-related content in the email body. (Note that the test address for this service uses a different parameter than the one used by the regular email service.)
  * The setup of the debug email content and workflow test address take place BEFORE the email is passed to the regular email service for processing.

This PR updates the setup for the action list and email content services, so that property-configured Production-style behavior can be enabled for both of them in a non-Production environment if necessary. That way, we can use a non-Production environment to test out what the email structure would look like in a Production-like setup.

db483: What values will we use for cu.actionlist.email.environment ?

cah292: By default, this PR sets up the "cu.actionlist.email.environment" property so that it equals the value of the "environment" property. If we wanted to force this setting to the PROD value in certain environments, we could have DevOps override the property to explicitly equal "kfs-prod".

db483: I find the use of this property to force prod like behavior is potentially confusing. Since we are adding custom code to force production like behavior should we use a more explicit way? Would a system param be better? The system parameter description could also better explain what the "production like" behavior is.

cah292: The old Cynergy code was using a config property to set up the Production-like behavior here. However, since we're changing the feature for KEW-to-KFS compatibility anyway, I agree that it makes more sense to control this setting via a parameter instead. I will work on implementing that change, and will also resolve the merge conflict that is being reported now on this PR.

cah292:
I've updated the code to use a parameter for configuring the workflow email mode, and also updated the email stylesheet to replace the help email with the help URL that Kris had mentioned. For the moment, the help email is only on the daily/weekly emails for consistency with the related Cynergy PROD setup, though it could easily be added to the immediate emails if needed.

This is ready for review again, and can be merged once this PR and the nonprod-sql PR have been approved.

src/main/java/edu/cornell/kfs/kew/mail/service/impl/CuEmailContentServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+        updateEmailEnvironmentIfNotRunningInProduction();
+        return super.generateDailyReminder(user, actionItems);
+    }
+
+    private void updateEmailEnvironmentIfNotRunningInProduction() {
+        if (isProductionEnvironment()) {
+            return;
+        }
+        if (isProductionWorkflowEmailModeEnabled()) {
+            setWorkflowEmailDeploymentEnvironmentFromProperty(Config.PROD_ENVIRONMENT_CODE);
```
db483, line 53: Does this also need to explicitly set the environment for 'kfs-prod' when running 'production mode' in a test environment?

cah292: Without this extra setup in the email content service, all non-Production environments will add some debug-related text to the email body. I wasn't sure whether we wanted to allow for suppressing that extra email text in a non-PROD environment.

Upon further consideration, I could update this code to avoid overwriting the setting on the content service (thus forcing non-PROD environments to always display that extra text in the email), and I could also customize that added email section to indicate what environment the email was sent from. Thoughts?

db483: Chad, let's chat after the standup. I was hoping the system parameter would make things clearer but did not realize it would also require the extra customizations. I thought the isProduction method was going to be used to determine behavior based on prod/non-prod env.

cah292: After further discussion with Daniela, I will remove this extra CuEmailContentServiceImpl code so that non-PROD workflow emails will always display the extra debug content in the email text. This will bring the customization more in line with how it behaves in current Cynergy code, except that we will now allow the behavior to be configured via a parameter instead of a config property.

db483: Thank you very much Chad!

cah292:
I've committed some more changes based on the discussion I had with Daniela earlier today. I also updated the email stylesheet to make the debug-related sections show the environment that sent the emails, and to also show debug sections in the daily/weekly reminder emails.

Note that cu-spring-kew.xml is still overriding the email content service bean, to fix an issue where base code was not injecting the environment code into the service. I created KFSPTS-22495 last week to get this bug reported back to KualiCo.

db483 (approved):
Changes look good.

PR #1176, opened 2021-09-14
KFSPTS-20966 Cleanup for Doc Requeuer Job

cah292:
This PR implements some extra minor revisions to our custom Document Requeuer Job. The bulk of this job's remediation was already dealt with in the KFSPTS-21563 ticket, so this PR just cleans up a few more items:

* Updated one of the main SQL sections to use a sub-query that selects a smaller number of results, which also required adding more table aliases to the SQL accordingly.
* Made some minor tweaks to the printing of ActionItemNoteDetailDto objects, to make their toString() values more readable in the application logs.
* Cleaned up the Spring configuration for the job's main service, such as by injecting the required services into the bean.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
+++ b/src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
@@ -37,15 +37,14 @@ public class DocumentMaintenanceDaoJdbc extends PlatformAwareDaoBaseJdbc impleme
     }
 
     private CuSqlQuery buildRequeueSqlQuery() {
-        CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk();
-        sqlChunk.append(" ORDER BY DH.DOC_HDR_ID ASC");
+        CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk(true);
         return sqlChunk.toQuery();
     }
 
-    private CuSqlChunk buildRequeueSqlQueryChunk() {
+    private CuSqlChunk buildRequeueSqlQueryChunk(boolean includeOrderByClause) {
         Collection<String> docTypeIds = findNonRequeueableDocumentTypes();
         Collection<String> roleIds = findRequeueableRoleIds();
-        return CuSqlChunk.of(
+        CuSqlChunk subQuery = CuSqlChunk.of(
                 "SELECT DH.DOC_HDR_ID FROM KFS.KREW_DOC_HDR_T DH ",
                 "WHERE DH.DOC_HDR_STAT_CD = ", CuSqlChunk.forParameter(KewApiConstants.ROUTE_HEADER_ENROUTE_CD),
                 " AND DH.DOC_TYP_ID NOT IN (", CuSqlChunk.forStringParameters(docTypeIds), ")",
@@ -55,6 +54,10 @@ public class DocumentMaintenanceDaoJdbc extends PlatformAwareDaoBaseJdbc impleme
                         "AND RQ.RSP_ID IN (",
                                 "SELECT RR.RSP_ID FROM KFS.KRIM_ROLE_RSP_T RR ",
                                 "WHERE RR.ROLE_ID IN (", CuSqlChunk.forStringParameters(roleIds), ")))");
+        if (includeOrderByClause) {
+            subQuery.append(" ORDER BY DH.DOC_HDR_ID ASC");
+        }
+        return subQuery;
     }
 
     @Override
@@ -77,7 +80,7 @@ public class DocumentMaintenanceDaoJdbc extends PlatformAwareDaoBaseJdbc impleme
                 "SELECT AI.PRNCPL_ID, AI.DOC_HDR_ID, AIE.ACTN_NOTE, AIE.LAST_UPDT_TS, AI.ACTN_ITM_ID ",
                 "FROM KFS.KREW_ACTN_ITM_T AI ",
                 "JOIN KFS.KREW_ACTN_ITM_EXT_T AIE ON AI.ACTN_ITM_ID = AIE.ACTN_ITM_ID ",
-                "WHERE AI.DOC_HDR_ID IN (", buildRequeueSqlQueryChunk(), ")");
+                "WHERE AI.DOC_HDR_ID IN (", buildRequeueSqlQueryChunk(false), ")");
     }
 
     private <T> List<T> queryForValues(CuSqlQuery sqlQuery, RowMapper<T> rowMapper) {
--- a/src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
+++ b/src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
@@ -38,7 +38,7 @@ public class DocumentMaintenanceDaoJdbc extends PlatformAwareDaoBaseJdbc impleme
 
     private CuSqlQuery buildRequeueSqlQuery() {
         CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk();
-        sqlChunk.append(" ORDER BY DOC_HDR_ID ASC");
+        sqlChunk.append(" ORDER BY DH.DOC_HDR_ID ASC");
         return sqlChunk.toQuery();
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
```diff
@@ -38,22 +38,23 @@
 
     private CuSqlQuery buildRequeueSqlQuery() {
         CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk();
-        sqlChunk.append(" ORDER BY DOC_HDR_ID ASC");
+        sqlChunk.append(" ORDER BY DH.DOC_HDR_ID ASC");
```
nkimble, line 41: All of the other methods that have a table alias defined keep that alias in the scope of the method. Even though this method is private to this class, I am concerned about the table alias being outside of the method where is is defined.  
What are your thoughts about the use of a boolean parameter "includeOrderBy" to the buildRequeueSqlQueryChunk method to keep this SQL within the method scope of the table alias definition which would also allow the buildRequeueSqlQueryChunk to be reusable?

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -46,8 +48,7 @@ public boolean requeueDocuments() {
     @Transactional
     private void requeueDocumentByDocumentId(List<ActionItemNoteDetailDto> noteDetails, String docId) {
         LOG.info("requeueDocumentByDocumentId: Requesting requeue for document: " + docId);
-        DocumentRefreshQueue documentRequeuer = (DocumentRefreshQueue) SpringContext.getService(CUKFSConstants.DOCUMENT_REFRESH_QUEUE_SERVICE_SPRING_CONTEXT_NAME);
-        documentRequeuer.refreshDocument(docId);
+        documentRefreshQueue.refreshDocument(docId);
```
nkimble, line 51: It looks like DOCUMENT_REFRESH_QUEUE_SERVICE_SPRING_CONTEXT_NAME is not longer used in the code base. Should this constant also be removed from CUKFSConstants?

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
```diff
@@ -107,4 +108,12 @@ public void setDocumentMaintenanceDao(DocumentMaintenanceDao documentMaintenance
...
     }
 
+    public void setDocumentRefreshQueue(DocumentRefreshQueue documentRefreshQueue) {
+        this.documentRefreshQueue = documentRefreshQueue;
+    }
+
+    public void setBusinessObjectService(BusinessObjectService businessObjectService) {
+        this.businessObjectService = businessObjectService;
+    }
+
```
nkimble, line 118: Shouldn't these two attributes also have getters defined and those getters used instead of directly accessing the attributes?

nkimble (changes requested):
I have completed my review. I have noted a few items to consider with these changes.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
cah292: Ok, I'll see if I can adjust this to get it closer to where the alias was defined. I was hoping to avoid the use of a boolean method parameter, but now that an alias has been added for that table, I agree it makes more sense to move it over to where the alias was defined.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
cah292: I'll remove the unused constant.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
cah292: In general, KualiCo's service beans usually have just the setter methods for injecting the required services, and then the injected services are accessed directly instead of going through a getter method. I was aiming to follow that setup here.

There are still some cases in base code that define both getters and setters (or that define protected getters to allow subclass access to private service variables), but they might be less frequent now due to KualiCo's refactoring efforts.

nkimble: I am still not sure I agree with that approach by KualiCo. There have been a number of times we have had to put in a customization to deal with a null pointer exception due a service attribute being null when the class attempts to use it directly. Class org.kuali.kfs.kim.impl.role.RoleServiceBase being just one example. If KualiCo had their standard be to create both class attribute accessors every time, then the amount of customization we would need to do would be encapsulated in the getter when the "null" issues have to be addressed. I will get off my soap box now. :-)

cah292: In the RoleServiceBase example that you gave, I believe that was a case where the kimTypeInfoService property was null and needed fixing. In that particular case, though, the kimTypeInfoService property was actually one of our customizations to RoleServiceBase, and we had failed to populate the service reference properly in our bean setup. It was a problem we had accidentally introduced, not KualiCo.

The issue with the initialize-if-necessary service getters is that it adds extra clutter to the service classes, and also implies that the service classes might not be fully initialized at startup. With the service injection approach, it allows the service to stipulate that its required services MUST be injected at application startup, and that a null service reference would indicate an initialization problem that should be fixed via the Spring bean setup instead of custom service-loading code.

cah292:
I've committed some changes based on the current feedback. If you still think that getters and/or service-initializing getters are needed, please let me know.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
```diff
@@ -37,23 +37,27 @@
     }
 
     private CuSqlQuery buildRequeueSqlQuery() {
-        CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk();
-        sqlChunk.append(" ORDER BY DOC_HDR_ID ASC");
+        CuSqlChunk sqlChunk = buildRequeueSqlQueryChunk(true);
```
nkimble, line 40: I understand now why you wanted to avoid using booleans in the parameter lists.

Do you think it would be clearer as to what the boolean values mean when we look at this code in a year from now if we had the following two method wrapper names for the boolean use as
- buildRequeueSqlQueryWithOrderBy() {buildRequeueSqlQueryChunk(true)}
- buildRequeueSqlQueryWithoutOrderBy() {buildRequeueSqlQueryChunk(false)}

where this private method on line 39 is renamed to buildRequeueSqlQueryWithOrderBy()
and the method call on line 83 is changed to buildRequeueSqlQueryWithoutOrderBy()?

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
nkimble: Thank you.

src/main/java/edu/cornell/kfs/sys/service/impl/DocumentMaintenanceServiceImpl.java
nkimble: Lets leave things as is for now. Thanks for the clarification.

src/main/java/edu/cornell/kfs/sys/dataaccess/impl/DocumentMaintenanceDaoJdbc.java
cah292: Yes, I agree that adding wrapper methods like those will help make the code's intent much clearer. I'll adjust this code accordingly.

cah292:
I've updated the code so that it uses some new "WithOrderBy"-style and "WithoutOutOrderBy"-style methods like what was suggested. This is ready for review again.

nkimble (approved):
My concerns have been addressed. Unit tests have passed. Merging.

PR #1177, opened 2021-09-14
KFSPTS-22105 Fix Document Reindexer Job

cah292:
This PR fixes a few bugs in the Document Reindexer Job, and also brings the code in line with our current coding practices.

Note that when the job fails, it will now throw an exception instead of returning false. I will create a separate user story to get our other cu-kfs batch jobs to follow this setup, since the KFS Javadocs suggest that throwing an exception is the correct approach for reporting a batch step failure.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
@@ -180,6 +180,9 @@ public class CUKFSConstants {
 
     public static final String LEGACY_PERMIT_NAMESPACE = "PERMIT";
 
+    public static final String DOCUMENT_REINDEX_FILE_NAME_PREFIX = "documentReindex";
+    public static final String DOCUMENT_REINDEX_FILE_NAME_EXTENSION = ".txt";
+
     public static final class FILE_EXTENSIONS {
         public static final String CSV_FILE_EXTENSION = ".csv";
     }
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
@@ -175,13 +175,13 @@ public class CUKFSConstants {
     public static final String DOCUMENT_REFRESH_QUEUE_SERVICE_SPRING_CONTEXT_NAME = "documentRefreshQueue";
 
     public static final String XML_FILE_EXTENSION = ".xml";
+    public static final String TEXT_FILE_EXTENSION = ".txt";
 
     public static final String LOCATION_SERVICE_BEAN_NAME = "locationService-fin";
 
     public static final String LEGACY_PERMIT_NAMESPACE = "PERMIT";
 
     public static final String DOCUMENT_REINDEX_FILE_NAME_PREFIX = "documentReindex";
-    public static final String DOCUMENT_REINDEX_FILE_NAME_EXTENSION = ".txt";
 
     public static final class FILE_EXTENSIONS {
         public static final String CSV_FILE_EXTENSION = ".csv";
--- a/src/main/java/edu/cornell/kfs/sys/batch/DocumentReindexStep.java
+++ b/src/main/java/edu/cornell/kfs/sys/batch/DocumentReindexStep.java
@@ -4,56 +4,70 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
-import java.util.ArrayList;
+import java.nio.charset.StandardCharsets;
 import java.util.Date;
-import java.util.Iterator;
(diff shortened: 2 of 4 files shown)


src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
```diff
@@ -180,6 +180,9 @@
 
     public static final String LEGACY_PERMIT_NAMESPACE = "PERMIT";
 
+    public static final String DOCUMENT_REINDEX_FILE_NAME_PREFIX = "documentReindex";
+    public static final String DOCUMENT_REINDEX_FILE_NAME_EXTENSION = ".txt";
```
nkimble, line 184: Should this constant have a more generic name such as "TEXT_EXTENSION"?
Asking if this would make sense as I can see where having the constant name defined as you currently have it is explicit for it's use but having it more generically defined would allow us to reuse/replace two other constants when that code is refactored.

src/main/java/edu/cornell/kfs/sys/batch/DocumentReindexStep.java
```diff
@@ -4,56 +4,70 @@
...
+        }
+    }
+
+    private void reindexDocuments(List<String> documentIds) {
+        for (String documentId : documentIds) {
+            try {
+                LOG.info("reindexDocuments, Indexing document " + documentId);
+                documentAttributeIndexingQueue.indexDocument(documentId);
+            } catch (Exception e) {
+                LOG.error("reindexDocuments, Unexpected error occurred while indexing document " + documentId, e);
```
nkimble, line 59: The exception generated here is not re-thrown. This will allow processing to be attempt for all documents listed in the file and it is good that the error is logged for each document that failed. 
Asking if more should be done to "alert" that there is an issue with a particular document so that these failures don't repeatedly happen and "get lost"?

nkimble (changes requested):
Changes look good. I noted a couple of things to consider if doing those types of changes makes sense in this case.

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
cah292: Earlier today when I went to CUKFSConstants to remove an unneeded constant in the other open PR, I noticed that we had actually started defining some constants for file extensions. I will rework this ".txt" constant to have a more generic name.

I also noticed that CUKFSConstants has two different strategies for defining those file extension constants. I will create a follow-up user story to get those consolidated into one spot.

src/main/java/edu/cornell/kfs/sys/batch/DocumentReindexStep.java
cah292: Apart from triggering a failure of the batch job as a whole, we could have the job send an email or generate a report file to indicate which documents failed to reindex (if any). However, I'm not certain whether it would be worth spending the effort to implement something like that at the present time; it might make more sense to have it as a future enhancement. Let me know what you think.

nkimble: Having this as a future enhancement makes sense. That will allow us to see what the behavior is like before we invest any more effort.

cah292: Sounds good, I'll create a follow-up enhancement user story accordingly. Thanks for the input!

nkimble:
My understanding (from quickly looking at the old opdoc) is that the functional users need to create the data file and upload it for this reindexing batch job to have data to process. Should we ask DFA whether there would be any value in an additional enhancement of automatically creating that data file ?

cah292:
I suppose we could investigate auto-generation of that file as another enhancement, but I'm not certain if there's a practical way to determine whether a document needs reindexing. The enhancement could look for some search attribute that is common to all or most docs and see if it exists and/or is non-blank, but that may or may not be enough information on its own to determine whether the document should indeed be a candidate for reindexing.

Also, I believe only the technical team has authorization to upload the required text files at the moment, but that could easily be opened up to the functionals via a permission change.

cah292:
I've committed some minor changes based on the current feedback. This is ready for review again.

se12:
I thought the initial iteration of this job did use an uploaded datafile; but we enhanced the process at some point to create the data file internally?

cah292:
I think the document REQUEUER job was updated to automatically determine which documents to requeue, but I'm not sure if we ever did that to the document REINDEXER job.

nkimble:
retest this please

nkimble:
retest this please

nkimble (approved):
My concerns have been dealt with. Additional JIRAs have been created to address items that were uncovered but are out of scope for this particular user story. Merging these changes.

PR #1178, opened 2021-09-16
KFSPTS-22552 add cookie processor setting to context.xml

jhulslander:
This change allows for better cookie persistence when using eshop

### DIFF
--- a/src/main/java/org/kuali/kfs/kns/util/CookieUtils.java
+++ b/src/main/java/org/kuali/kfs/kns/util/CookieUtils.java
@@ -1,0 +1,83 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2021 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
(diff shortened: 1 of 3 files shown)


chadbower (commented):
This is the value we tested manually

jhulslander:
Actually, please don't merge this yet

jhulslander:
retest this please

nkimble:
@jhulslander  Based on the Tomcat documentation I saw on-line, by specifying this setting with the value of "none", the same-site cookie attribute will be set and the cookie will always be sent in cross-site requests.  This is the behavior we want, correct?

jhulslander:
@chadbower the issue last night where the navigation pane would start to appear and then disappear.  I am experiencing the problem locally with Chrome as well.  Firefox does does work locally with this configuration.  This configuration will need some tweeks

jhulslander:
@nkimble  the goal of this change was to prevent the jsession and kualisession cookies from being deleted when in the eshop iframes.  With out configuration this cookie configuration change would only affect the eshop integration, which is fairly scoped down.  @chadbower do you have further details to Nancy's question?

db483:
@jhulslander This change looks similar to the fix KualiCo put in place for the Jaegger issue, take a look at the email from Lesley. They have a couple other changes in place.

jhulslander:
As @db483  mentioned, KualiCo had some other changes they implemented for Stevens.  I have added those changes to this PR

src/main/java/org/kuali/kfs/kns/util/CookieUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    public static final String AUTH_TOKEN_COOKIE_NAME = "authToken";
+    public static final String FIN_AUTH_TOKEN_COOKIE_NAME = "financialsAuthToken";
+
+    public Cookie createCookie(HttpServletRequest request, String name, String value, String path) {
+        Cookie cookie = new Cookie(name, value);
+        /*
+         * KFSPTS-22552
+         * Original base code was cookie.setSecure(request.isSecure()); 
+         */
+        cookie.setSecure(true);
```
nkimble, line 40: @db483 Do we need to know when "request.isSecure()" would return a false value?

nkimble (approved):
@db483 and I conversed regarding the logging question off-line. Additional questions will be posed to KualiCo in a subsequent meeting regarding why they choose their fix of the hard setting of the cookie secure value to true. Merging these changes.

PR #1179, opened 2021-09-20
KFSPTS-21721 Fix KSR maintenance doc conversion

cah292:
This PR fixes the maintenance XML conversion for older KSR Security Group and Security Provisioning documents.

Many of these objects' fields had Kuali numeric types that were later converted to regular Java numeric types. To simplify the required XML sub-element unwrapping that resulted from this change, I updated the maintenance XML converter to include a new rule type for doing so. This also required updating an existing unit test case that was relying upon KSR objects, and it allowed for simplifying the conversion of unrelated Nonresident Tax Percent objects.


db483 (approved):
Changes look good.

PR #1181, opened 2021-09-21
KFSPTS-22283 Fix KSR Provisioning validation

cah292:
This PR fixes some various parts of the Security Provisioning business rules validation:

* Added some Data Dictionary configuration that requires the role fields to reference roles that exist and are active.
* Fixed the business rule validation for Role Tab Order, to use the correct comparison to check for duplicate values. Also updated the related error message to provide a little more detail.
* Fixed the sections that add a role name to an error message, to make them null-safe.

Note that KFS auto-skips some portions of the Security Provisioning validation for inactive Provisioning Groups, though the Role Tab Order validation is still in effect in such cases. If you believe some of this active/inactive validation behavior needs to change, please let me know.


db483:
@cah292 If the current implementation doesn't validate the inactive provisioning groups then I think we should keep this the same. If functionals will decide that should be validated we can add that enhancement at that time. Thank you.

db483 (approved):
Changes look good.

PR #1182, opened 2021-09-22
KFSPTS-22620 Fix broken PDP-related batch jobs

cah292:
This PR fixes our PaymentDetail overlay, so that it correctly overrides the afterInsert() method instead of the beforeInsert() method. (Prior to the KEW upgrade, this class was overriding the postPersist() method, which KualiCo renamed to afterInsert().)


nkimble (approved):
Changes look appropriate. Merging.

PR #1184, opened 2021-09-29
KFSPTS-22284 Allow inactivating KSR dependent roles

cah292:
There are two PRs for this change: One in cu-kfs and one in nonprod-sql. Please make sure both are good to go before merging.

This PR adds an "active" flag to the KSR Dependent Roles objects, and adjusts the Provisioning and Security Request business rules to skip some of their validation for inactive ones. This allows for at least some minimal configuration of such objects, whose other fields are normally uneditable after creation (since most or all of their other fields are part of the primary key).


db483 (approved):
Code changes look good.

PR #1195, opened 2021-10-21
KFSPTS-21521 update account reversion fiscal year maker processing to update inactive account reversions to the new year

jhulslander:
Unfortunately fiscal year maker in over write mode doesn't run in a reasonable time frame locally.  I think this change will accomplish the goal.

In over-write mode, we only want to copy active account reversions.
in non-over-write mode, we want to update account reversions that have been made inactive in the previous FY.

I was not able to fully test this change locally.


cah292 (approved):
Changes look good, and unit tests pass. Given the difficulties in testing this locally, I'm okay with having further validation take place in DEV, to confirm that the changes behave correctly for both Fiscal Year Maker modes.

Merging.

PR #1199, opened 2021-10-26
KFSPTS-22784 Backport FINP-7585 context XML fix

cah292:
This PR backports the FINP-7585 enhancement, which moves the context.xml file to a different location and adjusts the maven-war-plugin config to get the file from its new location.

In my local testing, I confirmed that the context.xml file was successfully being added to the META-INF folder of the exploded WAR directory structure. However, this setup was also working locally before I had made the changes, so I'm not certain what prompted KualiCo to make this update. Perhaps it was needed to fix OS-specific or overlay-related issues.


deploy/tomcat/context.xml
```diff
@@ -0,0 +1,43 @@
...
+<!DOCTYPE doc>
+<!--
+    CU Customizations:
+
+    * Added settings for the "path" and "reloadable" attributes, as well as a DOCTYPE declaration.
+    * Added "CookieProcessor" settings to fix some Shibboleth integration issues.
+    * Backported FINP-7556 resource cache fix from the 2021-05-13 patch.
+    * Backported FINP-7585 context.xml bundling fix from the 2021-05-27 patch (which moved this file
+          to a new location and required adjusting the POM's maven-war-plugin configuration).
+ -->
```
nkimble, line 30: The comment above talks about adding a DOCTYPE declaration (which was in the file that was removed) but I do not see it in the XML code below.  Should the comment be fixed or does the XML need adjusting?

cah292: The DOCTYPE part is still in this file, though it's above the initial comment instead of below it. Sorry for the confusion.

If you want me to move that DOCTYPE declaration below the comment, please let me know.

nkimble: If it is not too much trouble, I think having the comment firs(as it talks about all the changes) followed by the changes would keep this closer to the existing mod being moved and prevent others from missing that single statement in the future.

cah292: Moved its position as requested.

cah292:
I've committed a minor change to move the DOCTYPE declaration below the main comment, as Nancy had suggested. This is ready for review again.

nkimble (approved):
My concerns have been addressed.

PR #1200, opened 2021-10-27
KFSPTS-22970 Fix invalid 'kr' URL usage

cah292:
Many of the KEW-to-KFS URLs should no longer have "kr" path sections (though some still need them). This PR updates the relevant areas of cu-kfs that should not be using the "kr" path sections anymore:

* The biggest problem was in DisbursementVoucherAction, where the functionals were getting "Our Apologies" messages on DVs because we had forgotten to add some minor KEW-to-KFS changes into our overlay.
* I also noticed an issue where the Favorite Account sections on certain documents were using the wrong KEW-to-KFS URL to load the "add" button image.
* Some of the code for the Create Done Batch File lookup was using an unneeded "kr" part as well. However, its prior setup was still generating the URLs correctly, so this PR just cleans up the unnecessary "kr" usage.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/businessobject/actions/CreateDoneBatchFileActionsProvider.java
+++ b/src/main/java/edu/cornell/kfs/sys/businessobject/actions/CreateDoneBatchFileActionsProvider.java
@@ -1,15 +1,11 @@
 package edu.cornell.kfs.sys.businessobject.actions;
 
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/sys/businessobject/actions/CreateDoneBatchFileActionsProvider.java
```diff
@@ -27,7 +27,6 @@
```
nkimble, line 27: There are a number of unused imports in this file. Should we consider cleaning them up?

cah292: The imports were not in the scope of this change, but I suppose there would be no harm to include their cleanup here. I can work on removing the unneeded imports.

cah292:
I've committed some changes to cleanup unused imports and an unused variable. (In the DisbursementVoucherAction overlay, there was an unused import that was not present in base code, so I removed it to keep the file a little closer to the base one.)

nkimble (approved):
Changes look good and my concerns have been addressed. Merging.

PR #1201, opened 2021-10-28
KFSPTS-21860 Remove unneeded GLPE service overlay

cah292:
This PR removes our overlay of the GeneralLedgerPendingEntryServiceImpl class. As of the 2021-01-28 KEW-to-KFS financials release, all of our customizations to that class should now be present in base code.


nkimble (approved):
These changes are in base code as of the KEW upgrade. It is appropriate to remove this local customization.

PR #1203, opened 2021-11-02
KFSPTS-23032: Log additional information to analyze NullPointerException email.

nkimble:


src/main/java/org/kuali/kfs/sys/document/workflow/FinancialSystemSearchableAttribute.java
```diff
@@ -215,8 +215,16 @@
         if (LOG.isDebugEnabled()) {
             LOG.debug("extractDocumentAttributes( " + extensionDefinition + ", " + documentWithContent + " )");
         }
-        List<DocumentAttribute> searchAttrValues = super.extractDocumentAttributes(extensionDefinition,
+        List<DocumentAttribute> searchAttrValues = null;
+        try {
+            searchAttrValues = super.extractDocumentAttributes(extensionDefinition,
                 documentWithContent);
+        } catch (NullPointerException npe) {
```
jhulslander, line 222: Do you think we should catch a more generic object, like RunTimeException or Exception?  That way if something more low-level is happening, we can catch that too.

jhulslander (commented):
Found one things to consider

src/main/java/org/kuali/kfs/sys/document/workflow/FinancialSystemSearchableAttribute.java
nkimble: I will change to catching Exception so that we are at the lowest level. With the stack trace being output to the log we should then have whatever is happening.

nkimble:
Requested changes have been made and unit tests are running.

jhulslander (approved):

PR #1204, opened 2021-11-03
KFSPTS-21561 Removed unneeded custom AD portions

cah292:
This PR removes various Advance Deposit (AD) document customizations that we no longer need. We used to have some approval routing set up for the AD docs, and we customized it accordingly to grant the appropriate level of access to the approvers. However, we don't route these documents for approval anymore, so we're removing the relevant parts of the customization as per the request from the functionals.

Note that a few AD customizations will still remain in place, such as disabling the ability to error-correct AD docs.


src/main/java/edu/cornell/kfs/fp/document/authorization/CuAdvanceDepositDocumentPresentationController.java
```diff
@@ -1,61 +0,0 @@
...
-	            editModes.add(KFSConstants.AMOUNT_TOTALING_EDITING_MODE);
-	        }
-
-	        editModes.add(KFSConstants.BANK_ENTRY_VIEWABLE_EDITING_MODE);
-	        
-	        AccountingDocument accountingDocument = (AccountingDocument) document;
-	        WorkflowDocument workflowDocument = accountingDocument.getDocumentHeader().getWorkflowDocument();
-
-	        if (workflowDocument.isInitiated() || workflowDocument.isSaved()) {
-	            editModes.add(CUKFSAuthorizationConstants.AdvanceDepositEditMode.EDITABLE_ADVANCE_DEPOSITS);
```
nkimble, line 55: Is this another one of our constants that will no longer be used? Should it be removed as well?

nkimble (changes requested):
The changes made look good. I did find a couple of items we may want to consider including with these changes.

src/main/java/edu/cornell/kfs/fp/document/authorization/CuAdvanceDepositAccountingLineAuthorizer.java
```diff
@@ -1,221 +0,0 @@
...
-     * 
-     * @param accountingLine
-     * @return true if editable, false otherwise
-     */
-    private boolean isOrgReviewEditable(AccountingLine accountingLine) {
-        boolean orgReviewEditable = false;
-        OrganizationService organizationService = SpringContext.getBean(OrganizationService.class);
-
-        Person currentUser = GlobalVariables.getUserSession().getPerson();
-        String roleId = getRoleService().getRoleIdByNamespaceCodeAndName(KFSConstants.CoreModuleNamespaces.FINANCIAL, CUKFSConstants.SysKimApiConstants.ADVANCE_DEPOSIT_ORGANIZATION_REVIEWER_ROLE_NAME);
```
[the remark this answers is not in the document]
cah292: I agree, I don't see any other areas in cu-kfs that are actively using this constant. I will remove it.

src/main/java/edu/cornell/kfs/fp/document/authorization/CuAdvanceDepositDocumentPresentationController.java
cah292: This constant also seems unused by the rest of cu-kfs, so I will remove it.

cah292:
I've removed the unused constants as requested. This is ready for review again.

nkimble (approved):
Additional changes look good. My concerns have been addressed. Merging.

PR #1208, opened 2021-11-08
KFSPTS-22261 Cleanup KSR provisioning validation

cah292:
This PR simplifies the business rules of the Security Provisioning document, to perform cleaner handling of the property paths in the validation error messages. The addLine validation already adds appropriate path prefixes in the superclass or other related classes, allowing for a simpler handling of prefixes when validating the existing line items.


cah292:
UPDATE: Please don't merge this PR yet; I recently noticed some areas in the modified class that are not null-safe, so I'm including cleanup of those areas as part of this work. I will post a follow-up comment once this is ready for further review.

nkimble:
I had a couple of questions about these changes.
Could local variable tempProvisioningGroup itself ever be null?
Could method parameter securityProvisioningGroup itself ever be null?

Both classes are accessed in various methods to obtain attribute values so does a null check need to be added as the first thing in if-statements prior to their attribute values being accessed?

cah292:
No, I would not expect those provisioning group variables to ever be null. Their values are being derived from either the provisioning group list (which should not contain null elements) or from the addLine object (where the code is implicitly doing a null check via the "instanceof" condition). It's a similar case for the Dependent Role variables as well.

nkimble (approved):
Thank you for the clarification. Changes look appropriate. Merging.

PR #1211, opened 2021-11-09
KFSPTS-21963 Fix Security Group validation

cah292:
There was a section of the Security Group business rules that was trying to auto-correct the Group Tab data, but which was ultimately resulting in more problems. This PR removes the problematic sections that attempt to repopulate or remove Group Tabs at validation time.

In addition, I performed some cleanup to make the various business rule checks null-safe, and simplified the addLine logic so that it doesn't have to re-validate the existing Group Tabs at that time.

### DIFF
--- a/src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
+++ b/src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.kuali.kfs.kns.document.MaintenanceDocument;
 import org.kuali.kfs.kns.maintenance.rules.MaintenanceDocumentRuleBase;
 import org.kuali.kfs.krad.bo.PersistableBusinessObject;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
```diff
@@ -32,18 +33,15 @@ public boolean processCustomAddCollectionLineBusinessRules(MaintenanceDocument d
...
-                            GlobalVariables.getMessageMap().putError(KSRPropertyConstants.SECURITY_GROUP_TAB_NAME, KSRKeyConstants.ERROR_SECURITY_GROUP_TAB_NAME_UNIQUE);
-                            success = false;
-                        }
+            if (CollectionUtils.isNotEmpty(securityGroup.getSecurityGroupTabs())) {
+                for (SecurityGroupTab tab : securityGroup.getSecurityGroupTabs()) {
+                    if (tab.getTabOrder() != null && tab.getTabOrder().equals(tempTab.getTabOrder())) {
+                        GlobalVariables.getMessageMap().putError(KSRPropertyConstants.SECURITY_GROUP_TAB_ORDER, KSRKeyConstants.ERROR_SECURITY_GROUP_TAB_ORDER_UNIQUE);
+                        success = false;
+                    }
+                    if (StringUtils.isNotBlank(tab.getTabName()) && StringUtils.equals(tab.getTabName(), tempTab.getTabName())) {
```
nkimble, line 42: Do we need to be concerned about case with this compare?

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
```diff
@@ -87,8 +85,8 @@ protected boolean dataDictionaryValidate(MaintenanceDocument document) {
         SecurityGroup temp = retrieveSecurityGroupByName(securityGroup.getSecurityGroupName());
         if (ObjectUtils.isNotNull(temp)) {
             if (securityGroup.getSecurityGroupId() != null) {
-                if (temp.getSecurityGroupName().equals(securityGroup.getSecurityGroupName())
-                        && (!temp.getSecurityGroupId().equals(securityGroup.getSecurityGroupId()))) {
+                if (StringUtils.equals(temp.getSecurityGroupName(), securityGroup.getSecurityGroupName())
```
nkimble, line 88: Same question, would case be an issue with this string compare?

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
```diff
@@ -131,29 +129,18 @@ public boolean validateSecurityGroupTabs(MaintenanceDocument document) {
...
-					if (successTabOrder
+					if (successTabOrder && tab.getTabOrder() != null
 							&& tab.getTabOrder().equals(tempTab.getTabOrder())) {
 						GlobalVariables.getMessageMap().putErrorForSectionId(KSRConstants.SECTION_SECURITY_TABS, KSRKeyConstants.ERROR_SECURITY_GROUP_TAB_ORDER_UNIQUE);
 						successTabOrder = false;
 					}
-					if (successTabName
-							&& tab.getTabName().equals(tempTab.getTabName())) {
+					if (successTabName && StringUtils.isNotBlank(tab.getTabName())
+							&& StringUtils.equals(tab.getTabName(), tempTab.getTabName())) {
```
nkimble, line 143: Same question with this one too, would string case be an issue?

nkimble (commented):
I had the same question in three different locations. Other that that, changes look good.

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
cah292: Yes, I agree that it makes sense to convert the noted comparisons to be case-insensitive. I will work on the appropriate updates.

cah292: Updated to be case-insensitive.

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
cah292: Updated to be case-insensitive.

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityGroupRule.java
cah292: Updated to be case-insensitive.

cah292:
I've committed some changes to make the appropriate string comparisons case-insensitive. Note that this also required updating one section to use the CriteriaLookupService instead of BusinessObjectService, to allow for case-insensitive DB searches without implementing a custom DAO class.

nkimble (approved):
Changes look good. My concerns have been addressed. Merging.

PR #1215, opened 2021-11-19
KFSPTS-23162 Fix Account Delegate Global from Model page

cah292:
One of the overrides from the SEC (Access Security) module was interfering with our CuAccountDelegateGlobal customization for the "Account Delegate Global from Model" page. This PR fixes the issue by overriding the problematic lookupableHelperService bean, so that it uses an appropriate CU-specific class with the CuAccountDelegateGlobal feature in place, while still preserving the appropriate SEC module functionality. (As for why this is not a problem in DEV/TEST/PROD, it could be that our overrides were structured to overwrite the problematic SEC one in our before-KEW setup.)

NOTE: I did notice errors like the following in the logs, but such errors already appear to be present in our other environments and they don't appear to be interfering with the Account Delegate Global document:

`ERROR org.kuali.kfs.krad.util.ObjectUtils :: error getting property value for  class org.kuali.kfs.coa.businessobject.AccountDelegateGlobalDetail.accountDelegate.versionNumber Unknown property 'versionNumber' on class 'class org.kuali.kfs.kim.impl.identity.PersonImpl'`


db483 (approved):
Changes look good.

PR #1216, opened 2021-11-19
KFSPTS-23138 Fix handling of older KSR documents

cah292:
The current KEW-to-KFS version of the Security Request Document assumes that each doc has a Request Role object for every provisioned group from the related Security Group. However, this prevents older KSR docs from being opened if a new Provisioning Group item was added after the Security Request was created.

This PR fixes the issue above so that if an unmatched Provisioning Group is discovered, the Security Request Document will log the condition instead of throwing an exception, and will omit the Provisioning Group's role from the document. The Cynergy-side KRAD version of the document already has a similar workaround (but without the logging part).


db483 (approved):
Changes look good.

PR #1217, opened 2021-11-22
KFSPTS-23120 add logging to log in filter to report when remote user is not available

jhulslander:
After review a couple of the NPE email from prod, it looks like this might be happening due to the logged in person is null.   This PR adds logging to the CornellLoginFilter to add more logging if the remote user can not be retrieved.   As apart of this review, please take a look at my note on the parent user story.  If you think anything else should be logged, please let me know.


db483:
Jay,

The changes look good. Based on your comments on the parent user story, I am wondering if logging should also be added to MaintenanceUtils and KualiAccountLookupableHelperServiceImpl? Thank you.

jhulslander:
This is ready for review again.  I figured out my problem was a local configuration type issue

db483 (approved):
Changes look good.

PR #1219, opened 2021-11-29
KFSPTS-22469 Fix KFS cookie error messages

cah292:
Certain KFS browser cookies contain commas in the values, even though many cookies are not supposed to contain such content. This PR overlays the JavaScript file that manages the problematic cookies, to replace the commas with underscores instead. This helps prevent some invalid-cookie errors from appearing in the Catalina logs.

NOTE: If the reviewer decides to test this fix locally, you may have to clear your browser cache beforehand so that the browser will correctly use the overlayed JavaScript file.

### DIFF
--- a/src/main/webapp/scripts/core.js
+++ b/src/main/webapp/scripts/core.js
@@ -1,0 +1,474 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2021 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


cah292:
NOTE: This PR only fixes the issue on the KFS side. If you think we should add this fix to the Cynergy side as well, please let me know and I'll create a PR for that.

nkimble:
These changes look good and the error no longer displays in my local environment. 

I am ok with only doing this for KFS but I have included @db483 on the review so she can have the opportunity to provide her opinion as to whether Cynergy should be similarly adjusted.  

I will merge as appropriate once she has had time to comment.

nkimble:
Daniela and I spoke about the Cynergy side fix.  I am currently not encountering this issue in Cynergy so we are both ok with deferring this change in that environment for now.

nkimble (approved):

PR #1223, opened 2021-12-13
KFSPTS-22566: Cleanup references to Rice and Cynergy in cu-kfs project.

nkimble:


cah292 (approved):
Changes look appropriate based on my review from yesterday, and the unit tests pass. These are good to move forward now so that we can proceed with a kfs-kew-upgrade rebase; if we discover issues later then we can address them in a follow-up ticket.

Merging.

PR #1224, opened 2021-12-13
KFSPTS-23381 update to use log4j 2.15

jhulslander:


jhulslander:
retest this please

pom.xml
```diff
@@ -994,6 +1506,284 @@
             <artifactId>gson</artifactId>
             <version>${gson.version}</version>
         </dependency>
+        
+        <dependency>
+            <groupId>org.kuali.rice</groupId>
+            <artifactId>rice-core-impl</artifactId>
```
cah292, line 1512: This PR is explicitly adding in some of the Rice dependencies and then making Log4J adjustments to those, rather than making the adjustments to only the existing Rice dependencies. Can these extras be removed so that the exclusions can be kept to the existing imports?

If you do need to keep them, then make sure they have the other exclusions that the existing imports have (xapool, etc.).

pom.xml
```diff
@@ -234,6 +235,13 @@
                             	<exclude>WEB-INF/lib/log4j-*-2.11.*.jar</exclude>  
                             	<exclude>WEB-INF/lib/log4j-1.2-api-2.11.0.jar</exclude>
                             	<exclude>WEB-INF/lib/httpclient-4.5.5.jar</exclude>
+                            	<exclude>WEB-INF/lib/log4j-1.2-api-2.13.2.jar</exclude>
```
cah292, line 238: When I build this branch locally, I'm seeing the Log4J 2.13.3 JARs in the exploded WAR file. I would suggest updating these WAR exclusions to match "2.13.*" instead of "2.13.2".

cah292 (changes requested):
I found a few things that need to be changed.

pom.xml
jhulslander: ok I will set them to 2.13.*

pom.xml
jhulslander: thanks for noticing, I added these manually, they are not needed, I shall remove these entries

jhulslander:
I updated the web-inf exclusions, and removed the manually added entries that are no longer needed.  The commit message now references the parent user story

cah292 (approved):
Changes look good, unit tests pass, and my concerns have been addressed. I will update the commit message with the appropriate KFSPTS ticket number as well.

Merging.

PR #1232, opened 2021-12-20
KFSPTS-23465 Follow-up changes from 2021-12-20 rebase

cah292:
Some minor class import tweaks were needed after the 2021-12-20 rebase.

Note that KFSPTS-23465 is the sub-task of KFSPTS-23315. This PR was named after the sub-task because we already have a separate commit for KFSPTS-23315 itself (as a result of last week's rebase).


db483 (approved):
Changes look good and unit tests pass. I will merge these changes.

PR #1233, opened 2021-12-23
KFSPTS-21662 remove unneeded duplicates of base System Spring beans

db483:


cah292 (approved):

PR #1238, opened 2022-01-06
KFSPTS-23439: Backport FINP-5295 onto KualiCo patch 2021-01-21 versions of the same files.

nkimble:
Initial attempt to see if unit tests pass or fail for the FINP-5295 backport.


db483 (approved):
Changes look good, unit tests pass and the application starts successfully locally.

PR #1239, opened 2022-01-11
KFSPTS-23538 Follow-up changes for 2022-01-11 rebase

cah292:
This PR updates the rebased "kfs-kew-upgrade" code with the necessary changes for KEW-to-KFS compatibility. The follow-up changes are very minor, which mostly involve tweaking class imports and a few other minor adjustments.

There were no merge conflicts as a result of the cu-kfs rebase.

Note that on the "AccountsReceivableValidators.xml" changes, there was one bean that was missing the "abstract='true'" setting, which should have been present prior to the KEW upgrade as well. This PR adds that setting back in, but if the Production code ends up needing that change sooner, then we may need another Production patch. (Or, if you think that change should be made in a separate user story and then we bring it into the KEW upgrade code via a rebase, please let me know.)

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceImpl.java
@@ -6,7 +6,7 @@ import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.kuali.rice.core.api.config.property.ConfigContext;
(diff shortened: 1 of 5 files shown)


db483:
@cah292 you are correct, that bean is missing the "abstract='true'" setting. We should create a separate user story to fix that on develop and then rebase. Thank you for noticing this.

nkimble:
@cah292 I am currently creating theuser story for this sprint and fixing that bean issue so we can get it fixed on develop asap.

cah292:
Now that the bean issue has been merged to the "develop" branch, I will proceed with updating the rebased KEW-to-KFS code accordingly. I'm closing this PR since I will likely end up creating a new rebased feature branch to deal with these follow-up changes.

PR #1241, opened 2022-01-12
KFSPTS-23538 Follow-up changes for 2022-01-11 rebase

cah292:
This is an updated version of the #1239 PR. The only notable difference is that the addition of the "abstract='true'" setting in the bean file should no longer be a part of this changeset, since the change was already implemented for KFSPTS-23633, and was automatically brought into the "kfs-kew-rebase-2022-01-11" branch after performing another rebase from "develop".


db483 (approved):
Changes look good.

PR #1242, opened 2022-01-18
KFSPTS-23675 Add CGB excluded-award and thru-curr-date features

cah292:
This PR implements Janet's requested CGB enhancements; namely, allowing manual invoices to be invoiced through the current date, and allowing manual invoices to use excluded-from-invoicing Awards. Portions of this code are based off of the KFSPTS-14970 enhancement that was removed by the KFSPTS-21538 work.

This code implements the invoice-through-current-date feature in a different way than the old implementation did. The previous code used a non-persisted field on the Awards to temporarily hold the Creation Process Type enum value, whereas this new code adds the value as an argument to the relevant method signatures. I believe the latter way is a cleaner way of adding the feature back in.

In addition, I noticed that KualiCo's excluded-from-invoicing suspension category gets removed in a future release (2021-03-04). I've created KFSPTS-23689 to implement the appropriate follow-up work once we're ready to migrate to that patch or later.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/ar/service/impl/CuContractsGrantsInvoiceCreateDocumentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/ar/service/impl/CuContractsGrantsInvoiceCreateDocumentServiceImpl.java
@@ -18,6 +18,7 @@ import org.kuali.kfs.integration.cg.ContractsAndGrantsBillingAwardAccount;
 import org.kuali.kfs.krad.util.ErrorMessage;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.module.ar.ArConstants;
+import org.kuali.kfs.module.ar.ArKeyConstants;
 import org.kuali.kfs.module.ar.ArConstants.ContractsAndGrantsInvoiceDocumentCreationProcessType;
 import org.kuali.kfs.module.ar.ArPropertyConstants.ContractsGrantsInvoiceDocumentErrorLogLookupFields;
 import org.kuali.kfs.module.ar.businessobject.ContractsGrantsInvoiceDocumentErrorLog;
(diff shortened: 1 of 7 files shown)


src/main/java/edu/cornell/kfs/module/ar/service/impl/CuContractsGrantsInvoiceCreateDocumentServiceImpl.java
```diff
@@ -213,19 +214,91 @@ protected boolean isExpenditureSubFund(String subFundGroupCode) {
...
-    
+
+    /*
+     * CUMod: KFSPTS-23675
+     * 
+     * Allow for using excluded-from-invoicing awards when creating a manual invoice.
+     * Also include creationProcessType in the validateBillingFrequency() method call.
+     * 
+     * TODO: The related suspension category for the former change (AwardSuspendedByUserSuspensionCategory)
+     * was removed by KualiCo in the 2021-03-04 patch. When we upgrade to this patch or later,
```
jhulslander, line 241: would it be worth also noting the KuaiCo Jira number in this comment?

jhulslander (changes requested):
I had one small note about adding the KualiCO Jira number.  I looked in base code: https://github.com/KualiCo/financials/search?q=VerifyBillingFrequencyService ... Does InvoiceBillingPeriodSuspensionCategory need to be in the CU Overlay as well?  I see it uses getStartDateAndEndDateOfPreviousBillingPeriod

As we bring in KualiCo patches, we will need to be careful about any changes to classes that use VerifyBillingFrequencyService as we may need to overlay it to make use of creationProcessType.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuContractsGrantsInvoiceCreateDocumentServiceImpl.java
cah292: I've updated the comment to note the FINP ticket number.

cah292:
I've committed a change to update the comment as requested.

As for InvoiceBillingPeriodSuspensionCategory, it looks like that class was first added in the 2021-10-14 financials patch, so we don't need to worry about modifying it yet. I'll create another follow-up KFSPTS ticket to look into customizing that class accordingly when we upgrade to 2021-10-14 or later.

jhulslander:
My concerns have been addressed.  Once the unit tests pass, I shall merge

jhulslander (approved):

PR #1243, opened 2022-01-19
KFSPTS-23580 Follow-up changes for 2022-01-19 KEW rebase

cah292:
This PR makes one minor follow-up change to the rebased KEW upgrade code; namely, a minor update to a class import.

The rebase process did not encounter any merge conflicts.

Note that this is keyed according to the sub-task, since user story KFSPTS-23538 already has an associated commit.


db483 (approved):

db483:
Changes look good.

PR #1250, opened 2022-02-03
KFSPTS-23878 Backport redis fixes on FINP-8169

db483:
This backports changes for FINP-8169 and removes most of changes related to the fix on KFSPTS-23531.
KualiCo missed updating the getAttributeDefinitionByName on AttributeValidationHelper.java so I kept our fix in place. Please let me know if you think we should update the local fix to be more similar to the KualiCo changes. KualiCo created FINP-8210 to fix this method in base code.

### DIFF
--- a/src/main/java/org/kuali/kfs/kim/document/rule/AttributeValidationHelper.java
+++ b/src/main/java/org/kuali/kfs/kim/document/rule/AttributeValidationHelper.java
@@ -47,6 +47,7 @@ import java.util.Map;
  * CU Customization (KFSPTS-23531):
  * 
  * Updated the generation of the cache keys for improved compatibility with Redis.
+ * Backported redis fix on FINP-8169
  */
 public class AttributeValidationHelper {
 
@@ -54,7 +55,6 @@ public class AttributeValidationHelper {
     private static final String DOCUMENT_PROPERTY_PREFIX = KRADConstants.DOCUMENT_PROPERTY_NAME + ".";
 
     // CU Customization (KFSPTS-23531): Add MessageFormat-related constants for generating various cache keys.
-    private static final String ATTRIBUTE_BY_ID_CACHE_KEY_PATTERN = "'{getAttributeDefinitionById}'-id={0}";
     private static final String ATTRIBUTE_BY_NAME_CACHE_KEY_PATTERN = "'{getAttributeDefinitionByName}'-name={0}";
 
     protected BusinessObjectService businessObjectService;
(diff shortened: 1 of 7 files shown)


src/main/java/org/kuali/kfs/kim/impl/role/RoleServiceImpl.java
```diff
@@ -346,8 +341,9 @@ protected void putRoleInCache(RoleLite role) {
         return new ArrayList<>(roleMap.values());
     }
 
-    // bacport redis annotation
-    @Cacheable(cacheNames = Role.CACHE_NAME, key = "'{getRoleByNamespaceCodeAndName}-namespaceCode=' + #p0 + '|' + 'name=' + #p1")
+    // bacport redis annotation; backport fix on FINP-8169
+    @Cacheable(cacheNames = Role.CACHE_NAME,
+            key = "'{" + Role.CACHE_NAME + "}-namespaceCode=' + #p0 + '|' + 'name='+ #p1")
```
cah292, line 346: The KualiCo fix leaves the hyphen in this cache key, even though the related namespace-and-name keys in the other methods above don't have this hyphen. The cache will still work properly, it's just that this method's cache entries won't be interchangeable with those from the relevant get/put methods farther up.

If KualiCo is still working on the FINP-8210 follow-up fixes, we could see if they could include changing this key as part of that work.

src/main/java/org/kuali/kfs/kim/document/rule/AttributeValidationHelper.java
```diff
@@ -47,14 +47,14 @@
...
+ * Backported redis fix on FINP-8169
  */
 public class AttributeValidationHelper {
 
     private static final Logger LOG = LogManager.getLogger();
     private static final String DOCUMENT_PROPERTY_PREFIX = KRADConstants.DOCUMENT_PROPERTY_NAME + ".";
 
     // CU Customization (KFSPTS-23531): Add MessageFormat-related constants for generating various cache keys.
-    private static final String ATTRIBUTE_BY_ID_CACHE_KEY_PATTERN = "'{getAttributeDefinitionById}'-id={0}";
     private static final String ATTRIBUTE_BY_NAME_CACHE_KEY_PATTERN = "'{getAttributeDefinitionByName}'-name={0}";
```
cah292, line 58: I'm fine with keeping this custom cache key in place until we're ready to backport KualiCo's follow-up FINP-8210 fixes.

cah292 (approved):
Changes look good, and unit tests pass. I added a couple line-level comments, but they involve items that are related to KualiCo's follow-up FINP-8210 fix and which don't require Cornell code changes at this time.

Merging.

PR #1254, opened 2022-02-10
KFSPTS-23970 Add script-loading fix based on FINP-7386

cah292:
The FINP-7386 fix from the 2021-02-25 financials patch corrected some script file path references that were causing issues in the KEW-to-KFS upgrade. Paths starting with "../" were not being resolved properly, but removing that prefix allowed KFS to find the files successfully. This PR fixes up similar "../" file paths within the cu-kfs code.

In addition, this PR adds some temporary overlays of MaintenanceDocumentEntry and TransactionalDocumentEntry that will automatically strip the "../" paths from existing script files as needed. This allows for fixing up the base financials files that have this issue, but without having to override/overlay the ~34 files affected by the FINP-7386 fix. The two overlays can be easily removed with one of our subsequent KFS upgrades.

If you think it would be better to override/overlay the other base financials files instead of using the temporary code overlays, please let me know.

### DIFF
--- a/src/main/java/org/kuali/kfs/kns/datadictionary/MaintenanceDocumentEntry.java
+++ b/src/main/java/org/kuali/kfs/kns/datadictionary/MaintenanceDocumentEntry.java
@@ -1,0 +1,386 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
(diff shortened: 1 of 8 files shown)


db483 (approved):
Changes look good.

PR #1255, opened 2022-02-15
KFSPTS-23996 Fix IWNT-to-REQS phone number formatting

cah292:
We discovered some cases where phone numbers copied from the IWNT to the REQS document were in a format that the REQS docs considered to be invalid. As a minimal-effort workaround, this PR updates the generate-REQS-from-IWNT code so that it automatically alters the appropriate phone numbers to use the format expected by the Requisition.

I also noticed that the needed phone number formatting was already being performed in certain IWNT-to-REQS cases if the IWNT did not fill out the delivery section. This PR expands that formatting to encompass the cases where the IWNT delivery section has been filled out, and also cleans up the coding in the existing phone formatting case.


db483 (approved):
Changes look good.

PR #1261, opened 2022-03-03
KFSPTS-24216 backport changes for FINP-8249

db483:


src/main/java/org/kuali/kfs/sys/document/authorization/AccountingDocumentAuthorizerBase.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Map;
+
+/**
+ * DocumentAuthorizer containing common, reusable document-level authorization code for financial (i.e. Transactional)
+ * documents
+ */
+/**
+ * 
+ * CU customization: Backport FINP-9249. The backport has been added to a copy of this file is from the 28-01-2021 version of financials.
```
cah292, line 37: Should this comment reference FINP-8249 instead of FINP-9249?

cah292 (changes requested):
I noticed one minor spot in one of the comments that would be good to clean up for clarity. Other than that, the backported changes look good.

src/main/java/org/kuali/kfs/sys/document/authorization/AccountingDocumentAuthorizerBase.java
db483: Good catch. I am updating it.

cah292 (approved):

PR #1263, opened 2022-03-04
KFSPTS-24218 backport FINP-8270

db483:


src/main/java/org/kuali/kfs/module/purap/document/service/impl/PaymentRequestServiceImpl.java
```diff
@@ -0,0 +1,2108 @@
...
+	public Collection<PaymentRequestDocument> getPaymentRequestsToExtractByCM(String campusCode,
+			VendorCreditMemoDocument cmd) {
+		LOG.debug("getPaymentRequestsByCM() started");
+		Date currentSqlDateMidnight = dateTimeService.getCurrentSqlDateMidnight();
+		List<PaymentRequestDocument> paymentRequestIterator = paymentRequestDao.getPaymentRequestsToExtract(campusCode,
+				null, null, cmd.getVendorHeaderGeneratedIdentifier(), cmd.getVendorDetailAssignedIdentifier(),
+				currentSqlDateMidnight);
+
+		return filterPaymentRequestByAppDocStatus(paymentRequestIterator, PaymentRequestStatuses.APPDOC_AUTO_APPROVED,
+				PaymentRequestStatuses.APPDOC_DEPARTMENT_APPROVED);
```
cah292, line 159: This overlay contains several formatting changes that do not line up with the 01/28/2021 version of the file. For example, it looks like this "return" statement is meant to span three lines instead of two. Can you please verify that the correct file version was overlayed?

src/main/java/org/kuali/kfs/module/purap/document/service/impl/PaymentRequestServiceImpl.java
```diff
@@ -0,0 +1,2108 @@
...
+	 *
+	 * @param vendorHeaderGeneratedId The vendor header generated id.
+	 * @param vendorDetailAssignedId  The vendor detail assigned id.
+	 * @return List of payment request documents.
+	 */
+	/*
+	 * CU customization: backport FINP-8270
+	 */
+	@Override
+	public List<PaymentRequestView> getPaymentRequestsByVendorNumber(Integer vendorHeaderGeneratedId,
```
cah292, line 498: It seemed strange that this method's return type was altered without also changing the PaymentRequestService interface (and likewise with KualiCo's changes), but then I saw that the interface declares this method with a raw return type instead of one using generics.

We may want to notify KualiCo about the raw-type issue on the PaymentRequestService interface if they're not aware of it.

cah292 (changes requested):
One of the overlay files might need some minor adjustments, and I found a small issue that we may want to report to KualiCo. Other than that, the changes look good.

src/main/java/org/kuali/kfs/module/purap/document/service/impl/PaymentRequestServiceImpl.java
db483: Chad, thank you for noticing. I updated the file.

src/main/java/org/kuali/kfs/module/purap/document/service/impl/PaymentRequestServiceImpl.java
db483: Good point Chad, I will notify KualiCo. Thank you.

cah292 (approved):

PR #1264, opened 2022-03-10
KFSPTS-24344 Backport FINP-8283 dup-invoice-check fix

cah292:
This PR backports the FINP-8283 fix, which is a follow-up fix for the FINP-8270 changes that we had backported in a previous PR.


ajd299 (approved):
Changes look good, merging

PR #1268, opened 2022-03-25
KFSPTS-24369 Improve caching via Ehcache and Redis

cah292:
This is ready for initial review, but please do NOT merge it yet. More testing is needed in Pre-KEW to confirm that it's working as expected and that it isn't somehow introducing new slowness.

This PR updates KFS's Spring-based caching to use a hybrid of Ehcache and Redis. The code uses Ehcache to allow each KFS instance to maintain its own distinct local cache, and it uses Redis to communicate updates/removals/expirations of cache entries to each KFS node. The actual cache value for each key will only be stored in Ehcache; a placeholder value will be stored in Redis instead, so that Redis will be used as just a notification/expiration handler and not as the cache of record.

The new feature uses the same Redis library used by KualiCo (namely, Lettuce) for handling the Redis interactions. To use the needed Redis notification features, though, I had to update to a newer version of Lettuce that KualiCo started using in one of their post-add-Redis releases.

This PR also introduces asynchronous handling of the Redis interactions, in an attempt to improve performance of the various cache actions. If you notice something about that setup which could potentially slow things down, please let me know.


cah292:
Now that we've upgraded KFS to a newer Spring version, it's now possible to use a newer version of Spring Data Redis that integrates well with the needed version of the Lettuce library. I'm closing this PR so that I can open up a new one that has those changes in place.

PR #1272, opened 2022-04-04
KFSPTS-24203 PAAT Document Approval Workflow Enhancements

ajd299:


src/main/java/edu/cornell/kfs/pdp/document/CuPayeeACHAccountMaintainableImpl.java
```diff
@@ -64,12 +69,25 @@ public boolean answerSplitNodeQuestion(String nodeName) {
...
         }
-        
         return super.answerSplitNodeQuestion(nodeName);
     }
 
     protected boolean payeeACHAccountMaintenanceRequiresPdpApproval() {
+        try {
+            DocumentService documentService = SpringContext.getBean(DocumentService.class);
+            FinancialSystemMaintenanceDocument document = (FinancialSystemMaintenanceDocument) documentService.getByDocumentHeaderId(this.getDocumentNumber());
+            if (isDocumentInitiatorSystemUser(document)) {
```
cah292, line 79: The whole block for the is-sys-user check should be moved below the is-edit-action check. As per the requirements from the user story, PAAT edits should *ALWAYS* route for approval. You should adjust this method so that it short-circuits if the is-edit-action check returns true, and only proceed to the is-sys-user check if the is-edit-action check returns false.

src/main/java/edu/cornell/kfs/pdp/document/CuPayeeACHAccountMaintainableImpl.java
```diff
@@ -64,12 +69,25 @@ public boolean answerSplitNodeQuestion(String nodeName) {
...
+            }
+        } catch (Exception exception) {
+            LOG.error("payeeACHAccountMaintenanceRequiresPdpApproval " + exception.getMessage(), exception);
+        }
         return StringUtils.equalsIgnoreCase(KFSConstants.MAINTENANCE_EDIT_ACTION, getMaintenanceAction());
     }
 
+    protected boolean isDocumentInitiatorSystemUser(FinancialSystemMaintenanceDocument document) {
+        Person documentInitiator = KimApiServiceLocator.getPersonService().getPerson(document.getDocumentHeader().getWorkflowDocument().getInitiatorPrincipalId());
+        return documentInitiator != null && documentInitiator.getPrincipalName().equalsIgnoreCase(KFSConstants.SYSTEM_USER);
```
cah292, line 90: For better null safety, you could use StringUtils.equalsIgnoreCase() to compare the principal names.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,27 @@
...
+import org.kuali.kfs.pdp.document.authorization.PayeeACHAccountMaintenanceDocumentAuthorizer;
+
+public class CuPayeeACHAccountMaintenanceDocumentAuthorizer extends PayeeACHAccountMaintenanceDocumentAuthorizer {
+    private static final Logger LOG = LogManager.getLogger();
+
+    @Override
+    public boolean canApprove(Document document, Person user) {
+
+        try {
+            String documentInitializerPrincipalId = document.getDocumentHeader().getWorkflowDocument().getInitiatorPrincipalId();
```
cah292, line 16: For consistency with KualiCo's terminology, I would suggest renaming this variable to say "Initiator" instead of "Initializer".

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,27 @@
...
+            if (user.getPrincipalId().equals(documentInitializerPrincipalId)) {
```
cah292, line 17: Another comparison that could be switched to using StringUtils.equalsIgnoreCase() if necessary.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,27 @@
...
+                return false;
```
cah292, line 18: As per the user story's requirements, there are certain initiators that may still need to be allowed to approve the document, such as the KFS system user and those with superuser privileges (and possibly those with blanket-approve privileges). Please adjust this code accordingly to account for that.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,27 @@
...
+            }
+        } catch (Exception ex) {
+            LOG.error("canApprove", ex.getMessage(), ex);
```
cah292, line 21: You should adjust this logging statement to either directly concatenate the exception's message with the literal or to not manually print the exception message. Otherwise, Log4J might think you're preparing a parameterized message, as opposed to a custom message followed by the exception's stack trace.

Also, if an exception occurs during the is-initiator check, it might be safer to either return false or rethrow the exception, to ensure that the initiator can't take the action somehow.

cah292 (changes requested):
I found a few areas that should be adjusted, especially some spots that still need to implement more of the user story's requirements.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,46 @@
...
+public class CuPayeeACHAccountMaintenanceDocumentAuthorizer extends PayeeACHAccountMaintenanceDocumentAuthorizer {
+    private static final Logger LOG = LogManager.getLogger();
+
+    @Override
+    public boolean canApprove(Document document, Person user) {
+
+        try {
+            if (StringUtils.equalsIgnoreCase(user.getPrincipalId(), KFSConstants.SYSTEM_USER)) {
+                return true;
+            } else if (canCurrentUserBlanketApprove(document)) {
```
cah292, line 24: In addition to always allowing blanket approvers to approve the document, you may also want to allow superusers to do so, as per the requirements from the user story. (You can double-check with the functionals to confirm which sets of initiators should still be allowed to approve these docs.) The service returned by KEWServiceLocator.getDocumentTypeService() should have some methods to help determine whether a given user is a superuser for the document.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
cah292: See the other comments on my latest review for further details on the can-approve setup.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,46 @@
...
+                return true;
+            }
+
+            String documentInitiatorPrincipalId = document.getDocumentHeader().getWorkflowDocument().getInitiatorPrincipalId();
+            if (StringUtils.equalsIgnoreCase(user.getPrincipalId(), documentInitiatorPrincipalId)) {
+                return false;
+            }
+        } catch (Exception ex) {
+            LOG.error("canApprove" + ex.getMessage(), ex);
```
cah292, line 33: The "canApprove" literal should have an extra space added at the end, so that there's some separation between the literal string and the concatenated message.

src/main/java/edu/cornell/kfs/pdp/document/authorization/CuPayeeACHAccountMaintenanceDocumentAuthorizer.java
```diff
@@ -0,0 +1,46 @@
...
+            return false;
+        }
+
+        return super.canApprove(document, user);
+    }
+
+    private boolean canCurrentUserBlanketApprove(Document document) {
+        DocumentPresentationController documentPresentationController = KNSServiceLocator.getDocumentHelperService().getDocumentPresentationController(document);
+        Set<String> documentActions = documentPresentationController.getDocumentActions(document);
```
cah292, line 42: I'm not sure if you need to perform the presentation controller check at this phase, and such a check would not be sufficient on its own. You should have this code call the inherited canBlanketApprove() method on the authorizer itself.

cah292 (changes requested):
I found a few more areas that could use revisions, but it's looking good in general.

cah292 (approved):

PR #1273, opened 2022-04-05
KFSPTS-24517 Upgrade to latest Spring 5.3.x release

cah292:
This is the KFS-side equivalent of the Spring version upgrade that was recently performed on the Cynergy side.

This PR upgrades KFS to use the latest Spring 5.3.x release, and adds various other changes based on KualiCo's FINP-7916 ticket. The major changes include the following:

* Upgraded to version 5.3.18 of the Spring framework. (Note that certain other Spring JARs that are not part of the basic Spring framework have been left at their current versions.)
* For consistency with the Cynergy changes, the slf4j-api JAR has been updated to a newer version. Most of the rest of KualiCo's FINP-7989 logging-related changes have NOT been added here; we will deal with those when we upgrade to the related financials patch.
* Overlayed ReportGenerationServiceImpl to include the related KualiCo code changes.
* Updated and/or overlayed various Spring XML files to update the beans with "local" attributes to use "bean" attributes instead, since the former ones are no longer supported.

Also note the following:

* FINP-7916 made an enormous number of changes to XSD references within the various Spring XML files. Those changes have not been included in this PR, since KFS still seems to be able to read and process the XML files without such updates, and those changes would have resulted in massive code clutter.
* FINP-7916 updates the KualiInitializeListener class, but prior to the KEW upgrade, that class is still in Rice and doesn't have a KFS-specific equivalent. An overlay for that has not been included in this PR, since the related overlay for it from cornell_customizations should automatically get carried over. (When we prepare the KEW-to-KFS equivalent of these changes, we will add the appropriate KFS-specific overlay at that time.)
* This PR is being merged into an emergency fix branch. After the changes are merged, the emergency branch can be renamed as needed without affecting what was merged. Also, a separate PR will be used to handle the merging of the changes into the develop branch, after they have been adapted for KEW-to-KFS compatibility.
* I made a minor change to the Jenkins unit test job, so that it will use the appropriate Cynergy emergency JARs when running the tests against an emergency branch. Hopefully the tests will still pass.

PR #1274, opened 2022-04-06
KFSPTS-24517 Upgrade to latest Spring 5.3.x release

cah292:
This PR contains the KEW-to-KFS equivalent of the #1273 PR's changes for upgrading KFS to Spring 5.3.x. Below is a summary of how the KEW-to-KFS variant differs:

* For the dependencies that were removed for the KEW-to-KFS upgrade (such as Rice and Cynergy), the related setup for such dependencies was not included.
* The KEW-to-KFS Redis-enabled code has a "spring-oxm" dependency that also had to be updated to the latest Spring version.
* Another transitive Redis-related dependency ("spring-data-commons") was updated to the latest release that has the same major and minor version number as before, since Maven had reported a vulnerability with the one previously used.
* While updating the dependency above, I also noticed some minor vulnerabilities with the "guava" version in use by our KEW-to-KFS code. Adjustments have been made to upgrade to the "guava" version in use by the latest financials code. (Note that our current KEW-to-KFS code was using an even older Guava version than our current Production KFS version, and that older Guava version appeared to have more issues than the one in our Production code.)
* The overlays for "CoreServiceLocalSpringBeans.xml" and "krad-servlet.xml" were not included, since those files do not exist in the KEW-to-KFS code.
* Overlays of various other Spring XML files have been added or updated as needed, including those that weren't in scope for the emergency patch variant of the fix.
* The KEW-to-KFS versions of KualiInitializeListener and ReportGenerationServiceImpl have been overlayed to include the appropriate Spring upgrade changes.


jhulslander (approved):
this PR looks good

PR #1275, opened 2022-04-12
KFSPTS-24369 Improve caching via Ehcache and Redis

cah292:
This PR adds a cleaner implementation of the Ehcache-and-Redis hybrid approach from PR #1268. It starts by upgrading cu-kfs to the newer Spring Data Redis and Lettuce setup introduced by FINP-7985, and then uses those upgraded libraries to help implement the new caching solution.

The actual cached values are stored locally on the KFS instance via Ehcache. KFS only uses Redis for managing the cache flush notifications; it creates/deletes expirable placeholder values in Redis for each key, and uses the Redis "client tracking" feature to listen for notifications.

This code relies upon Spring Data Redis to manage the lifecycle of a shared Redis connection, which frees us from the complexities of managing the connection ourselves like in the older #1268 code. However, I wasn't able to find any features in Spring Data Redis for turning on client tracking, so I've added subclasses/implementations of the appropriate connection provider and factory classes that will use lower-level Lettuce calls to enable client tracking accordingly. All the other Redis data interactions, though, are being handled through the higher-level RedisTemplate from Spring Data Redis.

The Ehcache setup relies on an Ehcache Event Listener implementation to update the Redis-side entries in response to local cache changes. The listeners are added programmatically at cache startup to simplify the setup. Also, this code allows for certain caches (like BatchFile) to be excluded from the Redis-related management.

I've tried to set things up to prevent the KFS cache instances from getting stuck in a cache-flushing loop. The Redis client tracking is being enabled with the "noloop" option, so that the KFS instance will not receive notifications for Redis cache changes that it triggers. On the Ehcache side, the Ehcache listeners are configured to only listen for locally-triggered cache changes, and the Redis-triggered updates to Ehcache are specifying an extra flag that should prevent "local" Ehcache listeners from being notified.

This code still needs more thorough testing to ensure that the Ehcache and Redis interactions are working as intended, especially in a multi-node KFS environment.


src/main/java/edu/cornell/kfs/sys/cache/ClientTrackingLettuceConnectionFactory.java
```diff
@@ -0,0 +1,67 @@
...
+        Objects.requireNonNull(eventListener, "eventListener cannot be null");
+        this.eventListener = eventListener;
+        this.initializedProviderForTracking = new AtomicBoolean(false);
+    }
+
+    @Override
+    protected LettuceConnectionProvider doCreateConnectionProvider(AbstractRedisClient client,
+            RedisCodec<?, ?> codec) {
+        LettuceConnectionProvider connectionProvider = super.doCreateConnectionProvider(client, codec);
+        if (initializedProviderForTracking.compareAndSet(false, true)) {
```
jhulslander, line 36: I am not sure I understand how this works.  initializedProviderForTracking is defaulted to false, so compareAndSet would be true, because we define the expected value is false, and then set the value to true.  However, initializedProviderForTracking is also final, so that means the value should not be changeable.

How does the else condition get executued?

src/main/java/edu/cornell/kfs/sys/cache/CuCompositeCacheKey.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public static String buildRedisCacheKey(String cacheName, String localCacheKey) {
+        return buildRedisCacheKeyInternal(StringUtils.defaultIfBlank(cacheName, KFSConstants.EMPTY_STRING),
+                StringUtils.defaultIfBlank(localCacheKey, KFSConstants.EMPTY_STRING));
+    }
+
+    private static String buildRedisCacheKeyInternal(String cacheName, String localCacheKey) {
+        if (StringUtils.isBlank(cacheName) || StringUtils.isBlank(localCacheKey)) {
+            throw new IllegalArgumentException("Cache name and/or local key cannot be blank");
```
jhulslander, line 63: would it be better to include this check in the constructor to prevent this condition early?

src/main/java/edu/cornell/kfs/sys/cache/CuEhCacheCacheManager.java
```diff
@@ -0,0 +1,191 @@
...
+        Objects.requireNonNull(ehcacheManager, "Internal Ehcache Manager cannot be null");
+        Objects.requireNonNull(redisEventListenerProxy, "redisEventListenerProxy cannot be null");
+        Objects.requireNonNull(ehcacheEventListener, "ehcacheEventListener cannot be null");
+        Objects.requireNonNull(cacheNames, "cacheNames cannot be null");
+        Objects.requireNonNull(cachesIgnoringRedisEvents, "cachesIgnoringRedisEvents cannot be null");
+        Objects.requireNonNull(cachesToClearOnRedisConnectionChange,
+                "cachesToClearOnRedisConnectionChange cannot be null");
+        
+        if (!cacheNames.containsAll(cachesIgnoringRedisEvents)) {
+            throw new IllegalStateException("cachesIgnoringRedisEvents references caches not found in cacheNames");
```
jhulslander, line 44: This might be an improvement for later, but would it be helpful have debug  log statements specifying the cachesIgnoringRedisEvents or cachesToClearOnRedisConnectionChange that aren't in cacheNames

src/main/java/edu/cornell/kfs/sys/cache/EhcacheEventListenerForUpdatingRedis.java
```diff
@@ -0,0 +1,180 @@
...
+                        + redisCacheKey);
+            }
+            redisTemplate.opsForValue().setIfAbsent(redisCacheKey, KFSConstants.ACTIVE_INDICATOR,
+                    timeToLive, TimeUnit.SECONDS);
+        });
+    }
+
+    private long determineTimeToLiveForCache(Ehcache cache) {
+        long timeToLive = cache.getCacheConfiguration().getTimeToLiveSeconds();
+        return (timeToLive > 0) ? timeToLive : defaultTimeToLive;
```
jhulslander, line 78: if the time t live is less than 0, should that be a warning or the like so we can fix the configuration?

jhulslander (commented):
I had a few questions and some other changes to consider

src/main/java/edu/cornell/kfs/sys/cache/ClientTrackingLettuceConnectionFactory.java
cah292: initializedProviderForTracking is an AtomicBoolean, which is a wrapper object and not a primitive boolean. The declaration of "final" only makes the reference to the wrapper object immutable; it does not make its internal boolean value immutable.

The superclass will call the doCreateConnectionProvider() method at least twice: Once to initialize a regular shared connection, and a second time to initialize a shared connection to be used with the "reactive" API (which this PR does not use at all). The setup of the "reactive" connection should get handled by the "else" block, and the DEBUG message in this code explains why we don't want client tracking on both connections.

src/main/java/edu/cornell/kfs/sys/cache/CuCompositeCacheKey.java
cah292: This particular method returns a String and not a CuCompositeCacheKey instance. There are some areas of the code that rely upon having a CuCompositeCacheKey object that has at least a blank key, and which should have safeguards in place to prevent such objects from reaching this error condition.

As a future enhancement (or a change in this PR), we could refactor this code to help make the intent and error handling clearer.

src/main/java/edu/cornell/kfs/sys/cache/CuEhCacheCacheManager.java
cah292: Yes, having that addition as an enhancement makes sense.

src/main/java/edu/cornell/kfs/sys/cache/EhcacheEventListenerForUpdatingRedis.java
cah292: If the value is 0 or potentially negative in Ehcache, that means Ehcache does not specify a time to live. This code just enforces a time to live in such cases. (Note that the caches that are excluded from Redis handling, like BatchFile, will not be affected by this code.)

We can use a follow-up user story to reevaluate this setup, unless it needs to be addressed now.

src/main/java/edu/cornell/kfs/sys/cache/ClientTrackingLettuceConnectionFactory.java
jhulslander: Thanks for clarification

src/main/java/edu/cornell/kfs/sys/cache/EhcacheEventListenerForUpdatingRedis.java
jhulslander: later sounds reasonable

jhulslander (approved):
My concerns will be addressed in a follow up user stories

PR #1278, opened 2022-04-25
KFSPTS-24695: Fix DV issue where Saved DVs without Payee Selected do not present buttons to SuperUser.

nkimble:


src/main/java/org/kuali/kfs/fp/document/DisbursementVoucherDocument.java
```diff
@@ -1468,6 +1474,13 @@ protected boolean isTaxReviewRequired() {
      * @return true if the payee is a vendor and has withholding dates therefore should receive tax review, false otherwise
      */
     protected boolean isPayeePurchaseOrderVendorHasWithholding() {
+        /* CU Customization: Null checks added to prevent NPE when payee is not selected on saved DV causing buttons to not present for a super user.
+         */
+        if (ObjectUtils.isNull(this.getDvPayeeDetail()) || 
+                ObjectUtils.isNull(this.getDvPayeeDetail().getDisbursementVoucherPayeeTypeCode())) {
```
cah292, line 1480: There's some code in the isTaxReviewRequired() method (both in the overlay and in the CU subclass) that expects getDvPayeeDetail() and getDisbursementVoucherPayeeTypeCode() to return non-null values.

I also just noticed something. It appears that DisbursementVoucherDocument and CuDisbursementVoucherDocument have separate "dvPayeeDetail" instance variables, so perhaps only one of them is being populated properly due to our getter/setter overrides. Could CuDisbursementVoucherDocument be updated to use the same "dvPayeeDetail" variable as the parent class (or, alternatively, to change the setter method setup to update both variables)?

cah292 (changes requested):
While reviewing this PR, I discovered a potential bug in our CuDisbursementVoucherDocument code that could be the root cause of the issue. Please see my inline comments below and make adjustments to the CuDisbursementVoucherDocument class accordingly. (Also, we may not need the isPayeePurchaseOrderVendorHasWithholding() method changes if the CU class gets fixed appropriately.)

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -962,6 +970,10 @@ protected ParameterEvaluatorService getParameterEvaluatorService(){
     	return SpringContext.getBean(ParameterEvaluatorService.class);
     }
 
+    @Override
+    public void setDvPayeeDetail(DisbursementVoucherPayeeDetail dvPayeeDetail) {
+        this.dvPayeeDetail = (CuDisbursementVoucherPayeeDetail) dvPayeeDetail;
+    }
 
     public void setDvPayeeDetail(CuDisbursementVoucherPayeeDetail dvPayeeDetail) {
```
cah292, line 978: I see that you added an override of the superclass's "setDvPayeeDetail" setter, but the subclass-specific setter is still present. Can the code be updated so that it only has one dvPayeeDetail setter method?

Also, this subclass still contains its own "dvPayeeDetail" variable. Can the subclass's variable be removed so that the code can use the one inherited from the parent class?

src/main/java/org/kuali/kfs/fp/document/DisbursementVoucherDocument.java
```diff
@@ -1468,6 +1474,13 @@ protected boolean isTaxReviewRequired() {
      * @return true if the payee is a vendor and has withholding dates therefore should receive tax review, false otherwise
      */
     protected boolean isPayeePurchaseOrderVendorHasWithholding() {
+        /* CU Customization: Null checks added to prevent NPE when payee is not selected on saved DV causing buttons to not present for a super user.
+         */
+        if (ObjectUtils.isNull(this.getDvPayeeDetail()) ||
```
cah292, line 1479: Can this overlay's extra null checks be removed? If the code can be updated so that DisbursementVoucherDocument and CuDisbursementVoucherDocument share the same "dvPayeeDetail" variable, the custom null checks might not be necessary.

cah292 (changes requested):
I found a few areas that could be adjusted further to simplify the solution. Sorry if my suggestions from yesterday's review and today's meeting(s) weren't clear about what I was proposing.

nkimble:
The requested changes were made. Please be aware that there were additional changes required in order to get the code to compile with those code review requested changes. IMHO, these changes should undergo full regression testing of the DV document as well as the batch job(s) that generate DVs based upon those required additional changes.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -613,13 +614,14 @@ protected DocumentHelperService getDocumentHelperService() {
         return documentHelperService;
     }
 
-    public CuDisbursementVoucherPayeeDetail getDvPayeeDetail() {
+    @Override
+    public DisbursementVoucherPayeeDetail getDvPayeeDetail() {
```
cah292, line 618: Now that the dvPayeeDetail variables have been consolidated, this getter seems to be the same as in the superclass and isn't adding any value anymore. I would suggest removing it.

If you would rather defer this cleanup to a future user story so that we can keep these changes moving forward, please let me know.

cah292 (commented):
I found one minor spot that would be nice to clean up, but otherwise the change look good. Please let me know if you want the suggested cleanup deferred to a future ticket.

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
nkimble: This getter has been removed. Doing the cleanup now reduces the burden of validation testing for the users.

cah292 (approved):

PR #1283, opened 2022-05-09
KFSPTS-24844 Fix bulk-routing many PREQs for same PO

cah292:
We sometimes encounter OptimisticLockExceptions when the eInvoice Route Job routes multiple PREQs for the same PO in quick succession. These documents are all trying to make updates to the same PO document during routing, and some of the threads appear to be retrieving and updating the PO before one of the other threads has finished doing so.

This PR fixes the issue by leveraging KEW's lock-route-header functionality. KEW allows for a document to also lock some other documents during routing, which can be specified via a particular method. An appropriate override of this method has been added to the PREQ document so that it will also lock the document representing the "current" version of the Purchase Order.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
@@ -3,11 +3,16 @@ package edu.cornell.kfs.module.purap.document;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kuali.kfs.kew.api.WorkflowDocument;
+import org.kuali.kfs.kew.framework.postprocessor.ActionTakenEvent;
+import org.kuali.kfs.kew.framework.postprocessor.DocumentRouteStatusChange;
 import org.kuali.kfs.kns.document.authorization.DocumentAuthorizer;
 import org.kuali.kfs.kns.service.DocumentHelperService;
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
```diff
@@ -272,6 +275,19 @@ public void doActionTaken(ActionTakenEvent event) {
             }
         }
 
+    @Override
+    public List<String> getWorkflowEngineDocumentIdsToLock() {
+        Stream<String> otherDocIdsToLock = Stream.of(super.getWorkflowEngineDocumentIdsToLock())
+                .flatMap(idList -> (idList != null) ? idList.stream() : Stream.empty());
+
```
nkimble, line 282: Base code has special processing in place for split PO's in class org.kuali.kfs.module.purap.document.PurchaseOrderDocument line 605. Do we need to be concerned about that here?

That class also has code in place on line 616 to allow the output of the documentNumber when debug processing is turned on. Should we consider having that in place in this method as well?

nkimble (commented):
These changes look appropriate.
I did see two items when looking at base code that we may want to consider if we believe these changes may possibly cause a database deadlock error as stated in the main user story. I am bringing these items up as strictly questions that I think we should think about before the code merge. They are not concrete requests for changes.

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
cah292: I don't anticipate any changes being needed when a PO Split Document is involved. If possible, though, it would be good to have such a situation tested to make sure the PREQs handle it accordingly, and then we can make future adjustments if issues are discovered there.

As for the document number logging, the StandardWorkflowEngine.lockAdditionalDocuments() method already logs the document IDs of the docs that it attempts to lock. However, if we decide to update the fix to lock more than one document, then we could look into adding some debug logging to show all the relevant doc IDs.

nkimble (approved):
My concerns have received reasonable explanations of how they should be addressed should they arise.

PR #1284, opened 2022-05-13
KFSPTS-22326 Add support for Request V4 API

cah292:
There are PRs for this in both cu-kfs and nonprod-sql. Please make sure both are good to go before merging. Also, please do not merge this yet until I state otherwise; I need Tami to create some more test travel requests so that I can properly test out my latest code revisions.

This is an updated version of the #1259 PR's changes. Here are some notable new changes to the setup since then:

* In response to Jay's suggestion, the parameters for from/to/last-processed date have been replaced by a num-days-old parameter instead. The query's "to" date will be the current date, and the query's "from" date will be the date that is the given number of days before the current date.
* I've updated the request processing so that it no longer uses the confusing BiFunction setup. It's been refactored to be simpler, and it should hopefully be easier to understand now. (Note, though, that I had to make some unit test adjustments to find another way of tracking the retrieved dummy search results.)
* I've made several other miscellaneous updates to simplify the code and to address the other items from the #1259 feedback. Please let me know if you notice any new issues or if you notice something from the previous PR that has not yet been addressed here.


jhulslander:
retest this please

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
```diff
@@ -0,0 +1,344 @@
...
+    protected Stream<ConcurEventNotificationProcessingResultsDTO> processTravelRequestsSubset(
+            String accessToken, Map<String, String> testUserIdMappings, ConcurRequestV4ListingDTO requestListing) {
+        if (CollectionUtils.isEmpty(requestListing.getListItems())) {
+            return Stream.empty();
+        }
+        Stream.Builder<ConcurEventNotificationProcessingResultsDTO> subResults = Stream.builder();
+        
+        for (ConcurRequestV4ListItemDTO requestAsListItem : requestListing.getListItems()) {
+            if (StringUtils.isBlank(requestAsListItem.getId())) {
+                throw new IllegalStateException("Found a request item with a blank ID; this should NEVER happen");
```
jhulslander, line 134: I agree this should never happen, but if it does happen, is there a way we can add some details here so we can identify the "bad" request.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    protected boolean requestHasAppropriateOwner(ConcurRequestV4ListItemDTO requestAsListItem,
+            Map<String, String> testUserIdMappings) {
+        ConcurRequestV4PersonDTO owner = requestAsListItem.getOwner();
+        if (ObjectUtils.isNull(owner) || StringUtils.isBlank(owner.getId())) {
+            return false;
+        }
+        boolean isTestUser = testUserIdMappings.containsKey(owner.getId());
+        return isProduction() != isTestUser;
```
jhulslander, line 160: I think this logic is correct, but it is a bit confusing.  Can it be refactored?
When in production, only allow users that aren't test users
when in non production, only allow users that are test users.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    protected ConcurEventNotificationProcessingResultsDTO processTravelRequest(String accessToken,
+            ConcurRequestV4ListItemDTO requestAsListItem) {
+        String requestUuid = requestAsListItem.getId();
+        if (!StringUtils.isAlphanumeric(requestUuid)) {
+            throw new RuntimeException("Found a request with an unexpected non-alphanumeric ID: " + requestUuid);
```
jhulslander, line 167: Are these types of checks needed?

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
```diff
@@ -0,0 +1,344 @@
...
+        
+        return resultsDTO;
+    }
+
+    protected ConcurAccountInfo buildAccountInfo(ConcurRequestV4ReportDTO request) {
+        ConcurAccountInfo accountInfo = new ConcurAccountInfo();
+        accountInfo.setChart(defaultCodeToEmptyIfAbsent(request.getChart()));
+        accountInfo.setAccountNumber(defaultCodeToEmptyIfAbsent(request.getAccount()));
+        accountInfo.setSubAccountNumber(defaultCodeToEmptyIfAbsent(request.getSubAccount()));
+        accountInfo.setObjectCode(concurBatchUtilityService.getConcurParameterValue(
```
jhulslander, line 203: don't we want to get the object code from the request, and if that is empty, then default to the parameter value

jhulslander (changes requested):
I didn't get through the testing side of the PR, but I found a few things to consider in the non-test code

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
cah292: Ok, I'll update this to log some more detailed of the invalid request.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
cah292: I'll update this to use the conditional/ternary operator or something similar; that should help make the intent clearer.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
cah292: I agree that this alphanumeric check is not needed, given that other areas of the code are already checking for a non-blank ID, and that the constructed web URLs are encoding the ID appropriately. I'll remove the alphanumeric check.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
cah292: Based on the existing code in ConcurReportsServiceImpl.extractAccountInfoFromTravelRequestDetails(), the Travel Request processing always seems to set the object code to the value from the parameter. I would need to double-check with Tami to confirm whether that should continue to remain the case. (Looking at some manually-retrieved travel requests from my earlier testing, there might be an object code in the "custom12" field, but I would need more analysis to confirm that.)

jhulslander: Yes, please double check with Tami.  I think the expense report processing looks at the object code of the report, and only uses the default if the report doesn't have an object code.  If we should just use the default object code, the expense report processing may need to be updated.  That can be done on a separate user story.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
```diff
@@ -0,0 +1,754 @@
...
+import edu.cornell.kfs.concur.rest.jsonObjects.ConcurRequestV4OperationDTO;
+import edu.cornell.kfs.concur.rest.jsonObjects.ConcurRequestV4PersonDTO;
+import edu.cornell.kfs.concur.rest.jsonObjects.ConcurRequestV4ReportDTO;
+import edu.cornell.kfs.concur.rest.jsonObjects.ConcurRequestV4StatusDTO;
+import edu.cornell.kfs.concur.service.ConcurAccountValidationService;
+import edu.cornell.kfs.sys.CUKFSConstants;
+import edu.cornell.kfs.sys.service.impl.TestDateTimeServiceImpl;
+import edu.cornell.kfs.sys.web.mock.MockRemoteServerExtension;
+
+@Execution(ExecutionMode.SAME_THREAD)
```
jhulslander, line 72: I am able to run this unit test class without this execution statement.  Is it still needed?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
```diff
@@ -0,0 +1,754 @@
...
+            KFSConstants.BLANK_SPACE,
+            "Name1=",
+            "=Value1",
+            "Name1=Value1;Name2=;Name3=Value3",
+            "Name1=Value1;Name2=Value2;=Value3"
+    })
+    void testParsingFailureForInvalidTestUserParameterValue(String invalidValue) {
+        overrideParameter(ConcurParameterConstants.REQUEST_V4_TEST_USERS, invalidValue);
+        assertThrows(RuntimeException.class, () -> requestV4Service.getRequestV4TestUserMappingsFromParameter(),
+                "The parsing of the test users parameter should have failed when its value is blank or malformed");
```
jhulslander, line 362: Should the value of invalidValue be included in this message, it might help if this test breaks in the future.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
```diff
@@ -0,0 +1,754 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("travelRequests")
```
jhulslander, line 366: I am confused how this works (this may be a good subject for a bookclub).  I think this corresponds to the static function Stream<Arguments> travelRequests(), but this function takes a RequestV4DetailFixture.  How does this work?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4Backend.java
```diff
@@ -0,0 +1,204 @@
...
+        Date modifiedAfter = getExistingDateParameter(queryParameters, ConcurApiParameters.MODIFIED_AFTER);
+        Date modifiedBefore = getExistingDateParameter(queryParameters, ConcurApiParameters.MODIFIED_BEFORE);
+        String sortField = getExistingParameter(queryParameters, ConcurApiParameters.SORT_FIELD);
+        String sortOrder = getExistingParameter(queryParameters, ConcurApiParameters.SORT_ORDER);
+        
+        Comparator<ConcurRequestV4ListItemDTO> requestSorter = Comparator.comparing(
+                ConcurRequestV4ListItemDTO::getStartDate, Comparator.reverseOrder());
+        
+        if (!StringUtils.equalsIgnoreCase(RequestV4Views.APPROVED, view)) {
+            throw new IllegalArgumentException(
```
jhulslander, line 82: Is this error handling needed?  Since this backend  is just for testing, I am wondering if it could be simplified.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4Backend.java
```diff
@@ -0,0 +1,204 @@
...
+
+    private String getExistingParameter(Map<String, String> queryParameters, String key) {
+        String value = queryParameters.get(key);
+        if (StringUtils.isBlank(value)) {
+            throw new IllegalArgumentException("Query has blank or missing parameter: " + key);
+        }
+        return value;
+    }
+
+    private static final class RequestEntry {
```
jhulslander, line 192: what is the purpose of this class, I am not understanding how it is used.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4ServiceEndpoint.java
```diff
@@ -0,0 +1,222 @@
...
+    @Override
+    protected void onServerInitialized(String baseUrl) {
+        super.onServerInitialized(baseUrl);
+        baseRequestV4Url = baseUrl + ParameterTestValues.REQUEST_V4_RELATIVE_ENDPOINT;
+        mockBackendServer = new MockConcurRequestV4Backend(baseRequestV4Url, initialRequestDetails);
+    }
+
+    public MockConcurRequestV4Backend getMockBackendServer() {
+        if (ObjectUtils.isNull(mockBackendServer)) {
+            throw new IllegalStateException("Mock back-end server was null; this endpoint may not "
```
jhulslander, line 106: similarly, can some of this error handling code be removed to simplify this mock service?

src/test/java/edu/cornell/kfs/sys/web/mock/MockRemoteServerExtension.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.http.HttpHost;
+import org.apache.http.localserver.LocalServerTestBase;
+
+import edu.cornell.kfs.sys.web.mock.MockServiceEndpointBase;
```
jhulslander, line 13: can this import be removed, it doesn't seem to be used?

src/test/java/edu/cornell/kfs/sys/web/mock/MockRemoteServerExtension.java
```diff
@@ -0,0 +1,75 @@
...
+
+// TODO: Convert this into a true JUnit 5 test extension class.
```
jhulslander, line 15: should we create a follow up user story and remove this comment?

jhulslander (changes requested):
I found a few other things to consider in the test code

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
cah292: I was concerned about whether the tests would encounter problems if parallel execution was enabled, but I agree that they seem to run fine locally with such a setup. I'll remove the annotation and see if Jenkins handles the parallel test runs properly.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
cah292: Good point, I'll add the value to the failure message.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImplTest.java
cah292: This is a JUnit 5 feature. JUnit will invoke the method referenced by the MethodSource annotation, and that method has to return a Stream of values (which can be represented as JUnit Arguments objects). Each element of that Stream will be used for a separate run of this unit test case, and the element's components will be passed into the test as method arguments.

There are some good details on this feature in the official JUnit 5 documentation, though we can also discuss this in a future bookclub meeting if desired.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4Backend.java
cah292: Some of the parameters with dynamic values could end up causing unexpected results or failures if the wrong values are passed in. For some of the parameters that are always expected to have the same values (like the "view" parameter), I could remove the error checks or reduce the exceptions to warnings.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4Backend.java
cah292: This class is meant to encapsulate the travel requests that are present on the mock backend server. Each instance contains three objects for a specific travel request: Its fixture, its list item DTO, and its report/detail DTO. It was meant to avoid reconstructing the DTOs on each query call, and to lay the groundwork for testing updates to the DTOs when a future user story implements the travel request workflow handling.

Upon further consideration, though, I might be able to simplify this by setting up some lazy DTO construction and removing this helper class altogether. I'll look into cleaning this up if possible.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4ServiceEndpoint.java
cah292: In the case of this getMockBackendServer() method, it's not actively used by the tests and won't be needed until we implement the workflow travel request handling in a future user story. I'll go ahead and remove this getter for now.

As for the getBaseRequestV4Url() method, it's extremely important that it returns the properly initialized URL; otherwise the calling code in the unit test class won't be able to invoke the local endpoint. I would prefer to keep that one in place for now.

src/test/java/edu/cornell/kfs/sys/web/mock/MockRemoteServerExtension.java
cah292: Good catch, I'll remove this import.

src/test/java/edu/cornell/kfs/sys/web/mock/MockRemoteServerExtension.java
cah292: I'll remove this comment, and update the title of sub-task KFSPTS-23928 to have an appropriate follow-up user story created for this.

cah292:
I've committed some more changes based on the most recent feedback. However, I still haven't gotten a response from Tami yet concerning the object code setup, so I may need to implement more changes depending on her reply. I also need to perform another round of local testing once any object-code-related changes are in place and Tami creates some more test requests.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/MockConcurRequestV4ServiceEndpoint.java
jhulslander: OK, thanks for explaining

jhulslander:
I think this PR looks good.  Waiting on further action till you can do the local validation you mentioned

jhulslander (approved):
changes look good, unit tests pass, and the SQL is good as well.  I shall merge

PR #1285, opened 2022-05-17
KFSPTS-24843: Remove checkSubFundGroup validation from our custom overlay of AccountGlobalRule to fix GACC not saving or submitting accounts associated with CAPITAL_SUB_FUND_GROUPS.

nkimble:


db483 (approved):
Changes look good.

PR #1289, opened 2022-05-23
KFSPTS-24821 Allow certain processes to run without Quartz

cah292:
KFS has the ability to retry certain document routing operations if they fail unexpectedly (such as due to an OptimisticLockException), but the code that does so depends on Quartz being enabled. Such processes are failing across our environments because of this, now that the KEW-upgraded code no longer prepares a KSB-related Quartz scheduler. This PR works around the problem by allowing the SchedulerService to handle the appropriate retries, and by customizing the SchedulerService to be usable without Quartz.

This PR also adds some basic code to allow a KFS instance to run batch jobs locally without Quartz, since I was relying upon a particular batch job to recreate the situation that would trigger the routing retries. (The run-local-batch-job feature is disabled by default, but it can be enabled via a config property.) The run-batch-locally feature is currently meant for developer convenience; if we want to update it to the point where we could potentially use it to rework our existing batch job configuration, then we should use a separate user story to do so.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
@@ -199,4 +199,8 @@ public class CUKFSConstants {
         public static final String LOOKUP = BASE_PATH + "/lookup";
     }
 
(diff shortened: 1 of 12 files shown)


ajd299 (approved):
Code looks good, comments are valuable and succinct.

PR #1292, opened 2022-05-25
KFSPTS-25017: Override toString method to restrict output of data values to server logs when failures occur.

nkimble:


dp65:
A minor suggestion about the language. I think I would understand "RestrictedDataMasked" or "RestrictedDataWasPresent" more readily than "RestrictedDataPresent" when I see it in the context of papertrail logs.  The data has been masked from the logs; it used to be present but isn't any more. I'm really not trying to nit-pick; I'm just saying how my brain interprets those words.

nkimble:
For consistency, I would also want to modify the language in constant PaymentWorksConstants.OUTPUT_RESTRICTED_DATA_PRESENT to match whatever language is decided on for the constant @dp65 referenced above. I am making this recommendation due to the fact that the coding change in this PR was directly modeled after the logic and wording surrounding that constant.

cah292:
@dp65 With the changes that Nancy has here, the object's toString() output will look something like the following:

```
edu.cornell.kfs.pdp.businessobject.PayeeACHAccountExtractDetail@12a34b5[
  bankAccountNumber=RestrictedDataPresent
  id=12345
  createDate=2022-05-25
  ...
]
```

@nkimble I'm fine with the output saying "RestrictedDataPresent", though I agree with Donivan that the constant's name of "OUTPUT_RESTRICTED_DATA_PRESENT" is a little confusing. Perhaps a constant name like "RESTRICTED_DATA_PLACEHOLDER" would make the intent clearer, and its value could still be set to "RestrictedDataPresent".

If needed, we can discuss this further at the next team standup.

nkimble:
@cah292 How about this as a compromise to deal with the wording issues and to use ONE constant for both locations with the same logic:

- Name the constant RESTRICTED_DATA_CONCEALED with the data value output to the log file being RestrictedDataConcealed
- Place that constant in class CUKFSConstants
- Create a followup user story to replace PaymentWorksConstants.OUTPUT_RESTRICTED_DATA_PRESENT with new constant CUKFSConstants.RESTRICTED_DATA_CONCEALED so that there is consistency in the server logs for the same logic.

db483:
Could we output just RestrictedData or replace the value with Xs ?

nkimble:
Code changes have been made as discussed in today's standup. Unit tests are running. This is ready for review again.

cah292 (approved):

PR #1294, opened 2022-06-03
KFSPTS-23422: Make KFS-PURAP component code PurchaseOrderAmendment inactive so that corresponding component annotation changes on class files correctly create derived components resulting in a single active component being returned by component lookup.

nkimble:


nkimble:
There is a corresponding SQL PR that needs to be merged at the same time as these code changes.

jhulslander (commented):
Nancy and I spoke about this and I understand it a little bit better.  These types objects automatically load parameters into the dynamically loaded table krcr_drvd_cmpnt_t.  This new annotation makes our class have the same component code as the parent kualico object.

The related SQL inactivates our manually added component to avoid end user confusion of duplication.

After further discussion, both Nancy and I aren't sure if this the best approach.  @cah292 or @db483  could you take a look at this PR as well.

cah292 (commented):
The changes look good from my perspective. This setup appears to be consistent with how we've addressed similar derived component issues on some of our other custom document classes (such as CuRequisitionDocument and CuDisbursementVoucherDocument). As long as KFS correctly retrieves only the derived version of the component once the related SQL inactivates the other component, I believe this fix will work properly.

@jhulslander Please let us know if you still want Daniela to review this before merging.

db483:
This solution looks good and is consistent with the way we have implemented this in other parts of the code, as Chad pointed out.

cah292 (approved):
Code changes look good, unit tests pass, and the related SQL PR is also ready. Merging.

PR #1298, opened 2022-06-22
KFSPTS-24770 Fix KSR read-only and superuser handling

cah292:
Security Request documents were relying upon a request-level "readOnly" variable to handle the read-only state on the downstream KSR tags, but that was causing other fields (like the note text field) to unexpectedly become read-only on uneditable documents. These documents were also missing the tag that adds the superuser tab, along with a few other misc. tags. This PR adds fixes for both of those issues.


ajd299 (approved):
Code looks good, merging

PR #1299, opened 2022-06-24
KFSPTS-25220 add chart and org qualifications

db483:


cah292 (commented):
Here are my initial comments on the changes thus far. I agree that the DD routing type attributes are fine as-is since they affect responsibilities and not permissions; sorry about putting too much emphasis on that in my earlier conversations with Daniela. To get this working as expected, though, it might make more sense to use a new simpler role instead of an org-based role, given the complexities noted in my comments below. Also, if doc-submission permissions need to be limited based on the orgs from the retirement details, then I would agree with Daniela's suggestion of adding a business rule to limit the specified assets appropriately.

db483:
retest this please

cah292:
I'm closing this PR since I'm about to cut a new develop branch from master. This PR is several months old, but if you believe its changes should still be moved forward at some point, a new PR can be opened once the new develop branch is ready and this feature branch has been rebased or recreated accordingly.

PR #1300, opened 2022-07-05
KFSPTS-23855: Remove unused modification to KFS re Incoming Wires.

nkimble:


nkimble:
non-prod SQL PR #1091 must be merged at the same time this code change is merged.

cah292:
Thanks for checking. I see that Cathy responded on the user story about this behavior being acceptable, so we should be good to proceed.

cah292 (approved):
Changes look good, unit tests pass, my concerns have been addressed and the related SQL PR is also ready. Merging.

PR #1303, opened 2022-07-21
KFSPTS-25351 Limit Bank Name to 40 Characters

ajd299:


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -338,7 +339,18 @@ protected void cleanPayeeACHAccountExtractDetail(PayeeACHAccountExtractDetail de
             if (!StringUtils.isNumeric(detail.getBankAccountNumber())) {
                 LOG.error(logMessageStarter + " is not numeric after cleaning");
             }
+
+            detail.setBankName(extractBankName(detail));
```
cah292, line 343: This statement is currently inside the "if" block that checks for a non-numeric bank account number. Should this line be moved outside of that "if" block so that the cleaning of the bank name does not depend on the account number formatting?

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -338,7 +339,18 @@ protected void cleanPayeeACHAccountExtractDetail(PayeeACHAccountExtractDetail de
...
                 LOG.error(logMessageStarter + " is not numeric after cleaning");
             }
+
+            detail.setBankName(extractBankName(detail));
+        }
+    }
+
+    private String extractBankName(PayeeACHAccountExtractDetail detail) {
+        String bankName = ObjectUtils.isNull(detail) ? StringUtils.EMPTY : detail.getBankName();
+        if (StringUtils.length(bankName) > 40) {
```
cah292, line 349: I would suggest moving the max length value to a constant or a KFS parameter.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -338,7 +339,18 @@ protected void cleanPayeeACHAccountExtractDetail(PayeeACHAccountExtractDetail de
...
+            LOG.info("cleanPayeeACHAccountExtractDetail truncating bank name to 40 characters");
```
cah292, line 350: This log statement should be updated to reference the "extractBankName" method that it's actually inside of.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -338,7 +339,18 @@ protected void cleanPayeeACHAccountExtractDetail(PayeeACHAccountExtractDetail de
...
+            bankName = bankName.substring(0, 40);
         }
+        return bankName.trim();
```
cah292, line 353: Should the bank name also be trimmed prior to performing the length check? Or do you only need it here?

Also, I think StringUtils has a trim() method on it, if you want to make this null-safe.

src/test/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImplTest.java
```diff
@@ -744,65 +741,87 @@ public void markRowAsInvalid() {
...
         payeeACHAccountExtractService.cleanPayeeACHAccountExtractDetail(detail);
         
         assertEquals(expectedCleanedBankAccountNumber, detail.getBankAccountNumber());
     }
+    private void validateCleanPayeeACHAccountExtractDetailBankName(String bankName, String expectedCleanedBankName) {
+        PayeeACHAccountExtractDetail detail = new PayeeACHAccountExtractDetail();
+        detail.setBankName(bankName);
+        payeeACHAccountExtractService.cleanPayeeACHAccountExtractDetail(detail);
+
+        assertEquals(detail.getBankName(), expectedCleanedBankName);
```
cah292, line 824: In this JUnit test assertion, the first argument should be the expected value and the second argument should be the actual value.

cah292 (changes requested):
Changes look good in general, though I found some various areas that could use cleanup.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
dp65: Ah, nice! A null-safe class method. That's an elegant solution to an unrelated issue, in a different language, that I'm having. I'm glad you suggested it here, and I'm glad I read this.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -339,6 +339,19 @@ protected void cleanPayeeACHAccountExtractDetail(PayeeACHAccountExtractDetail de
...
+
+        detail.setBankName(extractBankName(detail));
+    }
+
+    private String extractBankName(PayeeACHAccountExtractDetail detail) {
+        String bankName = ObjectUtils.isNull(detail) ? StringUtils.EMPTY : StringUtils.trim(detail.getBankName());
+
+        if (StringUtils.length(bankName) > CUPdpConstants.PAYEE_ACH_ACCOUNT_MAX_BANK_NAME_LENGTH) {
+            LOG.info("extractBankName truncating bank name to " + CUPdpConstants.PAYEE_ACH_ACCOUNT_MAX_BANK_NAME_LENGTH + " characters");
+            bankName = bankName.substring(0, CUPdpConstants.PAYEE_ACH_ACCOUNT_MAX_BANK_NAME_LENGTH);
```
cah292, line 351: Does the bank name need to be trimmed a second time after being truncated? There might be cases where the 40th character is a space between words.

cah292 (commented):
I had one more quick question about the truncation setup, but otherwise the revised changes look good.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
ajd299: This should be fixed now

PR #1302, opened 2022-07-18
KFSPTS-25480 new jaggaer role to varous links mapping

jhulslander:
This adds the framework for Jaggaer roles to links from KFS

there is a related non-prod-sql PR that needs to be reviewed at the same as this

PR #1304, opened 2022-07-21
KFSPTS-24959 Add JAXB DTOs for Jaggaer CXML Login

cah292:
NOTE: This PR's changes conflict with those from the #1302 PR. If the other PR gets merged first, then I will need to rebase this feature branch accordingly.

This PR adds the JAXB DTOs for creating a CXML-based Jaggaer Login request. It also adds a service for generating the XML and a bare-bones service for deriving the Jaggaer roles. (The implementation of the latter service will be handled by follow-up user stories.) The changes also add the service beans for these, but the services are not yet in active use.

I've set this up to try and follow most of the formatting from the XML generated by CuB2BShoppingServiceImpl, though I did make some changes to account for JAXB processing and to bring things more in line with the examples from Jaggaer's documentation. Also, for the main service method that builds the DTO structure, I tried to set it up in a way that helps convey the layout of the resulting XML. Please let me know if you think that part needs rework for simplicity or clarity.

The DTD-to-POJO processing of the related .dtd file resulted in almost 400 DTO classes, so I removed most of the unneeded DTOs and set up a dummy DTO to act as a placeholder for the removed parts. If you think there are still too many DTOs for the type of XML that we're creating, I could look into an alternative means of generating the CXML (such as by creating a handful of custom simplified JAXB DTOs and using an XSL stylesheet to transform them into the final XML). Please let me know your thoughts about the DTO setup.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/CUPurapConstants.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/CUPurapConstants.java
@@ -268,4 +268,11 @@ public class CUPurapConstants {
         public static final String ROOM_NUMBER_PREFIX = "Room ";
         public static final String VENDORS = "vendors";
     }
+
+    public enum JaggaerRoleSet {
+        ESHOP,
+        CONTRACTS_PLUS,
+        ADMINISTRATOR
+    }
+
 }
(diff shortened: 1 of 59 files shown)


cah292:
As per our recent team meeting, we have decided not to pursue the DTO solution at this time. We will instead continue to use the string concatenation like in base financials, though we will customize that code accordingly to allow for adding an arbitrary number of roles. I will create a different feature branch to implement such a feature.

PR #1305, opened 2022-07-26
KFSPTS-24959 Allow for adding multiple roles to Jaggaer CXML

cah292:
This PR adjusts the creation of the Jaggaer Login XML so that it allows for specifying an arbitrary number of roles. It also adds some minimal code to allow deriving the appropriate roles based on a specified JaggaerRoleSet, though the full implementation of that functionality is being deferred to a future user story. The JaggaerRoleSet-based derivation is not actively used by the current code here.

Our original plan was to create custom JAXB DTOs for building the Login CXML in a cleaner fashion, but that approach required creating an unexpectedly large number of DTOs. After further discussion with the team, we decided to instead make the appropriate modifications to the existing string-concatenation-based process, and we'll handle converting this to a cleaner setup once KualiCo implements a better solution on their end. If you think we should go ahead and implement our own custom JAXB/XSLT/etc. setup anyway, please let me know.

Note that this PR currently does not change the part that builds the DTD reference and the root CXML element. Please let me know if you think those sections should be updated to specify an explicit DTD URL and an explicit CXML "version" attribute, like what is shown in the sample CXML from Jaggaer's documentation.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
@@ -7,6 +7,6 @@ import edu.cornell.kfs.module.purap.CUPurapConstants.JaggaerRoleSet;
 
 public interface CuB2BShoppingService extends B2BShoppingService {
 
-    String getPunchOutUrl(Person user, JaggaerRoleSet roleSet);
+    String getPunchOutUrlForRoleSet(Person user, JaggaerRoleSet roleSet);
 
 }
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
@@ -1,0 +1,12 @@
+package edu.cornell.kfs.module.purap.document.service;
+
+import org.kuali.kfs.kim.api.identity.Person;
+import org.kuali.kfs.module.purap.document.service.B2BShoppingService;
+
+import edu.cornell.kfs.module.purap.CUPurapConstants.JaggaerRoleSet;
+
+public interface CuB2BShoppingService extends B2BShoppingService {
+
+    String getPunchOutUrl(Person user, JaggaerRoleSet roleSet);
+
+}
--- a/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
@@ -13,6 +13,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.kuali.kfs.coreservice.framework.parameter.ParameterService;
+import org.kuali.kfs.kim.api.identity.Person;
+import org.kuali.kfs.kim.api.services.KimApiServiceLocator;
 import org.kuali.kfs.krad.service.BusinessObjectService;
 import org.kuali.kfs.krad.service.DocumentService;
 import org.kuali.kfs.krad.service.PersistenceService;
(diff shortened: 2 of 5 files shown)


cah292:
@jhulslander This is ready for initial review. Be aware that this solution no longer lines up exactly with what was proposed on the user story and in the Confluence documentation, due to our team discussion about modifying the existing string-based process instead. (The user story and documentation have not yet been updated fully from today's discussion.) Please let me know if you think the changes here will still fulfill the appropriate Jaggaer XML and role requirements, or if you think we should still attempt one of the alternate XML strategies anyway.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -217,47 +225,83 @@ public void setUserFavoriteAccountService(
...
+
+    @Override
     public String getPunchOutSetupRequestMessage(Person user, B2BInformation b2bInformation) {
+        String preAuthRole = getPreAuthValue(user.getPrincipalId());
+        String viewRole = getViewValue(user.getPrincipalId());
+        List<String> roles = List.of(preAuthRole, viewRole);
+        return getPunchOutSetupRequestMessage(user, b2bInformation, roles);
+    }
+
+    protected String getPunchOutSetupRequestMessage(Person user, B2BInformation b2bInformation, List<String> roles) {
```
jhulslander, line 304: instead of taking in List<String> roles, it could take in JaggaerRoleSet roleSet

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -217,47 +225,83 @@ public void setUserFavoriteAccountService(
...
+        PunchOutSetupResponse posr = B2BParserHelper.getInstance().parsePunchOutSetupResponse(response);
+        return posr.getPunchOutUrl();
+    }
+
+    @Override
     public String getPunchOutSetupRequestMessage(Person user, B2BInformation b2bInformation) {
+        String preAuthRole = getPreAuthValue(user.getPrincipalId());
+        String viewRole = getViewValue(user.getPrincipalId());
+        List<String> roles = List.of(preAuthRole, viewRole);
+        return getPunchOutSetupRequestMessage(user, b2bInformation, roles);
```
jhulslander, line 301: this is the old eShop link function, so it could pass in JaggerRoleSet.eShop

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
```diff
@@ -312,8 +356,9 @@ public String getPunchOutSetupRequestMessage(Person user, B2BInformation b2bInfo
         // KFSPTS-1720
         cxml.append("      <Extrinsic name=\"FirstName\">").append(user.getFirstName()).append("</Extrinsic>\n");
         cxml.append("      <Extrinsic name=\"LastName\">").append(user.getLastName()).append("</Extrinsic>\n");
-        cxml.append("      <Extrinsic name=\"Role\">").append(getPreAuthValue(user.getPrincipalId())).append("</Extrinsic>\n");
```
jhulslander, line 315: I suggest doing roles the "old" way for the roleSet is eShop.  Otherwise use jaggaerRoleService get a list of roles, and then loop as you have coded below

jhulslander (changes requested):
I had a few ideas to help us make sure we don't change how eShop works while we add the other links.  Let me know what you think

src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
```diff
@@ -0,0 +1,12 @@
+package edu.cornell.kfs.module.purap.document.service;
+
+import org.kuali.kfs.kim.api.identity.Person;
+import org.kuali.kfs.module.purap.document.service.B2BShoppingService;
+
+import edu.cornell.kfs.module.purap.CUPurapConstants.JaggaerRoleSet;
+
+public interface CuB2BShoppingService extends B2BShoppingService {
+
+    String getPunchOutUrl(Person user, JaggaerRoleSet roleSet);
```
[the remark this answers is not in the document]
cah292: Perhaps a name like "getPunchOutUrlForRoleSet" would help?

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
cah292: I'll adjust this to accept a role set instead.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
cah292: I'll update this to pass in the role set instead.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuB2BShoppingServiceImpl.java
cah292: I'll rework this to implement the old/new split like you had suggested.

src/main/java/edu/cornell/kfs/module/purap/document/service/CuB2BShoppingService.java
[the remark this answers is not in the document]
jhulslander: Yes that would work

cah292:
I've committed some changes to address the items that were brought up in the initial feedback. This is ready for review again.

jhulslander (commented):
These changes look good, I will merge after the unit tests pass

jhulslander (approved):

PR #1306, opened 2022-07-28
KFSPTS-25158 Fix display of Jaggaer roles in KIM

cah292:

While preparing the Jaggaer KIM role SQL, I had discovered that some missing Data Dictionary configuration was preventing the Jaggaer Role attribute from displaying properly on the Permission lookup. This PR adds the needed DD beans to fix that, and bundles it with the main KIM-role-related SQL changes for that user story.

If you think this fix should be separated into a separate ticket, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/identity/CuPurapKimAttributes.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/identity/CuPurapKimAttributes.java
@@ -1,0 +1,20 @@
+package edu.cornell.kfs.module.purap.identity;
+
+import org.kuali.kfs.module.purap.identity.PurapKimAttributes;
+
+public class CuPurapKimAttributes extends PurapKimAttributes {
+    private static final long serialVersionUID = 5268027448509178843L;
+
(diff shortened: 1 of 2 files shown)


jhulslander (commented):
These changes look fine and can be merged when the SQL is ready.  
I just want to confirm I understand this thought.  Is the point of this BO and data dictionary XML, to validate the jaggaerRole entered on permission edit screen based on the rules in the DD?  Does this do anything else?

cah292:
The addition of this attribute to the data dictionary handles at least two things. It allows for proper business rule validation of the attribute on Permission docs that use the related KIM type, without which it might not be possible to add/edit Jaggaer permissions through the UI. It also fixes the labeling of the permissions' Jaggaer Role attributes when viewing them through the Permission lookup.

cah292:
In that case, the prior setup must have had enough configuration to allow for the business rules to pass. The changes from this PR should still fix the display issues on the Permission lookup, though.

jhulslander (approved):

PR #1308, opened 2022-08-01
KFSPTS-25667 update pom file to no longer reference nexus.kuali.org and use nexus.kuali.co instead

jhulslander:


pom.xml
```diff
@@ -364,7 +364,7 @@
         <repository>
             <id>foundation.nexus</id>
             <name>Kuali Foundation Nexus Group</name>
-            <url>http://nexus.kuali.org/content/groups/public</url>
```
db483, line 367: Looking at the KualiCo financials repo, I see that KualiCo used the http version of the url. Also, with https://github.com/KualiCo/financials/pull/4488/files they removed this entry in Sept 2021: FINI-8974: Upgrade maven to 3.8.2. I wonder if we can also eliminate it or if we have to keep it if it should be the http version? Thank you.

dp65: Hope you don't mind me sticking my nose in here. I am unable to browse to a https version. https://nexus.kuali.org/content/groups/public
I get a "this site cannot be reached" error. Does Kuali support anything but the http version?

db483: Good catch Donivan. This https://nexus.kuali.co/nexus/content/groups/public/ works which is what KualiCo uses for KualiCo Nexus Group. However for the Foundation group they only used http.

jhulslander: @dp65 my original goal was to switch from kuali.org to kuali.co version version of this entry.  I thought it would make sense to switch to HTTPS since the co version works with HTTPS.

@db483  I will try remove the entire entry, and see what happens

jhulslander:
removing the entry entirely seemed to be just fine, so I think we can proceed like this, if the unit tests pass .... this is ready for review again

db483 (approved):
Changes look good.

PR #1310, opened 2022-08-03
KFSPTS-25260: Generic ISO-to-FIPS Country code conversion mod.

nkimble:
There are PRs for this change in both cu-kfs and nonprod-sql. Please make sure both PRs are good to go before merging.

This PR adds maintenance documents for adding/editing the generic ISO countries as well as one for mapping between these generic ISO countries and the existing FIPS countries.  The new KualiCo framework was used for these maintenance documents. The access security from the existing (FIPS) country maintenance document was use for these document's so that the same functional users would have access to both the FIPS and ISO versions.

Active indicators were added to both generic ISO country maintenance documents. The current PaymentWorks version of this country mapping does not have/use active indicators so that nuance was not included in the generic version of the service implementation logic but utilizing that flag could be added if we found it was needed.

The existing PaymentWorks mapping table contents was used to establish the baseline generic ISO-to-FIPS mapping to keep things as consistent as possible. 

I am still working through some data differences related to the ISO countries in the related SQL PR so please verify with me that all SQL is present before merging.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
@@ -97,5 +97,6 @@ public class CUKFSPropertyConstants {
     public static class ISOFIPSCountryMap {
         public static final String ISO_COUNTRY_CODE = "isoCountryCode";
         public static final String FIPS_COUNTRY_CODE = "fipsCountryCode";
+        public static final String ACTIVE = "active";
     }
 }
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
@@ -93,4 +93,9 @@ public class CUKFSPropertyConstants {
 
     public static final String DOCUMENT_ID = "documentId";
     public static final String FINALIZED_DATE = "finalizedDate";
+
+    public static class ISOFIPSCountryMap {
+        public static final String ISO_COUNTRY_CODE = "isoCountryCode";
+        public static final String FIPS_COUNTRY_CODE = "fipsCountryCode";
+    }
 }
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
@@ -97,6 +97,6 @@ public class CUKFSPropertyConstants {
     public static class ISOFIPSCountryMap {
         public static final String ISO_COUNTRY_CODE = "isoCountryCode";
         public static final String FIPS_COUNTRY_CODE = "fipsCountryCode";
-        public static final String ACTIVE = "active";
+        public static final String ACTIVE = KFSPropertyConstants.ACTIVE;
     }
 }
--- a/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
@@ -1,0 +1,20 @@
+package edu.cornell.kfs.sys.service;
+
+/*
+ * This is the generic solution create by Cornell to deal with converting between
+ * many ISO country codes mapping to a single FIPS country code. 
+ * Details of this implmentation are documented in KFSPTS-25260.
+ */
+public interface ISOFIPSConversionService {
+
+    /*
+     * ISO-to-FIPS mappings are many-ISO-to-one-FIPS.
+     */
+	public String convertISOCountryCodeToFIPSCountryCode(String isoCountryCode);
+
+	/*
+	 * FIPS-to-ISO mappings are one-FIPS-to-many-ISO
+	 */
+	public String convertFIPSCountryCodeToISOCountryCode(String fipsCountryCode);
+
+}
--- a/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
@@ -1,15 +1,17 @@
 package edu.cornell.kfs.sys.service;
 
+import java.util.List;
+
+import edu.cornell.kfs.sys.businessobject.ISOFIPSCountryMap;
+
 public interface ISOFIPSConversionService {
 
-    /*
-     * ISO-to-FIPS mappings are many-ISO-to-one-FIPS.
-     */
-	public String convertISOCountryCodeToFIPSCountryCode(String isoCountryCode);
+    public String convertISOCountryCodeToActiveFIPSCountryCode(String isoCountryCode);
 
-	/*
-	 * FIPS-to-ISO mappings are one-FIPS-to-many-ISO
-	 */
-	public String convertFIPSCountryCodeToISOCountryCode(String fipsCountryCode);
+    public String convertFIPSCountryCodeToActiveISOCountryCode(String fipsCountryCode);
+
+    public List<ISOFIPSCountryMap> findManyActiveISOCountryCodesForFIPSCode(String fipsCountryCode);
+
+    public List<ISOFIPSCountryMap> findManyActiveFIPSCountryCodesForISOCode(String isoCountryCode);
 
 }
--- a/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
@@ -1,10 +1,5 @@
 package edu.cornell.kfs.sys.service;
 
-/*
- * This is the generic solution create by Cornell to deal with converting between
- * many ISO country codes mapping to a single FIPS country code. 
- * Details of this implmentation are documented in KFSPTS-25260.
- */
 public interface ISOFIPSConversionService {
 
     /*
--- a/src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
@@ -1,8 +1,5 @@
 package edu.cornell.kfs.sys.service.impl;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
 
 import org.apache.log4j.LogManager;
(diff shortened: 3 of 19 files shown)


cah292 (changes requested):
Here are my initial thoughts and suggestions concerning the ISO-to-FIPS mapping code.

nkimble (commented):
Please do not merge these changes yet. I am still verifying locally but wanted to give you the opportunity to review again.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        
+        for (Iterator iter = fipsCodesFound.iterator(); iter.hasNext();) {
+            ISOFIPSCountryMap fipsCodeMapping = (ISOFIPSCountryMap) iter.next();
+            mappingsFound.add(fipsCodeMapping);
+        }
+        
+        if (mappingsFound.isEmpty()) {
+            LOG.error("convertISOCountryCodeToFIPSCountryCode: No ISO-to-FIPS Country mapping found for ISO Country code : " + isoCountryCode);
+            throw new RuntimeException("No ISO-to-FIPS Country mapping found for ISO Country code : " + isoCountryCode);
+        } else if (mappingsFound.size() > 1) {
```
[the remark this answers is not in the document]
cah292: Thanks for the explanation. I'm fine with further functional or technical adjustments being handled in future user stories.

src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
```diff
@@ -0,0 +1,15 @@
+package edu.cornell.kfs.sys.service;
+
+public interface ISOFIPSConversionService {
+
+    /*
+     * ISO-to-FIPS mappings are many-ISO-to-one-FIPS.
```
cah292, line 6: In the service interface's comments, the "one" side and "many" side are the opposite of what you stated in the service impl Javadocs. Can you please update or remove these interface comments accordingly?

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+ *****************************************************************************/
+
+public class ISOFIPSConversionServiceImpl implements ISOFIPSConversionService {
+    
+    private static final Logger LOG = LogManager.getLogger(ISOFIPSConversionServiceImpl.class);
+    
+    private ISOFIPSCountryMapDao isoFipsCountryMapDao;
+    
+    /*
+     * ISO-to-FIPS mappings are many-ISO-to-one-FIPS.
```
cah292, line 69: The "many" and "one" sides in this comment seem to contradict the direction stated in the larger Javadoc comment. Please update or remove this comment accordingly.

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOCountry.xml
```diff
@@ -0,0 +1,192 @@
...
+            <list>
+                <ref bean="ISOCountry-code"/>
+                <ref bean="ISOCountry-name"/>
+                <ref bean="ISOCountry-alternateCode"/>
+                <ref bean="ISOCountry-active"/>
+            </list>
+        </property>
+    </bean>
+
+    <bean id="ISOCountry-inquirySectionDefinition" parent="ISOCountry-inquirySectionDefinition-parentBean"/>
```
cah292, line 123: If you want to remove the older-style inquiry setup, you should remove the legacy "[BOName]-inquirySectionDefinition" bean as well as the "inquirySections" property setting from the related Inquiry Definition bean. One of your intermediate commits removed the entire Inquiry Definition bean(s) by mistake, which would explain why the KFS-startup-dependent unit tests failed until you added those beans back in.

Similar adjustments can also be made to the country-code-mapping BO's beans.

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOFIPSCountryMap.xml
```diff
@@ -0,0 +1,191 @@
...
+          parent="BusinessObjectEntry"
+          p:actionsProvider-ref="businessObjectActionsProvider"
+          p:businessObjectAdminService-ref="defaultBoAdminService"
+          p:businessObjectClass="edu.cornell.kfs.sys.businessobject.ISOFIPSCountryMap"
+          p:inquiryDefinition-ref="ISOFIPSCountryMap-inquiryDefinition"
+          p:lookupDefinition-ref="ISOFIPSCountryMap-lookupDefinition"
+          p:name="ISOFIPSCountryMap"
+          p:objectLabel="ISOFIPSCountryMap"
+          p:searchService-ref="defaultSearchService"
+          p:titleAttribute="ISOFIPSCountryMap"
```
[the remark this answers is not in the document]
cah292: Does this "titleAttribute" still need adjustment to reference an actual or derived property? Or is the BO still fully working as expected without such a change?

src/main/resources/edu/cornell/kfs/sys/cu-ojb-sys.xml
```diff
@@ -273,5 +273,26 @@
...
         <field-descriptor name="lastUpdatedTimestamp" column="LAST_UPDT_TS" jdbc-type="TIMESTAMP" index="true"/>
     </class-descriptor>
-
+    
+    <class-descriptor class="edu.cornell.kfs.sys.businessobject.ISOFIPSCountryMap" table="CU_ISO_FIPS_CNTRY_MAP_T">
+        <field-descriptor name="isoCountryCode" column="ISO_POSTAL_CNTRY_CD" jdbc-type="VARCHAR" primarykey="true" indexed="true"/>
+        <field-descriptor name="fipsCountryCode" column="FIPS_POSTAL_CNTRY_CD" jdbc-type="VARCHAR" primarykey="true" indexed="true" />
+        <field-descriptor name="objectId" column="OBJ_ID" jdbc-type="VARCHAR" />
+        <field-descriptor name="versionNumber" column="VER_NBR" jdbc-type="BIGINT" locking="true" />
+        <field-descriptor name="active" column="ACTV_IND" jdbc-type="varchar"
```
cah292, line 282: You may want to uppercase "VARCHAR" for consistency.

src/main/resources/edu/cornell/kfs/sys/cu-ojb-sys.xml
```diff
@@ -273,5 +273,26 @@
         <field-descriptor name="active" column="ACTV_IND" jdbc-type="VARCHAR" conversion="org.kuali.kfs.core.framework.persistence.ojb.conversion.OjbCharBooleanConversion"/>
         <field-descriptor name="lastUpdatedTimestamp" column="LAST_UPDT_TS" jdbc-type="TIMESTAMP" index="true"/>
     </class-descriptor>
-
+    
+    <class-descriptor class="edu.cornell.kfs.sys.businessobject.ISOFIPSCountryMap" table="CU_ISO_FIPS_CNTRY_MAP_T">
+        <field-descriptor name="isoCountryCode" column="ISO_POSTAL_CNTRY_CD" jdbc-type="VARCHAR" primarykey="true" indexed="true"/>
+        <field-descriptor name="fipsCountryCode" column="FIPS_POSTAL_CNTRY_CD" jdbc-type="VARCHAR" primarykey="true" indexed="true" />
+        <field-descriptor name="objectId" column="OBJ_ID" jdbc-type="VARCHAR" />
```
cah292, line 280: This property should set the indexed="true" attribute.

cah292 (changes requested):
Changes look good overall, though I did find a few more things that should be cleaned up.

src/main/java/edu/cornell/kfs/sys/service/ISOFIPSConversionService.java
nkimble: I will remove.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
nkimble: I will remove.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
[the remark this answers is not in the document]
nkimble: I thought about this a little more and I do have a concern that future user stories will put the logic that should be in their service implementation class (i.e. Jaggaer, Tax, PaymentWorks,...) into this generic data lookup class making this generic solution too specific thus tying it directly to those other implementations which we do not want.

Suggestion : What if I did the following? 

1. Created 4 exceptions that were extended from RuntimeException (NoISOtoFIPSMappingException, ManyISOtoFIPSMappingException, NoFIPStoISOMappingException, ManyFIPStoISOMappingException).

2. Created two new methods that returned the list of country code mappings when many were found (findManyISOCountryCodesForFIPSCode, findManyFIPSCountryCodesForISOCode).

This would allow the calling service implementations (i.e. Jaggaer, Tax, PaymentWorks,...) to identify those specific conditions when they happened, have the ability to request the data they need, and have specific/unique logic to deal with those anomalies for that part of the functionality when those exceptions occurred since the logic may be different for each area.   Thoughts??

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOCountry.xml
nkimble: Fixing both.

src/main/resources/edu/cornell/kfs/sys/cu-ojb-sys.xml
nkimble: Fixing.

src/main/resources/edu/cornell/kfs/sys/cu-ojb-sys.xml
nkimble: Fixing.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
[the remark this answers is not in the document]
cah292: Yes, the solution you suggested sounds like a good idea. Thanks!

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOFIPSCountryMap.xml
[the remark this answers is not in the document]
nkimble: The local tests I have performed are showing the BO working as expected thus far with that value for the titleAttribute.  Would it make sense to keep it as it is for now and make adjustments when it shows to not work?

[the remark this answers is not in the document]
cah292: Given the successes you reported, I'm okay with leaving this as-is for now. (Some other KFS BOs might also still be using similar configuration.) If we do end up finding problems with it down the road, we can revisit it in a future user story. Thanks for checking.

nkimble:
This PR and the corresponding SQL PR are both ready for review again.

src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
```diff
@@ -93,4 +93,10 @@ public class CUKFSPropertyConstants {
 
     public static final String DOCUMENT_ID = "documentId";
     public static final String FINALIZED_DATE = "finalizedDate";
+
+    public static class ISOFIPSCountryMap {
+        public static final String ISO_COUNTRY_CODE = "isoCountryCode";
+        public static final String FIPS_COUNTRY_CODE = "fipsCountryCode";
+        public static final String ACTIVE = "active";
```
cah292, line 100: If you wanted to simplify the setup of the property constants, you could replace this custom "active" property reference with the use of the KFSPropertyConstants.ACTIVE constant.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private ISOFIPSCountryMapDao isoFipsCountryMapDao;
+    
+    /**
+     * Return the single ACTIVE FIPS country code mapped to the ISO country code input parameter;
+     * Otherwise throw a NoISOtoFIPSMappingException or ManyISOtoFIPSMappingException.
+     *
+     * Active status of the MAPPING IS USED for this data retrieval.
+     */
+    public String convertISOCountryCodeToActiveFIPSCountryCode(String isoCountryCode) {
```
cah292, line 86: You should put the "Override" annotation on the various service method implementations.

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOCountry.xml
```diff
@@ -0,0 +1,169 @@
...
+                <ref bean="ISOCountry-code"/>
+                <ref bean="ISOCountry-name"/>
+                <ref bean="ISOCountry-alternateCode"/>
+                <ref bean="ISOCountry-active"/>
+            </list>
+        </property>
+    </bean>
+
+    <!-- Business Object Lookup Definition -->
+    <bean id="ISOCountry-inquiryDefinition" parent="ISOCountry-inquiryDefinition-parentBean"/>
```
cah292, line 124: This comment is out of place. The comment should be moved over to the lookup bean instead of the inquiry bean.

cah292 (commented):
I found a few minor cleanup items, but otherwise the changes look good. If you would rather defer their cleanup to the new follow-up user story you had created, that's fine with me.

src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
nkimble: adjusting

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSConversionServiceImpl.java
nkimble: adding annotation

src/main/resources/edu/cornell/kfs/sys/businessobject/datadictionary/ISOCountry.xml
nkimble: moved

cah292 (approved):
Changes look good, unit tests pass, and the related SQL changes are also ready. Merging.

PR #1312, opened 2022-08-05
KFSPTS-25494 enable contracts plus link

jhulslander:
Please note, there is also a related PR in non-prod-sql, please don't merge this PR till both are ready


cah292 (approved):
Changes look good, unit tests pass, and the related SQL PR is also ready. Merging.

PR #1314, opened 2022-08-10
KFSPTS-25538 SubFundGroup & HigherEducationFunction Long Description

ajd299:


cah292:
Moving the problematic fields to the bottom of the inquiry might not work if the text is too long. I noticed that the inquiry/details page for Message of the Day has the same problem, and even though the long field is at the bottom, it's getting truncated in DEV and TEST so that you can't see the full value. (I've noticed this in both Edge and Chrome; I haven't tried Firefox yet.) Please let me know whether or not you're seeing those truncation problems on the two updated inquiries; if you are then please look at my notes below, if not then your workaround could be enough for now.

On the user story's analysis sub-task, I noticed that Jay identified the KualiCo PR for FINP-7304 that introduced the breaking change. Upon further analysis, it seems KualiCo added a follow-up change for FINP-7455 that may have fixed the problem. If possible, we may want to use a variant of the KualiCo follow-up fix instead.

I don't think we can override the related "field" CSS class in "style.css" directly, due to how KualiCo's build process is bundling/packing the new UI configuration. Using my browser's developer tools, though, I think the bundled/packaged version of that CSS class is named "r3eT5kjcRPI47GrH-1GV9". If we could override that CSS class in an appropriate CSS file (either a new one or an overlay of an existing one), we might be able to put in our own temporary fix.

KualiCo's FINP-7455 fix changes that CSS class's "height: 3.1rem;" setting to "height: fit-content;" instead. If that doesn't work, you could omit the "height" setting or rename it to "min-height". If you're still not able to get this to work, or if you believe your existing workaround is still able to avoid truncation issues for long descriptions, please let me know.

Below is how the CSS class shows up in my browser's developer tools, plus a suggested alternate override:

```css
.r3eT5kjcRPI47GrH-1GV9 {
    border-bottom: 1px solid #AAAAAA;
    height: 3.1rem;
    margin-bottom: 16px;
    padding-bottom: 4px;
}

/* If adding and modifying the entry above doesn't work, you could try something like the one below: */

div.r3eT5kjcRPI47GrH-1GV9 {
    height: fit-content !important;
}
```

ajd299:
@cah292 Is there an existing .css file in cu-kfs that I could try putting this code in?

cah292:
It looks like we're not adding or overlaying any CSS files in cu-kfs at the moment. Given that, you could try one of the following setups:

* Overlay the "newPortal.css" file and add your changes there.

--**OR**--

* Create a new CU-specific CSS file under "src/main/webapp/css" and add your changes there.
* Overlay the "webapp.jsp" file and add a "link" element in the head section that loads the custom CSS file.

PR #1315, opened 2022-08-10
KFSPTS-25823 Update old Concur event job to use new authentication

cah292:
This PR fixes our ongoing Concur Event Notification Job (CN-07-01) problems by reconfiguring it to use Concur's newer OAuth2 setup.

To minimize the number of required code changes, I added a variable to ConcurReportsService to temporarily store the appropriate token. This variable will get populated when CN-07-01 starts and will get cleared when CN-07-01 finishes.

I'm currently only able to test out a portion of the changes locally, to prevent accidental updates to the failed events queue. I was at least able to confirm that the failed events queue can be successfully queried using the newer OAuth2 token. The rest of the process might not be fully testable in a non-PROD system; we may have to move things into PROD to verify that the Concur workflow updates to expense reports and travel requests are working as expected.

I already verified locally that I can cherry-pick this PR's commit to a feature branch based off of our 2022-08-07 KFS tag. These changes should be ready for a manual merge into an emergency fix branch once merged into develop.

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/batch/ConcurEventNotificationProcessingStep.java
+++ b/src/main/java/edu/cornell/kfs/concur/batch/ConcurEventNotificationProcessingStep.java
@@ -2,13 +2,17 @@ package edu.cornell.kfs.concur.batch;
 
 import java.util.Date;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.kuali.kfs.sys.batch.AbstractStep;
 
 import edu.cornell.kfs.concur.batch.service.ConcurEventNotificationProcessingService;
 import edu.cornell.kfs.concur.service.ConcurReportsService;
 
 public class ConcurEventNotificationProcessingStep extends AbstractStep {
-    
+
+    private static final Logger LOG = LogManager.getLogger();
+
     protected ConcurEventNotificationProcessingService concurEventNotificationProcessingService;
     protected ConcurReportsService concurReportsService;
 
@@ -17,6 +21,9 @@ public class ConcurEventNotificationProcessingStep extends AbstractStep {
         try {
             concurReportsService.initializeTemporaryAccessToken();
             concurEventNotificationProcessingService.processConcurEventNotifications();
+        } catch (RuntimeException e) {
+            LOG.error("execute, An unexpected error occurred during event notification processing", e);
+            throw e;
         } finally {
             concurReportsService.clearTemporaryAccessToken();
         }
--- a/src/main/java/edu/cornell/kfs/concur/batch/ConcurEventNotificationProcessingStep.java
+++ b/src/main/java/edu/cornell/kfs/concur/batch/ConcurEventNotificationProcessingStep.java
@@ -3,24 +3,23 @@ package edu.cornell.kfs.concur.batch;
 import java.util.Date;
 
 import org.kuali.kfs.sys.batch.AbstractStep;
-import org.kuali.kfs.sys.context.SpringContext;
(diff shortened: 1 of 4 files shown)


src/main/java/edu/cornell/kfs/concur/service/impl/ConcurReportsServiceImpl.java
```diff
@@ -31,22 +33,60 @@
...
 import edu.cornell.kfs.sys.util.CURestClientUtils;
 
 public class ConcurReportsServiceImpl extends DisposableClientServiceImplBase implements ConcurReportsService, DisposableBean {
-    
-	private static final Logger LOG = LogManager.getLogger(ConcurReportsServiceImpl.class);
-    protected ConcurAccessTokenService concurAccessTokenService;
+
+    private static final Logger LOG = LogManager.getLogger(ConcurReportsServiceImpl.class);
+
+    private static final String BEARER_AUTHENTICATION_SCHEME = "Bearer";
```
jhulslander, line 46: we have a constant for this: ConcurConstants.BEARER_AUTHENTICATION_SCHEME

jhulslander (changes requested):
These changes look good for the most part, I found a few items to consider.  I may have some testing ideas, let me take a closer look

src/main/java/edu/cornell/kfs/concur/batch/ConcurEventNotificationProcessingStep.java
```diff
@@ -3,24 +3,23 @@
...
 
     @Override
     public boolean execute(String jobName, Date jobRunDate) throws InterruptedException {
-        concurAccessTokenService.refreshAccessToken();
-        concurEventNotificationProcessingService.processConcurEventNotifications();
-       
+        try {
+            concurReportsService.initializeTemporaryAccessToken();
+            concurEventNotificationProcessingService.processConcurEventNotifications();
+        } finally {
```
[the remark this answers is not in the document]
cah292: Good point, I'll add some error logging here.

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurReportsServiceImpl.java
cah292: The constant ConcurConstants.BEARER_AUTHENTICATION_SCHEME was added by the KFSPTS-22326 changes, which are not in PROD yet and will be going out with the 08/28/2022 patch instead. Since I intend to use this PR's commit for the emergency patch branch, I would prefer to use this separate constant to avoid merge conflicts. We could clean this up at a future date if needed.

Please let me know if you have a suggestion on an alternate way to handle this.

jhulslander:
Upon further investigation, I am not sure I have really good answers to testing.  If you look in ConcurEventNotificationServiceImpl.findConcurFailedEventQueueProcessingMode, you can see the different supported modes for looking at the failed event queue.

Based on the changes in this PR, I feel fairly confident that if we can call the failed event queue with the authentication, we should be able to do that with expense and travel

src/main/java/edu/cornell/kfs/concur/service/impl/ConcurReportsServiceImpl.java
jhulslander: that makes total sense, I agree with sticking with this local constant.  Thanks for explaining

cah292:
I've committed some changes based on the given feedback. Also, I've prepared some tentative validation steps on the KFSPTS-25855 sub-task.

jhulslander:
Changes look good, I will merge once the unit tests complete

jhulslander (approved):

PR #1316, opened 2022-08-17
KFSPTS-25538 SubFundGroup Fix Long Description Overflow

ajd299:


src/main/webapp/css/newPortal.css
```diff
@@ -0,0 +1,530 @@
...
+    }
+    .content.fullwidth {
+        margin-left: 0;
+    }
+    #view_div {
+        padding: 10px;
+    }
+}
+
+/* KFSPTS- prevent long description overflow (Cornell customization) */
```
cah292, line 524: Please update the comment to include the rest of the JIRA number.

cah292 (commented):
There's one comment that could use some minor cleanup, but other than that the changes look good.

I also noticed that a 2022 version of this file is being overlayed. However, it looks like the only thing KualiCo changed on it after the 06/17/2021 patch was the license header, so I'm fine with the non-customized parts of this file remaining as-is.

cah292 (approved):

PR #1321, opened 2022-08-29
KFSPTS-25728 Fix Security Model multi-Definition issue

cah292:
There seems to be a bug on the Security Model document, where the duplicate-Model-Definition check fails unexpectedly when multiple such BOs have both the same Definition ID and the same default null/uninitialized primary key (Model Definition ID). This PR adds a fix that auto-populates the latter ID when one of those BOs is newly added on the document.

We would like to notify KualiCo of this issue so that they can potentially fix it on their end, but our custom fix should be enough in the meantime.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sec/document/CuSecurityModelMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/sec/document/CuSecurityModelMaintainableImpl.java
@@ -1,13 +1,53 @@
 package edu.cornell.kfs.sec.document;
 
+import java.util.List;
+
+import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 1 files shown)


db483 (approved):
Changes look good.

db483:
@cah292 changes look good. Please could you provide steps for reproducing this in monsters and any other details you think may be helpful so that I can report it to KualiCo? Thank you.

cah292:
@db483 I've added a comment to KFSPTS-25728 with details on how to tentatively recreate the issue, as well as a short explanation as to why the error appears to be happening. Note that the KualiCo Monsters environment might not have the appropriate workflow doc types in place to initiate a Security Model maintenance document, but one of KualiCo's other environments could potentially be used instead.

PR #1322, opened 2022-08-31
KFSPTS-25897 Remove CXF Web Service

ajd299:


ajd299:
retest this please

src/main/java/edu/cornell/kfs/fp/document/service/impl/CULegacyTravelServiceImpl.java
```diff
@@ -193,41 +85,13 @@ public String getTravelUrl() {
...
-                cuLegacyTravelIntegrationInterace.setTripAssociationStatusCode(TRIP_ASSOCIATIONS.IS_NOT_TRIP_DOC);
-                cuLegacyTravelIntegrationInterace.setTripId(StringUtils.EMPTY);
-            } else {
-                cuLegacyTravelIntegrationInterace.setTripAssociationStatusCode(TRIP_ASSOCIATIONS.IS_TRIP_DOC);
-                cuLegacyTravelIntegrationInterace.setTripId(tripId);
-            }
-        }
-        boolean retVal = StringUtils.equals(TRIP_ASSOCIATIONS.IS_TRIP_DOC, 
-                cuLegacyTravelIntegrationInterace.getTripAssociationStatusCode());
-        return retVal;
```
cah292, line 230: If it's not practical to remove the last few methods of the CULegacyTravelService, then I would suggest keeping this boolean check in place for compatibility with older legacy-travel-related docs (unless the functionals don't care about seeing the related travel links on old DVs or DIs), and also create a follow-up user story to look into removing the last few legacy travel remnants from KFS.

cah292 (changes requested):
Changes look good overall, though I did notice a few areas for additional cleanup.

cah292:
@ajd299 If the unit tests are passing when those closeQuietly() calls are removed, then it's possible that the Jersey framework is auto-closing the Response object when the Client gets closed, which is different than the old CXF behavior.

There are two different workarounds I can think of:

1. Refactor the code so that the Client stays open until after the Response has been processed, then close the Response, then close the Client. ConcurAccessTokenV2ServiceImpl has an example of this approach.
   * If the PaymentWorks services are required to maintain very different ClientConfig setups and can't share a common one with all the relevant features already enabled, then you may have to go with this approach instead of the one below.
2. We have a base class called DisposableClientServiceImplBase, which lazily initializes the JAX-RS Client object when necessary and then auto-closes it when the Spring context shuts down. You could update the appropriate PaymentWorks service(s) to extend from this class and have them rely on the inherited getClient() method to retrieve the Client instance. One example area that's actively using this feature is ConcurReportsServiceImpl.
   * Note that the base class may need changes to separate its ClientConfig setup into a different overridable method, so that you can include the extra MultiPartFeature setup that the PaymentWorks service requires.
   * In addition, you will likely have to update the relevant PaymentWorks unit tests so that in the after-test phase, the service's destroy() method gets explicitly invoked to clean up the Client instance.

Regardless of which workaround is chosen, the web-related PaymentWorks unit tests are simple enough to run directly from within Eclipse, so you should be able to quickly test the changes locally before committing them.

ajd299:
Thank you for the direction, please see my latest commit

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -48,9 +44,8 @@
 import edu.cornell.kfs.sys.CUKFSConstants;
 import edu.cornell.kfs.sys.service.WebServiceCredentialService;
 import edu.cornell.kfs.sys.util.CURestClientUtils;
-import edu.cornell.kfs.sys.web.CuMultiPartWriter;
 
-public class PaymentWorksWebServiceCallsServiceImpl implements PaymentWorksWebServiceCallsService, Serializable {
+public class PaymentWorksWebServiceCallsServiceImpl extends DisposableClientServiceImplBase implements PaymentWorksWebServiceCallsService, Serializable {
```
cah292, line 48: Now that this service relies on the lazy-loaded Client feature, please update the corresponding PaymentWorks unit tests so that they call the destroy() method afterwards to clean up the Client instance.

cah292 (changes requested):
There's a little more cleanup that should be done as a result of the PaymentWorks service refactoring, but otherwise that part of the changes looks good.

Also, since it seems you've decided to remove the rest of the legacy travel service, I noticed some additional areas that will need cleanup.

src/main/java/edu/cornell/kfs/fp/document/web/struts/CuDisbursementVoucherForm.java
```diff
@@ -34,36 +33,13 @@ public boolean isAlumni() {
         CuDisbursementVoucherPayeeDetail dvPayeeDetail = (CuDisbursementVoucherPayeeDetail) disbursementVoucherDocument.getDvPayeeDetail();
         return dvPayeeDetail.isAlumni();
     }
-    public boolean getCanViewTrip() {
```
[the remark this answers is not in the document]
ajd299: Okay I have asked Cathy/Dave/Kris. Is struts automatically converted "canViewTrip" to the getter getCanViewTrip()?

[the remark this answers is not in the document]
cah292: Thanks for the follow-up with the functionals. And yes, the Struts "canViewTrip" property reference should automatically link to the getCanViewTrip() getter accordingly.

ajd299:
Please see my changes in response to your feedback.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -198,18 +190,10 @@ private URI buildPaymentWorksNewVendorUpdateStatusURI() {
...
-            ClientConfig clientConfig = new ClientConfig();
-            client = ClientBuilder.newClient(clientConfig);
-            Invocation request = buildJsonClientRequest(client, uri, jsonString);
-            response = request.invoke();
-            response.bufferEntity();
-            return response;
-        } finally {
-            CURestClientUtils.closeQuietly(client);
-        }
+        Invocation request = buildJsonClientRequest(getClient(MultiPartFeature.class), uri, jsonString);
```
cah292, line 193: This PaymentWorks service has a mix of `getClient()` and `getClient(MultiPartFeature.class)` calls. To ensure that the cached Client always has the MultiPartFeature registered, please update the code so that `getClient(MultiPartFeature.class)` will always be invoked accordingly.

To make the setup much simpler, you could have this class override the `getClient()` method so that it calls `getClient(MultiPartFeature.class)` explicitly. That way, the rest of the code can call the simpler `getClient()` method and still register the appropriate class.

src/main/java/edu/cornell/kfs/sys/service/impl/DisposableClientServiceImplBase.java
```diff
@@ -17,7 +18,7 @@ public void destroy() throws Exception {
         closeClientQuietly();
     }
     
-    protected Client getClient() {
+    protected Client getClient(Class classToRegister) {
```
cah292, line 21: I would suggest declaring this variable with the `Class<?>` type instead.

src/main/java/edu/cornell/kfs/sys/service/impl/DisposableClientServiceImplBase.java
```diff
@@ -27,13 +28,20 @@ protected Client getClient() {
                 if (jerseyClient == null) {
                     ClientConfig clientConfig = new ClientConfig();
                     jerseyClient = ClientBuilder.newClient(clientConfig);
+                    if (classToRegister != null) {
+                        jerseyClient.register(classToRegister);
```
cah292, line 32: The ClientConfig object also has a register() method, if you want to register this class in advance at the ClientConfig level.

cah292 (changes requested):
I noticed a few more minor items to be cleaned up, but other than that the changes look good.

Hopefully early next week we can get the needed functional feedback on the DV/DI-related questions, so that the remaining changes for those parts can be finalized.

db483:
retest this please

db483:
retest this please

src/main/java/edu/cornell/kfs/fp/document/web/struts/CuDisbursementVoucherForm.java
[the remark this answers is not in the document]
ajd299: Based on the functionals response we can remove the links, but keep the trip ID and canViewTrip fields

cah292:
@ajd299 Now that the 11/17/2021 patch has been merged to develop, please rebase your feature branch for these changes to resolve a couple merge conflicts. The KFS patch has temporarily commented out the web-related PaymentWorks unit tests, so your changes to those classes need to be adjusted accordingly. (If it's not practical to put commented-out variants of the changes in place yet, then please add the appropriate information to the KFSPTS-26117 user story so that the changes don't get lost.)

ajd299:
I have resolved the merge conflicts

cah292:
@ajd299 It looks like the KFSPTS-26287 fix introduced another conflict, this time in the POM. Please rebase again when you get a chance. Sorry about the extra work the 11/17 patch is causing.

ajd299:
retest this please

ajd299:
Please see the 2 PRs that are failing, and if any other changes are needed?

cah292 (approved):
Changes look good, and the only failing unit tests are the ones related to the known ongoing Jaggaer issue, so we should be good to move these changes forward.

Merging.

PR #1325, opened 2022-09-01
KFSPTS-22873 Switch to base inquiries for Person and Role

cah292:
This PR adjusts our cu-kfs setup to use KualiCo's new Person and Role inquiries, now that KualiCo has updated them to use the table-based layouts. The following modifications were made on our end:

* Removed our remaining code and JSP/tag/JavaScript files related to the older document-like Person and Role inquiries, now that KualiCo has fully removed them.
* Removed our customized Person field authorization checks (and related Data Dictionary setup), which were part of our original attempt to allow authorized users to view suppressed Person fields on the new inquiries.
* Added some Person-inquiry-specific handling to allow appropriate users to view suppressed fields on such inquiries. This functionality was present on KualiCo's old document-like Person inquiry but hasn't been added to the base version of the new inquiry yet. (Hopefully KualiCo will add this for their KIM rework and/or future KIM inquiry migrations.)
* Moved the relevant part of the KFSPTS-19308 customization from the PersonInquirableImpl overlay to this PR's new CuPersonInquirableImpl class. This allowed for removing the PersonInquirableImpl overlay.


src/main/java/edu/cornell/kfs/kim/inquiry/CuPersonInquirableImpl.java
```diff
@@ -0,0 +1,136 @@
...
+
+        if (person != null && person instanceof PersonImpl) {
+            ((PersonImpl) person).populateMembers();
+        }
+        return person;
+
+    }
+
+    @Override
+    public List<Section> getSections(BusinessObject businessObject) {
```
db483, line 83: From your comment on the PR I understand that this code was present in the old base code but not in the new implementation. Should we report this to KualiCo?

db483 (changes requested):
Changes look good. I just had one question about some changes that were present in the old code and were brought forward as a customization. Thank you.

src/main/java/edu/cornell/kfs/kim/inquiry/CuPersonInquirableImpl.java
cah292: Just to clarify, this particular inquiry-section-modifying code is not was present in base code before. What was present before was the ability for the old document-style inquiry to unmask suppressed person fields to appropriate users. This custom inquiry-section-modifying code is trying to replicate that behavior using the regular inquiry setup.

We might have reported this issue to KualiCo already; there's a comment on KFSPTS-19335 stating that Request 12480 was created for it. Not sure if KualiCo created a FINP ticket for it yet, though. I do think it would be worth following up with KualiCo again on this, to see if they're changing the behavior of this field masking as part of their KIM Person rework.

db483: Thank you for clarifying this.

db483 (approved):
Changes look good.

src/main/java/edu/cornell/kfs/kim/inquiry/CuPersonInquirableImpl.java
db483: https://kualico.atlassian.net/browse/FINP-6977 was closed as Won't Fix.

PR #1333, opened 2022-09-20
KFSPTS-25858: Add Business Rules for Country, ISO Country, ISO FIPS Country Mapping Maintenance Documents.

nkimble:


src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * CU Generic ISO/FIPS Country modification
+ */
+const ISO_COUNTRY_CODE_FIELD = 'document.newMaintainableObject.isoCountryCode';
+const ISO_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.isoCountry.name.div';
+const FIPS_COUNTRY_CODE_FIELD = 'document.newMaintainableObject.fipsCountryCode';
+const FIPS_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.fipsCountry.name.div';
+const EMPTY_STRING = '';
+var previousFipsCountryCode = "";
+var previousIsoCountryCode = "";
```
cah292, line 10: Are these previous-value variables needed? I don't see any areas in this file that are actively reading their values; they're only being written to.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,82 @@
...
+        clearCountryName(targetFieldName);
+    } else {
+        var dwrReply = {
+                 callback:function(data) {
+                    if (data === null || data.includes("NOT FOUND")) {
+                        setRecipientValue(targetFieldName, wrapError("Country Name not found"), true);
+                    } else {
+                        setFieldToSpecifiedValue(targetFieldName, data);
+                    } },
+                    errorHandler:function(errorMessage ) {
```
cah292, line 46: You might want to adjust the formatting of the "dwrReply" setups, to provide a clearer distinction between the "callback" part and the "errorHandler" part.

cah292 (changes requested):
Changes look good in general, though there are some areas that may need further cleanup.

nkimble:
retest this please

nkimble:
retest this please

nkimble:
retest this please

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,82 @@
...
+function clearCountryName(countryNameFieldName) {
+    dwr.util.setValue(countryNameFieldName, EMPTY_STRING)
+}
+
+function setFieldToSpecifiedValue(fieldName, fieldValue) {
+    dwr.util.setValue(fieldName, fieldValue)
+}
+
+function ensurePrimaryKeyIsUppercase(countryCodeField) {
+    return dwr.util.getValue(countryCodeField).trim().toUpperCase();
```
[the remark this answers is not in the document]
cah292: In my prior questions, what I meant was to ask whether the first argument to the getValue() and setValue() methods is an element ID/name or the element object itself. Sorry for the confusion.

I think the first argument is intended to be an element ID, based on the various other usage of these methods in our custom JavaScript files, though there might be a few areas in base code that are passing in the actual elements instead. You may want to verify that the element ID is what's always being passed into the methods and make adjustments accordingly.

In the case of read-only fields, I believe their IDs have a ".div" suffix on the HTML form, which it looks like you had already accounted for in the country-name-handling sections.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * CU Generic ISO/FIPS Country modification
+ */
+const ISO_COUNTRY_CODE_FIELD = 'document.newMaintainableObject.isoCountryCode';
```
cah292, line 4: If you're not using the ISO_COUNTRY_CODE_FIELD or FIPS_COUNTRY_CODE_FIELD constants, you could remove them.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,83 @@
...
+const ISO_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.isoCountry.name.div';
+const FIPS_COUNTRY_CODE_FIELD = 'document.newMaintainableObject.fipsCountryCode';
+const FIPS_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.fipsCountry.name.div';
+const EMPTY_STRING = '';
+
+function onblur_lookupFipsCountry(fipsCountryCodeField) {
+    let primaryKeyValue = ensurePrimaryKeyIsUppercase(fipsCountryCodeField);
+    if (primaryKeyValue === null || primaryKeyValue === '') {
```
cah292, line 12: Due to how boolean checks work in JavaScript, you could potentially write this file's various null/empty checks in the format `!primaryKeyValue` or `!primaryKeyValue || primaryKeyValue === ''`, which might help catch more null/empty-value cases. Sorry for not mentioning this earlier.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,83 @@
...
+        return countryCodeField;
+    }
+    return dwr.util.getValue(countryCodeField).trim().toUpperCase();
+}
+
+function updateCountryName(primaryKeyValue, targetFieldName) {   
+    if (primaryKeyValue === '') {
+        clearCountryName(targetFieldName);
+    } else {
+        var dwrReply = {
```
cah292, line 39: Can the various dwrReply variables be declared using "const" or "let" instead of "var"?

cah292 (changes requested):
I found a few more minor cleanup items that would be good to address (in addition to a few follow-up comments I had submitted earlier), but other than that the changes look good.

src/main/java/edu/cornell/kfs/sys/businessobject/ISOFIPSCountryMap.java
```diff
@@ -3,6 +3,7 @@
 import org.kuali.kfs.core.api.mo.common.active.MutableInactivatable;
 import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
 import org.kuali.kfs.sys.businessobject.Country;
+import edu.cornell.kfs.sys.businessobject.ISOCountry;
```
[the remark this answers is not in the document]
cah292: Just to be clear, this BO can continue to use the ISOCountry, it just doesn't need to explicitly import it if it's in the same package.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
cah292: Ok, thanks for explaining. I'm okay with leaving these particular variables as "var" ones for now.

src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
```diff
@@ -4,13 +4,16 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.lang.StringUtils;
```
cah292, line 7: Please use lang3 StringUtils instead.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOCountryServiceImpl.java
```diff
@@ -4,12 +4,15 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.lang.StringUtils;
```
cah292, line 7: Another spot that should use lang3 StringUtils.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSCountryMapServiceImpl.java
```diff
@@ -8,10 +9,13 @@
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang.StringUtils;
```
cah292, line 12: Another spot that should use lang3 StringUtils.

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * CU Generic ISO/FIPS Country modification
+ */
+const ISO_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.isoCountry.name.div';
+const FIPS_COUNTRY_NAME_FIELD = 'document.newMaintainableObject.fipsCountry.name.div';
+const EMPTY_STRING = '';
+
+function onblur_lookupFipsCountry(fipsCountryCodeField) {
+    let primaryKeyValue = ensurePrimaryKeyIsUppercase(dwr.util.getValue(fipsCountryCodeField));
```
cah292, line 9: Is "fipsCountryCodeField" the field object itself, or just the field's ID? That's what I was trying to explain in the related conversation on this PR. If it's the former but the script is still working fine, then this could be left as-is for now, though we may want to reevaluate the setup at some future point.

The same goes for the similar spot in the  "onblur_lookupIsoCountry" function.

cah292 (changes requested):
I noticed a few more minor items for cleanup and/or future analysis, but otherwise the changes look good. This can be merged once those items have been addressed or deferred accordingly.

src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
nkimble: ok

src/main/java/edu/cornell/kfs/sys/service/impl/ISOCountryServiceImpl.java
nkimble: ok

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSCountryMapServiceImpl.java
nkimble: ok

src/main/webapp/scripts/sys/isoFipsCountryMapScript.js
nkimble: Browser console logging showed this for the fipsCountryCodeField and isoCountryCodeField    [object HTMLInputElement]    so it is the object.

I have changed the code back to the previous version so that there is only one point of modification for the "dwr.util.getValue" call if that is needed in the future. From what I can see running locally, document is working as expected.

cah292 (approved):
Changes look good, unit tests pass and the related SQL changes are also ready.

There's some more elementObject-vs-elementId cleanup that would be nice to take care of in the new JavaScript file at some point. However, since @nkimble has confirmed that the script is still working locally, I'm fine with deferring that cleanup to a future user story so that we can keep the FIPS-to-ISO changes moving forward.

Merging.

PR #1334, opened 2022-10-06
KFSPTS-25850 Security Request Document Validate Department

ajd299:


src/main/java/edu/cornell/kfs/ksr/KSRKeyConstants.java
```diff
@@ -23,6 +23,7 @@ public class KSRKeyConstants {
     public static final String ERROR_SECURITY_REQUEST_DOC_DEPENDENT_ROLE_MISSING = "error.ksr.securityrequestdocument.dependentRole.missing";
     public static final String ERROR_SECURITY_REQUEST_DOC_CHANGE_MISSING = "error.ksr.securityrequestdocument.change.missing";
     public static final String ERROR_SECURITY_REQUEST_DOC_SERVICE_EXCEPTION = "error.ksr.securityrequestdocument.service.exception";
+    public static final String ERROR_SECURITY_REQUEST_DOC_PRIMARY_DEPARTMENT = "error.ksr.securityrequestdocument.service.primary.department";
```
cah292, line 26: This new constant doesn't seem to be used by this code at all, and I don't think it's been added to the app resource properties either. Can it be removed?

src/main/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityRequestDocumentRule.java
```diff
@@ -56,8 +57,8 @@ protected Map<String,String[]> getRequiredQualificationsMap() {
 
     @Override
     public boolean isDocumentAttributesValid(Document document, boolean validateRequired) {
-        boolean success = true;
         SecurityRequestDocument securityRequestDocument = (SecurityRequestDocument) document;
+        boolean success = validateDepartmentCode(securityRequestDocument.getPrimaryDepartmentCode());
```
cah292, line 61: Regex-based validation of properties can be configured at the Data Dictionary level, and it looks like such a configuration is already present on the "primaryDepartmentCode" property. I would suggest using just the DD-based validation instead, if possible.

It looks like the "primaryDepartmentCode" DD config is specifying the "validCharactersConstraint" instead of the "validationPattern". Several other KSR beans also have such a config; this must have been part of the KNS-to-KRAD conversion back on the Cynergy side, which we now have to undo. Fixing this DD config accordingly might be enough to fix the problem.

src/test/java/edu/cornell/kfs/ksr/document/validation/impl/SecurityRequestDocumentRuleTest.java
```diff
@@ -0,0 +1,54 @@
+package edu.cornell.kfs.ksr.document.validation.impl;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class SecurityRequestDocumentRuleTest {
```
cah292, line 9: If the KSR document gets updated to only use the DD-based validation for the Primary Department Code, then this unit test can be removed.

cah292 (changes requested):
After a quick analysis, I think I've identified some Data-Dictionary-related configuration that is actually at the root of the validation problem. Fixing the DD-based config would be a much cleaner approach.

Given that Andrew will be out of office for a couple days, I might go ahead and implement the DD-based approach on a new feature branch.

cah292:
I'm closing this PR since I've prepared an updated one that addresses the Data Dictionary issues appropriately. If we decide to proceed with Andrew's changes instead, we can reopen this PR accordingly.

PR #1336, opened 2022-10-06
KFSPTS-26287 Fix critical conflicting dependencies in 2021-11-17 patch

cah292:
This PR excludes some conflicting dependencies that somehow went unnoticed during the 11/17/2021 patch preparation.

The biggest culprits were the Guava JAR, the Netty JARs and the Spring-related JARs. There are still some HttpClient-related JARs that have not been cleaned up yet, but their version numbers only seem to defer at the patch-number level and they don't interfere with launching KFS locally. To keep DEV moving forward, I would like to defer the HttpClient cleanup to another user story unless others disagree.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -254,16 +254,20 @@
                                 <exclude>WEB-INF/lib/commons-logging-*.jar</exclude>
                                 <exclude>WEB-INF/lib/gson-*.jar</exclude>
                                 <exclude>WEB-INF/lib/lettuce-*.jar</exclude>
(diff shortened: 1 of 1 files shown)


jhulslander (approved):
I agree with deferring the HTTP libraries so we can get kfs-dev up with the new kualico patch.  I suggest updating PR commit message to reference the kualico patch that caused the conflicts.
this PR can be merged once the unit tests pass, and the commit message is updated

cah292:
Updated PR title to reference the 2021-11-17 patch.

PR #1338, opened 2022-10-07
KFSPTS-26146: Fix stacktraces happening on GACC and GDLG when looking up multiple accounting lines by FO for both closed and open accounts.

nkimble:


nkimble:
retest this please

db483:
retest this please

src/main/java/edu/cornell/kfs/coa/businessobject/lookup/AccountGlobalSearchLookupableHelperServiceImpl.java
```diff
@@ -35,6 +36,8 @@
 
 public class AccountGlobalSearchLookupableHelperServiceImpl extends KualiAccountLookupableHelperServiceImpl {
     private AccountGlobalSearchDao accountGlobalSearchDao;
+    private IdentityService identityService;
+    private PermissionService permissionService;
```
cah292, line 40: This subclass doesn't seem to be using the permissionService variable, so this variable and the additions or overrides of the related getters/setters can be removed.

cah292 (commented):
There's one minor spot that would be nice to clean up, but otherwise the changes look good. This cleanup could be performed at a later date if needed, if we want to move these changes forward sooner.

src/main/java/edu/cornell/kfs/coa/businessobject/lookup/AccountGlobalSearchLookupableHelperServiceImpl.java
cah292: Given that Nancy is out of office until next week, and that the extra permissionService variable/getter/setter appear to be set up properly, and that we already have follow-up JIRAs to potentially move our various lookups to the new framework anyway, I'm fine with deferring the permissionService cleanup to a future user story.

cah292 (approved):
Code changes look good, and the permissionService variable cleanup will be deferred to the KFSPTS-25387 ticket. All unit tests passed except for the two known problematic Jaggaer tests, so this code is good to move forward.

Merging.

nkimble:
@cah292 Even though the PermissionService is not used by the subclass it must be declared in the subclass and the setting of the attribute in the super class needs to be performed; otherwise, the same type of stack trace reported by this JIRA will result when the "Closed?" "Yes" radio button is selected in the Lookup which uses the ParameterService in the super class. This change and fix for the additionally discovered stack trace is the reason why I added the validation for the radio buttons.

Yes, I agree this should not be the case but for some reason the spring wiring is not working as expected in this subclass-superclass case. Can you please note the need for this change in whatever additional JIRA you were going to remove these ParameterService changes from so that we do not reintroduce a bug. An alternative approach may be possible but at a minimum, if any code is removed/changed all of those radio buttons on the Account Global Lookup need to be validated.

cah292:
@nkimble Thanks for the clarification. I've updated KFSPTS-25387 to note why that override is still in place.

PR #1339, opened 2022-10-07
KFSPTS-25850 Fix Dept Code validation on KSR docs

cah292:
This is an alternative version of the #1334 fix that addresses the underlying cause of the reported problem. It appears we already had some KSR Data Dictionary configuration for validating the Primary Department Code, but it was configured for KRAD-UI-based document validation instead. This PR updates the configuration to work with the KNS-based setup, and also adjusts it to fit with how KualiCo has configured similar validation beans.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/ksr/document/datadictionary/SecurityRequestDocument.xml
+++ b/src/main/resources/edu/cornell/kfs/ksr/document/datadictionary/SecurityRequestDocument.xml
@@ -60,10 +60,6 @@
   <bean id="SecurityRequestDocument-requestPerson.principalName" parent="SecurityRequestDocument-requestPerson.principalName-parentBean"/>
   <bean id="SecurityRequestDocument-requestPerson.principalName-parentBean" abstract="true" parent="PersonImpl-principalName">
     <property name="name" value="requestPerson.principalName"/>
-    <property name="validationPattern"><null /></property>
-    <property name="validCharactersConstraint">
-      <bean parent="AlphaNumericPatternConstraint" p:lowerCase="true" />
-    </property>
     <property name="description" value="The person's Cornell NetId.&lt;br /&gt;&lt;br /&gt;If partial matches are desired, then wildcards must be explicitly included in the search expression." />
   </bean>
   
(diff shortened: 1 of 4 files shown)


db483:
retest this please

cah292:
retest this please

src/main/resources/edu/cornell/kfs/ksr/document/datadictionary/SecurityRequestDocument.xml
```diff
@@ -60,10 +60,6 @@
   <bean id="SecurityRequestDocument-requestPerson.principalName" parent="SecurityRequestDocument-requestPerson.principalName-parentBean"/>
   <bean id="SecurityRequestDocument-requestPerson.principalName-parentBean" abstract="true" parent="PersonImpl-principalName">
     <property name="name" value="requestPerson.principalName"/>
-    <property name="validationPattern"><null /></property>
```
db483, line 63: Was this causing an issue as well? Based on the jira description only an issue with the department code was reported.
I see a similar change was made for the principalNameForSearch. Was that also needed? Thank you.

db483 (changes requested):
Changes look good but I had a couple of questions about some changes unrelated to the primary dept code. Thank you.

src/main/resources/edu/cornell/kfs/ksr/document/datadictionary/SecurityRequestDocument.xml
cah292: While I was preparing the changes for this, I noticed that several KSR attributes were using the "validCharactersConstraint" setup instead of the "validationPattern" setup. I went ahead and switched the other attributes from the constraint setup to the pattern setup for consistency, especially since that transition seemed to resolve the validation problems on the department code. This also helped bring our configuration more in line with base code, which is almost entirely using the "validationPattern" setup throughout the various modules.

db483: I think that is already set up and inherited from KimBaseBeans-principalName so that is why this is not needed? Thank you.

cah292: Yes, for the other cases in this file, they have KimBaseBeans-principalName as an ancestor and are inheriting its validation pattern. Sorry for the confusion.

db483: Thank you.

db483 (approved):
Changes look good and the only failing unit tests are the Jaggaer related ones.

PR #1340, opened 2022-10-12
KFSPTS-25523 backport FINP-8341

db483:


cah292:
retest this please

cah292 (commented):
I had one question before moving this to DEV, but other than that the changes look appropriate:

* Is the "spring-kew.xml" overlay needed? We're already overriding the relevant beans in "cu-spring-kew.xml" so we may not need the overlay. If you think the overlay is still needed, then please update it to put the "documentRouteHeaderService" bean definition back in; it looks like the overlay is excluding it even though the base file still has it. ("cu-spring-kew.xml" already has an override of that bean, which is likely masking the problem.)

cah292 (commented):
Changes look good with the "spring-kew.xml" overlay removed. Please confirm if this is ready to be merged or if you were still planning to make a few more follow-up changes first.

db483:
@cah292 this is not ready to be committed, I am doing some testing in prekew. Thank you.

db483:
@cah292 I have finished my testing. This can be reviewed and merged now. Thank you.

cah292 (approved):

PR #1344, opened 2022-10-24
KFSPTS-23690: Code adjustments to retain CU customization to allow manual billing to override billing period end date when CINV is created. CINV edit or view will follow base code processing and not the CU mod since the CINV exists with the correct end date. Needed due to base code creating new suspension category with FINP-7443.

nkimble:
Functionality has been verified and if reviewer approves these changes, they can be merged.


nkimble:
retest this please

src/main/java/org/kuali/kfs/integration/cg/ContractsAndGrantsBillingAward.java
```diff
@@ -0,0 +1,391 @@
...
+import org.kuali.kfs.module.ar.ArConstants.ContractsAndGrantsInvoiceDocumentCreationProcessType;
+
+import java.sql.Date;
+import java.sql.Timestamp;
+import java.util.List;
+
+/**
+ * Integration interface for Award (specific to CGB functionality)
+ */
+public interface ContractsAndGrantsBillingAward extends Billable, ContractsAndGrantsAward {
```
jhulslander, line 34: Would it be cleaner to make this a CU interface:
CuContractsAndGrantsBillingAward extends ContractsAndGrantsBillingAward

and just have CuContractsAndGrantsBillingAward define the two new functions:
ContractsAndGrantsInvoiceDocumentCreationProcessType getCreationProcessType();
void setCreationProcessType(ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType);

src/main/java/org/kuali/kfs/module/cg/businessobject/Award.java
```diff
@@ -172,7 +173,7 @@ public class Award extends PersistableBusinessObjectBase implements MutableInact
     private String letterOfCreditFundGroupCode;
     private LetterOfCreditFundGroup letterOfCreditFundGroup;
     private transient String invoiceLink;
-    private transient String cgInvoiceDocumentCreationProcessTypeCode;
+    private transient ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType;
```
jhulslander, line 176: should this variable have a comment that it is a CU Customization?

jhulslander (commented):
This change looks good to me.  My primary thought for improvement involves how the interface changes are being done.

src/main/java/org/kuali/kfs/integration/cg/ContractsAndGrantsBillingAward.java
jhulslander: it looks like these two new interface functions are only used in ContractsGrantsInvoiceCreateDocumentServiceImpl and VerifyBillingFrequencyServiceImpl.  You would only need to cast the award to CuContractsAndGrantsBillingAward when you need to call these two new functions.

src/main/java/org/kuali/kfs/module/cg/businessobject/Award.java
jhulslander: thanks, that comment works, no action needed on this comment.

jhulslander:
Nancy and I had a zoom conversation about the Interface comments I had.  The root problem is the way KualiCo built CGB, we have seen inconsistent results in the object graph returned based on some "interesting" things they do with Spring.  Based on previous problems in this space, currently the safest approach will be to overlay their classes rather than extend them.

jhulslander (approved):

PR #1347, opened 2022-10-27
KFSPTS-26091 add locking document ID to cache

jhulslander:


jhulslander:

src/main/java/edu/cornell/kfs/coa/document/CuAccountGlobalMaintainableImpl.java
```diff
@@ -42,6 +47,13 @@ public List<MaintenanceLock> generateMaintenanceLocks() {
                 .filter(account -> isClosingAccount(account, oldAccountClosedStatuses))
                 .flatMap(this::generateTrickleDownMaintenanceLocks);
         
+        Cache cache = MaintenanceUtils.getBlockingCache();
+        String cacheKey = MaintenanceUtils.buildLockingDocumentCacheKey(getDocumentNumber());
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("generateMaintenanceLocks, clear cache id: " + cacheKey);
+        }
+        cache.evictIfPresent(cacheKey);
```
cah292, line 55: I'm wondering if the cache entry also needs to be cleared when the Account Global adds or removes collection lines, if this current mechanism is not enough to perform such clearing. If such an addition is needed, perhaps the `processAfterPost()` method could potentially be used for such entry clearing when it detects an add-line or remove-line operation, both single-valued and/or multi-valued.

src/main/java/edu/cornell/kfs/coa/document/CuSubObjectCodeMaintainableImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    
+    @Override
+    public List<MaintenanceLock> generateMaintenanceLocks() {
+        List<MaintenanceLock> maintenanceLocks = super.generateMaintenanceLocks();
+        
+        Cache cache = MaintenanceUtils.getBlockingCache();
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("generateMaintenanceLocks, clear all blocking cache ");
+        }
+        cache.clear();
```
cah292, line 24: You should also add this clear-all-entries logic to the Account Reversion maintainable.

src/main/resources/edu/cornell/kfs/coa/document/datadictionary/SubObjectCodeMaintenanceDocument.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
+    Copyright 2005-2021 Kuali, Inc.
```
cah292, line 6: Does this file need the KualiCo license header? It's not overlaying the related file from base code.

src/main/resources/kfs.ehcache.xml
```diff
@@ -0,0 +1,77 @@
...
+
+    <cache name="ParameterType" maxEntriesLocalHeap="2500" timeToLiveSeconds="3600"/>
+    <cache name="NamespaceType" maxEntriesLocalHeap="500" timeToLiveSeconds="3600"/>
+
+    <cache name="DocumentTypeType" maxEntriesLocalHeap="25000" timeToLiveSeconds="3600"/>
+    <cache name="RoutePathType" maxEntriesLocalHeap="25000" timeToLiveSeconds="3600"/>
+    <cache name="ExtensionDefinitionType" maxEntriesLocalHeap="5000" timeToLiveSeconds="3600"/>
+    
+    <!-- CU custom cache -->
+    <cache name="LockingIdCache" maxEntriesLocalHeap="10000"/>
```
cah292, line 76: You may want to specify an explicit time-to-live value for this cache.

cah292 (changes requested):
I found a few areas that might need more revisions, but overall the changes look good.

src/main/java/edu/cornell/kfs/coa/document/CuAccountGlobalMaintainableImpl.java
jhulslander: I am confused at what you're saying here.  Should this clear cache logic be moved to processAfterPost()?  If so, should this logic be moved to processAfterPost() in the other maintainable classes?

src/main/resources/kfs.ehcache.xml
jhulslander: OK, I will set timeToLiveSeconds="1800", 30 minutes seems reasonable for this particular cache

src/main/java/edu/cornell/kfs/coa/document/CuSubObjectCodeMaintainableImpl.java
jhulslander: OK, added code to AccountReversionMaintainableImpl

src/main/java/edu/cornell/kfs/coa/document/CuAccountGlobalMaintainableImpl.java
cah292: If the user edits the Account Global doc between submission attempts, that may affect what downstream objects need maintenance locks during routing. This spot will at least clear the cache when the locks are being persisted to the database, but if this spot doesn't affect adding/removing accounts from the global doc, then appropriate cache clearing might need to be added to `processAfterPost()` or a similar hook.

The other maintainables don't need to override the `processAfterPost()` logic, since they only need to clear the cache when they store or delete locks that may affect Account Global. However, after considering this a little further, the other maintainables may want to move the cache clearing from their `generateMaintenanceLocks()` methods to their `doRouteStatusChange()` methods, so that the cache gets cleared when the objects start or finish their routing.

Similarly, it might make sense for the Account Global maintainable to move its single-entry clearing from `generateMaintenanceLocks()` to `doRouteStatusChange()` for clarity, unless that messes up the new caching feature.

jhulslander:
I've mad some additional changes based on the last comments.  I want to test these changes in prekew before they merged into develop.  If you want to review again, please do so, but hold off on merging till I validate  some more.

src/main/java/edu/cornell/kfs/coa/document/AccountReversionMaintainableImpl.java
```diff
@@ -257,4 +263,15 @@ protected boolean shouldIncludeField(Field field) {
         return includeField;
     }
     
+    @Override
+    public void doRouteStatusChange(DocumentHeader documentHeader) {
```
cah292, line 267: I noticed that the various `doRouteStatusChange()` overrides are set up to clear the cache on every status change. During your Pre-KEW testing, you may want to verify whether the cache clearing should be limited to certain specific status changes (ENROUTE, PROCESSED, DISAPPROVED, etc.).

cah292 (commented):
I did a quick scan of the revised code, and noticed something that may or may not need further attention. I will perform a more thorough review tomorrow.

cah292 (commented):
I noticed one spot that may or may not need revising, but otherwise the latest revisions look good. Also, the upcoming Pre-KEW testing should confirm whether additional cache clearing is needed when the Account Global doc adds or removes collection lines.

src/main/java/org/kuali/kfs/krad/maintenance/MaintenanceUtils.java
```diff
@@ -64,10 +74,72 @@ public static void checkForLockingDocument(MaintenanceDocument document, boolean
...
+    public static Cache getBlockingCache() {
+        if (blockingCache == null) {
+            CacheManager cm = CoreImplServiceLocator.getCacheManagerRegistry()
+                    .getCacheManagerByCacheName(LOCKING_ID_CACHE_NAME);
+            blockingCache = cm.getCache(LOCKING_ID_CACHE_NAME);
+        }
+        return blockingCache;
+    }
+    
+    public static void clearAllBlockingCache() {
```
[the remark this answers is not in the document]
jhulslander: I can change it

jhulslander:
I did run into a problem with adding detail lines.  I've added "clear cache" logic to addMultipleValueLookupResults() and addNewLineToCollection().  I need to add "clear cache" to removing a detail line.  I haven't found a.good approach to doing that yet

jhulslander:
I believe the bug with the approver adding removing details was fixed with clearing cache on prepareForSave().  This is ready for review again

src/main/java/edu/cornell/kfs/coa/document/CuAccountGlobalMaintainableImpl.java
```diff
@@ -45,6 +52,29 @@ public List<MaintenanceLock> generateMaintenanceLocks() {
...
+        String cacheKey = MaintenanceUtils.buildLockingDocumentCacheKey(getDocumentNumber());
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("clearBlockingCacheForCurrentDocument, clear cache id: " + cacheKey);
+        }
+        cache.evictIfPresent(cacheKey);
+    }
+    
+    @Override
+    public void prepareForSave() {
+        clearBlockingCacheForCurrentDocument();
```
cah292, line 75: Can you please confirm if you're still seeing the expected performance gains locally or in Pre-KEW? If you're putting a cache-clearing operation in the `prepareForSave()` method, I'm wondering if that would cause the cache to be cleared too frequently and potentially nullify the intended performance improvements.

cah292 (commented):
I had a question about the placement of one of the cache-clearing calls, but otherwise the revised changes look good.

src/main/java/edu/cornell/kfs/coa/document/CuAccountGlobalMaintainableImpl.java
jhulslander: my pre-kew testing has good performance results

cah292: Thanks for confirming. I'm okay with moving the changes forward to DEV for more in-depth validation.

cah292 (approved):

PR #1349, opened 2022-10-31
KFSPTS-26460 re-enable AccountVerificationWebService

jhulslander:
reimplementation of AccountVerificationWebService it was recently removed with https://github.com/CU-CommunityApps/cu-kfs/commit/3d57d1faa734f4aa575079beebfa58d7af8969c6


cah292 (commented):
I'm still reviewing this, but there's something I wanted to note quickly:

I noticed that our CXF removal also removed a couple wss4j dependencies. You might need to add those back in as well.

jhulslander:
I put back in the ws4j libraries, and I believe I have fixed this paymentworks test

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
```diff
@@ -349,8 +356,18 @@ public int uploadVendorsToPaymentWorks(InputStream vendorCsvDataStream) {
     }
 
     private Response performSupplierUpload(InputStream vendorCsvDataStream) {
-        Invocation request = buildMultiPartRequestForSupplierUpload(getClient(), buildSupplierUploadURI(), vendorCsvDataStream);
-        return request.invoke();
+        Client client = null;
+
+        try {
+            ClientConfig clientConfig = new ClientConfig();
+            clientConfig.register(CuMultiPartWriter.class);
```
cah292, line 363: The CXF removal changed the Client object setup to rely upon a superclass for handling the Client object's lifecycle. (I think this was due to the Response getting closed prematurely with Jersey instead of CXF, if the Client was closed before reading the response.) To keep the configuration consistent, I would suggest replacing this manual Client creation with an override of the appropriate getClient() method.

This class already overrides the getClient() method to call a variant that registers a multi-part handler, you could try changing that override so that it registers CuMultiPartWriter.class instead of MultiPartFeature.class

cah292 (commented):
I noticed one more minor item that might need cleanup.

cah292 (commented):
The rest of the changes look good, this can be merged once the PMW-Client-related concern has been addressed and the unit tests pass.

src/main/java/edu/cornell/kfs/pmw/batch/service/impl/PaymentWorksWebServiceCallsServiceImpl.java
jhulslander: OK, I have adjusted as requested

cah292 (commented):

PR #1350, opened 2022-11-01
KFSPTS-26393: Fix Contracts Grants Invoice Documents Job so that it presents an error when there is an invalid customer address instead of failing the whole job.

nkimble:


src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    
+    private static final Logger LOG = LogManager.getLogger();
+    
+    /*
+     * CU Customization: KFSPTS-26393
+     * When creationProcessType is BATCH, validate that the customer number on the award as part if the customer
+     * address composite key matches the customer number associated to the agency associated to the award.
+     */
+    public boolean agencyCustomerMatchesAwardCustomer(ContractsAndGrantsBillingAward award,
+            ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType) {
```
db483, line 22: Should the check in the creation process type be done outside of this method?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+    /*
+     * CU Customization: KFSPTS-26393
+     * When creationProcessType is BATCH, validate that the customer address identifier on the award exists.
+     */
+    public boolean customerAddressIdentifierExists(ContractsAndGrantsBillingAward award,
+            ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType) {
+        
+        if (ContractsAndGrantsInvoiceDocumentCreationProcessType.BATCH == creationProcessType) {
```
db483, line 53: Should the check in the creation process type be done outside of this method?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    /*
+     * CU Customization: KFSPTS-26393
+     * When creationProcessType is BATCH, validate that the customer number on the award as part if the customer
+     * address composite key matches the customer number associated to the agency associated to the award.
+     */
+    public boolean agencyCustomerMatchesAwardCustomer(ContractsAndGrantsBillingAward award,
+            ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType) {
+        
+        if (ContractsAndGrantsInvoiceDocumentCreationProcessType.BATCH == creationProcessType) {
+            if (ObjectUtils.isNotNull(award.getAgency().getCustomer().getCustomerNumber()) &&
```
db483, line 25: I think StringUtils.isNotBlank is checking for null also. I wonder if this should instead check that the agency and customer objects are not null

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+                    ObjectUtils.isNotNull(award.getCustomerNumber()) &&
```
db483, line 26: I am not sure this check is needed if StringUtils.isNotBlank is already checking for null

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+                    StringUtils.isNotBlank(award.getAgency().getCustomer().getCustomerNumber()) &&
+                    StringUtils.isNotBlank(award.getCustomerNumber()) &&
+                    award.getAgency().getCustomer().getCustomerNumber().equalsIgnoreCase(award.getCustomerNumber())) {
+                return true;
+            } else {
+                LOG.info("agencyCustomerMatchesAwardCustomer: Mismatched Customer records detected for " +
+                         "Award.Agency and Award.CustomerAddress : Award.Agency.CustomerNumber = " +
+                            (ObjectUtils.isNull(award.getAgency()) ? "null" : (ObjectUtils.isNull(award.getAgency().getCustomer().getCustomerNumber()) ? "null" : award.getAgency().getCustomer().getCustomerNumber())) +
```
db483, line 34: Does this need to check on the customer object being null?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+                        " Award.CustomerNumber = " + award.getCustomerNumber() +
+                        " Award.CustomerAddressIdentifier = " +
```
db483, line 36: this method is not doing any checks on CustomerAddressIdentifier. Does this value need to be logged?

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    
+    /*
+     * CU Customization: KFSPTS-26393
+     * When creationProcessType is BATCH, validate that the customer address identifier on the award exists.
+     */
+    public boolean customerAddressIdentifierExists(ContractsAndGrantsBillingAward award,
+            ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType) {
+        
+        if (ContractsAndGrantsInvoiceDocumentCreationProcessType.BATCH == creationProcessType) {
+            if (ObjectUtils.isNotNull(award.getCustomerAddressIdentifier())) {
```
db483, line 54: Is this only supposed to be a null check? Or is it supposed to do a check against the database?

db483 (changes requested):
Code changes look good in general but I did have some questions. Thank you.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
nkimble: My understanding after talking with the functional user was to report validation errors in a way that they could correct the data. Later downstream CGB CINV processing performs the existence check for the address. This check was purposely code to not be a database check (just prevent NPEs) so we were not incurring more processing overhead.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
db483: Based on the fact that this is modeled on base code I am ok with the code as is.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
db483: After Nancy tested the ObjectUtils.isNotNull method it confirmed my suggestion that it does not check on all the objects in the chain and that if agency or customer are null then it would throw a null pointer exception. I see that the code has been updated to account for this. Thank you.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
db483: This has been updated as well after testing the ObjectUtils.isNull method. Thank you.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
db483: While this method does not deal with the CustomerAddressIdentifier itself, I understand that it can be helpful to also have that data logged.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
db483: This has now been updated to also check against the DB to see if the address exists.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+     * obtained by downstream processing in ContractsGrantsInvoiceCreateDocumentServiceImpl.buildInvoiceAddressDetails.
+     */
+    public boolean invoicingCustomerAddressExists(ContractsAndGrantsBillingAward award,
+            ContractsAndGrantsInvoiceDocumentCreationProcessType creationProcessType) {
+        
+        CustomerAddress customerAddress = null;
+        
+        if (ContractsAndGrantsInvoiceDocumentCreationProcessType.BATCH == creationProcessType) {
+            if (ObjectUtils.isNotNull(award.getCustomerAddressIdentifier()) &&
+                    ObjectUtils.isNotNull(award.getCustomerNumber())) {
```
db483, line 60: Should this check be updated to use StringUtils.isNotBlank to ensure it checks on whitespaces too? Sorry for not asking this earlier.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                customerAddress = customerAddressService.getByPrimaryKey(award.getCustomerNumber(),
+                        award.getCustomerAddressIdentifier());
+            }
+            if (ObjectUtils.isNull(customerAddress) && 
+                    ObjectUtils.isNotNull(award.getCustomerNumber())) {
```
db483, line 65: Same here, should this check be updated to use StringUtils.isNotBlank to ensure it checks on whitespaces too? Sorry for not asking this earlier.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
nkimble: Good catch.

src/main/java/edu/cornell/kfs/module/ar/service/impl/CuCustomerAddressHelperServiceImpl.java
nkimble: Good catch, updating.

nkimble:
Latest changes have been made. Unit tests have passed. Local validation performed in prekew/sandbox environment with successful batch job completion. New validation checks detected and output the error condition to the batch validation error file indicating this mod is working as expected from a technical perspective. This PR is ready for review again.

PR #1351, opened 2022-11-01
KFSPTS-26538 update spring-security-core to 5.6.9 and spring-core to 5.3.23

jhulslander:


cah292 (changes requested):
The version change looks good, but it seems that the newer JAR is also pulling in some slightly newer versions of the regular Spring JARs ("5.3.23"). You might also need to update the "spring.version" property to match, unless you think we should stay at our current version (which would require adding exclusions to the security JAR dependency).

jhulslander:
KualiCo is currently at 5.3.20.
I think it would make sense to update spring-core to 5.3.23 to be in line with what came with spring-security.  My primary thought would be to get ahead of potential future security vulnerabilities

jhulslander:

cah292 (approved):

PR #1354, opened 2022-11-11
KFSPTS-24580 Improvements to Ehcache-and-Redis setup

cah292:
This PR makes the following enhancements to our Ehcache-and-Redis setup, which were identified during the KEW-to-KFS work but were deferred until now:

1. Reworked the helper code that handles our cu-kfs cache key parsing and formatting. The relevant wrapper class was only being used for element removal operations, so that class has been renamed and revised for clarity. The relevant cache key construction/deconstruction code has also been moved into its own helper class.
2. Improved the error logging when an invalid cache configuration is detected. If initialization fails due to one of the cache name sets not being a sub-set of the all-cache-names set, the offending elements will now be printed to the logs.
3. Adjusted the setup of default time-to-live settings for Redis-aware caches. The forced setup of defaults will now take place during the initialization of the cache manager, and a warning will be logged when an override is performed on a given cache. (The "kfs.ehcache.xml" file is not yet being updated by this PR to explicitly set more timeouts; see the note below.)
   * NOTE: It looks like KualiCo adjusted some of its Redis cache time-to-live settings to zero/infinite in the 2022-05-18 patch, for consistency with how the related caches were configured in Ehcache. At some point we should reevaluate our time-to-live setup to determine if we should also use the appropriate zero/infinite durations, which would require adjusting our code to allow for such values. I was planning to create a follow-up user story to address that for the 2022-05-18 upgrade; if you think it should be handled for this PR instead, please let me know.
   * NOTE: It seems that KualiCo is not keeping its "kfs.ehcache.xml" settings completely up-to-date. A few of the caches there (such as "AssetObjectCode") do not appear to be in use anymore. I'm planning to create a follow-up ticket to report this back to KualiCo.


db483 (approved):
Changes look good.

db483:
Please create follow up user stories for reviewing the time-to-live settings in the 2022-05-18 patch and for the KualiCo ticket related to  keeping the "kfs.ehcache.xml" settings up to date. Thank you.

PR #1356, opened 2022-11-17
KFSPTS-26549: Remove DataObjectRestServiceController and related wiring as it is no longer being used.

nkimble:


cah292 (approved):

PR #1359, opened 2022-11-30
KFSPTS-26773 backport FINP-8908 to fix issue with Action List in Firefox

db483:


jhulslander (approved):
backport looks good, I shall merge

PR #1362, opened 2022-12-05
KFSPTS-23719 Add new CNCR Expense workflow handling

cah292:

This PR updates the new Concur Event Notification V2 Job so that it will now perform workflow actions against the processed Expense Reports. Note that the new workflow handling for Travel Requests is not being added yet; that will be handled in a subsequent user story.

I ended up using an endpoint from the Expense V4 API to handle the workflow operations, since the Expense V3 API didn't appear to have one. (If you happen to find such an endpoint for Expense V3, please let me know.) Also, during local testing I discovered that the send-back workflow action is case-sensitive and has to be specified as "sendBack" in the URL; that's why it's defined that way in the PR's code.

In addition, I made some adjustments to some previously-introduced testing infrastructure in an attempt to speed up the new unit tests a little bit. If you think such adjustments should be deferred to a later time, then I can modify the unit tests to use the previous setup.


cah292:
I'm closing this PR since I'm about to cut a new develop branch from master. I'll open a new PR once the changes have been migrated to an appropriate updated feature branch.

PR #1364, opened 2022-12-07
KFSPTS-23719 Add new CNCR Expense workflow handling

cah292:

(NOTE: This is an updated version of the #1362 PR's changes. It has been updated for compatibility with the new develop branch.)

This PR updates the new Concur Event Notification V2 Job so that it will now perform workflow actions against the processed Expense Reports. Note that the new workflow handling for Travel Requests is not being added yet; that will be handled in a subsequent user story.

I ended up using an endpoint from the Expense V4 API to handle the workflow operations, since the Expense V3 API didn't appear to have one. (If you happen to find such an endpoint for Expense V3, please let me know.) Also, during local testing I discovered that the send-back workflow action is case-sensitive and has to be specified as "sendBack" in the URL; that's why it's defined that way in the PR's code.

In addition, I made some adjustments to some previously-introduced testing infrastructure in an attempt to speed up the new unit tests a little bit. If you think such adjustments should be deferred to a later time, then I can modify the unit tests to use the previous setup.


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceImpl.java
```diff
@@ -144,11 +153,47 @@ protected void validateExpenseAllocations(String accessToken, List<ConcurEventNo
...
+        String logMessageDetail = MessageFormat.format(expenseV4WorkflowMessageFormat, workflowAction, reportId);
+        
+        ConcurWebRequest<Void> webRequest = ConcurWebRequestBuilder.forRequestExpectingEmptyResponse()
+                .withUrl(workflowActionUrl)
+                .withHttpMethod(HttpMethod.PATCH)
+                .withJsonBody(workflowDTO)
+                .build();
         
-        LOG.info("updateStatusInConcur, update of status in Concur not implemented yet");
+        concurEventNotificationV2WebserviceService.callConcurEndpoint(
```
ajd299, line 179: updateStatusInConcur is somewhat of a long method. Would abstracting the building of the web request make the method shorter and cleaner?

src/main/java/edu/cornell/kfs/concur/rest/jsonObjects/ConcurV4WorkflowDTO.java
```diff
@@ -0,0 +1,38 @@
...
+ * If further separation is needed between the two in the future, then please refactor this class accordingly.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ConcurV4WorkflowDTO {
+
+    @JsonProperty("comment")
+    private String comment;
+
+    public ConcurV4WorkflowDTO() {
+    }
```
ajd299, line 20: Is an explicit default constructor required?

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceUpdateReportStatusTest.java
```diff
@@ -0,0 +1,362 @@
...
+        concurExpenseV3Service.setConcurAccountValidationService(Mockito.mock(ConcurAccountValidationService.class));
+        concurExpenseV3Service.setSimulateProduction(true);
+        
+        mockEndpoint.addReportsAwaitingAction(ConcurTestConstants.REPORT_ID_1,
+                ConcurTestConstants.REPORT_ID_2, ConcurTestConstants.REPORT_ID_3);
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        concurExpenseV3Service = null;
```
ajd299, line 97: Is the "throws Exception" needed for all or any of the methods within this test class?

src/test/java/edu/cornell/kfs/sys/web/mock/MockMvcWebServerExtension.java
```diff
@@ -52,31 +57,59 @@
...
+ * the ResettableController interface and then this extension will reset them after each run
+ * of a test method. (Note that this auto-reset feature is ignored if the extension was set up
+ * as an instance variable or if the MVC configuration was set up via the setMockMvc() method.)
+ * 
+ * Note that if this extension is set up at the class/static level, it is strongly recommended
+ * that the test class be configured for SAME_THREAD execution mode so that the test methods
+ * don't interfere with each other.
  */
-public class MockMvcWebServerExtension implements BeforeEachCallback, BeforeTestExecutionCallback, AfterEachCallback {
+public class MockMvcWebServerExtension implements BeforeEachCallback, BeforeTestExecutionCallback, AfterEachCallback,
```
ajd299, line 85: This is interesting, maybe you could demo how to use this class in the Developer's Bookclub?

ajd299 (changes requested):
Code looks good, just a few comments

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceImpl.java
cah292: I'll look into refactoring this method to make it shorter.

src/main/java/edu/cornell/kfs/concur/rest/jsonObjects/ConcurV4WorkflowDTO.java
cah292: The Jackson framework might need a default constructor in order to build this DTO from a JSON value. Even though this PR's new Production code is only doing DTO-to-JSON conversion for this clas, the unit test's MVC controller is implicitly relying on Jackson's JSON-to-DTO conversion for the class.

src/test/java/edu/cornell/kfs/concur/batch/service/impl/ConcurExpenseV3ServiceUpdateReportStatusTest.java
cah292: Many of our other unit tests already have a convention of adding `throws Exception` to methods annotated with `@Before`/`@BeforeEach` or `@After`/`@AfterEach`. I would prefer to keep that part of the declaration in place.

src/test/java/edu/cornell/kfs/sys/web/mock/MockMvcWebServerExtension.java
cah292: Good idea, I've added a note to the main Bookclub Confluence page to discuss this further at one of our future Bookclub meetings.

cah292:
I've made some changes based on the initial feedback. This is ready for review again.

ajd299 (approved):
Looks good, merging to develop

PR #1366, opened 2022-12-08
KFSPTS-26698 Requisition Item Description Length 254

ajd299:


cah292 (approved):
Changes look good, unit tests pass and my concerns from the previous version of the PR have been addressed. Merging.

PR #1367, opened 2022-12-08
KFSPTS-26901 Update receiptProcessing to create .done file

cah292:
This PR modifies the Receipt Processing Update Job so that when it writes the CSV output files, it will also create corresponding .done files. Such an update is needed to support Donivan's KFSPTS-19965 changes.

The affected file-writing area of the code was a bit outdated and was not using the proper setup for closing the file stream. That section has been updated to adhere to our clean coding practices; further cleanup of the affected class(es) can be performed in a follow-up user story.

I also did some minor rework of a few file-extension-related constants to simplify the setup. We can use a separate user story to perform more in-depth cleanup of our various file extension constants.

In addition, I revised the relevant integration test to look for the generated files afterwards, and updated the related unit/integration tests to rename the sample files to fit the convention expected by our code. Note that it was not very practical for me to validate the integration test locally, so please make sure all of the tests pass before merging.

### DIFF
--- a/src/main/java/edu/cornell/kfs/coa/batch/service/impl/CreateClosedAccountsCsvServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/coa/batch/service/impl/CreateClosedAccountsCsvServiceImpl.java
@@ -230,7 +230,7 @@ public class CreateClosedAccountsCsvServiceImpl implements CreateClosedAccountsC
     private String generateCsvOutputFileName() {
         SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US);
         StringBuilder filename = new StringBuilder(CuCoaBatchConstants.ClosedAccountsFileCreationConstants.OUTPUT_FILE_NAME);
-        filename.append("_").append(sdf.format(getDateTimeService().getCurrentDate())).append(CUKFSConstants.FILE_EXTENSIONS.CSV_FILE_EXTENSION);
+        filename.append("_").append(sdf.format(getDateTimeService().getCurrentDate())).append(CUKFSConstants.FileExtensions.CSV);
         return filename.toString();
     }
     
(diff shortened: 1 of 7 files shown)


cah292:
@dp65 I had at least one question pertaining to the Receipt Processing .done file feature you requested:

Our existing KFS code will create the "CIT_OUT_*" CSV files using the system's default character encoding. For the new .done creation feature, I'm planning to update that code to write such CSV files using the UTF-8 encoding instead. Can you please confirm whether your handling of these CSV files needs the CSVs to use the system default encoding, or if it's okay for them to be created with the UTF-8 encoding instead?

Thanks!

dp65:
@cah292 
I don't really know what the system's default character encoding is, so I can't say. In general, I think it's better if we explicitly specify UTF-8, like you are planning. We will be performing end-to-end testing. If PerceptiveContent has any problems reading the UTF-8 encoded CSV file, we should detect it during that testing.
Actually, I found an email thread from 2016 which indicates that PerceptiveContent expects UTF-8.  I'll forward that email chain to the team via email, and make further comments there.
For the purpose of this PR, I agree that proceeding by explicitly using UTF-8 is the appropriate direction.

nkimble (approved):
The changes in this PR are appropriate to address the immediate need to also create a .done file when the .csv file is generated by the Receipt Processing routines. Approving these changes to achieve that goal.

We definitely need a follow-up ticket to address other issues that I am seeing that are outside the scope of the present user story. IMHO, I believe that all of these additional changes should be performed at the same time so that we can get common logic in place and reduce our code foot print that we need to maintain. Suggested improvements are:

1. Create an abstract class that provides the same base csv file processing being duplicated across classes: CreateClosedAccontsCsvServiceImpl, CreateWorkdayOpenAccountsCsvServiceImpl, JaggaerGeneratedContractsPartyCsvServiceImpl, and ReceiptProcessingServiceImpl
2.  Convert constants hard coded or declared in the service implementations into constants defined in the appropriate common or specific constants classes OR convert those values to system parameters for easier maintenance.
3. Perform general code and import class cleanup.

cah292:
@dp65 Thanks for the feedback. I'm fine with letting this code produce UTF-8 output based on your response, and that addresses my DevOps-related concerns for this. Also, regarding your Byte Order Mark (BOM) comment from the related email thread, I'm not certain whether our code is able to auto-handle it or not; we can perform further investigation in the follow-up ticket.

@nkimble Thanks for your input. I will proceed with creating a follow-up ticket to address the items you brought up as well as Donivan's BOM question. (If the new ticket's scope ends up becoming too large, we can split it up later.) This PR can be merged if you are satisfied with the changes.

nkimble:
Unit tests have passed. My concerns will be addressed in the follow-up work. Merging.

PR #1372, opened 2022-12-13
KFSPTS-23784 fix unit tests after utest refresh

jhulslander:
I changed the user to bs622, who has made recent salary transfer documents in the "R" account range


cah292 (approved):

PR #1373, opened 2022-12-14
KFSPTS-26979 Fix tax PDP handling for chart-to-campus rename

cah292:
NOTE: I'm still wrapping up local validation of this. Please don't merge until I add a comment indicating it's safe to do so.

One of the recent KualiCo patches renamed the PDP Customer Profile's Chart field to Campus instead, but our custom Tax Processing Job did not get updated yet to compensate for that. This PR fixes the issue by making the following changes:

* In the PDP data query that still needs to reference the renamed field, that section and the related areas have been updated to correctly reference the new Campus field.
* In the section that validates the account and org data from the PDP transaction, that section has been updated to derive the Chart from the payment's accounting details instead of from the renamed field. (The functionals have requested this change; see the user story.) This also allowed for removing a hack that temporarily stored the old renamed field inside one of the other data fields.

### DIFF
--- a/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
+++ b/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
@@ -516,7 +516,7 @@ final class TaxSqlUtils {
         appendQuery(selectSql,
                 SqlText.SELECT, docNoteRow.orderedFields, SqlText.FROM, docNoteRow.tables,
                 SqlText.WHERE,
-                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM FS_DOC_HEADER_T WHERE FDOC_NBR = ?)");
+                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM KFS.FS_DOC_HEADER_T WHERE FDOC_NBR = ?)");
         
         // Log and return the query.
         if (LOG.isDebugEnabled()) {
--- a/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
+++ b/src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
@@ -516,7 +516,7 @@ final class TaxSqlUtils {
         appendQuery(selectSql,
                 SqlText.SELECT, docNoteRow.orderedFields, SqlText.FROM, docNoteRow.tables,
                 SqlText.WHERE,
-                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM KRNS_DOC_HDR_T WHERE DOC_HDR_ID = ?)");
+                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM FS_DOC_HEADER_T WHERE FDOC_NBR = ?)");
         
         // Log and return the query.
         if (LOG.isDebugEnabled()) {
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowPdpBuilder.java
```diff
@@ -285,6 +282,7 @@ protected String findincomeClassCode(ResultSet rs, PdpSourceRow pdpRow, String d
     @Override
     void insertNullsForTransactionRow(PreparedStatement insertStatement, TransactionDetailRow detailRow, int offset) throws SQLException {
         super.insertNullsForTransactionRow(insertStatement, detailRow, offset);
+        insertStatement.setString(detailRow.documentTitle.index - offset, null);
```
jhulslander, line 285: Just to confirm, we want the document title to be null?

jhulslander (commented):
Changes look good, I just had one question

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TransactionRowPdpBuilder.java
cah292: If I remember correctly, the tax processing code does two passes for setting up the transaction detail rows. On the first pass, if the main data setup method doesn't explicitly set a value for a given field, then this `insertNullsForTransactionRow()` method needs to explicitly set those unset fields to null so that the SQL insert query will succeed. The code for the second pass should set these fields (including the doc title) to non-null values as needed.

jhulslander: OK, thanks

cah292:
I've added a change to fix another issue that I discovered in local testing:

* Fixed a section of the code that was hard-coded to reference the KRNS_DOC_HDR_T table, which was consolidated into FS_DOC_HEADER_T in a recent KFS upgrade. The section has been updated to point to FS_DOC_HEADER_T instead; we can refactor this further in our future revisions to this batch job.

I'm running another quick local test to verify that this fix works and that there are no further issues.

cah292:
My local run of the job in 1099 mode succeeded with the changes above in place. This is ready to merge once approved.

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
```diff
@@ -516,7 +516,7 @@ static String getDocNotesSelectSql(DocumentNoteRow docNoteRow) {
         appendQuery(selectSql,
                 SqlText.SELECT, docNoteRow.orderedFields, SqlText.FROM, docNoteRow.tables,
                 SqlText.WHERE,
-                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM KRNS_DOC_HDR_T WHERE DOC_HDR_ID = ?)");
+                        docNoteRow.remoteObjectIdentifier, SqlText.EQUALS, "(SELECT OBJ_ID FROM FS_DOC_HEADER_T WHERE FDOC_NBR = ?)");
```
ajd299, line 519: Does the SQL need a schema reference? (kfs.)

cah292: It ran fine without it, but I can add it and then we can validate further in DEV.

ajd299: okay given that the schema reference wasn't there before I will merge.

ajd299 (approved):
Code looks good, merging

src/main/java/edu/cornell/kfs/tax/dataaccess/impl/TaxSqlUtils.java
cah292: Schema reference has been added. Further cleanup of schema references in this process will be handled in separate tickets as appropriate.

PR #1374, opened 2022-12-15
KFSPTS-26956 update CXF from 3.3.10 to 3.4.10 and reconfigure how ehcache is setup in pom file

jhulslander:
The dependabot update to cxf 3.4.10 is failing with a compile time problem where ehcache classes could not be found.  Upon reviewing the POM, it looks like we inherited ehcache from KualiCo.  I tried excluding ehcache from KualiCo jars and adding a new dependency for ehcache (2.10.6), the same version as we had pulled in.  I also found cxf-rt-security had ehcache 3.8.1, which was also excluded.

I am not sure if this is the best solution to doing this but it seems to work out.  Please let me know if you think there is a better approach.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2331,6 +2331,10 @@
                     <groupId>javax.activation</groupId>
                     <artifactId>activation</artifactId>
                 </exclusion>
+                <exclusion>
+                  <groupId>com.sun.activation</groupId>
+                  <artifactId>jakarta.activation</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


cah292 (commented):
I realize this is still a work in progress, but in case this helps I did notice a few things:

1. The unit tests were complaining about not being able to load or use certain JAXB-related classes. You may want to check if the new version of CXF is pulling in some conflicting JAXB dependencies.
2. We're currently using the Ehcache 2.x line that KualiCo also uses at the moment, but one of the newer CXF dependencies is pulling in Ehcache 3.x instead. There are major differences between the Ehcache 2.x and 3.x lines, so we should stay on 2.x for now and only exclude the 3.x ones that CXF brings in. Note, though, that the 3.x version has the "org.ehcache" group ID instead, and I'm not certain whether there's a feasible way to disable Ehcache 3.x usage for CXF.

jhulslander:
retest this please

cah292 (approved):
Changes look good and unit tests pass. We can perform more thorough testing of the web service calls and Ehcache usage in DEV, to ensure that these changes didn't break anything.

I did notice some inconsistencies in the formatting, but we already have a follow-up ticket for cleaning up the POM formatting, so I'm okay with deferring such cleanup to that ticket.

Merging.

PR #1376, opened 2022-12-19
KFSPTS-26275 Add new Concur Request workflow handling

cah292:
Please do not merge this yet. I need to do more local testing but Tami needs to stage some non-PROD Travel Requests first.

This PR updates the Concur Event Notification V2 Job so that workflow actions will now be taken on processed Travel Requests. The Requests will either be approved or sent back, depending on the results of the account validation.

Once Tami stages some Travel Requests I can test the workflow actions further. I especially need to test the "sendback" action, since it needed to be mixed-case for the KFSPTS-23719 Expense changes but I'm not certain whether that mixed-case value is acceptable for the Request API.

To keep the unit testing simpler, the existing Request V4 unit test class has been updated to avoid performing workflow actions, and a new unit test class has been created for explicitly testing the Request workflow calls. (Note that the new test uses some of the same helper objects as the existing test, in addition to using some previously-introduced helper code for simplifying the setup of a mock remote server.) If you notice any areas that seem incorrect or could use further simplification, please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
@@ -22,17 +22,21 @@ import org.kuali.kfs.core.api.datetime.DateTimeService;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.krad.util.UrlFactory;
 import org.kuali.kfs.sys.KFSConstants;
+import org.springframework.http.HttpMethod;
 
 import edu.cornell.kfs.concur.ConcurConstants;
 import edu.cornell.kfs.concur.ConcurConstants.ConcurApiOperations;
 import edu.cornell.kfs.concur.ConcurConstants.ConcurApiParameters;
 import edu.cornell.kfs.concur.ConcurConstants.ConcurEventNoticationVersion2EventType;
 import edu.cornell.kfs.concur.ConcurConstants.ConcurEventNotificationVersion2ProcessingResults;
+import edu.cornell.kfs.concur.ConcurConstants.ConcurWorkflowActions;
 import edu.cornell.kfs.concur.ConcurConstants.RequestV4Status;
 import edu.cornell.kfs.concur.ConcurConstants.RequestV4Views;
 import edu.cornell.kfs.concur.ConcurKeyConstants;
 import edu.cornell.kfs.concur.ConcurParameterConstants;
 import edu.cornell.kfs.concur.ConcurUtils;
+import edu.cornell.kfs.concur.batch.ConcurWebRequest;
+import edu.cornell.kfs.concur.batch.ConcurWebRequestBuilder;
 import edu.cornell.kfs.concur.batch.service.ConcurBatchUtilityService;
 import edu.cornell.kfs.concur.batch.service.ConcurEventNotificationV2WebserviceService;
 import edu.cornell.kfs.concur.batch.service.ConcurRequestV4Service;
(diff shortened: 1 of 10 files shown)


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
```diff
@@ -224,8 +231,102 @@ protected void updateRequestStatusInConcur(
...
+        
+        String requestId = resultsDTO.getReportNumber();
+        String workflowAction = isValid ? ConcurWorkflowActions.APPROVE : ConcurWorkflowActions.SEND_BACK;
+        String logMessageDetail = buildLogMessageDetailForRequestWorkflowAction(
+                workflowAction, requestId, requestUuid);
+        
+        ConcurWebRequest<ConcurRequestV4ReportDTO> webRequest = buildWebRequestForTravelRequestWorkflowAction(
+                workflowAction, requestUuid, resultsDTO);
+        
+        ConcurRequestV4ReportDTO updatedTravelRequest = concurEventNotificationV2WebserviceService.callConcurEndpoint(
```
jhulslander, line 248: I think it makes total sense to use this service.  I wonder if we should change the name of the service since it is no longer used just for event notification.  That may be something we should do in a follow up user story

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotificationProcessingResultsDTO.java
```diff
@@ -25,6 +25,11 @@ public ConcurEventNotificationProcessingResultsDTO(ConcurEventNoticationVersion2
         this(eventType, processingResults, reportNumber, travelerName, travelerEmail, new ArrayList<String>());
     }
     
+    public ConcurEventNotificationProcessingResultsDTO(ConcurEventNotificationProcessingResultsDTO dtoToCopy) {
```
jhulslander, line 28: this may be another example of a class that should be renamed to reflect it just isn't used with event notification

jhulslander (commented):
Changes look good, I noted a couple of things that may be handled on a follow up user story if it isn't feasible to do now.

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurRequestV4ServiceImpl.java
cah292: I thought the service was still being used for event-notification-related purposes, both for reading reports/requests and performing workflow actions on reports/requests. Nevertheless, I agree that the service's name is confusing and that it makes sense to update it in the future. I'll create a follow-up user story for that.

src/main/java/edu/cornell/kfs/concur/businessobjects/ConcurEventNotificationProcessingResultsDTO.java
cah292: I'll add this class to the scope of the user story for renaming the event notification V2 web service.

cah292:
Tami staged some appropriate invalid Travel Requests, but I haven't been able to get them "sent back" as expected. Based on some manual test web calls (via curl/Postman/etc.), I believe the action needs to be "sendback" as stated in the Request V4 docs instead of "sendBack", but Concur is returning "Permission denied" errors despite allowing the "approve" actions in earlier testing. I went ahead and made the "sendback" change in the code, but please do NOT merge until we sort out the potential permission issue with Concur and get the code to successfully perform a send-back action.

cah292:
As per the related discussion on the user story, we are planning to stop using the Travel Request API for Cornell's processing. Thus, the changes from this PR should not be moved forward. We will open follow-up PRs as needed for updating or removing the Travel Request side of the code.

PR #1378, opened 2022-12-21
KFSPTS-27104 adjust 1099 header name box numbers

jhulslander:
This PR is ready for review, it adjusts the header row values for the 1099 MISC.  Local validation shows the NEC file header is not altered.  I was expecting to have to update cu-kfs java code to be in line with changes made to the XML, but it doesn't look like it is required.


db483 (approved):
The changes look good. I ran the tax job in the prekew environment and the results look good to me. I think we should deploy to DEV and have Lori test the changes.

PR #1379, opened 2022-12-22
KFSPTS-26536 Allow for running just unit tests locally

cah292:
This PR adds a POM profile to explicitly exclude the integration tests when running the test suite. This will allow developers to easily run the entire unit/micro test suite while testing out code changes locally, and then they can use the PR-related unit test process to run the unit and integration tests together.

Below are the notable changes for this:

* Added a profile to allow for running only the unit tests. Specifically, the profile will avoid running any tests whose classnames end with "IntegTest" or "IntegrationTest". This is similar (but not exact) to the strategy used in base financials.
* For the existing integration tests that did not already have one of the suffixes above, they have been renamed to include the "IntegTest" suffix.
* `CuAttachmentServiceImplTest` is a micro-test but I was having problems getting it to run locally. I updated the test to properly close its various input streams, and now it seems to be working as expected.
* The test related to CuPaymentDetailInquirable relies upon some base financials code that uses SpringContext calls, but those calls are located in private methods that cannot be overridden. For now I've marked it as an integration test to keep things moving forward, and I've created a follow-up user story to report the limitation to KualiCo. If you're fine with this test being marked as an integration one, then I'll create another follow-up ticket to get it converted back to a micro-test once KualiCo addresses the other issue.

You can use a command like the following to run the unit-test-only suite locally:

`mvn clean test -Prun-only-unit-tests -DskipTests=false`

Note that you may see some PowerMock-related stack traces when running the tests locally, even though the related tests still appear to work as expected. I've created a follow-up user story to look into that further.


cah292:
I saw that one of the integration tests failed due to unexpected empty results from some searches. I've committed a potential fix and will see if the test passes now.

Problematic test is `AccountGlobalSearchLookupableHelperServiceImplTest` (renamed to `AccountGlobalSearchLookupableHelperServiceImplIntegTest` by this PR).

I'm still puzzled as to why it didn't fail before, though, since this PR doesn't have related SQL changes. I also noticed that for other unit tests runs from other recent PRs, they don't appear to be running the problematic class's tests at all. Not sure why that happened or why this PR's renaming of that class would suddenly cause its tests to start running again.

If the recent tweaks do fix the test, I propose creating a follow-up user story to reevaluate the problematic test further, unless you think it needs to be addressed now in this PR.

cah292:
It looks like the problematic integration test is working again for now. This PR should now be ready for review again.

I also created a follow-up user story to reevaluate that integration test class further.

jhulslander:
retest this please

jhulslander (approved):
I agree with your note about the classes
I confirmed the renamed integrationTest classes are running in Jenkins, and I picked a few unit test classes that run locally and they are also being run in Jenkins.

It is very confusing that AccountGlobalSearchLookupableHelperServiceImplTest wasn't running in Jenkins, but it is running after it was re-named.

We have a book club scheduled for Tuesday.  I think we should discuss this then.  I am wondering if there is something that we can do to make sure all of our tests are running

PR #1382, opened 2023-01-12
KFSPTS-25387: Change the following lookups and inquires to use the new frameworks: ContractGrantReportingCode, AppropriationAccount, AccountReversion.

nkimble:
There are dependent SQL changes (PR # 1190) for these code changes. 
Please make sure both PR's are ready before merging.

@cah292 and @db483, 
These are the changes that I am trying to move forward.
I have observed three things while validating these changes locally today.
I wanted to bring them to your attention at the start of the code review in the event you felt they were show stoppers for the functional users.

1.  Financials Parameter with parameter name RESULTS_LIMIT appears to no longer be used by the base code new frame work. This parameter restricts the number of rows returned by the old style look-ups at the system level so that all look-ups only returned 500 rows of data. It appears that this data restriction now needs to be set FOR EACH lookup in each business object's XML. This feels like a step backwards to me.

2. It appears that the lookup definition bean values used to restrict the number of rows returned by the lookup are not working correctly (or at least for me they are not in my local environment when the look-up is configured for the defaultSearchService). I attempted to restrict the results returned by the lookup to "500" by setting resultSetLimit and/or multipleValuesResultSetLimit on each respective lookup bean and that restriction was not adhered to (meaning more than 500 rows of data were returned). I used them each separately and together to no avail on the three business objects upgraded by this code. I did not leave those values in the code as they were not working appropriately. This may be an Oracle vs MySQL issue I am stumbling on or may be a configuration issue on my part. Please advise as to how you think we should proceed. IMHO we are looking at performance problems if we cannot get this base code lookup restriction to work. The values I used are listed below.
p:multipleValuesResultSetLimit="500"
p:resultSetLimit="500"

3. The sort order for the attributes are not the same between the old and new frameworks even when the bean setting is maintained. I saw this anomaly for all three look-ups that were converted by comparing the lookup search results from DEV (or INTDEV  for the CG item) with my local environment when the same search attributes were used. This may be another Oracle vs MySQL issue we are encountering or may be a configuration issue on my part. Please advise as to how you think we should proceed.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AccountReversion.xml
+++ b/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AccountReversion.xml
@@ -1,218 +1,265 @@
-<?xml version="1.0" encoding="UTF-8"?><beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://www.springframework.org/schema/p" xsi:schemaLocation="http://www.springframework.org/schema/beans         http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
-<!--
- Copyright 2006-2008 The Kuali Foundation
- 
- Licensed under the Educational Community License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
- 
- http://www.opensource.org/licenses/ecl2.php
- 
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- See the License for the specific language governing permissions and
- limitations under the License.
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans" 
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+       xmlns:p="http://www.springframework.org/schema/p" 
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
+
+    <import resource="classpath:org/kuali/kfs/sys/sys-lookup-beans.xml"/>
+    <import resource="classpath:edu/cornell/kfs/sys/cu-sys-attribute-beans.xml"/>
+    
+    <bean id="AccountReversion" parent="AccountReversion-parentBean"/>
+    <bean id="AccountReversion-parentBean" abstract="true" parent="BusinessObjectEntry"
+          p:actionsProvider-ref="businessObjectActionsProvider"
+          p:businessObjectAdminService-ref="defaultBoAdminService"
+          p:businessObjectClass="edu.cornell.kfs.coa.businessobject.AccountReversion"
+          p:inquiryDefinition-ref="AccountReversion-inquiryDefinition"
+          p:lookupDefinition-ref="AccountReversion-lookupDefinition"
+          p:objectLabel="Account Reversion"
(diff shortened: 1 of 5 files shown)


cah292:
@nkimble,

I haven't reviewed the code changes in detail yet, but here's what I've found so far while reviewing your initial notes:

1. I agree that the new lookups do not appear to be using RESULTS_LIMIT to restrict the returned rows. However, the new lookups do have an alternate setup for limiting rows; see the next item below.

2. Instead of using the parameter above or the lookup-bean-specific search limit properties, it looks like the new lookups have a different strategy for handling result limits. The new lookup web requests accept `skip` and `limit` attributes, which control how many initial result rows to skip and how many non-skipped rows to return, respectively. Also, the associated code in `LookupDaoOjb.findObjects()` appears to be performing two queries: One to retrieve the items for the current page of results and one to return the count of all rows matching the criteria. Thus, it seems that the new lookups are returning the actual result count instead of the truncated row count, and are also attempting to limit the items to just the current page of results. As to whether the underlying SQL query is truly limiting the results or is just retrieving everything before discarding out-of-range items, we would need to analyze the search's OJB behavior further.

3. The new lookups appear to be sorting on only one field at a time and are doing so at the SQL and/or OJB level. The old lookups might be sorting on the results after retrieving them instead of during the query, though more analysis is needed to confirm for certain. That could explain the sorting discrepancies, but if we wanted to make the new strategy sort like the old one, I'm not sure what our options would be.

cah292:
@nkimble,

There's another item I wanted to note on this update-COA-lookups PR. It looks like the feature branch does not contain the 02/23/2022 financials patch changes. If you encounter unexpected problems when trying to boot up the new "sandbox" environment with this code, you may need to rebase the feature branch first (or prepare a new branch).

nkimble:
Rebased develop branch into/onto this feature branch to obtain latest kualico patch code.

cah292:
@nkimble,

I've been doing some more analysis on this while reviewing the sandbox environment behavior and digging into the code further. I agree that the new lookup's row-limiting behavior is not limiting the queried data enough. Here's where I think the main sizing discrepancies are between the old and new ways:

* The old way: This seems to use the `LookupDaoOjb.findCollectionBySearchHelper()` method. One of its downstream calls will add the expected size-limiting criteria as needed, where the appropriate helper class will generate some DB-platform-specific query for limiting the number of rows returned. (Since we're using Oracle, this would be the `OracleDatabasePlatform.applyLimitSql()` method.)
  * Because of where the row-limiting setup is located, such queries generally have their values sorted AFTER limiting the result size.
  * Also, navigating between result pages seems to be using the cached result rows. Such functionality might not be present on the new lookups.
* The new way: This seems to use the `LookupDaoOjb.findObjects()` method. It uses OJB-query-related methods to set the start and end indexes for the search. Based upon further review of the OJB code, it looks like such queries will return a "scrollable" ResultSet that can potentially iterate over all matching rows (ignoring size limit), and then OJB will use scrolling/paging to return only a given number of rows starting from a given position.
  * Because of where the row-limiting setup is located, such queries generally have their values sorted BEFORE limiting the result size. (Some in-memory sorting of the returned results is also possible, though, if the appropriate configuration is performed.) Also, as noted during earlier analysis, the new way only supports sorting on one field at a time.

If we want to improve the efficiency of the new lookup functionality, here are a few potential options:

1. Update the new lookup querying to forcibly include the row-limiting functionality from the old lookup framework. This might be the easiest to implement, but it might not allow the lookups to report the search's actual count.
2. Update the new lookup querying to wrap the main query in a row-limiting one. (Not sure what the query efficiency would be for such a solution.) This would be a bit complex to implement, since it may involve using the DAO's underlying PersistenceBroker (and its helper services) to generate the SQL manually and then potentially wrap it in an OJB `QueryBySQL` object. The wrapped query might look something like this:
```sql
SELECT * FROM (actual_selection_query) WHERE ROWNUM BETWEEN x AND y;
```

PR #1392, opened 2023-01-30
KFSPTS-26953: ISO 20022 Changes to Payee ACH Extract Job and Create New Payee ACH Job to populate the Standard Entry Class.

nkimble:


nkimble:
The only difference between the reverted commit that was previous reviewed and this PR are the changes to file PaymentWorksVendorAchDataProcessingIntoKfsServiceImpl.java

cah292 (approved):

PR #1393, opened 2023-01-30
KFSPTS-22413 Remove custom Benefit Rate Category rules

cah292:

This PR removes our custom Benefit Rate Category validation from the Account doc, because base financials code now has equivalent validation in place. The appropriate base validation will occur when the related boolean parameter is enabled (which is already enabled across our environments). Our addition of the error message property has also been removed since it seems to duplicate the one found in base code.

Note that the parameter referenced in the error message is actually supposed to have an "_IND" suffix. I will create a follow-up user story to report that naming discrepancy back to KualiCo.


ajd299:
The code properly removes duplication of functionality and a property/constant that now exist in the KFS base code.

ajd299 (approved):

PR #1395, opened 2023-01-31
KFSPTS-26897 This is only a test! DO NOT MERGE!

cah292:
This PR is only for testing purposes. DO NOT MERGE!


cah292:
retest this please

cah292:
I confirmed in the Jenkins job console output that the appropriate variables were overridden as a result of an emergency branch being chosen as the target. The unit tests ran successfully except for a few Jaggaer-related ones, but that is likely the result of a Jaggaer-side issue in their environment.

Testing was successful.

PR #1396, opened 2023-01-31
KFSPTS-26897 TEST Upgrade to 01/01/2099 version of financials TEST

cah292:
This is only a test! DO NOT MERGE!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ Cornell University Customizations Project
 ### About
 This project provides customizations to the Kuali Financial System (KFS) to support Cornell University business processes and requirements.
 
+### Next Test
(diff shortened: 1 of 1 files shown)


cah292:
I confirmed in the Jenkins console output that the job successfully ran the financials upgrade prop injection, due to the PR's title indicating that it was intended for a financials upgrade. One of the unit tests failed as a result of some unexpected table structure in the database, but that is likely the result of the ongoing work that Daniela is doing for the next actual financials upgrade.

This seems to be working as intended, although the logs mentioned that the selected config file has "kew" in the name. We may want to clean up that naming convention, if not in the related DevOps user story then in a future DevOps ticket.

PR #1398, opened 2023-01-31
KFSPTS-27209: I Want Doc update language on ad hoc approver screen.

nkimble:


nkimble:
The language changes for these changes were provided by the functional users and are located on sub-task KFSPTS-27210 of this user story.

src/main/webapp/WEB-INF/tags/module/purap/iWantAdHocRecipients.tag
```diff
@@ -17,9 +17,9 @@
                         <strong>NOTE:</strong>
                     </p>
                     <ul>
-                        <li>Do not enter your Net ID in the box below.</li>
-                        <li>If additional approvals are required prior to submitting to your Service Center for processing, please list the individual's Net ID below.</li>
-                        <li>Please click the <strong>Submit</strong> button to route to the named approver or your Service Center.</li>
+                        <li>Do <strong>not</strong> enter your Net ID in the Approver's box below.</li>
+                        <li>If you did not enter an <strong>account number</strong> in Step 2/Account Information, enter the Net ID of the person in your department who can provide that information. </li>
```
cah292, line 21: Should "Net ID" be written as "NetID" (1 word)? I know the instructions previously said "Net ID" but Mary's comments on the sub-task suggest using "NetID" instead.

cah292 (commented):
I had one minor question on the new wording, but otherwise the changes look good.

src/main/webapp/WEB-INF/tags/module/purap/iWantAdHocRecipients.tag
nkimble: The wording provided by Mary had the term used to represent network identifier as both NetId and NetID.  

The label for the text box where that information should be entered on that same panel right below where these change are going uses the term "Net ID". I went with the term that I did for consistency and due to the fact that we are not changing that label.

cah292: Ok, thanks for explaining. Given that, I'm okay with the form saying "Net ID" for now; we can clean that up in a future ticket if necessary.

dp65: Sorry for chiming in late.  For reference: officially (https://it.cornell.edu/netids), Cornell uses "NetID" (Capitalized "Net", no space, and "ID" in all caps). I agree that consistency with items on the same panel is a good thing. Also, if this usage is in reference to the Cornell network identifier, which I think it is, I vote for cleaning it up at some point to be consistent with Cornell's spelling.

PR #1401, opened 2023-02-06
KFSPTS-27518 Remove expanded textarea from IWNT

cah292:
The 2022-04-13 financials upgrade removed the expanded textarea feature from KFS, but we accidentally forgot to remove the references to it from the IWNT document. This PR removes the expanded textarea icons from the IWNT document so that it will render successfully once more. The "wrap-break-word" CSS class has also been added, for consistency with similar removals of this feature from other documents.


jhulslander (approved):
Changes look good and can be merged once the unit tests complete

PR #1403, opened 2023-02-07
KFSPTS-27531 Fix PREQ date check for Oracle compatibility

cah292:
The 2022-04-13 financials upgrade added an Invoice Date check to the PREQ creation process, but the related code is passing a `java.time.LocalDate` parameter to the query instead of a `java.sql.Date` parameter. This seems to work fine in KualiCo's environment(s) since they're likely using a newer MySQL JDBC driver; however, our environments are encountering exceptions because our Oracle JDBC driver doesn't recognize the `java.time.*` classes as being valid for SQL date-time purposes.

This PR overrides the problematic method to convert the `java.time.LocalDate` object into a `java.sql.Date` object when adding the Invoice Date criteria to the query. That will at least fix the problem for the short term. In the long term, though, we may want to investigate updating our Oracle JDBC driver version, in case KualiCo starts using `java.time.*` objects in query parameters more frequently. I will create a follow-up user story for further analysis.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/dataaccess/impl/CuPaymentRequestDaoOjb.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/dataaccess/impl/CuPaymentRequestDaoOjb.java
@@ -1,22 +1,31 @@
 package edu.cornell.kfs.module.purap.document.dataaccess.impl;
 
 import java.sql.Date;
+import java.time.LocalDate;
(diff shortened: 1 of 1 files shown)


cah292:
It looks like we already have KFSPTS-18715 to look into updating the driver version. I'll update that ticket with notes about the date-time issue.

db483 (approved):
Changes look good.

PR #1404, opened 2023-02-09
KFSPTS-25387: Change the following lookups and inquires to use the new frameworks: ContractGrantReportingCode, AppropriationAccount, AccountReversion.

nkimble:


nkimble:
SQL PR https://github.com/CU-CommunityApps/nonprod-sql/pull/1207 depends on these code changes.
Please make sure both are good to go before merging.

Also be aware that there is a section of code introduced as a comment by this PR in the AccountReversion.xml file. Further analysis is needed before these changes can be successfully implemented. We did not want the effort already expended to be lost that is why that section is commented out. The analysis and coding to implement will be dealt with in future user stories in order to constrain the scope for this current body of work.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/ContractGrantReportingCode.xml
```diff
@@ -1,168 +1,128 @@
...
-        <property name="attributeNames">
-          <list>
-            <value>chartOfAccountsCode</value>
-            <value>code</value>
-          </list>
+    <bean id="ContractGrantReportingCode-lookupDefinition" parent="ContractGrantReportingCode-lookupDefinition-parentBean"/>
+    <bean id="ContractGrantReportingCode-lookupDefinition-parentBean" abstract="true" parent="LookupDefinition"
+          p:title="CG Reporting Code Lookup">
+        <property name="defaultSort">
+            <bean parent="SortDefinition" p:sortAscending="true">
```
cah292, line 102: Another lookup whose default sort was previously configured for descending order but is not configured for ascending order.

cah292 (changes requested):
I found a few minor items that should be addressed, but otherwise the changes look good. Also, I see that a separate user story has already been created to clean up the Account Reversion Inquiry further, so I'm fine with waiting until then to clean up the commented inquiry XML config.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AppropriationAccount.xml
```diff
@@ -1,164 +1,147 @@
...
-    <property name="title" value="Appropriation Account Lookup"/>
-    
-    <property name="defaultSort">
-      <bean parent="SortDefinition">
-        <property name="attributeNames">
-          <list>
-            <value>appropriationAccountNumber</value>
-          </list>
+    <bean id="AppropriationAccount-lookupDefinition" parent="AppropriationAccount-lookupDefinition-parentBean"/>
+    <bean id="AppropriationAccount-lookupDefinition-parentBean" abstract="true" parent="AppropriationAccount-lookupDefinition-base-parentBean"/>
```
cah292, line 116: Upon further review, I noticed you added another parent bean layer to the Appropriation Account lookup, likely as a result of following the setup for the Account lookup. I'm okay with this moving forward as such since I didn't find anything else in my second review that needs attention. In the future, though, we may want to remove the third bean if we don't intend to override this lookup's setup in our other module configs.

cah292 (approved):
Changes look good, unit tests pass and my main concerns have been addressed. The rest of the follow-up cleanup and fixes will be handled in separate user stories and/or by future KualiCo changes. (Also, the related SQL changes are good to go.)

Merging.

PR #1405, opened 2023-02-09
KFSPTS-23443 Make ClamAV properties configurable

cah292:
There are changes for this in cu-kfs, cornell_customizations, cynergy-server-customizations and cynergy-adf-core-lib. Please make sure all 4 PRs are good to go before merging. Also, it would be helpful to test at least the cu-kfs changes in the sandbox environment before merging.

This PR updates KFS to allow for specifying which ClamAV server to point to, rather than only allowing a local ClamAV server. The app will still point to a local ClamAV server by default.


jhulslander (commented):
This seems like a reasonable approach

cah292:
As per the discussion at today's standup, we can now proceed with merging these changes once they have been approved.

jhulslander (approved):

PR #1406, opened 2023-02-13
KFSPTS-27313 fixing bug with CuUiDocumentServiceImpl when the role can not be retrieved yet

jhulslander:


cah292 (approved):

PR #1407, opened 2023-02-13
KFSPTS-26954 Add CU parm to control ISO 20022 output

cah292:
There are PRs for this change in cu-kfs and nonprod-sql. Please make sure both are good to go before merging.

This PR adjusts the Check/ACH extraction process so that it will analyze a CU-specific parameter to determine whether to use ISO 20022 output. The base financials parameter normally controls both the extraction output and the business rules validation on certain docs, but with this change we can keep those two scopes separated until we finish the ISO 20022 migration. We intend to remove the CU-specific parameter after the ISO 20022 transition is complete.

Note that, due to how some of the method visibility has been set up in base code, I added an intermediate abstract helper class to assist with some of the configuration. The new class should help keep some of the customized code to a minimum for now, but if/when we need to overlay the ISO 20022 extractor class in a future ticket, we can remove the helper class at that time. If you have suggestions on a better setup for this, please let me know.


ajd299 (commented):
Code looks good

ajd299 (approved):

PR #1408, opened 2023-02-14
KFSPTS-27595 Backport FINP-9050 and FINP-8777 route log fixes

cah292:
This PR backports the FINP-9050 feature that allows certain special document actions (like a manual doc requeue) to add a message to the Route Log. This also required backporting a portion of the FINP-8777 fix, so that the messages would actually get written to the Route Log. (Much of FINP-8777 consisted of code formatting changes, so I only backported the part that contains the critical `saveActionTaken()` call.)

Note that FINP-9050's new REQS message in `RequisitionAction` contained a minor typo, which has been fixed in our overlay for the moment. Furthermore, during my testing I also discovered an existing issue, where document types with nested "split" nodes cannot be exported successfully. (The latter was likely caused by FINP-7919 from the 2022-01-12 patch.) I will create follow-up user stories to report both of these back to KualiCo; we might also need a local fix for the latter issue, but we can handle that in a separate user story.

### DIFF
--- a/src/main/java/org/kuali/kfs/kew/actionrequest/service/impl/ActionRequestServiceImpl.java
+++ b/src/main/java/org/kuali/kfs/kew/actionrequest/service/impl/ActionRequestServiceImpl.java
@@ -1,0 +1,836 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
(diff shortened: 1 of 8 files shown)


db483 (approved):
Backport looks good.

PR #1415, opened 2023-03-01
KFSPTS-26964 Upgrade ISO 20022 library version

cah292:
This PR updates cu-kfs to use a newer version of the related ISO 20022 library; specifically, the version that base financials upgraded to for its 2022-08-17 patch. We can potentially remove this change in the future once we've upgraded to the 2022-08-17 patch or later.


jhulslander (approved):
changes look good.  I shall merge.  We should create a follow up user story to remove this customization when we get to appropriate version of KualiCo financials

PR #1416, opened 2023-03-01
KFSPTS-18715 Upgrade database driver

cah292:
Please do ***NOT*** merge this yet! If these changes move forward mostly as-is, then we'll need to coordinate with DevOps to add another system property setting when migrating to our various KFS environments. It would also be helpful to test this in the sandbox environment first.

This PR upgrades the Oracle JDBC driver to the latest stable one that's compatible with our Oracle version. Note that, in order for `java.sql.Date` objects to work properly with the new driver, we need to add the following JVM property (as previously noted on the user story):

```
-Doracle.jdbc.DateZeroTime=true
```

This PR also removes a temporary workaround from the Payment Request creation process, where we were manually converting a `java.time.LocalDate` into a `java.sql.Date`. Such a conversion is not necessary with the new driver, since it seems to be able to handle the `java.time` objects properly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
                     </dependency>
                 </dependencies>
                 <configuration>
-                    <argLine>-Xms1024m -Xmx1024m</argLine>
+                    <argLine>-Xms1024m -Xmx1024m -Doracle.jdbc.DateZeroTime=true</argLine>
                     <excludes>
                         <exclude>${test.excludes}</exclude>
                     </excludes>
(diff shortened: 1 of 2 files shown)


db483 (approved):
Changes look good. I will merge after we get confirmation that this works as expected in sandbox.

cah292:
I've committed a change to have KFS use a not-as-new JAR that may have better compatibility with our current setup. I'm awaiting testing from the functionals to see if it improves things.

cah292:
I've rebased this PR from the 07/13-upgraded develop branch. Please make sure DevOps completes their DEV load testing for the 07/13 patch **BEFORE** you merge these changes.

jhulslander (approved):
We are ready to get these changes into develop now

PR #1417, opened 2023-03-06
KFSPTS-27397 Upgrade to 2022-06-29 version of financials

cah292:
Please make sure the PRs for this in cu-kfs, nonprod-sql and cu-kfs-upgrades get merged in conjunction with each other.

Also, please do *NOT* merge this until we're ready to coordinate with DevOps on some required config file changes. See the Redis-related entry below for details.

Here are some important notes about this upgrade:

* KualiCo updated the way in which the caching configuration is set up. Instead of using a config property to import either the Ehcache bean file or the Redis bean file, KFS now relies on a new Spring profile named `cache-redis` to control the configuration. (Specifying that profile will cause the Redis config to be loaded; not specifying it will cause the regular Ehcache config to be loaded.)
  * As a result of this, we will need to coordinate with DevOps to remove the `spring.source.files` property override from any environments containing the 2022-06-29 financials code. Similarly, all developers will need to remove that property override from their local configs. (As far as I can tell, the unit test config files don't need further adjustments for this.)
  * At the moment, this PR configures `web.xml` to include the `cache-redis` Spring profile, enabling it for anyone using the upgraded code (at least in non-unit/integration-test cases). If we still want to allow developers to run the Ehcache-only setup locally, we may need to come up with some other way to enable the appropriate settings (such as using a combination of Maven profiles and WAR filtering to inject the setting into `web.xml`). Please let us know your thoughts on this.
* While cleaning up the POM, it was discovered that KFS and CXF were including slightly different versions of `jakarta.activation-api`. For the moment, we're sticking with the 1.2.1 version from base financials, though CXF is using version 1.2.2 instead. Please let us know if you think we should use the 1.2.2 version instead.
* We removed the method override  `CuAutoDisapproveDocumentsServiceImpl.checkIfDocumentEligibleForAutoDispproval()`, because base code has now added doc type parent checking similar to what our customization had introduced. (And yes, base code does have that minor typo in the method name.) Please let us know if you notice anything further that needs to be fixed about that.
  * NOTE: Upon further analysis, it was discovered that our code had a bug in method `checkIfRunDateParameterExists()`, where one part was accidentally validating the wrong parameter. This was a pre-existing problem in our code, so a separate user story will be created to address it.
* We had to add the `KSBThreadPoolImpl` overlay back in to implement a fix, because KualiCo accidentally reintroduced a core/max size ordering bug. See KFSPTS-27768 for more details.
* This upgrade introduced some changes that ended up breaking all of the integration tests. To fix this, we had to backport the FINP-8690 changes from the 2022-07-20 financials patch.


cah292:
retest this please

db483:
@cah292, regarding the redis cache configuration change in FINP-8536, will setting the spring.profiles.active environment variable to redis-cache/leave epmty(for Ehcache as I understand it) the way KualiCo suggests (https://kuali-financials.zendesk.com/hc/en-us/articles/4407897132443-Distributed-Caching) not work for us?

db483:
If the spring_profiles_active env variable works as expected then I am not sure why we need to configure the Spring profile in the web.xml

cah292:
After further review and consideration, it looks like we can insert the Spring profile through a VM argument. That would make it much easier to customize the setting, and we'll already need DevOps involvement for the config property change so we can just add this VM argument change on top of that.

I've removed the `spring.profiles.active` setting from the `web.xml` file. I also confirmed that adding the following to the Tomcat VM arguments will properly enable Redis in KFS:

```
-Dspring.profiles.active=cache-redis
```

pom.xml
```diff
@@ -57,13 +57,13 @@
...
-        <plugin.dependency.version>3.2.0</plugin.dependency.version>
-        <plugin.war.version>3.3.1</plugin.war.version>
+        <plugin.surefire.version>3.0.0-M6</plugin.surefire.version>
+        <plugin.dependency.version>3.3.0</plugin.dependency.version>
+        <plugin.war.version>3.3.2</plugin.war.version>
         <jaxws-ri.version>2.3.0</jaxws-ri.version>
         <jta.version>1.1</jta.version>
         <log4j.version>2.17.2</log4j.version>
-        <ehcache.version>2.10.6</ehcache.version>
+        <ehcache.version>2.10.10.3.24</ehcache.version>
```
db483, line 66: Sorry for not noting this during my initial review. For the ehcache version change on FINP-8521, KualiCo added an entry to the pom file prefixed with `<!-- The most recent EhCache 2.x releases are not published to Maven Central for some reason. -->`
Do we need to add this to our pom as well or are we getting the correct ehcache jars as it is? Thank you

db483:
@cah292 Thank you for making the changes related to the spring profile. That looks good. I noticed one other thing related to the pom changes. Thank you.

pom.xml
cah292: When compiling locally, I was able to get the appropriate EhCache JAR downloaded without adding extra repository handling. I agree that this particular EhCache version is not available in the central Maven repository, as KualiCo's comment indicated; however, based on the info in my local Maven repo, my local machine was able to download the JAR successfully from KualiCo's financials release repository.

db483: Thank you for checking this. Since KualiCo had an issue with this, would it make sense to check the dev and test wars to make sure they are using the correct version after the 6/29 upgrade?

cah292: We could manually check the DEV and TEST WARs for the dependency version if needed, but since we're referencing the newer version explicitly in our POM, I suspect the DEV/TEST WARs would fail to build if the dependency couldn't be found.

Also, since we now seem to be using the same EhCache version as KualiCo, we could look into removing the EhCache dependency setup from our POM in a future user story.

db483 (commented):
Changes look good and my concerns have been addressed. I will merge when we are ready to move to DEV.

db483 (approved):

PR #1421, opened 2023-03-15
KFSPTS-27197: IWant Document Show Full Page and Show Multiple Pages buttons added that allows a user to switch the presentation when edoc status is Initiated or Saved.

nkimble:


src/main/java/edu/cornell/kfs/module/purap/document/IWantDocument.java
```diff
@@ -50,6 +50,8 @@ public class IWantDocument extends FinancialSystemTransactionalDocumentBase impl
     private static final long serialVersionUID = 1L;
 
     private String step;
+    private String presentationMode;
+    private boolean usersRequestingPresentationModeChange;
```
cah292, line 54: I see that these are both non-persisted fields, but I also see that you defined them on both the document itself and the document form. Can you please explain why you need them on both objects and not just on the form? (I see that the existing `step` variable is in both places, but do the other two variables also need to be on both?)

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentForm.java
```diff
@@ -37,7 +39,8 @@ public IWantDocumentForm() {
         newSourceLine = new IWantAccount();
         this.setDeptOrgKeyLabels(new ArrayList<KeyValue>());
         editingMode = new HashMap<Object,Object>();
-
+        presentationMode = CUPurapConstants.IWantPresentationModes.MULTIPLE_PAGE_MODE;
+        usersRequestingPresentationModeChange = true;
```
cah292, line 43: I'm only seeing `usersRequestingPresentationModeChange` being set to true. Is it supposed to lose its value and revert to false if the user takes a document action that doesn't call the related setter? If so, could the variable be renamed to make its intent clearer, or could this be implemented in a clearer fashion?

cah292 (changes requested):
Changes look good overall, though I did have a few questions and suggestions.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentForm.java
```diff
@@ -205,8 +211,38 @@ public List<ExtraButton> getExtraButtons() {
...
+                isMultiplePagesAllowed && isUsersRequestingPresentationModeChange()) {
+            extraButtons.add(createShowMultiplePagesButton());
+        }
+    }
+    
+    protected ExtraButton createShowFullPageButton() {
+        ExtraButton showFullPageButton = new ExtraButton();
+        showFullPageButton.setExtraButtonProperty("methodToCall.showFullPage");
+        showFullPageButton.setExtraButtonSource("${" + KFSConstants.EXTERNALIZABLE_IMAGES_URL_KEY + "}buttonsmall_continue.gif");
+        showFullPageButton.setExtraButtonAltText("Show Full Page");
```
cah292, line 235: This button name doesn't line up with the yellow-highlighted one that was mentioned in the new user story. Can you please update this if necessary?

The same goes for the other new button.

cah292 (changes requested):
I noticed one other thing; sorry for not mentioning it in the main review.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentForm.java
nkimble: Thanks for asking about this boolean flag. It was used during initial development of this feature but subsequent logic adjustments have resulted in it no longer being needed. I have removed it.

nkimble (commented):
Requested changes have been made. This is ready to be reviewed again.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
```diff
@@ -67,18 +67,35 @@ protected void loadDocument(KualiDocumentFormBase kualiDocumentFormBase) {
...
+                iWantForm.setStep(CUPurapConstants.IWantDocumentSteps.CUSTOMER_DATA_STEP);
+                iWantDocument.setStep(CUPurapConstants.IWantDocumentSteps.CUSTOMER_DATA_STEP);
+
+                if (StringUtils.isNotBlank(iWantDocument.getCurrentRouteToNetId())) {
+                    iWantForm.getNewAdHocRoutePerson().setId(iWantDocument.getCurrentRouteToNetId());
+                }
+
+            } else if (!iWantDocument.getDocumentHeader().getWorkflowDocument().isInitiated()) {
+              iWantForm.setStep(CUPurapConstants.IWantDocumentSteps.REGULAR);
+              iWantDocument.setStep(CUPurapConstants.IWantDocumentSteps.REGULAR);
```
cah292, line 84: There are a couple lines here where the indentation is still off a bit.

cah292 (changes requested):
I noticed one very minor spot that would be good to clean up, but otherwise the revisions look good.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
nkimble: fixed.

nkimble: Formatting has been fixed and unit tests are running.

cah292 (commented):
Changes look good, I'll merge this once the unit tests pass.

cah292 (approved):

PR #1422, opened 2023-03-17
KFSPTS-27774 Fix parameter validation bug in Auto Disapprove Documents Job

db483:


jhulslander (approved):
changes look good.  I confirmed the variable has a value of YEAR_END_AUTO_DISAPPROVE_DOCUMENT_RUN_DATE and we do have that parameter defined in kfs

PR #1426, opened 2023-03-22
KFSPTS-27882 fix EIRT match process to only retrieve active vendors by DUNS  number

db483:


jhulslander (approved):
this PR is what we put into emergency. this should be fine for develop

PR #1427, opened 2023-03-23
KFSPTS-27905 fix PREQ exception

db483:


jhulslander (approved):
this is the change that got validated in emergency.  Moving to develop should be fine

PR #1430, opened 2023-03-24
KFSPTS-27950 upgrade spring version from 5.3.20 to 5.3.26

db483:


pom.xml
cah292, line 1: The `spring-jcl` dependency still needs to be updated to the newer version. It looks like we're currently excluding it from our `spring-core` dependency since KualiCo was initially doing that in their POM; however, KualiCo removed that exclusion from their POM, as part of their FINP-7989 changes for the 2021-12-01 financials patch. (We must not have noticed the need for removing our exclusion, since KualiCo was using the same or almost the same version of Spring at the time.)

In addition to removing the `spring-jcl` exclusion from the `spring-core` dependency, you may need to add exclusions for it on the KFS dependencies and/or the `maven-war-plugin` configuration.

cah292 (changes requested):
I noticed one transitive dependency that may still need to be updated. Other than that, the changes look good.

pom.xml
db483: Thank you for the feedback. I will clean up the spring-jcl dependency.

PR #1434, opened 2023-03-29
KFSPTS-27973 fix null pointer exception

db483:


jhulslander (approved):
parameter string change is correct as I can look up the parameter in prod
The logic change looks reasonable too

I shall merge

PR #1435, opened 2023-03-29
KFSPTS-26270 Add DTOs to handle Jaggaer Contract XML

cah292:
**NOTE:** Before merging these changes, we should determine whether Jay's #1418 changes can be merged first. I had structured my unit tests to avoid potential interference with Jay's XMLUnit-related additions, but if Jay's changes get merged then I could potentially update some of my tests to take advantage of XMLUnit.

This PR introduces the DTOs needed for handling the Jaggaer Contract XML exported by Jaggaer. There are two sets of DTOs: Those representing the XML received from Jaggaer, and those representing the XML response to be sent back to Jaggaer. (Some of the DTOs, such as the "Header" one, are used by both sets.) The actual processing for these DTOs will be added by subsequent tickets.

Note that the DTOs don't include all of the attributes/elements/sub-elements that are defined in the related Jaggaer DTD files. The DTOs have been simplified to primarily include the needed items, plus any items that may provide helpful reference information when processing or validating the data.

The PR also introduces a new date-related XML adapter for handling certain date-time values. Since KualiCo seems to be moving away from the Joda Time library (based on their recent release notes), I set it up to use the `java.time` objects instead, while still being interoperable with the legacy `Date` objects for now. Please let me know if you think this needs to be structured differently.

A few basic XML marshalling/unmarshalling unit tests have been included, to verify that the DTOs are created properly despite ignoring several unneeded XML chunks, plus to verify that the basic response XML can be built correctly. Only one XML input test case is included at the moment, since I'm currently only able to export the `<Contract>` element (and its children) from Jaggaer without fully setting up the export process; I had to manually construct the root and header sections, along with some dummy Base64-encoded attachment content. Further test cases can be added by future tickets, once we set up more of the process and can see what the actual full Contract XML will look like.

Note that this code aims to import/export the DTD declarations in the XML, for consistency with some sample XML presented in Jaggaer's documentation. To avoid auto-DTD-validation problems when unmarshalling and to simplify outputting the DTD declaration when marshalling, some appropriate helper methods have been added to the  `CUMarshalService` class.


cah292:
For more information on the new Jaggaer Contract XML handling, see the following Confluence page:

https://confluence.cornell.edu/x/Uoz2Gg

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Attachment.java
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.xml.bind.annotation.XmlElement;
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = {
+    "attachmentFile"
+})
+@XmlRootElement(name = "Attachment")
+public class Attachment {
```
jhulslander, line 14: There is an Attachment object in Jaggaer supplier, the two are very different.  I wonder if the two should be renamed so they are context specific?  Like this one being AttachmentContract, and the other being AttachmentSupplier?

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Authentication.java
```diff
@@ -0,0 +1,39 @@
...
+import jakarta.xml.bind.annotation.XmlRootElement;
+import jakarta.xml.bind.annotation.XmlType;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = {
+    "identity",
+    "sharedSecret"
+})
+@XmlRootElement(name = "Authentication")
+public class Authentication {
```
jhulslander, line 15: I think this is a duplicate of edu.cornell.kfs.module.purap.jaggaer.supplier.xml.Authentication. Can contacts use the already existing object?  If so, maybe the existing one should be moved to a different package indicating it is used by multiple Jaggaer integrations

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Contract.java
```diff
@@ -0,0 +1,162 @@
...
+    public void setSummary(String summary) {
+        this.summary = summary;
+    }
+
+    public KualiDecimal getContractValue() {
+        return contractValue;
+    }
+
+    public void setContractValue(KualiDecimal contractValue) {
+        this.contractValue = contractValue;
```
jhulslander, line 127: On the the supplier side, the List object getters instantiate the List if it is null.  Should we do that here?

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/CustomField.java
```diff
@@ -0,0 +1,58 @@
...
+import jakarta.xml.bind.annotation.adapters.NormalizedStringAdapter;
+import jakarta.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = {
+    "internalName",
+    "customFieldValues"
+})
+@XmlRootElement(name = "CustomField")
+public class CustomField {
```
jhulslander, line 21: Supplier has a similar object called CustomElement, however they contain different values. Should we consider re-naming here to avoid confusion?

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Header.java
```diff
@@ -0,0 +1,56 @@
...
+import jakarta.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
+
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlType(name = "", propOrder = {
+    "messageId",
+    "timestamp",
+    "authentication"
+})
+@XmlRootElement(name = "Header")
+public class Header {
```
jhulslander, line 20: Supplier has a Header object as well, it contains all the fields here, plus relatedMessageId.  Can supplier and contract share?

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/ManagerList.java
```diff
@@ -0,0 +1,31 @@
...
+    @XmlElementWrapper(name = "UserList")
+    @XmlElement(name = "User")
+    private List<String> users;
+
+    public List<String> getUsers() {
+        return users;
+    }
+
+    public void setUsers(List<String> users) {
+        this.users = users;
```
jhulslander, line 28: should this getter instantiate the List if it is null?

src/main/java/edu/cornell/kfs/sys/xmladapters/ZonedStringToJavaDateXmlAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public Date unmarshal(String value) throws Exception {
+        return StringUtils.isNotBlank(value) ? parseToDate(value) : null;
+    }
+
+    @Override
+    public String marshal(Date value) throws Exception {
+        return value != null ? formatDate(value) : null;
```
jhulslander, line 38: should this return an empty string rather than null?

jhulslander (changes requested):
PR looks good, I noted a few places that we may want to re-use Objects added for Jaggaer supplier, and a few places were we may want be consistent with how the supplier objects work.  

Also, I am not sure if it is worth changing now, but I found CuXMLUnitTestUtils to be helpful in debugging XML constructions as I worked through it.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Attachment.java
cah292: I'm wondering whether renaming the two objects would introduce too much clutter. Both Attachment objects are in separate package structures, so that should already help clarify the context of each object. (Also, in the code that actively uses the DTOs, it should theoretically be obvious which context the code is meant to use.)

I haven't found any cases in base code where two XML DTOs have the same name, but I did find some cases (such as with the Branch class) where the BO and an XML DTO have the same classname.

Perhaps we could discuss this further at the next bookclub meeting to get the team's input.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Authentication.java
cah292: I'm not sure whether it's safe to consolidate the two Authentication classes entirely, in case Jaggaer makes future changes to their authentication setup. However, since they currently have the same content, I could add a base class to hold the common portions and then update the supplier/contract variants to extend from the base class. That will make it easier to separate the two if Jaggaer ends up making changes to one API but not the other.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Contract.java
cah292: I will update this to auto-instantiate the list as needed, as per the discussion at one of our prior team meetings.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/ManagerList.java
cah292: I'll update the getter to instantiate the list as needed.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/CustomField.java
cah292: Given the results of the `Attachment` DTO discussion, I don't think this needs renaming at this time, due to the context separation in the package setup.

src/main/java/edu/cornell/kfs/module/purap/jaggaer/contract/xml/Header.java
cah292: I think there are too many differences between the two Header DTOs at this time. We may or may not be fully consolidating the Authentication sub-DTO (see the other comment), and my version of the DTO converts the timestamp string into a Java Date.

We could possibly look into consolidating the two DTOs at a future date.

src/main/java/edu/cornell/kfs/sys/xmladapters/ZonedStringToJavaDateXmlAdapter.java
cah292: If the Date is null, I think it would be safer to return null instead of an empty string, so that the JAXB marshaller knows that the element/attribute has no value and could potentially be omitted. Also, the various other Date-related adapters in cu-kfs and base financials are returning null when the Date to marshal is null; `StringToJavaDateAdapter` is one example.

cah292:
I've committed some initial changes and also rebased the branch, but this is not ready for the next review yet. I still want to look into revising the PR's unit test setup first.

cah292:
I think I've addressed the areas that needed to be modified. If you notice something that still needs changes or if you think more class renaming/consolidation is needed, please let me know.

I went ahead and modified the new marshalling unit test class to use the new XML utility that was introduced in an earlier PR. Note, though, that I made some minor adjustments to it so that the XML could come from sources other than a `File` object.

Also, these changes are now ready to be merged if they get approved.

cah292:
As per the announcement at today's standup, the functionals do not want these changes to move forward to DEV just yet. There are some additional details/requirements that they want to sort out first; there's even a possibility where we don't proceed with these changes at all.

I'll close this PR for now and will reopen it when/if we're ready to move forward again.

PR #1438, opened 2023-04-06
KFSPTS-28093 Revert PositionData.xml to FINP-8084 and Overlay With Customizations

ajd299:


jhulslander:
@ajd299 I am trying to understand this change.  When I compare this code to the KualiCO version of the FINP: https://github.com/KualiCo/financials/blob/08075c4cfbedb17f7b0765453e97be8a2138c902/kfs-ld/src/main/resources/org/kuali/kfs/module/ld/businessobject/datadictionary/PositionData.xml there are a lot of differences.  Should there be?

ajd299:
@jhulslander No there should not be that many differences (ideally only where there is a comment )

However, I am not using the current version of KualiCo PositionData.xml, the initial source file (before customization) is before the code included in the 7/13/2022 patch (updated on 1/3/22).

https://github.com/KualiCo/financials/blob/c927742bf34f6bd0ffcd7d7a8d3a95f856ceba15/kfs-ld/src/main/resources/org/kuali/kfs/module/ld/businessobject/datadictionary/PositionData.xml

jhulslander (approved):
I spoke with Andrew and I agree with which version of Financials code was used, and this  code looks reasonable to me.  Shall merge

PR #1440, opened 2023-04-10
KFSPTS-28105: Fix stack trace when trying to open Role docs created before KEW Upgrade.

nkimble:


nkimble:
A SQL approach to adjust the data that appears to have been accidentally misconfigured during one of the upgrades instead of a customized coded solution that would need to be continually maintained would be a better way to go. I have no objections to that approach if we can answer the following observation I had with the data (and the reason I took the coding approach).

My concern with the SQL approach when I initially saw what you are suggesting while investigating this was the difference  between the values for name space in table KRIM_ROLE_DOCUMENT_T column ROLE_NMSPC and table KIM_TYP_T column NMSPC_CD.

The value for NMSPC_CD in the KIM_TYP_T table is KFS-SYS. 
There are 21 distinct ROLE_NMSPC codes in table KRIM_ROLE_DOCUMENT_T where some are still referencing "KR".

Do we need to be concerned about this name space code value mismatch when 2076 rows in table KRIM_ROLE_DOCUMENT_T will be affected with a SQL update to the Default KIM_TYP_ID "KFSMI9562-TYP"?

cah292:
As far as I can tell, KRIM_ROLE_DOCUMENT_T.ROLE_NMSPC does not need to be updated by the potential SQL fix. The role's namespace does not necessarily have to match the namespace of the KIM role type that it uses, and it seems that the document-side role namespace is primarily used for various rule or permission checks. As long as there are no unusual rendering problems or unexpected stack traces, it should be safe to keep the role document namespaces as-is.

PR #1444, opened 2023-04-14
KFSPTS-28000 Improve maint lock search efficiency

cah292:
This is not ready to be merged yet; I would like to put this branch in the Sandbox environment for further testing before making it available in DEV.

This PR introduces bulk maintenance lock checking for KFS maintenance documents. KFS normally checks each lock one by one, which can result in thousands of DB calls for certain types of documents (especially Account Global and Account Delegate Global). By converting these checks to examine the locks in bulk, that should greatly improve the efficiency of opening/saving/submitting global maintenance documents.

Some maintenance documents introduce additional lock checks beyond the basic ones. The most problematic one is done by Account Delegate Global docs, which convert certain lock strings into wildcarded strings in an additional search operation. This PR reworks that check to do bulk checks of the wildcarded values, which should mitigate the slowness that such checks would normally introduce. (Note that, to simplify the setup of the bulk LIKE criteria, I introduced some new helper classes to allow for creating parameterized plain-SQL OJB sub-queries.)

There are still a few extra lock checks that are not currently addressed by this PR, such as the "trickle down" locks when an Account or Account Global doc inactivates downstream Sub-Accounts/Sub-Objects/Account Reversions. Given the extent of the current changes here, we may want to defer further bulking of lock checks to a future ticket, if the unconverted cases are running into problems.

Due to the limitations of what can be tested (and tested efficiently) in a local environment, I intend to move these changes into the Sandbox environment within the next few days. That should help us get a better idea as to whether the bulk lock checking is enough to address the efficiency or DB-memory concerns associated with the various global maintenance docs.


cah292:
The functionals have finished testing this branch's changes in the KFS Sandbox environment, and they are ready for the changes to move forward to KFS DEV now. This PR can be merged after being approved by a technical review.

src/main/java/edu/cornell/kfs/core/framework/persistence/ojb/CuStatementManager.java
```diff
@@ -0,0 +1,30 @@
...
+    public CuStatementManager(final PersistenceBroker pBroker) {
+        super(pBroker);
+    }
+
+    @Override
+    public int bindStatement(PreparedStatement stmt, Query query, ClassDescriptor cld, int param) throws SQLException {
+        int newIndex = param;
+        if (query instanceof QueryByPreparedSQL) {
+            QueryByPreparedSQL preparedQuery = (QueryByPreparedSQL) query;
+            newIndex = bindValues(stmt, preparedQuery.getParameters(), newIndex);
```
jhulslander, line 25: What is going on with this function, please explain

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/CuAccountDelegateGlobalDaoOjb.java
```diff
@@ -0,0 +1,92 @@
...
+                "SELECT LCK.MAINT_LOCK_ID FROM KFS.KRNS_MAINT_LOCK_T LCK ",
+                "WHERE ", createAndedDocIdCriteria(documentNumber),
+                "EXISTS (SELECT 1 FROM SEARCH_PATTERNS WHERE LCK.MAINT_LOCK_REP_TXT LIKE SEARCH_PATTERNS.LIKE_EXPR)"
+        );
+        return new QueryByPreparedSQL(MaintenanceLock.class, sqlQuery);
+    }
+
+    private CuSqlChunk createWithClauseForSearchPatterns(List<String> lockingRepresentations) {
+        CuSqlChunk caseBlock = new CuSqlChunk();
+        caseBlock.append("CASE LEVEL");
```
jhulslander, line 66: I understand SQL case statements, but I am confused by what is going on with LEVEL here.  Could this benefit for refactor for clarity?

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/CuAccountDelegateGlobalDaoOjb.java
```diff
@@ -0,0 +1,92 @@
...
+     * using a "CASE" block to support a larger number of pattern expressions):
+     * 
+     * https://asktom.oracle.com/pls/apex/asktom.search?tag=like-operator-with-in-clause#1688255100346222560
+     */
+    private QueryByPreparedSQL createPatternMatchSubquery(List<String> lockingRepresentations, String documentNumber) {
+        CuSqlQuery sqlQuery = CuSqlQuery.of(
+                createWithClauseForSearchPatterns(lockingRepresentations),
+                "SELECT LCK.MAINT_LOCK_ID FROM KFS.KRNS_MAINT_LOCK_T LCK ",
+                "WHERE ", createAndedDocIdCriteria(documentNumber),
+                "EXISTS (SELECT 1 FROM SEARCH_PATTERNS WHERE LCK.MAINT_LOCK_REP_TXT LIKE SEARCH_PATTERNS.LIKE_EXPR)"
```
jhulslander, line 59: I presume SEARCH_PATTERNS.LIKE_EXPR will get replaced in the parameterized processing, where does that happen?

jhulslander (commented):
I haven't completed a review yet, but I have a few questions so far

src/main/java/edu/cornell/kfs/core/framework/persistence/ojb/CuStatementManager.java
cah292: This line is calling a superclass function that binds the given parameters to the prepared SQL statement, using appropriate parameter indexing and returning the next index to use for subsequent param binds. Please let me know if you think this needs more comments or refactoring to explain its intent further.

jhulslander: I think some refactoring would be helpful.  I was wondering what this provided base library function didn't

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/CuAccountDelegateGlobalDaoOjb.java
jhulslander: OK, that makes more sense

src/main/java/edu/cornell/kfs/krad/util/MaintenanceLockUtils.java
```diff
@@ -0,0 +1,52 @@
...
+        throw new UnsupportedOperationException("do not call");
+    }
+
+    public static String doChunkedSearchForAnyLockingDocumentNumber(List<String> lockingRepresentations,
+            String documentNumber, BiFunction<List<String>, String, String> actualQueryOperation) {
+        if (CollectionUtils.isEmpty(lockingRepresentations)) {
+            LOG.debug("doChunkedSearchForAnyLockingDocumentNumber, No locking reps specified, skipping search");
+            return null;
+        }
+        LOG.debug("doChunkedSearchForAnyLockingDocumentNumber, Searching with {} lock reps{}",
```
jhulslander, line 28: should this block be in an IF statement LOG.isiDebugEnabled() so the string is only built if the log statement needs to be generated

src/main/java/edu/cornell/kfs/krad/util/MaintenanceLockUtils.java
```diff
@@ -0,0 +1,52 @@
...
+                        : KFSConstants.EMPTY_STRING);
+        
+        String lockingDocumentNumber = null;
+        int startIndex = 0;
+        
+        do {
+            int endIndex = Math.min(startIndex + BULK_LOCK_QUERY_LIMIT, lockingRepresentations.size());
+            List<String> locksSubList = lockingRepresentations.subList(startIndex, endIndex);
+            lockingDocumentNumber = actualQueryOperation.apply(locksSubList, documentNumber);
+            startIndex += BULK_LOCK_QUERY_LIMIT;
```
jhulslander, line 41: based on this code is structured, I don't think it matters, but wouldn't it mkae more sense to add endIndex to startIndex?  The next iteration starts with where the current iteration started plus where it ended.

src/main/java/edu/cornell/kfs/core/framework/persistence/ojb/CuStatementManager.java
```diff
@@ -0,0 +1,30 @@
+package edu.cornell.kfs.core.framework.persistence.ojb;
```
jhulslander, line 1: Is there a standard convention for extending OJB libraries like this?  I wonder if the class should be more similar to what is being extended.  Something like edu.cornell.kfs.core.ojb.broker.accesslayer

src/main/java/edu/cornell/kfs/krad/service/impl/CuMaintenanceDocumentServiceImpl.java
```diff
@@ -0,0 +1,27 @@
...
+    public String getLockingDocumentId(Maintainable maintainable, String documentNumber) {
+        List<MaintenanceLock> maintenanceLocks = maintainable.generateMaintenanceLocks();
+        List<String> lockingRepresentations = maintenanceLocks.stream()
+                .map(MaintenanceLock::getLockingRepresentation)
+                .collect(Collectors.toUnmodifiableList());
+        return getCuMaintenanceDocumentDao().getAnyLockingDocumentNumber(lockingRepresentations, documentNumber);
+    }
+
+    private CuMaintenanceDocumentDao getCuMaintenanceDocumentDao() {
+        return (CuMaintenanceDocumentDao) getMaintenanceDocumentDao();
```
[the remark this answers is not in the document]
cah292: Good point, I'll adjust the type checking on this.

src/main/java/edu/cornell/kfs/krad/util/MaintenanceLockUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.kuali.kfs.sys.KFSConstants;
+
+public final class MaintenanceLockUtils {
+
+    private static final Logger LOG = LogManager.getLogger();
+
+    private static final int BULK_LOCK_QUERY_LIMIT = 500;
```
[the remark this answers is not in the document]
cah292: I'll update this to use a derive value instead. I'll see if I can retrieve the value that OJB uses when it splits up large "IN" lists.

src/main/java/edu/cornell/kfs/krad/util/MaintenanceLockUtils.java
cah292: It shouldn't make much of a difference, but I can adjust this to set the start index to the end index.

src/main/java/edu/cornell/kfs/core/framework/persistence/ojb/CuStatementManager.java
cah292: KualiCo placed many of their custom OJB-related classes under the `org.kuali.kfs.core.framework.persistence.ojb` package, so I was trying to use a CU-specific variant of that.

jhulslander: Thanks, if KualiCo has a standard approach, I am fine with complying with that

src/main/java/edu/cornell/kfs/krad/util/MaintenanceLockUtils.java
jhulslander: interesting convention, but we should follow suit as you have done

cah292:
I've committed an initial set of changes based on the reviews thus far, but I still need to perform further validation of the revisions. Also, depending on the results of tomorrow's team discussion, I may need to make additional changes.

If we end up deciding that the proposed OJB customizations are too risky or are not worth the trouble, we may have to go with a different multi-LIKE strategy, such as one of the following:

1. Use OJB's criteria API as-is to add lots of ORed-together LIKE conditions. This should theoretically work, but OJB's Criteria API is notoriously cumbersome with regards to adding OR criteria, so I'm not certain whether this will result in too many criteria objects or in having too many sets of nested criteria parentheses (due to how OJB handles criteria concatenation).
2. Augment OJB's criteria API to provide a cleaner way of specifying bulk LIKE conditions, even if the resulting SQL simply has the format `field LIKE exp1 OR field LIKE exp2 OR ...`. This would result in cleaner SQL and would stick to using standard LIKE condition querying, but it would still require customizing OJB's setup (though in a different way than the current plain-SQL-subquery approach).

cah292:
As per the discussion with the technical team, we decided that my subquery-based bulk-LIKE-check approach was too complex, and that we should instead use a plain SQL approach that combines the exact-match and like-match checks into a single query. I will modify the code accordingly.

dp65:
Does OJB support regular expression based LIKE ? The underlying databases (Oracle, MySQL, etc) all do. Would it be more efficient and easier to read if the OR clauses are moved into an RE?  For example: `field LIKE exp1 OR field LIKE exp2 OR ...` becomes: `field REGEX_LIKE '(exp1|exp2|exp3|...)'`

cah292:
@dp65 Yes, Oracle has a `REGEXP_LIKE` function for regex-based matching. However, based on various forum comments I was reading online, regex-based searching seems to be slower than LIKE-condition searching.

cah292:
I've committed some changes to bring the fix in line with the results of yesterday's team discussion. The customized OJB sections have been removed and the Account Delegate Global's bulk lock checks have been consolidated into a plain SQL/JDBC query. Furthermore, I've removed the new `MaintenanceLockUtils` class that was performing custom split-ups of the lock expression list; any needed split-ups are now handled either explicitly in the new SQL-based area or implicitly by OJB's auto-split-up feature.

It looks like some of our plain-SQL-building code in the PMW and TAX modules is still using hard-coded constants for the max-IN-list sizes, so for now the new code uses a similar constant when building the plain-SQL query. I would like to defer consolidation and/or derivation of the max-size setting to a separate user story, unless you think it should be handled now.

I also made some adjustments to the custom `CuSqlChunk` class that was introduced a couple years ago. The most important changes revolved around adding convenience methods to make it easier to append query parameters. In addition, though, I went ahead and tweaked some of the class's state-based handling, to avoid constructing unneeded `StringBuilder` instances after the object gets processed. If you think that the latter cleanup should be moved into its own user story, please let me know.

This is ready for review again, but please don't merge it just yet. Due to the extent of the changes, we may want to have the functionals test this again in Sandbox before moving it to DEV. I'll ask for their input at the next team standup.

cah292:
The functionals have successfully validated the feature's latest revisions in the KFS Sandbox environment. This can be merged once it has been approved.

My previous comment has further high-level details on what revisions were made. Please let me know if you notice areas that could use further changes or if you believe some of the additional cleanup should be deferred to a separate ticket.

src/main/java/edu/cornell/kfs/sys/util/CuSqlChunk.java
```diff
@@ -15,19 +15,30 @@ public final class CuSqlChunk implements CharSequence {
 
     private StringBuilder queryString;
     private Stream.Builder<SqlParameterValue> parameters;
+    private CharSequence underlyingCharSequence;
```
jhulslander, line 18: Why is this needed?  I see this is set to queryString in the constructor, if something happens where queryString is set to a new object, the two might get out of synch

jhulslander (changes requested):
Changes look good, I had a few questions, and some things to consider

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
```diff
@@ -0,0 +1,95 @@
...
+                ") AND ROWNUM <= 1"
+        );
+    }
+
+    private CuSqlChunk createAndedCriteriaForDocumentIdExclusion(String documentNumber) {
+        if (StringUtils.isNotBlank(documentNumber)) {
+            return CuSqlChunk.of(
+                    "DOC_HDR_ID <> ", CuSqlChunk.forParameter(documentNumber), " AND ");
+        } else {
+            return new CuSqlChunk();
```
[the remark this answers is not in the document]
cah292: I structured the SQL generation so that if the doc-ID-limit condition ended up not being needed, the query would still have the correct format. A non-blank doc ID should result in a query like this:
```sql
SELECT DOC_HDR_ID FROM KFS.KRNS_MAINT_LOCK_T
WHERE DOC_HDR_ID <> ? AND (lock_exact_match_criteria OR lock_pattern_criteria)
AND ROWNUM <= 1;
```
A blank doc ID should result in a query like this:
```sql
SELECT DOC_HDR_ID FROM KFS.KRNS_MAINT_LOCK_T
WHERE (lock_exact_match_criteria OR lock_pattern_criteria)
AND ROWNUM <= 1;
```

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    private CuSqlChunk createOredCriteriaForLockMatching(List<String> lockingRepresentations) {
```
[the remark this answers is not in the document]
cah292: I'll rename the method for clarity, in addition to the related "anded" method that controls the doc ID criteria.

I'll also look into whether it's feasible to unit test the broken-up chunks.

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/CuAccountDelegateGlobalDaoOjb.java
```diff
@@ -0,0 +1,20 @@
...
+
+import edu.cornell.kfs.coa.dataaccess.CuAccountDelegateGlobalDao;
+
+public class CuAccountDelegateGlobalDaoOjb extends AccountDelegateGlobalDaoOjb implements CuAccountDelegateGlobalDao {
+
+    @Override
+    public List<String> createSearchPatternsFromLockingRepresentations(List<String> lockingRepresentations) {
+        return lockingRepresentations.stream()
+                .map(this::convertForLikeCriteria)
+                .distinct()
```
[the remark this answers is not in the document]
cah292: The duplication is just a result of how KualiCo creates the wildcarded versions of the locking representations; the non-wildcarded ones should already be unique. As per `AccountDelegateGlobalDaoOjb.convertForLikeCriteria()`, if the lock rep contains a doc type field, the whole part of the string after the field name will be replaced by a wildcard. This can result in a significant number of duplicate patterns, based on the lock-generating code in `AccountDelegateGlobalMaintainableImpl`.

Removing the duplicate patterns should greatly improve the efficiency of the constructed SQL query and also simplify its ORed section of the criteria.

src/main/java/edu/cornell/kfs/sys/util/CuSqlChunk.java
cah292: This is part of the cleanup I had hinted at in one of my earlier comments. I need the `CuSqlChunk` class to implement `CharSequence` because of how this SQL-building API has been configured, but I also want to clear out the underlying string content when the object is no longer needed. This PR will use this new field to handle the implemented `CharSequence` methods; it's initially set to the underlying `StringBuilder` but it will be replaced with an empty string after this object has been used up. (Prior to the change, this class would replace the `StringBuilder` with a fresh empty one after the object gets used up. I was trying to structure this to avoid the creation of new unnecessary `StringBuilder` objects.)

I could replace this with some other setup, such as having some of the relevant `CharSequence` methods throw exceptions if the object has already been used up. Thoughts?

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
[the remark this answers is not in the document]
jhulslander: Sorry, I was trying to ask under what application conditions would we need to check for locks without the document ID

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
```diff
@@ -0,0 +1,95 @@
...
+        List<String> lockRepPatterns = accountDelegateGlobalDao.createSearchPatternsFromLockingRepresentations(
+                lockingRepresentations);
+        CuSqlChunk lockMatchCriteria = new CuSqlChunk();
+        LOG.debug("createOredCriteriaForLockMatching, Performing search with {} lock representations and {} patterns",
```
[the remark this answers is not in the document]
jhulslander: oh, I see now

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/CuAccountDelegateGlobalDaoOjb.java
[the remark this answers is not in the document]
jhulslander: Thanks, that makes sense

src/main/java/edu/cornell/kfs/sys/util/CuSqlChunk.java
jhulslander: I am not sure what the best action is right now.  Having both queryString and underlyingCharSequence seems to add some confusion that I would like to avoid if possible.

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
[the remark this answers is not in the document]
cah292: If you're asking about the conditions where the document ID parameter might be blank, it's possible that we might not actually encounter such a situation, after further review of the calling code. However, KualiCo's lock-related DAOs contain logic to handle cases where a blank document ID was provided, so I had added that to our custom code for consistency.

If you're asking about **WHY** we need to add the not-equal-to-doc-ID check to the query in the first place, that's part of how most (if not all) of the lock checks work. When a particular document needs to check if a **DIFFERENT** document is holding one of the needed maintenance locks, it will pass its own document ID to the appropriate service/DAO method(s) so that the query ignores any locks that that document may already be holding.

src/main/java/edu/cornell/kfs/sys/util/CuSqlChunk.java
cah292: Ok, I'll refactor this to remove `underlyingCharSequence`.

src/main/java/edu/cornell/kfs/coa/dataaccess/impl/AccountDelegateGlobalLockDaoJdbc.java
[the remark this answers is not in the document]
jhulslander: sorry for being confusing, but you answered my question in the first paragraph, we don't expect to encounter the situation but base code handles it, so we should to in this class.  that works for me, thanks.

cah292:
I've committed some more changes to address the remaining concerns that were brought up. I also reworked the code that builds the bulk IN conditions to make it available on the `CuSqlChunk` class itself, so that future SQL-building user stories can take advantage of it. Appropriate unit tests for this new functionality have been added as well.

jhulslander (approved):

PR #1446, opened 2023-04-17
KFSPTS-28238 update spring core to 5.3.27

jhulslander:


cah292 (approved):

One quick note, though: Should this change be made in standalone Cynergy as well? (The Cynapps applications are too old to be upgraded, but standalone Cynergy is on the 5.x Spring line and should be eligible for the upgrade. Sorry for not noticing this during our other recent Spring upgrade.)

I'll go ahead and merge this PR since the KFS-side changes can proceed while we sort out Cynergy.

jhulslander:
good note about Cynergy, I will work on a PR for that

PR #1450, opened 2023-04-25
KFSPTS-27949 Attachment Size Limit Increase

ajd299:


src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
```diff
@@ -130,8 +130,8 @@
 public class PurchasingActionBase extends PurchasingAccountsPayableActionBase {
 
     private static final Logger LOG = LogManager.getLogger();
-    
-    private static final int SIZE_5MB =5242880;
+
+    private static final int SIZE_15MB = 15728640;
```
cah292, line 134: I see that this max-size check is a Cornell customization. I strongly recommend that you make this limit configurable via a parameter, as opposed to leaving it hard-coded like before. That will make it much easier to reconfigure the limit if we decide to change it in the future.

I would suggest creating a new parameter that is similar to the `MAX_FILE_SIZE_ATTACHMENT` parameter from base code.

cah292 (changes requested):
I had a suggestion on how to make the size limit setting more flexible.

cah292 (changes requested):
There are a few spots in the revised code that could be cleaned up, but overall the changes look good.

ajd299:
Please review the new changes in this PR and the related SQL PR

src/main/java/edu/cornell/kfs/module/purap/CUPurapConstants.java
```diff
@@ -487,4 +487,6 @@ public String getHeaderName() {
         }
         
     }
+
+    public static final String PURCHASE_ORDER_COMPONENT_CODE = "Purchase Order";
```
cah292, line 491: Please remove this constant if you don't need it.

src/main/java/edu/cornell/kfs/module/purap/CUPurapKeyConstants.java
```diff
@@ -23,15 +23,16 @@ public class CUPurapKeyConstants {
     public static final String ERROR_EXCEED_SQ_NUMBER_OF_ATT_LIMIT = "errors.exceed.sq.number.of.att.limit";
     public static final String ERROR_REASON_IS_REQUIRED = "errors.reason.is.required";
     public static final String ERROR_ATT_FILE_SIZE_OVER_LIMIT = "errors.att.filesize.over.limit";
-   
+    public static final String ERROR_ATT_SEND_TO_VENDOR_FILE_SIZE_OVER_LIMIT = "errors.att.send.to.vendor.filesize.over.limit";
```
cah292, line 26: I see you created a new property key to replace the old `errors.att.filesize.over.limit` property. I think the old property key was only used for this custom filesize check, so if you're not using it then you may want to remove it (or create a follow-up ticket to do so).

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -248,7 +248,7 @@ error.po.not.fully.received.and.invoiced=PO indicates receiving is required and
 
 	errors.add.new.note.sendToVendor.noAtt=Send to Vendor note must have an Attachment.
 	errors.exceed.sq.number.of.att.limit=Failed to {0} because the number of 'Send to Vendor' Attachments exceeds Sciquest limit of {1}. 
-	errors.att.filesize.over.limit=Attachment "{0}" file size over limit of {1} MB.
```
cah292, line 251: As I noted in my previous review, this old property key was defined in two different spots within this file. If you end up removing the property key entirely, make sure you remove its other reference from this file.

src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
```diff
@@ -1659,11 +1661,13 @@ public ActionForward insertBONote(ActionMapping mapping, ActionForm form, HttpSe
...
 				return mapping.findForward(KFSConstants.MAPPING_BASIC);
 			} else {
+                String fileSizeLimitParameterValue = getMaxPoSendToVendorFileSizeParameterValue();
 				if (isAttachmentSizeExceedSqLimit(form, "add")) {
 					return mapping.findForward(KFSConstants.MAPPING_BASIC);
-				} else if (attachmentFile.getFileSize() > SIZE_5MB) {
-					GlobalVariables.getMessageMap().putError(String.format("%s.%s",KRADConstants.NEW_DOCUMENT_NOTE_PROPERTY_NAME,
-							KRADConstants.NOTE_TOPIC_TEXT_PROPERTY_NAME), CUPurapKeyConstants.ERROR_ATT_FILE_SIZE_OVER_LIMIT, attachmentFile.getFileName(), "5");
+				} else if (attachmentFile.getFileSize() > FileUtil.getBytes(fileSizeLimitParameterValue)) {
+                    GlobalVariables.getMessageMap().putError(String.format("%s.%s",KRADConstants.NEW_DOCUMENT_NOTE_PROPERTY_NAME,
```
cah292, line 1668: Some of the edited lines' whitespace formatting seems a bit off.

cah292 (changes requested):
There are a few spots that may need a little more cleanup, but otherwise the revisions look good.

src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
ajd299: After some recent commits, looking at the file on the link below the commits made seem to have improved the formatting related . When viewing the method in the IDEs, the formatting looks correct (does it on your end?). However in the link below (viewing in GitHub) the code within the method (and elsewhere in the file) seems to have undesirable formatting. 

- Should we try to fix the formatting in the rest of the method or create another user story for the whole method/file?

https://github.com/CU-CommunityApps/cu-kfs/blob/027b8c49dc72670f1ae8fbd098be6c71668c0753/src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java

cah292: The formatting differences seem to be caused by the file using tabs in most places instead of spaces, where the file was likely created before our team standardized on the use of spaces. (Our IDEs might only be indenting tabs by 4 units even though the GitHub UI indents them by more than that, which is why the formatting may look correct in the IDE but not in GitHub.)

If it ends up being too complex to clean up this section of whitespace at this time, then I'm fine with deferring the whitespace cleanup to a separate user story.

ajd299: I think that it would be better to do it in a separate user story so it is slightly easier to discern what actually changed. I tried to fix the method but GitHub didn't show that was only indentation.

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -248,7 +247,7 @@ error.po.not.fully.received.and.invoiced=PO indicates receiving is required and
...
+	errors.att.send.to.vendor.filesize.over.limit=Attachment "{0}" file size over limit of {1}.
```
cah292, line 250: Just to confirm, is this property being read properly by KFS despite its leading whitespace? If it's still working as expected, then I'm okay with cleaning up that whitespace in a future ticket, since I'm planning to create one anyway for dealing with various duplicate properties in this file.

cah292 (commented):
Changes look good, I just had one quick question about the property file entry before merging.

src/main/resources/CU-ApplicationResources.properties
ajd299: I verified locally that the error message is displaying as desired before and after the most recent commit that was made.

image

cah292: Thanks for confirming. Your changes can proceed as-is and I'll create a follow-up user story to clean up the properties file further.

src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
```diff
@@ -1658,13 +1660,15 @@ public ActionForward insertBONote(ActionMapping mapping, ActionForm form, HttpSe
...
-				} else if (attachmentFile.getFileSize() > SIZE_5MB) {
-					GlobalVariables.getMessageMap().putError(String.format("%s.%s",KRADConstants.NEW_DOCUMENT_NOTE_PROPERTY_NAME,
-							KRADConstants.NOTE_TOPIC_TEXT_PROPERTY_NAME), CUPurapKeyConstants.ERROR_ATT_FILE_SIZE_OVER_LIMIT, attachmentFile.getFileName(), "5");
-					return mapping.findForward(KFSConstants.MAPPING_BASIC);					
+            } else {
+                String fileSizeLimitParameterValue = getMaxPoSendToVendorFileSizeParameterValue();
+                if (isAttachmentSizeExceedSqLimit(form, "add")) {
+                    return mapping.findForward(KFSConstants.MAPPING_BASIC);
+                } else if (attachmentFile.getFileSize() > FileUtil.getBytes(fileSizeLimitParameterValue)) {
+                    String errorMessage = String.format("%s.%s",KRADConstants.NEW_DOCUMENT_NOTE_PROPERTY_NAME, KRADConstants.NOTE_TOPIC_TEXT_PROPERTY_NAME);
```
cah292, line 1668: This is actually the property that the error message will be tied to; it's not the error message or message key itself. Please rename the variable to `propertyName` or something similar to make that clearer.

cah292 (changes requested):
I see you made one additional edit that wasn't related to my previous pending question, and I noticed something with it that should be updated for clarity.

src/main/java/org/kuali/kfs/module/purap/document/web/struts/PurchasingActionBase.java
ajd299: Okay I have fixed this, good catch thank you

cah292 (approved):
Changes look good, unit tests pass, my concerns have been addressed and the related SQL changes are also ready. Merging.

PR #1452, opened 2023-05-04
KFSPTS-27121: Cornell fix for bug introduced by FINP-8149.

nkimble:


nkimble:
Role qualifiers have their attributes defined to be of type String. The transactional system is presuming the fromAmount and toAmount role qualifier values entered by the users for money amount comparisons have been validated by the system and only contain numerals or a period that can be converted to a KualiDecimal value during downstream processing. This check will present an error on the Person document when the user attempts to "Add" an invalid from or to amount role qualifier.
Lines 478 to 495 are the only ones that differ from base code. 
The approach of overlaying the whole file an not just the method needing changes was taken to make it easier to remove our mod when KualiCo provides a fix for this bug.

nkimble:
retest please

src/main/java/org/kuali/kfs/kns/kim/type/DataDictionaryTypeServiceBase.java
```diff
@@ -0,0 +1,874 @@
...
+                } catch (NumberFormatException e) {
+                    // quash; this indicates that the DD contained a max for a non-numeric attribute
+                }
+            }
+            //Cornell Modification: KFSPTS-27121 Start
+            //Role qualifiers have their attributes defined to be of type String. The transactional system is presuming
+            //the fromAmount and toAmount role qualifier values entered by the users for money amount comparisons have
+            //been validated by the system and only contain numerals or a period that can be converted to a KualiDecimal
+            //value during downstream processing. This check will present an error on the Person document when the user
+            //attempts to "Add" an invalid from or to amount role qualifer.
```
cah292, line 483: Minor typo in comment: "qualifer" -> "qualifier"

src/main/java/org/kuali/kfs/kns/kim/type/DataDictionaryTypeServiceBase.java
```diff
@@ -0,0 +1,874 @@
...
+            if (errors.isEmpty() && StringUtils.isNotBlank(attributeName) && StringUtils.isNotBlank(attributeValue)
+                    && (StringUtils.equals(attributeName, KimAttributes.FROM_AMOUNT) 
+                            || StringUtils.equals(attributeName, KimAttributes.TO_AMOUNT))) {
```
cah292, line 486: If you want to simplify the ORed string comparison section, you could use the `StringUtils.equalsAny()` method.

src/main/java/org/kuali/kfs/kns/kim/type/DataDictionaryTypeServiceBase.java
```diff
@@ -0,0 +1,874 @@
...
+                try {
+                    KualiDecimal moneyAmount = new KualiDecimal(attributeValue);
+                }
+                catch (NumberFormatException nfe) {
```
cah292, line 490: The `catch` statement should be on the same line as the closing curly brace.

src/main/java/org/kuali/kfs/kns/kim/type/DataDictionaryTypeServiceBase.java
```diff
@@ -0,0 +1,874 @@
...
+                    errors.add(AttributeError.Builder.create(errorKey, DataDictionaryTypeServiceHelper
+                            .createErrorString(KFSKeyConstants.ERROR_CURRENCY, errorLabel, attributeValue.toString())).build());
```
cah292, line 492: Can you please double-check to make sure that the correct number of var-arg parameters are being passed to this `createErrorString()` call? Based on the current format string for the "error.currency" key, that particular message only seems to support one parameter, not two.

nkimble: Good catch. Fixing.

src/main/java/org/kuali/kfs/kns/kim/type/DataDictionaryTypeServiceBase.java
```diff
@@ -0,0 +1,874 @@
...
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * A base class for {@code KimTypeService} implementations which read attribute-related information from the Data
+ * Dictionary. This implementation is currently written against the KNS apis for Data Dictionary. Additionally, it
+ * supports the ability to read non-Data Dictionary attribute information from the {@link KimTypeInfoService}.
+ */
+public class DataDictionaryTypeServiceBase implements KimTypeService {
```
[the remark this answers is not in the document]
cah292: What I meant was that you could still place your new code in an override of the `validateAttributeFormat()` method, but instead of placing it in the base class, you could place it in the class of the specific role type that uses it (`AccountingOrganizationHierarchyReviewRoleTypeServiceImpl` in this case). That should still allow it to catch the bad KIM data before it gets stored in the database.

On the other hand, given that this is only a minor code change that executes quickly and shouldn't interfere with other role types, I'm okay with leaving the customization in the base class, so that it will automatically handle any future role types that include these from/to amounts. If we end up needing to make this more localized in the future, we can handle that in a separate ticket.

cah292 (approved):
Changes look good, unit tests pass and my concerns have been addressed. (The related SQL PR was already merged earlier, since we determined it could move forward without waiting on the code.) Merging.

PR #1461, opened 2023-06-07
KFSPTS-28620 Update ISO 20022 remittance handling

cah292:
This PR updates how remittance data is handled by the ISO 20022 extraction process, to meet the needs of both Cornell and the bank. It's not possible or practical to introduce the changes through the use of a subclass, so I had to overlay the `Iso20022FormatExtractor` class to implement the new behavior.

Now that the extractor class is being overlayed, I also made a minor adjustment to it that allowed for the removal of a related CU-specific helper class. This will help simplify our ISO 20022 setup.


src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -0,0 +1,1463 @@
...
+            return true;
+        } else if (extractTypeContext.isExtractionType(ExtractionType.ACH)) {
+            PayeeACHAccount payeeAchAccount = retrievePayeeACHAccount(templatePaymentGroup);
+            if (ObjectUtils.isNull(payeeAchAccount)) {
+                throw new IllegalStateException("Could not find Payee ACH Account for Payment Group '"
+                        + templatePaymentGroup.getId()  + "'; this should NEVER happen!");
+            }
+            return StringUtils.equals(payeeAchAccount.getStandardEntryClass(), StandardEntryClass.CTX.name());
+        } else {
+            throw new IllegalStateException("Extraction was for neither Check nor ACH; this should NEVER happen!");
```
jhulslander, line 887: I would suggest adding the extraction type value to this error to help debug the problem if it ever happens

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -0,0 +1,1463 @@
...
+            remittanceInformation.addStrd(structuredRemittanceInformation);
+        });
+
+        return remittanceInformation;
+    }
+
+    /*
+     * CU Customization: Marked this method as "unused" since the customized code no longer calls it.
+     */
+    @SuppressWarnings("unused")
```
jhulslander, line 1126: I am not sure the best approach but I am wondering if it would be better to comment out or remove the function entirely?  I am wondering what is the best way to help someone doing a kualico patch update to notice we don't use this function

jhulslander (dismissed):
I had a question and a suggestion

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: Ok, I'll adjust the error message.

cah292:
I've committed a few changes in response to the initial feedback:

* Updated the one exception message to include additional details about the unexpected situation.
* Modified the code to comment out the appropriate unused areas, in response to today's team meeting. I know we still need to review with Daniela on whether we truly want to take this approach, but at least now you can see what I had in mind for structuring such comments.

cah292:
@db483,

When you have a moment, we would like to get your input on a subject that came up in Friday's team meeting (and which impacts how this PR's code changes should be set up). This PR is overlaying a base KFS class in such a way that some of the class's private methods are no longer being used. We were debating on how to best handle this situation in the overlay, where we seemed to have 3 competing strategies for doing so (though all of these would still involve adding an extra explanatory comment to the code):

1. Add annotations to the unused methods to suppress compiler warnings.
2. Comment out the unused methods.
3. Remove the unused methods altogether.

Please let us know which of these strategies you would prefer. Thanks!

db483:
@cah292 Thank you for bringing this to my attention.
Of all options I think commenting out the code is the least effective solution when it comes to upgrades. It is easier to understand how the file was customized but, if the code is commented out then it is likely it will not be updated during an upgrade and I think it should be if we keep it in this file. Given this, the better options would be to either:

-  add annotations to the unused methods to suppress compiler warnings + add comments explaining that the methods are not used. During an upgrade this code will require updating even though the code in not used by Cornell. 
- or
- remove the methods altogether; from an upgrade perspective this will require the least amount of time because not time is needed to update the methods but some time might be required to understand why the methods are not present in our local copy of the file.

I would favor removing the methods altogether.

cah292:
@db483 Thanks for the input. I'll update this PR to remove the unused code.

db483 (approved):
Changes look good.

PR #1464, opened 2023-06-21
KFSPTS-26962: ISO20022 Update PDP to populate PaymentGroups with Country Code instead of Country Name.

nkimble:


cah292 (approved):
Changes look good and unit tests pass. Further changes to the country code/name handling will be addressed in follow-up tickets accordingly.

Merging.

PR #1465, opened 2023-06-22
KFSPTS-28795 fix file validation to use the same settings for csv rea…

db483:
…der as the parsing function


jhulslander (approved):
changes look good, with this change the validation has the same code as the parse

PR #1466, opened 2023-06-28
KFSPTS-26963 Improve country handling for ISO 20022

cah292:
This PR updates the ISO 20022 extraction so that it will properly convert country names and FIPS country codes into ISO country codes. As per the user story's requirements and comments, if a unique mapping doesn't exist or if the code/name is blank, then country code "ZZ" (Unknown) will be used by default.

I also added some helper methods that will be needed when we eventually perform the full FIPS-to-ISO transition at some future date. Some of those methods are not fully implemented yet since their functionality is not needed at this time.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
+++ b/src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
@@ -80,9 +80,16 @@ public class CUKFSPropertyConstants {
     public static final String FINALIZED_DATE = "finalizedDate";
 
     //** CU Generic ISO-FIPS Country modification items **
-    public static class ISOFIPSCountryMap {
+    public static class Location {
+        public static final String COUNTRY_CODE = "countryCode";
         public static final String ISO_COUNTRY_CODE = "isoCountryCode";
         public static final String FIPS_COUNTRY_CODE = "fipsCountryCode";
+        public static final String COUNTRY_NAME = "countryName";
+        public static final String ISO_COUNTRY_NAME = "isoCountryName";
+    }
+    public static class ISOFIPSCountryMap {
+        public static final String ISO_COUNTRY_CODE = Location.ISO_COUNTRY_CODE;
+        public static final String FIPS_COUNTRY_CODE = Location.FIPS_COUNTRY_CODE;
         public static final String ACTIVE = KFSPropertyConstants.ACTIVE;
     }
     private static final String NAME = "name";
--- a/src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
@@ -2,13 +2,21 @@ package edu.cornell.kfs.sys.service.impl;
 
 import java.text.MessageFormat;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.log4j.LogManager;
-import org.apache.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.kuali.kfs.core.api.config.property.ConfigurationService;
+import org.kuali.kfs.core.api.criteria.CriteriaLookupService;
+import org.kuali.kfs.core.api.criteria.GenericQueryResults;
+import org.kuali.kfs.core.api.criteria.PredicateFactory;
+import org.kuali.kfs.core.api.criteria.QueryByCriteria;
 import org.kuali.kfs.krad.service.BusinessObjectService;
+import org.kuali.kfs.krad.util.KRADConstants;
+import org.kuali.kfs.krad.util.KRADPropertyConstants;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.sys.KFSConstants;
 import org.kuali.kfs.sys.businessobject.Country;
@@ -23,10 +31,11 @@ import edu.cornell.kfs.sys.service.CountryService;
  */
 public class CountryServiceImpl implements CountryService {
 
-    private static final Logger LOG = LogManager.getLogger(CountryServiceImpl.class);
+    private static final Logger LOG = LogManager.getLogger();
  
     protected BusinessObjectService businessObjectService;
     protected ConfigurationService configurationService;
+    protected CriteriaLookupService criteriaLookupService;
     
     @Override
     public boolean isCountryActive(String countryCode) {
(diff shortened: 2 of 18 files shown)


src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
```diff
@@ -90,6 +99,21 @@ public String findCountryNameByCountryCode(String countryCode) {
         }
     }
     
+    @Override
+    public List<String> findCountryCodesByCountryName(String countryName) {
+        if (isBlank(countryName)) {
+            return List.of();
+        }
+        QueryByCriteria criteria = QueryByCriteria.Builder.fromPredicates(
+                PredicateFactory.equalIgnoreCase(CUKFSPropertyConstants.Country.NAME, countryName),
```
nkimble, line 108: Table SH_CNTRY_T contains active rows where country names are in mixed case as well as all uppercase.
Do we need to make this query case insensitive?
Should leading and trailing white space also be addressed?

src/main/java/edu/cornell/kfs/sys/service/impl/ISOCountryServiceImpl.java
```diff
@@ -88,6 +97,21 @@ public String findISOCountryNameByCountryCode(String isoCountryCode) {
         }
     }
 
+    @Override
+    public List<String> findISOCountryCodesByCountryName(String isoCountryName) {
+        if (isBlank(isoCountryName)) {
+            return List.of();
+        }
+        QueryByCriteria criteria = QueryByCriteria.Builder.fromPredicates(
+                PredicateFactory.equalIgnoreCase(CUKFSPropertyConstants.Country.NAME, isoCountryName),
```
nkimble, line 106: Table CU_ISO_SH_CNTRY_T contains active rows where country names are in mixed case.
Do we need to make this query case insensitive?
Should leading and trailing white space also be addressed?

nkimble (changes requested):
Code changes look good. There were just a couple of areas where we might want to consider adding processing for white space and case being insensitive when performing the string comparisons based upon the current data content in the country tables.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -770,6 +814,93 @@ private static PostalAddress6 constructPostalAddress(
...
+    }
+
+    private String convertISOCountryValueToISOCountryCode(String countryValue) {
+        throw new IllegalStateException("The ISO-value-to-ISO-code conversion should not be reachable at this time; "
+                + "we will fully implement this feature as part of the global FIPS-to-ISO conversion in KFS.");
+    }
+
+    private String convertFIPSCountryValueToISOCountryCode(String countryValue) {
+        String trimmedValue = StringUtils.trim(countryValue);
+        if (StringUtils.equals(trimmedValue, KFSConstants.COUNTRY_CODE_UNITED_STATES)) {
```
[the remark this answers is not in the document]
cah292: I'll update this to make it case insensitive. I also noticed some areas in `CountryService` and `ISOCountryServiceImpl` that are inconsistent about performing case-insensitive country code searches, so I'll update those areas as well.

src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
cah292: This `equalIgnoreCase` predicate already makes the search case-insensitive. That's why I had set up the name-based searches to use KualiCo's criteria API and the `CriteriaLookupService`.

As for the leading and trailing whitespace, the calling code already deals with that, but I'll go ahead and also add that here, in case we have future areas of the code that need to use this method.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOCountryServiceImpl.java
cah292: See my notes on the equivalent area in the `CountryServiceImpl` class. The name-based search is already case-insensitive, but I will add some explicit trimming of the name to deal with potential whitespace.

cah292:
I've made some changes to address the issues that were brought up. This is ready for review again.

cah292:
NOTE: The unit tests for this PR still need to be executed again. (They didn't auto-execute after the latest changes because of Jenkins issues.) Please run the unit tests again once Jenkins is stable.

nkimble:
retest this please

nkimble (commented):
Sorry I did not see this earlier. There are a few places where property values are hard coded strings. Should property constants be added for those strings?

src/main/java/edu/cornell/kfs/sys/service/impl/CountryServiceImpl.java
```diff
@@ -62,40 +72,63 @@ public boolean isCountryInactive(String countryCode) {
         }
     }
     
-    private boolean isBlank(String countryCode) {
-        if (StringUtils.isBlank(countryCode)) {
-            LOG.debug("isBlank: " + MessageFormat.format(getConfigurationService().getPropertyValueAsString(CUKFSKeyConstants.NULL_OR_BLANK_CODE_PARAMETER), "countryCode"));
+    private boolean isBlankCountryCode(String countryCode) {
+        return isBlank(countryCode, "countryCode");
```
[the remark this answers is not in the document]
cah292: Adding constants.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOCountryServiceImpl.java
```diff
@@ -60,40 +70,63 @@ public boolean isISOCountryInactive(String isoCountryCode) {
...
         if (ObjectUtils.isNotNull(isoCountryFound)) {
             return isoCountryFound.getName();
         } else {
             return MessageFormat.format(getConfigurationService().getPropertyValueAsString(CUKFSKeyConstants.NAME_NOT_FOUND_FOR_COUNTRY_CODE), CUKFSConstants.ISO, isoCountryCode);
         }
     }
 
+    @Override
+    public List<String> findISOCountryCodesByCountryName(String isoCountryName) {
+        if (isBlank(isoCountryName, "isoCountryName")) {
```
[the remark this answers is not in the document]
cah292: Adding new constant(s) accordingly.

src/main/java/edu/cornell/kfs/sys/service/impl/ISOFIPSCountryMapServiceImpl.java
```diff
@@ -34,14 +35,16 @@ public List<ISOFIPSCountryMap> findActiveMapsByISOCountryId(String isoCountryCod
         if (isBlank(isoCountryCode, "isoCountryCode")) {
             return new ArrayList<ISOFIPSCountryMap>(); 
         }
-        return performMappingConversion(CUKFSPropertyConstants.ISOFIPSCountryMap.ISO_COUNTRY_CODE, isoCountryCode);
+        String uppercaseCode = isoCountryCode.toUpperCase(Locale.US);
+        return performMappingConversion(CUKFSPropertyConstants.ISOFIPSCountryMap.ISO_COUNTRY_CODE, uppercaseCode);
     }
     
     public List<ISOFIPSCountryMap> findActiveMapsByFIPSCountryId(String fipsCountryCode) {
         if (isBlank(fipsCountryCode, "fipsCountryCode")) {
```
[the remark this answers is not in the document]
cah292: Adding new constant(s) accordingly.

cah292:
I've added some new constants based on the latest feedback. Please let me know if you think the new constants setup needs further adjustments for clarity.

src/main/java/edu/cornell/kfs/sys/CUKFSPropertyConstants.java
```diff
@@ -80,9 +80,16 @@ public class CUKFSPropertyConstants {
     public static final String FINALIZED_DATE = "finalizedDate";
 
     //** CU Generic ISO-FIPS Country modification items **
-    public static class ISOFIPSCountryMap {
+    public static class Location {
```
nkimble, line 83: Asking: Would using "Nation" instead of "Location" make this more self documenting when the strings are used elsewhere in the code?
I could go either way on this. Asking the question since my mind jumps to either asset location placement in buildings or location services at a street level and not the higher level concept of a country when I see that term.

nkimble (approved):
Changes look good. My concerns have been addressed.

PR #1467, opened 2023-06-30
KFSPTS-28364: Have bean cuContractsAndGrantsResponsibilityPlusPayPeriodRoleTypeService extend from parent bean contractsAndGrantsResponsibilityRoleTypeService so that the kimTypeInfoService is properly populated.

nkimble:


cah292 (approved):

PR #1468, opened 2023-07-05
KFSPTS-27393: Update checkstyle configuration to suppress the checks for CAM classes.

nkimble:


cah292 (approved):
Changes look good, unit tests pass and the checkstyle suppressions are working as expected in my local Eclipse app. Merging.

PR #1469, opened 2023-07-06
KFSPTS-28464: Duplicate static image renaming it to tinybutton-hide1.gif until KualiCo renames references to it without the number one in the file name.

nkimble:


jhulslander (approved):
in kfs-sandbox, I compared this file with /static/images/tinybutton-hide.gif and they seem to look like the same thing to me and download to the same size.  I think this is good to move forward with

PR #1471, opened 2023-07-12
KFSPTS-28624 Add ISO 20022 Immediate Check handling

cah292:
This is ready for initial review, but I wanted to perform some more testing before moving this out of the Draft state.

This PR makes the following changes that are required for the ISO 20022 conversion and the associated Cornell-side processing:

* Immediate Check payments will now be separated into a distinct ISO 20022 file. The resulting file will eventually be used by Cornell for printing immediate Checks, and it will be also sent to the bank to notify them of our locally-printed Checks.
* The queries for PDP Check payments can now be limited based on the immediate processing flag. Some additional overlay changes have also been introduced, to help distinguish between the immediate and non-immediate Check processing when generating the ISO 20022 files.
* To support the needs of our local Check printing, the following additional information will be added to the immediate Checks file. Note that the locations below are not the standard places for inserting this information in the XML, but it's the best we can do at the moment while still adhering to the XML schema version that the bank supports. (Newer schema versions have better spots for most of this info, but the bank doesn't support the newer schemas at this time.) We also confirmed with the bank that they will ignore the data in the locations below, so we don't have to worry about the extra info causing processing problems on the bank's end.
  * Added the associated KFS document number (if applicable) to the related "Issr" element within the structured remittance section.
  * Added the PDP payment notes as zero-amount payment adjustments.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentDetailServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentDetailServiceImpl.java
@@ -1,16 +1,13 @@
 package edu.cornell.kfs.pdp.service.impl;
 
-import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.kuali.kfs.core.api.criteria.CriteriaLookupService;
-import org.kuali.kfs.core.api.criteria.GenericQueryResults;
-import org.kuali.kfs.core.api.criteria.Predicate;
-import org.kuali.kfs.core.api.criteria.PredicateFactory;
-import org.kuali.kfs.core.api.criteria.QueryByCriteria;
+import org.kuali.kfs.krad.service.BusinessObjectService;
 import org.kuali.kfs.krad.util.KRADConstants;
 import org.kuali.kfs.pdp.PdpPropertyConstants;
 import org.kuali.kfs.pdp.businessobject.PaymentDetail;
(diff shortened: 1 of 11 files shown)


cah292:
I'm done with my local testing for now. More extensive testing is still needed, but that can be handled in DEV and/or Sandbox.

This can be merged once it has been approved.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentDetailServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+                PredicateFactory.equal(PdpPropertyConstants.PaymentDetail.PAYMENT_DISBURSEMENT_NUMBER,
+                        disbursementNumber),
+                PredicateFactory.equal(PdpPropertyConstants.PaymentDetail.PAYMENT_GROUP + "." +
+                        PdpPropertyConstants.PaymentGroup.PAYMENT_GROUP_PROCESS_ID, processId),
+                PredicateFactory.equal(PdpPropertyConstants.PaymentDetail.PAYMENT_GROUP + "." +
+                        PdpPropertyConstants.PaymentGroup.PAYMENT_GROUP_DISBURSEMENT_TYPE_CODE, disbursementType),
+                PredicateFactory.equal(PdpPropertyConstants.PaymentDetail.PAYMENT_GROUP + "." +
+                        PdpPropertyConstants.PaymentGroup.PAYMENT_GROUP_BANK_CODE, bankCode),
+                createProcessImmediateCondition(processImmediate));
+        GenericQueryResults<PaymentDetail> results = criteriaLookupService.lookup(PaymentDetail.class, criteria);
```
jhulslander, line 48: Base code is using businessObjectService to look up payment details, and Map<String, Object> fieldValues for the search criteria.  Would it be better to use base code style to make bringing in kualicp patches easier?

src/main/java/org/kuali/kfs/pdp/batch/service/impl/CheckExtractTypeContext.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class CheckExtractTypeContext extends AbstractExtractTypeContext {
+
+    private final PaymentUrgency urgency;
+
+    public CheckExtractTypeContext(
+            final Date extractBeginDate,
+            final PaymentStatus extractedStatus,
+            final PaymentProcess paymentProcess,
+            final PaymentUrgency urgency
```
jhulslander, line 45: would it be better to add a new constructor that includes payment urgency?  Removing base code constructor may break things if a new class starts using this context and we don't update it.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/ExtractTypeContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * ====
+ * CU Customization: Added method for limiting extraction based on the payment's urgency.
+ * ====
+ * 
+ * The state (i.e. context) of the extraction being performed, which can differ depending on the
+ * {@link ExtractionType}.
+ */
+public interface ExtractTypeContext {
```
jhulslander, line 36: Since we are adding a new function, would it be better to create CuPaymentUrgency that extends PaymentUrgency

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -451,6 +507,22 @@ private Set<PaymentInstructionInformation3> constructPaymentInstructionInformati
...
+     */
+    private boolean getExpectedProcessImmediateFlag(
+            final ExtractTypeContext extractTypeContext
+    ) {
+        if (extractTypeContext.isLimitedToPaymentsWithUrgency(PaymentUrgency.REGULAR)) {
+            return false;
+        } else if (extractTypeContext.isLimitedToPaymentsWithUrgency(PaymentUrgency.IMMEDIATE)) {
+            return true;
+        } else {
+            throw new IllegalArgumentException("Context of type " + extractTypeContext.getClass().getName()
```
jhulslander, line 521: Would it be better to default to false rather than error?

jhulslander (changes requested):
Code changes look good, I found a few things to consider

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentDetailServiceImpl.java
cah292: I wasn't sure whether we needed to handle the possibility of the `processImmediate` flag being null on the Payment Group. I haven't found any examples where this flag has been set to null in the database, though, so it might be safe to treat it as non-null, in which case I could switch this back to using the Business Object Service.

Thoughts?

src/main/java/org/kuali/kfs/pdp/batch/service/impl/CheckExtractTypeContext.java
cah292: Good point, I'll add a new constructor accordingly.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/ExtractTypeContext.java
cah292: `PaymentUrgency` is a CU-specific enum that is being introduced by this PR. Did you mean to suggest creating `CuExtractTypeContext` that extends `ExtractTypeContext`?

I had set up the new function as a "default" one, so that implementing classes are not required to override/implement it. That makes it easier to introduce the new functionality without breaking existing code. (Also, due to the areas of the code that will be using this, it's much simpler to overlay the base interface than to introduce a CU-specific one.)

If you have other thoughts or recommendations on how to improve this, please let me know.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: The enclosing `getExpectedProcessImmediateFlag()` method and the context's new `isLimitedToPaymentsWithUrgency()` method are only meant to be invoked when processing Checks. If an explicit Payment Urgency setting cannot be discerned in this situation, I would prefer to throw an exception here, since such a situation suggests a misconfiguration or other coding error.

If needed, I could rename the `getExpectedProcessImmediateFlag()` method, to make it clearer that it's only intended for Check processing situations.

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPaymentDetailServiceImpl.java
cah292: Given my comments above, I'm switching this back to using the Business Object Service, since it looks like the process-immediate flag is not being left as null (and since certain PaymentGroup-processing areas of KFS code already assume the flag is non-null). If you think the null-checking solution should be used instead, please let me know.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: I'm renaming the get-expected-flag method to make its intent clearer. If you still think the method should no longer have the exception-throwing part, please let me know.

cah292:
I've committed some follow-up changes based on the initial feedback. This is ready for review again.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/ExtractTypeContext.java
jhulslander: I think that makes sense, lets go with this approach

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
jhulslander: renaming sounds good, thanks

PR #1477, opened 2023-07-19
KFSPTS-28999 Fix PDP Payment Groups to use the correct XML element for reading the immediate-processing flag

db483:


jhulslander (approved):

PR #1478, opened 2023-07-21
KFSPTS-29020 Fix bean setup affecting e-SHOP

cah292:
This PR fixes a minor misconfiguration on one of our overridden beans. The change is necessary to get the e-SHOP integration working again for the 2022-10-19 financials upgrade.

We might be able to remove the bean override altogether, since I don't think it's introducing anything new on the cu-kfs side anymore. If you think it makes sense to remove it now instead of in a future story, please let me know.


db483 (changes requested):
@cah292 changes look good. I just wanted to check if the configuration of similar overridden file type beans was also checked as part of this jira or it should be done on a separate user story. Thank you.

cah292:
I had done a quick check of the other XML-related file type beans and didn't notice anything that needs changing now. I did find that we still have a CU-specific file type for some of our JAXB handling, but I can create a separate ticket to look into replacing that.

db483 (approved):

PR #1484, opened 2023-08-10
KFSPTS-29211 Fix ad hoc routed docs going to EXCEPTION

cah292:
This is ready for initial review, but I would like to do a little more local testing before it gets merged.

### DIFF
--- a/src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
+++ b/src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
@@ -31,7 +31,6 @@ public interface DocumentProcessingQueue {
      * when subsequently invoking the workflow engine.
      */
     default boolean shouldProceedWithDocumentProcessing(String documentId) {
-        throw new IllegalStateException("Method has not been overridden yet; this default implementation "
-                + "is only present to avoid potentially breaking existing code");
+        return true;
     }
 }
--- a/src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
+++ b/src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
@@ -1,0 +1,37 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2022 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.kew.api.document;
+
+public interface DocumentProcessingQueue {
+
+    void process(String documentId) throws IllegalArgumentException;
+
+    void processWithOptions(String documentId, DocumentProcessingOptions options) throws IllegalArgumentException;
+
+    /*
+     * CU Customization: Added a separate method for handling the FINP-8541 functionality (which checks whether
+     * a document has been routed and, if not, will add a Complete request if one doesn't exist). This is needed
+     * to fix a DB transaction issue, where merely retrieving the route header can cause it to remain stale
+     * when subsequently invoking the workflow engine.
+     */
+    default boolean shouldProceedWithDocumentProcessing(String documentId) {
+        throw new IllegalStateException("Method has not been overridden yet; this default implementation "
+                + "is only present to avoid potentially breaking existing code");
+    }
+}
--- a/src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
+++ b/src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
@@ -114,4 +114,17 @@
           p:workflowDocumentActionsService-ref="workflowDocumentActionsService"
     />
 
+    <!--
+        Override processing queue service to fix an issue that can cause the workflow engine
+        to process a stale copy of the route header and/or its downstream data.
+     -->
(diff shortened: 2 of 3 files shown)


cah292:
I've finished my testing efforts locally and in Sandbox for now; further validation can be performed in KFS DEV. This is now ready to be merged once it has been reviewed and approved.

src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
```diff
@@ -0,0 +1,37 @@
...
+    void processWithOptions(String documentId, DocumentProcessingOptions options) throws IllegalArgumentException;
+
+    /*
+     * CU Customization: Added a separate method for handling the FINP-8541 functionality (which checks whether
+     * a document has been routed and, if not, will add a Complete request if one doesn't exist). This is needed
+     * to fix a DB transaction issue, where merely retrieving the route header can cause it to remain stale
+     * when subsequently invoking the workflow engine.
+     */
+    default boolean shouldProceedWithDocumentProcessing(String documentId) {
+        throw new IllegalStateException("Method has not been overridden yet; this default implementation "
```
jhulslander, line 34: I don't understand why we want to throw this exception.  If the method hasn't been overriden, won't that break routing

src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
```diff
@@ -114,4 +114,17 @@
...
+        Override processing queue service to fix an issue that can cause the workflow engine
+        to process a stale copy of the route header and/or its downstream data.
+     -->
+    <bean id="documentProcessingQueue" class="org.kuali.kfs.kew.impl.document.DocumentProcessingQueueImpl"
+          p:workflowEngineFactory-ref="workflowEngineFactory"
+          p:actionRequestService-ref="actionRequestService"
+          p:documentAttributeIndexingQueue-ref="documentAttributeIndexingQueue"
+          p:routeHeaderService-ref="documentRouteHeaderService"
+          p:routeNodeService-ref="routeNodeService"
+          p:ksbMessageDeliveryMode="${message.delivery}"
```
jhulslander, line 127: When I look for message.delivery in kuali-puppet, I only see it in some bash SH files.  Is this setting needed elsewhere

jhulslander (commented):
I had a couple of questions

src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
cah292: I was trying to find a way to avoid breaking other potential compilation problems while still indicating that it would be best to explicitly implement this method. However, I agree that the thrown exception is confusing, so I'll update this method to default to true.

src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
cah292: In the financials project, the default KFS config file sets `message.delivery` to asynchronous mode, whereas the default integration test config file sets it to synchronous mode. These defaults should be appropriate and we don't need to override them elsewhere.

If you notice a spot where a non-unit-test environment is trying to set up synchronous processing instead, please let me know.

jhulslander: Take a look at https://github.com/search?q=repo%3ACU-CommunityApps%2Fkuali-puppet%20message.delivery&type=code .... looks like batch_job_up.sh is used, we should double check that

src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
```diff
@@ -0,0 +1,36 @@
...
+    void processWithOptions(String documentId, DocumentProcessingOptions options) throws IllegalArgumentException;
+
+    /*
+     * CU Customization: Added a separate method for handling the FINP-8541 functionality (which checks whether
+     * a document has been routed and, if not, will add a Complete request if one doesn't exist). This is needed
+     * to fix a DB transaction issue, where merely retrieving the route header can cause it to remain stale
+     * when subsequently invoking the workflow engine.
+     */
+    default boolean shouldProceedWithDocumentProcessing(String documentId) {
+        return true;
```
jhulslander, line 34: maybe we should log this event as you mentioned it shouldn't happen and is a sign of something awry.  We can set up an alert in datadog if the log entry is found

src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
jhulslander: looking at the code in https://github.com/CU-CommunityApps/kuali-puppet/blob/5bbd11f427df3d607ad9fd2f421c5de046edc016/files/batch_scripts/batch_job_up.sh#L22 I don't think it would ever set  mode="-Dmessage.delivery=synchronous"

cah292: I agree that I'm not seeing any setup that would result in the `-Dmessage.delivery=synchronous` option getting chosen, unless I missed something. Also, I confirmed in the KFS DEV logs that this morning's eInvoice Route Job was performing the `StandardWorkflowEngine` work in a different thread than the one that routed the document. It seems like we already have asynchronous processing set up in the KFS batch system.

src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
cah292: The problem with adding a log statement here is that this is an interface, not a class. As a result, a static `LOG` constant in this spot would have to be declared as public; Java doesn't let interfaces have private static constants.

If this default method somehow got invoked, I think the worst case is that the FINP-8541 feature won't be effective anymore. (That feature isn't a high priority; it just adds a Complete request to a SAVED document if such a request doesn't exist for some reason.) I doubt this default method would ever get invoked, and even if it did, it seems like the consequences of its execution would not be serious. Thus, we might not need to add logging here.

Thoughts?

cah292:
I committed a change to have the default method simply return true instead of throwing an exception. Please also see my other comments, where I agree that we don't need to worry about synchronous processing somehow being enabled, and where we might not need to add logging in that default method.

@jhulslander If you agree that no further changes are necessary, then this PR can be merged once it has been approved.

jhulslander (approved):
my concerns have been addressed, I shall merge

PR #1485, opened 2023-08-14
KFSPTS-29033 add default sort order for customer look up

jhulslander:


nkimble (approved):
Changes look good. Unit tests have passed. Merging.

PR #1487, opened 2023-08-18
KFSPTS-28621 Allow creating legacy Check XML while in ISO 20022 mode

cah292:
This PR updates the Check side of the PDP Extract process so that the legacy/proprietary Check XML will still get generated when creating ISO 20022 XML. That way, we can continue using the proprietary XML to handle Immediate Checks locally until we're fully ready to use the ISO 20022 XML for our local Checks.

Note that the FastTrack Check files are not needed for local Check printing, so they won't get generated when running in ISO 20022 mode. Also, these changes should only affect the Check file side of the PDP Extract; the ACH side should remain as-is.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
@@ -113,7 +113,8 @@ public class CuExtractPaymentServiceImpl extends ExtractPaymentServiceImpl {
 
     /**
      * Overridden to call a CU-specific version of the parameter-checking method to determine
-     * whether or not to use the ISO 20022 format.
+     * whether or not to use the ISO 20022 format, as well as to allow for generating
+     * the proprietary XML when the ISO 20022 format is enabled.
      */
     @Override
     public void extractChecks() {
@@ -125,10 +126,12 @@ public class CuExtractPaymentServiceImpl extends ExtractPaymentServiceImpl {
                         PdpConstants.PaymentStatusCodes.EXTRACTED
                 );
 
+        if (shouldCreateLegacyCheckFiles()) {
+            extractChecksToProprietaryFormat(extractedStatus);
+        }
+
         if (shouldUseIso20022Format()) {
             iso20022FormatExtractor.extractChecks(extractedStatus, directoryName);
-        } else {
-            extractChecksToProprietaryFormat(extractedStatus);
         }
 
         LOG.debug("extractChecks() - Exit");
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -184,6 +189,18 @@ private boolean isIso20022FormatParameterEnabled(String parameterName) {
                 KFSConstants.Components.ISO_FORMAT, parameterName, Boolean.FALSE);
     }
 
+    private boolean shouldCreateLegacyCheckFiles() {
+        return true;
```
jhulslander, line 193: Should this be driven off of a parameter rather than hard-coded?

jhulslander (commented):
Changes look good, I had some questions about the "should" methods.

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
cah292: I agree that it makes sense to tie this to a non-hard-coded value. I'm wondering whether a parameter would be ideal for it, though, given that the setting will only have a short-term lifetime. (We won't need the legacy Check code once our local Check printing migrates to ISO 20022, and KualiCo retires/removes that legacy code in one of its 2023 releases). A config property would be a little easier to maintain but would not be as flexible as a parameter.

I'm trying to discern whether this should be set up as a config property, or if it should be set up as a parameter that will eventually be deleted in a future release. Please let me know if you have any thoughts on this.

nkimble: @cah292: May this will help to make the decision:
If we had to turn something off or on, which option would there be:

-  the least amount of disruption to the users
- require no system down time
- no required code rebuild or redeploy and could be done at a moment's notice

Thinking about it that way, I would lean toward a parameter that could be cleanup at a later date.

cah292: @nkimble Thanks for the input. I was reluctant to create a parameter that I know we would be deleting in a few months, but upon further consideration, I'm fine with preparing a parameter-based approach instead. I will proceed with that option.

jhulslander: I would lean toward a parameter

se12: I also think a parameter is the way to go here.

cah292:
I've committed some changes to have the logic rely upon a new KFS parameter, as requested. An associated nonprod-sql PR has been created; please make sure both this PR and the other one are good to go before merging.

jhulslander (approved):
code changes look good, SQL is good too, I shall merge

PR #1488, opened 2023-08-22
KFSPTS-26965 Fix ISO 20022 XML namespacing and char escaping

cah292:
This PR makes a few improvements to the ISO 20022 file generation process:

* Adds XML escaping of a few extra characters that are not being escaped currently or consistently, such as quotes and apostrophes.
* Upgrades the ISO 20022 library to a newer version, to fix an issue where quotes could get duplicated in the output. (The version selected by this PR is a little newer than the one used by KualiCo's 2023-05-17 upgrade.)
* Adjusts the ISO 20022 marshalling to use the code library's mechanism for suppressing XML namespace prefixes, rather than using KualiCo's string substitution approach.


cah292:
I'm closing this PR since I've prepared an updated version of the changes on a different branch. I will open up the new PR shortly.

PR #1489, opened 2023-08-24
KFSPTS-29283 Fix missing Doc Reference Info for ISO 20022

cah292:
This PR fixes the ISO 20022 files so that the Invoice/Credit Note indicator will get filled in on payments sourced from PDP feeds. It also allows for placing the KFS doc number into the invoice number field when necessary, similar to (but not exactly like) the functionality in our FastTrack file generation code.

### DIFF
--- a/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
+++ b/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
@@ -1527,11 +1527,12 @@ public class Iso20022FormatExtractor {
         } else if (isCreditNote(paymentDetail)) {
             documentTypeCode = DocumentType5Code.CREN;
         } else {
-            LOG.warn(
-                    "determineDocumentTypeCode(...) - PaymentDetail has unexpected DocumentTypeCode : "
-                    + "paymentDetailFinancialDocumentTypeCode={}",
-                    paymentDetail::getFinancialDocumentTypeCode
-            );
+            /*
+             * CU Customization: Allow payments from PDP feeds to be marked as Invoices or Credit Notes.
+             */
+            documentTypeCode = paymentDetail.getNetPaymentAmount().isGreaterEqual(KualiDecimal.ZERO)
+                    ? DocumentType5Code.CINV
+                    : DocumentType5Code.CREN;
         }
         return documentTypeCode;
     }
--- a/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
+++ b/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
@@ -1484,7 +1484,10 @@ public class Iso20022FormatExtractor {
                 extractTypeContext);
         referredDocumentInformation.setTp(referredDocumentType);
 
-        final String documentNumber = paymentDetail.getInvoiceNbr();
+        /*
+         * CU Customization: Externalized the Referred Document Number setup to a separate method.
+         */
+        final String documentNumber = determineReferredDocumentNumber(paymentDetail);
         referredDocumentInformation.setNb(documentNumber);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1527,11 +1530,12 @@ private static DocumentType5Code determineDocumentTypeCode(
...
         } else {
-            LOG.warn(
-                    "determineDocumentTypeCode(...) - PaymentDetail has unexpected DocumentTypeCode : "
-                    + "paymentDetailFinancialDocumentTypeCode={}",
-                    paymentDetail::getFinancialDocumentTypeCode
-            );
+            /*
+             * CU Customization: Allow payments from PDP feeds to be marked as Invoices or Credit Notes.
+             */
+            documentTypeCode = paymentDetail.getNetPaymentAmount().isGreaterEqual(KualiDecimal.ZERO)
```
db483, line 1536: Is the FinancialDocumentTypeCode value not filled when the payment comes from the PDP feed? I was just wondering if there is a way to determine that the payment comes from the PDP feed or is enough to assume that the payment comes from the PDP feed if it is not invoice or credit note? Thank you

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1557,6 +1561,26 @@ private static boolean isCreditNote(
...
+    private static String determineReferredDocumentNumber(
+            final PaymentDetail paymentDetail
+    ) {
+        if (StringUtils.equalsAny(paymentDetail.getFinancialDocumentTypeCode(),
+                DisbursementVoucherConstants.DOCUMENT_TYPE_CHECKACH,
+                DisbursementVoucherConstants.DOCUMENT_TYPE_CODE)) {
+            if (StringUtils.isNotBlank(paymentDetail.getInvoiceNbr()) && paymentDetail.getInvoiceDate() != null) {
+                return paymentDetail.getInvoiceNbr();
+            } else {
+                return CuDisbursementVoucherConstants.DV_EXTRACT_EDOC_NUMBER_PREFIX_IDENTIFIER
```
db483, line 1576: what is the source of this payment?

db483 (changes requested):
Changes look good but I had a couple of questions. Thank you.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: As per the `StringUtils.isAny()` comparison a few lines above, the payment is sourced from a DV if execution reaches this marked line. As far as I can tell, DV-sourced payments appear to be the only ones that can potentially have a blank invoice number.

Please let me know if you need me to make the is-DV check clearer.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
db483: Given that the FastTrack file generation worked in a similar way I am ok with this assumption. Thank you.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
db483: Are these DV's created in a different manner? Or is just a condition that can appear on any DV? Thank you.

cah292: DVs have fields in the "Payment Information" section for specifying the Invoice Number and Invoice Date, but they appear to be optional fields. As a result, there are cases where a DV document does not have an Invoice Number.

db483 (approved):

PR #1491, opened 2023-08-29
KFSPTS-29330 RASS Job NullPointerException

ajd299:


ajd299:
@cah292 do you think this fix is viable?

cah292:
@ajd299,

Upon further analysis, I think the fixes from KFSPTS-18189 and KFSPTS-29211 are interfering with each other. The former forces the RASS job to explicitly invoke the `DocumentProcessingQueue` in a custom thread, but based on the stack traces, I'm wondering if the operation is somehow running in the same thread. At the time of that invocation, though, the transaction that created/routed the edoc has not been committed yet. Thus, when the `DocumentProcessingQueue` runs (either in the same thread or an on-demand thread), and when it reaches the KFSPTS-29211 fix, it will invoke the helper's `handleNotYetRoutedDocument()` method **in a new transaction** that is unaware of the previous transaction's changes. I believe that's why the problematic method can't find the route header.

We cannot remove the KFSPTS-29211 fix because that would reintroduce our recent ad hoc routing errors; I would also advise against adding a null check to `handleNotYetRoutedDocument()` since that might hide any future transactional problems that come up. Instead, perhaps it would make sense to remove the KFSPTS-18189 fix; see my notes below.

Document-routing actions already trigger an asynchronous call to the `DocumentProcessingQueue`, and such calls are set up so that they won't start executing until the caller's transaction has been committed. This should be enough to resolve the transactional issues noted above. As for why we added the custom `DocumentProcessingQueue` call for KFSPTS-18189, that was needed to work around an efficiency issue related to the Rice/KFS Service Bus. It was spending too much time trying to send cache-flush messages to Rice/KFS nodes that no longer existed in AWS, thus creating a backlog in the thread pool that's shared by the Service Bus calls and async service calls. However, now that the cache-flush messages are instead being handled by our hybrid EhCache-and-Redis setup (which is much better and also uses its own dedicated thread pool for sending messages to Redis), I don't think we need the KFSPTS-18189 changes anymore. Thus, I would suggest removing the custom `DocumentProcessingQueue` invocation to see if that fixes the problem.

To remove the KFSPTS-18189 changes, the `RassUpdateServiceImpl` should be modified to no longer explicitly invoke the `DocumentProcessingQueue` and to no longer use the custom `DocumentProcessingQueueRunner` class. Note that some of the unit tests will also need their configuration updated.

Please let me know if you need more clarity on the details above, or if you have further questions or concerns about my suggestions.

ajd299:
@cah292 I made some changes based on your recommendation. Please review and let me know your thoughts.

cah292 (changes requested):
I agree that the changes made thus far should be enough to fix the issue. However, I would suggest including the following cleanup as well:

* Remove the custom `DocumentProcessingQueueRunner` class if it's no longer being used anywhere.
* Clean up the imports in the `RassUpdateServiceImpl` class.
* Clean up the related RASS unit test configuration:
  * Remove the DocumentProcessingQueue references from the `RassMockServiceFactory` class.
  * Remove the "documentProcessingQueue" beans and ID references from the `cu-spring-rass-service-test.xml` file.

ajd299:
Okay, please see my cleanup changes

src/main/java/org/kuali/kfs/kew/api/document/DocumentProcessingQueue.java
cah292, line 1: Please put the `DocumentProcessingQueue` overlay back in place, since it's needed for the KFSPTS-29211 fix that we still intend to preserve. It's only the KFSPTS-18189 fix that needs to be removed.

src/main/java/org/kuali/kfs/kew/impl/document/DocumentProcessingQueueImpl.java
cah292, line 1: Please put the `DocumentProcessingQueueImpl` overlay back in place, since it's needed for the KFSPTS-29211 fix that we still intend to preserve. It's only the KFSPTS-18189 fix that needs to be removed.

src/main/resources/edu/cornell/kfs/kew/config/cu-spring-kew.xml
```diff
@@ -114,17 +114,4 @@
           p:workflowDocumentActionsService-ref="workflowDocumentActionsService"
     />
 
-    <!--
-        Override processing queue service to fix an issue that can cause the workflow engine
-        to process a stale copy of the route header and/or its downstream data.
-     -->
-    <bean id="documentProcessingQueue" class="org.kuali.kfs.kew.impl.document.DocumentProcessingQueueImpl"
```
cah292, line 121: This bean override needs to be put back in place for the KFSPTS-29211 fix, which we still intend to preserve.

cah292 (changes requested):
I noticed a few removed items that need to be put back in place, but otherwise the changes look good.

ajd299:
Please see my latest changes

cah292 (commented):
Latest changes look good, will wait for the unit tests to pass before merging.

cah292 (approved):

PR #1266, opened 2022-03-24
KFSPTS-24454: Backport FINP-8322 improve performance of action request update for responsibility change

nkimble:

PR #1493, opened 2023-09-01
KFSPTS-29360 Fix parallel routing config on route nodes

cah292:
When we added the Object Code routing to our documents locally, some of the doc type XML files left out the "Parallel" routing indicator on several route nodes, causing them to default to "Sequential" routing and to mess up the route behavior in Production. This PR fills in the gaps by updating the doc type XML to include the "Parallel" routing indicator on the appropriate nodes.

Note the following:

* There were no changes to runtime code or config files, only to the workflow XML files that we're keeping in the codebase for reference.
* The associated SQL changes will automatically update the relevant route nodes at the database level, so that we don't have to ingest the XML at this time. The changes were made on the PR below (which has already been merged):
  * https://github.com/CU-CommunityApps/nonprod-sql/pull/1266
* Even though the associated SQL mentions all the doc types that were updated as part of the Object Code routing work, only 8 of them actually needed their route nodes fixed in the codebase (and database). Furthermore, there were a couple nodes on the DV and PCDO docs that were purposely left in "Sequential" routing mode, for consistency with the base financials configuration; the related SQL's `WHERE` clause references those exact nodes.


ajd299 (approved):
Looks good, workflow changes xml

PR #1496, opened 2023-09-08
KFSPTS-29464 Improve ISO 20022 blank country handling

cah292:
This PR improves the country handling in the ISO 20022 extraction. If a transaction has a blank country code but still appears to reference a domestic state code and zip code, then the "US" country code will be used in the file entry. Otherwise, the "ZZ" (Unknown) code will be used instead.

Note that, similar to KualiCo's `PostalCodeValidationServiceImpl` class, the updated code is using a `ZipcodeValidationPattern` instance to assist with the zip code pattern checking.

### DIFF
--- a/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
+++ b/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
@@ -848,12 +848,12 @@ public class Iso20022FormatExtractor {
         } else if (stateAndPostalCodeAreWithinUnitedStates(templateCustomerProfile.getStateCode(),
                 templateCustomerProfile.getZipCode())) {
             LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code but seems to reference "
-                    + "a domestic address; defaulting to '{}'",
+                    + "a domestic address; defaulting to ISO country code '{}'",
                     templateCustomerProfile.getId(), KFSConstants.COUNTRY_CODE_UNITED_STATES);
             isoCountryCode = KFSConstants.COUNTRY_CODE_UNITED_STATES;
         } else {
             LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code but seems to reference "
-                    + "a foreign address; defaulting to '{}'",
+                    + "a foreign address; defaulting to ISO country code '{}'",
                     templateCustomerProfile.getId(), CUKFSConstants.ISO_COUNTRY_CODE_UNKNOWN);
             isoCountryCode = CUKFSConstants.ISO_COUNTRY_CODE_UNKNOWN;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -833,8 +845,15 @@ private PostalAddress6 constructPostalAddress(
...
+        } else if (stateAndPostalCodeAreWithinUnitedStates(templateCustomerProfile.getStateCode(),
+                templateCustomerProfile.getZipCode())) {
+            LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code but seems to reference "
+                    + "a domestic address; defaulting to '{}'",
+                    templateCustomerProfile.getId(), KFSConstants.COUNTRY_CODE_UNITED_STATES);
+            isoCountryCode = KFSConstants.COUNTRY_CODE_UNITED_STATES;
         } else {
-            LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code; defaulting to '{}'",
+            LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code but seems to reference "
+                    + "a foreign address; defaulting to '{}'",
```
ajd299, line 856: Suggestion: "defaulting to" => "defaulting iso country code to"

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -879,8 +898,15 @@ private PostalAddress6 constructPostalAddress(
             isoCountryCode = wasPaymentGroupCreatedPriorToISOCountryConversion(templatePaymentGroup)
                     ? convertFIPSCountryValueToISOCountryCode(countryValue)
                     : convertISOCountryValueToISOCountryCode(countryValue);
+        }  else if (stateAndPostalCodeAreWithinUnitedStates(templatePaymentGroup.getState(),
```
ajd299, line 901: is there an extra space here (before "else if")?

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1027,6 +1053,27 @@ private String convertFIPSCountryNameToISOCountryCode(String fipsCountryName) {
         return isoCountryCode;
     }
 
+    private boolean stateAndPostalCodeAreWithinUnitedStates(String stateCode, String postalCode) {
```
ajd299, line 1056: I think this method is well written and easy to understand.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -833,8 +845,15 @@ private PostalAddress6 constructPostalAddress(
         String isoCountryCode;
         if (StringUtils.isNotBlank(countryCode)) {
             isoCountryCode = convertFIPSCountryCodeToISOCountryCode(countryCode);
+        } else if (stateAndPostalCodeAreWithinUnitedStates(templateCustomerProfile.getStateCode(),
+                templateCustomerProfile.getZipCode())) {
+            LOG.warn("constructPostalAddress, Customer Profile {} has a blank country code but seems to reference "
+                    + "a domestic address; defaulting to '{}'",
+                    templateCustomerProfile.getId(), KFSConstants.COUNTRY_CODE_UNITED_STATES);
+            isoCountryCode = KFSConstants.COUNTRY_CODE_UNITED_STATES;
```
ajd299, line 853: I think extracting this code to a method would add complexity to the log statement so I am okay with leaving it as is.

ajd299 (changes requested):
Please see my comments

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: I'll update the wording accordingly. I think the phrase "defaulting to ISO country code" would work a little better; please let me know if you would prefer your suggested wording instead.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: Good catch, I'll fix that.

cah292:
I believe I've addressed all of the appropriate follow-up items. This is ready for review again.

ajd299 (approved):
Changes look good, merging

PR #1499, opened 2023-09-15
KFSPTS-26965 Fix ISO 20022 XML namespacing and char escaping

cah292:
This PR makes a few additional changes to the ISO 20022 file generation:

* Sets up a custom `EscapeHandler` implementation (used by the relevant ISO 20022 Java library) that performs the char escaping (or non-escaping) requested by the new bank. The new class also works around a bug in the library's code that could cause quotes to be duplicated in the ISO 20022 file output.
* Adjusts the file generation to use the Java library's standard means of controlling the XML namespacing in the file output, rather than using KualiCo's string substitution hack.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuEscapeHandler.java
+++ b/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuEscapeHandler.java
@@ -5,7 +5,7 @@ import java.util.Map;
 import com.prowidesoftware.swift.model.mx.EscapeHandler;
 
 /*
- * CU-specific EscapeHandler implementation that is similar to Prowide's DefaultEscapeHandler,
+ * CU-specific EscapeHandler implementation that behaves similarly to Prowide's DefaultEscapeHandler,
  * but with some important differences:
  * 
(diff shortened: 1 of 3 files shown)


nkimble (commented):
Changes look good. 

Asking if this makes sense: Should we have a unit test for the new CuEscapeHandler? 
I am not suggesting that it use the data or logic from the JPMC changes. I am asking whether we should have simple happy-path/unhappy-path unit tests with bogus data that exercise the logic we think should be happening within this class to ensure future upgrades do not affect it.

cah292:
The main format extractor class may be too complex to feasibly write unit tests for it, but I agree that the escape handler is simple enough to support some simple tests. I'll look into preparing a few quick test cases for it.

cah292:
I've added some quick unit tests to verify the escape handler's functionality. (Note that in one of the test cases, the `\u00ED` value represents an accented lowercase "i" character.) This is ready for review again.

nkimble (approved):
Code looks good. Unit tests have passed. My concerns have been addressed. Approving changes and merging to develop branch.

PR #1502, opened 2023-09-19
KFSPTS-27527 Fix handling of IWNT amounts and addresses

cah292:
This PR introduces a few fixes to the IWant Document:

* Fixed a bug that was preventing the Requestor's address from auto-updating when entering a different Requestor NetID.
* Fixed an issue where Requestor or Deliver To addresses could end up with the word "null" on address lines containing null data. (Note that this only affects the auto-population of addresses from KIM data; any pre-stored IWNT "default" delivery addresses are beyond the scope of this PR.)
* Fixed some bugs that were interfering with real-time updates to Item and Account totals on the IWNT form. Some cases required specifying the `readOnlyAlternateDisplay` attribute on the relevant tags, to prevent KFS from writing the initial displayed value to a location that is not affected by the `setRecipientValue()` function.


db483 (approved):
Changes look good.

PR #1503, opened 2023-09-22
KFSPTS-29602 Replace newlines in ISO 20022 check stubs

cah292:
The new bank doesn't want the check stubs to contain newlines, so this PR changes the ISO 20022 process to replace them.

The current solution doesn't string-normalize the text; if you have a strong opinion about adding that, please let me know.

### DIFF
--- a/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
+++ b/src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import javax.xml.datatype.DatatypeConfigurationException;
 import javax.xml.datatype.DatatypeConstants;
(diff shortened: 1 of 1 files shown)


src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -169,6 +170,11 @@ public class Iso20022FormatExtractor {
 
     private static final int VENDOR_NUM_MAX_LENGTH = 20;
 
+    /*
+     * CU Customization: Added a helper regex for replacing newlines.
+     */
+    private static final Pattern NEWLINE_PATTERN = Pattern.compile("(\\r\\n|\\r|\\n)");
```
ajd299, line 176: Should the string be moved to a a constant? I am fine with the way it is if that is simpler for readability.

ajd299 (commented):
Just one question

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: We usually don't create a separate constant for a static regex string, especially if it's only used in one spot. This static field is already declared as a constant and the `Pattern` class is immutable, so I would prefer to keep this definition as-is.

ajd299 (approved):
Okay sounds good, merging

PR #1504, opened 2023-09-27
KFSPTS-28557 remove Award node for purchase order amendment documents

jhulslander:
I have made changes to the route node logic in the document type XML.  I wasn't sure if I should remove the category statuses "Disapproved C and G" and "Awaiting C and G Approval".  I was leaning toward not to avoid search related problems, but I am not certain.
The user story references some responsibilities.  Should I add sub tasks for Nicole to deactivate those?

### DIFF
--- a/src/main/resources/edu/cornell/kfs/module/purap/document/workflow/PurchaseOrderAmendmentDocument.xml
+++ b/src/main/resources/edu/cornell/kfs/module/purap/document/workflow/PurchaseOrderAmendmentDocument.xml
@@ -82,16 +82,7 @@
                         <join name="JoinNewUnorderedItems"/>
                     </split>
                     <role name="SubAccount" nextNode="Account" nextAppDocStatus="Awaiting Fiscal Officer"/>
-                    <role name="Account" nextNode="RequiresAwardReview" nextAppDocStatus="Awaiting C and G Approval"/>
-                    <split name="RequiresAwardReview" nextNode="Commodity">
-                        <branch name="True">
-                            <role name="Award" nextNode="JoinRequiresAwardReview" nextAppDocStatus="Awaiting Commodity Code Approval"/>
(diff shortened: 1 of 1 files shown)


cah292:
I think it's fine to keep the "Disapproved C and G" and "Awaiting C and G Approval" status definitions in place, for consistency with the other PO document variants.

As for whether to add responsibility inactivation sub-tasks, I'll add a follow-up note with more details.

cah292 (commented):
The POA workflow XML changes look good. There are a few follow-up tasks that are still needed, though, but we'll need more functional input as to the timing for implementing them:

1. I agree with the note in the PR's description that the KIM responsibilities for the removed nodes should be inactivated.
2. Also, I would suggest updating the `CuPurchaseOrderAmendmentDocument.answerSplitNodeQuestion()` method to remove the references to the deleted nodes.

For the existing Enroute POA documents that predate this workflow change, do the functionals expect them to be finalized prior to (or at the time of) ingesting this workflow XML into PROD? If such Enroute docs are expected to take several days/weeks to finish routing, then the code and responsibility cleanup items above may need to be deferred to a follow-up ticket.

jhulslander:
Chad,
Thanks for the feedback, I will ping Janet concerning the thoughts

jhulslander:
@cah292  Based on Janet's feed back, I created KFSPTS-29735 to handle the permission changes and code changes to be done after POA documents created on the old workflow have been finalized.  I added a note to KFSPTS-29666 to capture a list of documents when the new workflow XML is ingested to prod.

Please update the follow up user story if you think I missed any details.

cah292 (approved):
Changes look good and I'm fine with the appropriate follow-up items being deferred to a future user story. Sorry for the delay in re-reviewing this.

Merging.

PR #1505, opened 2023-09-28
KFSPTS-29711 add missing activationType 'P'

db483:


cah292 (approved):
Workflow XML changes look good and unit tests pass. Merging.

PR #1509, opened 2023-10-12
KFSPTS-29735 Removed unused split node

ajd299:
Remove code in the method CuPurchaseOrderAmendmentDocument.answerSplitNodeQuestion corresponding to the split node removal in https://github.com/CU-CommunityApps/cu-kfs/pull/1504


ajd299:
Please let me know if I missed anything?

cah292 (commented):
Changes look good, but I had a couple questions that Jay or the functionals should weigh in on first:

1. There's some related KIM data that also needs to be inactivated. Are we planning to have Nicole do the inactivation manually, or do we want to use SQL to do so?
2. Is it okay to proceed with moving the code and KIM inactivations forward to at least DEV/TEST now? Or do we want to wait until we're certain the appropriate old POA documents have finished routing in PROD?

cah292 (approved):
As per the discussion at today's team standup, the KIM changes will be performed manually by the users and the older POA documents have now been finalized in Production. Thus, my concerns have been addressed and this code is now ready to be merged.

PR #1511, opened 2023-10-13
KFSPTS-29789 fix jaggaer XML job PO mode to find vendors paid on one PO

jhulslander:


src/main/java/edu/cornell/kfs/module/purap/batch/dataaccess/impl/JaggaerUploadDaoJdbc.java
```diff
@@ -245,7 +245,7 @@ private CuSqlChunk buildPurchaseOrderLimitSubQuery(Date processingDate) {
                 "FROM KFS.PUR_PO_T ",
                 "WHERE PO_LST_TRNS_DT > ",
                 CuSqlChunk.forParameter(Types.DATE, processingDate),
-                " GROUP BY VNDR_HDR_GNRTD_ID HAVING COUNT(1) > 1)");
+                " GROUP BY VNDR_HDR_GNRTD_ID HAVING COUNT(1) >= 1)");
```
cah292, line 248: changing this condition from `> 1` to `>= 1` might make the whole `HAVING` clause redundant, though I agree that writing it this way seems to make the business logic clearer. Please let me know if you'd rather leave it like this or if you think it should be changed.

cah292 (commented):
I wanted to confirm one quick item before merging.

src/main/java/edu/cornell/kfs/module/purap/batch/dataaccess/impl/JaggaerUploadDaoJdbc.java
jhulslander: I think the improved clarity of leaving the code like this is worth it, would you agree?

cah292: Leaving it like this is fine with me. Thanks for confirming!

cah292 (approved):

PR #1512, opened 2023-10-18
KFSPTS-29740 Fix rare role document bug

cah292:
KualiCo's KIM Role updates are replacing certain objects' OJB collection/list proxies with different list instances, but such replacement can sometimes interfere with OJB's bulk collection pre-fetching behavior and result in a ClassCastException. This PR works around the issue by forcibly loading the contents of the problematic list proxies in advance, so that the bulk pre-fetching will occur prior to KualiCo's list replacement code.


db483 (approved):
Changes look good. I will merge once the unit tests pass.

PR #1516, opened 2023-10-25
KFSPTS-29958 update log4j to 2.21.1 and add log4j-layout-template-json

jhulslander:
This PR and a new log4j dependency log4j-layout-template-json and  updates log4j to version 2.21.1 per @dp65  needs to DataDog.

We were previously at 2.18 so this update is only a few minor patches newer.  Also KualiCo is 2.20: https://github.com/KualiCo/financials/blob/master/pom.xml .  So it seems safe to do this update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2333,7 +2333,7 @@
             <artifactId>log4j-web</artifactId>
             <version>${log4j.version}</version>
         </dependency>
-        <!-- This is needed to support JSON logging output for DataDog -->
+        <!-- CU customization to support JSON logging for DataDog. Log4j minimum version 2.21.1 required for this feature. -->
         <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-layout-template-json</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2333,6 +2333,12 @@
             <artifactId>log4j-web</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+        <!-- This is needed to support JSON logging output for DataDog -->
```
nkimble, line 2336: I agree we need to have a comment here but IMHO this looks to much like a KualiCo comment and not a Cornell customization. I think it needs to be more like the comment style on lines 29-31 of this file. Maybe something like the following which would notate why we took the log4j version bump so that is not lost during future KualiCo upgrades (since we will be ahead of the version they are using).

_CU customization to support JSON logging for DataDog. Log4j minimum version 2.21.1 required for this feature._

Thoughts?

nkimble (changes requested):
Changes look good. I did have a suggestion to consider.

pom.xml
jhulslander: no objections to comment change, I shall change

jhulslander:

nkimble (approved):
My concerns have been addressed. Approving and merging to develop.

PR #1519, opened 2023-10-31
KFSPTS-29608: Fix IWantDocument Requestor lookup to properly load the new address information when the Requestor is changed by the lookup.

nkimble:


src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
```diff
@@ -1035,6 +1028,26 @@ public ActionForward refresh(ActionMapping mapping, ActionForm form, HttpServlet
         return actionForward;
     }
 
+    private String obtainDeliverToAddressForRefresh(IWantDocument iWantDocument) {
```
jhulslander, line 1031: Two thoughts on this:
1) These functions could be combined with something like findCampusAddressFromNetId(String netId)

2) It looks like this approach will over write a manually edited address with the campus address of the person.  Is that the desired outcome?

jhulslander (commented):
I had a couple of thoughts

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
nkimble: Regarding your two points:

1. I have combined the two functions into one parameterized function as requested.
 
2. Per today's standup discussion with the functional user, the behavior of overwriting any manually entered address is the desired outcome.

nkimble (commented):

jhulslander (approved):

PR #1520, opened 2023-11-07
KFSPTS-29034 JaggaerXMLInputFileType using KualiCo JAXB KualiCo 

ajd299:
Part 1 in our attempt to conform our code to be similar to base code

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
@@ -1,0 +1,128 @@
+package edu.cornell.kfs.module.purap.batch;
+
+import edu.cornell.kfs.module.purap.CUPurapParameterConstants;
+import edu.cornell.kfs.module.purap.batch.service.impl.JaggaerUploadSupplierXmlFileDetailsDto;
+import edu.cornell.kfs.sys.CUKFSConstants;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+import org.kuali.kfs.core.api.config.property.ConfigurationService;
+import org.kuali.kfs.coreservice.framework.parameter.ParameterService;
+import org.kuali.kfs.module.purap.PurapConstants;
+import org.kuali.kfs.module.purap.PurapKeyConstants;
+import org.kuali.kfs.sys.batch.XmlBatchInputFileTypeBase;
+
+import java.io.File;
+import java.util.Locale;
+
+public class JaggaerXMLInputFileType extends XmlBatchInputFileTypeBase<JaggaerUploadSupplierXmlFileDetailsDto> {
+
+    protected static final DateTimeFormatter DATE_FORMATTER_FOR_FILE_NAME = DateTimeFormat
+            .forPattern(CUKFSConstants.DATE_FORMAT_yyyyMMdd_HHmmssSSS).withLocale(Locale.US);
+
+    protected String reportPath;
+    protected String reportPrefix;
+    protected String reportExtension;
+//    protected String jaggaerXmlDirectory;   replace with base class property
+    protected ParameterService parameterService;
+    protected ConfigurationService configurationService;
+    protected String fileNamePrefix;
+
+    /**
+     * Returns the identifier of the electronic invoice file type
+     *
+     * @return the electronic invoice file type identifier
+     */
+    @Override
+    public String getFileTypeIdentifier() {
+        return PurapConstants.ELECTRONIC_INVOICE_FILE_TYPE_INDENTIFIER;
+    }
+
+    @Override
+    public boolean validate(final Object parsedFileContents) {
+        return true;
+    }
+
+    @Override
+    public String getTitleKey() {
+        return PurapKeyConstants.MESSAGE_BATCH_UPLOAD_TITLE_EINVOICE;
+    }
+
+    @Override
+    public String getAuthorPrincipalName(final File file) {
+        return null;
+    }
+
+    @Override
+    public String getFileName(String principalName, Object parsedFileContents, String fileUserIdentifier) {
+        return "";
+//        String outputFileName = jaggaerXmlDirectory + findOutputFileNameStarter()
+//                + DATE_FORMATTER_FOR_FILE_NAME.print(dateTimeService.getCurrentDate().getTime())
+//                + CUKFSConstants.XML_FILE_EXTENSION;
+//
+//        final String fileName = ((ElectronicInvoice) parsedFileContents).getFileName();
+//        if (fileName == null) {
+//            return fileUserIdentifier;
+//        }
+//        final int whereDot = fileName.lastIndexOf('.');
+//
+//        return fileName.substring(0, whereDot);
+    }
+
+    protected String findOutputFileNameStarter() {
+        return getParameterValueString(CUPurapParameterConstants.JAGGAER_DEFAULT_SUPPLIER_OUTPUT_FILE_NAME_STARTER);
+    }
+
+    public String getReportPath() {
+        return reportPath;
+    }
+
+    public void setReportPath(final String reportPath) {
+        this.reportPath = reportPath;
+    }
+
+    public void setReportExtension(final String reportExtension) {
+        this.reportExtension = reportExtension;
+    }
+
+
+    public String getReportExtension() {
+        return reportExtension;
+    }
+
+    public String getReportPrefix() {
+        return reportPrefix;
+    }
+
+    public void setReportPrefix(final String reportPrefix) {
+        this.reportPrefix = reportPrefix;
+    }
+
+    protected String getParameterValueString(String parameterName) {
+        return parameterService.getParameterValueAsString(JaggaerGenerateSupplierXmlStep.class, parameterName);
+    }
+
+    public ConfigurationService getConfigurationService() {
+        return configurationService;
+    }
+
+    public void setConfigurationService(ConfigurationService configurationService) {
+        this.configurationService = configurationService;
+    }
+
+    public String getFileNamePrefix() {
+        return fileNamePrefix;
+    }
+
+    public void setFileNamePrefix(String fileNamePrefix) {
+        this.fileNamePrefix = fileNamePrefix;
+    }
+
+    public ParameterService getParameterService() {
+        return parameterService;
+    }
+
+    public void setParameterService(ParameterService parameterService) {
+        this.parameterService = parameterService;
+    }
+
+}
--- a/src/main/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerGenerateSupplierXmlServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerGenerateSupplierXmlServiceImpl.java
@@ -14,6 +14,7 @@ import java.util.Locale;
 import java.util.UUID;
 import java.util.regex.Pattern;
 
+import edu.cornell.kfs.module.purap.batch.JaggaerXMLInputFileType;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -71,7 +72,7 @@ public class JaggaerGenerateSupplierXmlServiceImpl implements JaggaerGenerateSup
             .forPattern(CUKFSConstants.DATE_FORMAT_yyyy_MM_dd_T_HH_mm_ss_SSS_Z).withLocale(Locale.US).withZoneUTC();
     protected static final DateTimeFormatter DATE_FORMATTER_FOR_FILE_NAME = DateTimeFormat
             .forPattern(CUKFSConstants.DATE_FORMAT_yyyyMMdd_HHmmssSSS).withLocale(Locale.US);
-    
+
     private Pattern numberPattern = Pattern.compile("-?\\d+(\\.\\d+)?");
 
     private String jaggaerXmlDirectory;
@@ -84,6 +85,7 @@ public class JaggaerGenerateSupplierXmlServiceImpl implements JaggaerGenerateSup
     protected ParameterService parameterService;
     protected WebServiceCredentialService webServiceCredentialService;
     protected ConfigurationService configurationService;
+    protected JaggaerXMLInputFileType jaggaerXMLInputFileType;
     protected JaggaerGenerateSupplierXmlReportService jaggaerGenerateSupplierXmlReportService;
 
     @Override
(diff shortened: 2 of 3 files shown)


jhulslander:
retest this please

src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.commons.lang.StringUtils;
+import org.kuali.kfs.core.api.datetime.DateTimeService;
+import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.batch.XmlBatchInputFileTypeBase;
+
+import java.io.File;
+
+import static org.kuali.kfs.pdp.PdpConstants.FILE_NAME_PART_DELIMITER;
+
+public class JaggaerXMLInputFileType extends XmlBatchInputFileTypeBase<SupplierSyncMessage> {
```
jhulslander, line 15: Should we add additional unit testing so we can validate this is doing what we expect?

src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        String authorPrincipalName = filename.substring(filename.indexOf(fileNamePrefix + fileNamePrefix.length() + 1));
+        authorPrincipalName = filename.substring(0, authorPrincipalName.indexOf(FILE_NAME_PART_DELIMITER));
+        return authorPrincipalName;
+    }
+
+    @Override
+    public String getFileName(String principalName, Object parsedFileContents, String fileUserIdentifier) {
+        String fileName = new StringBuilder()
```
jhulslander, line 49: The origninal implementation included a very fine grained time element to make sure the batch job created unique files when it had to create multiple files
I am OK proceeding with this, but we should run the batch job in sandbox in PO mode, without uploads turned off to make sure file names are good.

jhulslander (dismissed):
I need to do further review, but I had some thoughts to consider.  We may want to add some more unit testing to make sure the new code is working as expected.

We should run this code in sandbox in PO mode with uploads turned off to make sure file names are unique enough

nkimble:
retest this please

src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
ajd299: I've added a few unit tests

src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
```diff
@@ -0,0 +1,76 @@
...
+                .append(fileNamePrefix)
+                .append(principalName)
+                .append(StringUtils.isBlank(fileUserIdentifier) ? StringUtils.EMPTY : FILE_NAME_PART_DELIMITER + fileUserIdentifier)
+                .append(FILE_NAME_PART_DELIMITER)
+                .append(dateTimeService.toDateTimeStringForFilename(dateTimeService.getCurrentDate()))
```
ajd299, line 54: This was taken from JAXBXmlBatchInputFileBase

jhulslander: This particular implementation will need to have a granular enough date time attribute to make sure the batch job can produce multiple files without over writing anything.
Testing in sandbox should help us answer that question.

ajd299: I have built (new-jenkins) and deployed (old jenkins) sandbox. I'm not sure how to stage the files though? Normally I would use S3.

src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerXMLInputFileTypeTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import edu.cornell.kfs.module.purap.batch.JaggaerXMLInputFileType;
+import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.kuali.kfs.sys.ConfigureContext;
+import org.kuali.kfs.sys.context.KualiIntegTestBase;
+import org.kuali.kfs.sys.context.SpringContext;
+
+@ConfigureContext
+public class JaggaerXMLInputFileTypeTest extends KualiIntegTestBase {
```
jhulslander, line 11: To avoid having to load up the entire KFS spring context, you could having this test class extend SpringEnabledMicroTestBase.  You should just need to add an @LoadSpringFile annotation to JaggaerXMLInputFileTypeTest and have the annotation point to an XML file that just has enough for this testing class needs.

jhulslander, line 11: Did you want to add tests for getAuthorPrincipalName() and getFileName() since those have more interesting code.

src/main/java/edu/cornell/kfs/module/purap/batch/JaggaerXMLInputFileType.java
```diff
@@ -0,0 +1,76 @@
...
+    protected DateTimeService dateTimeService;
+
+    @Override
+    public String getFileTypeIdentifier() {
+        return CUPurapConstants.JAGGAER_XML_INPUT_FILE_TYPE_IDENTIFIER;
+    }
+
+    @Override
+    public boolean validate(final Object parsedFileContents) {
+        return true;
```
jhulslander, line 27: I am not sure, just asking, should this validate method do some checking?

jhulslander (commented):
I found a few things to consider and left some comments.

I am thinking about confusion.  We have code that creates Jaggaer XML files in the "generate step" with a custom naming convention.  Those files are then picked up in the "up load" step and uses JaggaerXMLInputFileType to help find and read those tiles.  JaggaerXMLInputFileType seems to specify a naming convention that is not used in the "generate step".  The naming convention difference doesn't seem to cause a functional problem, it just seems to add confusion for someone new to the XmlBatchInputFileTypeBase framework.  I am not sure what to think about this yet.

src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerXMLInputFileTypeTest.java
ajd299: I added tests for those methods

ajd299:
@jhulslander I have tried to address your feedback. 

From my understanding, the user story is intended to remove the dependency on JAXBXmlBatchInputFileTypeBase for the Jaggaer portion. 
I will create 2 user stories to remove the dependency on JAXBXmlBatchInputFileTypeBase for Rass and CreateAcctDoc.

src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerXMLInputFileTypeTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+
+import java.io.File;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@LoadSpringFile("edu/cornell/kfs/module/purap/jaggaer/xml/cu-spring-jaggaer-test.xml")
+public class JaggaerXMLInputFileTypeTest extends SpringEnabledMicroTestBase {
```
cah292, line 15: Ideally, it would be best to consolidate this test class with the `JaggaerUploadFileServiceImplTest` class, especially since all or most of this class's assertions could be consolidated into 1 test case. The Spring-loading feature could also be used to initialize the Jaggaer file type instance of the other test class.

If you wanted to try combining the test classes, you would either need to create a JUnit 5 version of the Spring-loading feature, or create a subclass of `SpringEnabledMicroTestBase` that can act as a pseudo-test-extension instance but has fully custom control of its initialization and configuration. The former would be preferable if it's not too complex.

src/test/resources/edu/cornell/kfs/rass/batch/cu-spring-jaggaer-test.xml
cah292, line 1: Please relocate or remove the Jaggaer test XML Spring file that's in the RASS folder structure.

src/test/resources/edu/cornell/kfs/module/purap/jaggaer/xml/cu-spring-jaggaer-test.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://www.springframework.org/schema/p"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+        http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
+
+    <bean id="dateTimeService" class="edu.cornell.kfs.sys.service.impl.TestDateTimeServiceImpl"/>
+
+    <bean id="jaggaerXMLInputFileType"
```
cah292, line 9: To simplify the bean setup for these tests and to minimize bean duplication, I would suggest structuring the file as follows:

1. Add `<import>` elements for each non-unit-test Spring file that contains beans needed for your test.
2. Add this tag:
   ```xml
   <import resource="classpath:edu/cornell/kfs/sys/cu-spring-base-test-beans.xml"/>
   ```
3. (Optional) If you want to automatically set up a few common test overrides of base beans (including a test "dateTimeService" override), then add this tag:
   ```xml
   <import resource="classpath:edu/cornell/kfs/sys/cu-spring-sys-test.xml"/>
   ```
4. If you have other test-only Spring XML files that you want to import, add their imports after the ones from above.
5. If the needed beans contain property substitution placeholders (in `${propname}` format), add an override of the "propertyPlaceholderConfigurer" bean and merge in the property key-value pairs that you want to use. (See `cu-spring-rass-service-test.xml` for an example.)
6. Create new beans or override imported beans as needed.
7. To limit which beans will actually be loaded by the unit test, add an override of the "beanFilterPostProcessor" bean and merge in a set of bean names that you want the test to preserve. Make sure you specify `merge="true"` on the overridden set, since the parent bean already contains some entries for preserving the base test beans. (See `cu-spring-rass-service-test.xml` for an example.)

cah292 (changes requested):
I have some suggestions on how to reorganize the XML file types and unit tests.

src/test/resources/edu/cornell/kfs/module/purap/jaggaer/xml/cu-spring-jaggaer-test.xml
ajd299: All I need for the test is dateTimeService and jaggaerXMLInputFileType. Can I import the cu-spring-purap.xml file from the non-test resources directory?

cah292: Yes, you can import `cu-spring-purap.xml`, just make sure the import precedes those for the test XML files.

ajd299:
@cah292 Please review what I have so far

src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerXMLInputFileTypeTest.java
cah292: I see that the two test classes are still separate at the moment. I believe that's fine for now, given the separate test superclasses that are involved.

In the future, we should replace the `CuLocalServerTestBase` and `SpringEnabledMicroTestBase` superclasses with JUnit 5 extensions. That should simplify our unit test setup and also make it easier to consolidate these Jaggaer unit test classes. I can create a JIRA for further analysis.

src/main/java/edu/cornell/kfs/sys/batch/CuXmlBatchInputFileTypeBase.java
```diff
@@ -0,0 +1,61 @@
+package edu.cornell.kfs.sys.batch;
+
+import org.apache.commons.lang.StringUtils;
```
cah292, line 3: Please use lang3 StringUtils instead.

src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerXMLInputFileTypeTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String EXPECTED_AUTHOR_NAME_AJD299 = "ajd299";
+    private static final String EXAMPLE_FILENAME = "jaggaerSupplierUploadFile_ajd299_20231025_092052250.xml";
+    private static final String EXPECTED_FILENAME_PREFIX = "jaggaerSupplierUploadFile_ajd299_";
+    private static final String JAGGAER_STAGING_RELATIVE_PATH = "purap/jaggaer/xml";
+    private static final String PRINCIPAL_NAME_AJD299 = "ajd299";
+
+    private JaggaerXMLInputFileType jaggaerXMLInputFileType;
+
+    @Override
+    @BeforeEach
```
cah292, line 27: Because this test class is currently using JUnit-4-related classes and annotations, please use the `@Before` annotation instead for consistency, unless it's causing problems.

src/test/resources/edu/cornell/kfs/module/purap/jaggaer/xml/cu-spring-jaggaer-test.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://www.springframework.org/schema/p"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+        http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
+
+    <import resource="classpath:edu/cornell/kfs/sys/cu-spring-base-test-beans.xml"/>
+    <import resource="classpath:edu/cornell/kfs/sys/cu-spring-sys-test.xml"/>
+    <import resource="classpath:edu/cornell/kfs/module/purap/cu-spring-purap.xml"/>
```
cah292, line 9: To avoid potential bean override problems, I would suggest moving the `cu-spring-purap.xml` import above the other two.

cah292 (changes requested):
I noticed a few more minor items to be cleaned up, but overall the changes look good.

ajd299:
@cah292 Please see my latest changes to determine if this is ready for develop?

cah292 (approved):

PR #1522, opened 2023-11-20
KFSPTS-29953 Handle Concur errors while validate expense reports and travel requests

jhulslander:
This PR updates concur event notification processing to allow a results report to be generated if there are web service errors while trying to validate subsequent expense reports 

Please do not merge this PR yet, I noticed there is a problem running the event notification in non prod that needs to be fixed before we merge this code


jhulslander:
retest this please

PR #1526, opened 2023-12-01
KFSPTS-30289 limit Account Delegate multi values search results

db483:


src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AccountDelegate.xml
```diff
@@ -0,0 +1,33 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+  <bean parent="DataDictionaryBeanOverride">
+       <property name="beanName" value="AccountDelegate-lookupDefinition" />
+       <property name="fieldOverrides">
+           <list>
+               <bean parent="FieldOverrideForValueReplace">
+                   <property name="propertyName" value="multipleValuesResultSetLimit" />
+                   <property name="value" value="10000" />
```
cah292, line 28: Since you're just overriding the value of a single easy-to-reach property, and since the target bean follows KualiCo's parent bean convention, It might be better to just override the `AccountDelegate-lookupDefinition` bean directly and put the property override there.

Also, I'm wondering whether 10000 is small enough. I know that the regular Account Delegate maintenance document allows up to 10000 changes, but that limit represents the doc's number of delegates multiplied by the doc's number of accounts (meaning the document itself likely won't display all 10000 of the resulting changes). Would 1000 be a more reasonable limit, like what our custom `AccountGlobalSearch` lookup uses? Or do the functionals absolutely need the 10000 limit instead?

cah292 (changes requested):
I had some suggestion for cleaning this up, and I had some more questions concerning the proposed result set limit.

src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/AccountDelegate.xml
db483: Cathy has requested that the value be 10000 specifically.
Looking into overriding the lookupDefinition directly.

db483:
This is ready for review again. I also noticed that the PR had the wrong jira number in the title. Please ensure the commit message references KFSPTS-30289 and not KFSPTS-30283. Thank you!

cah292 (approved):
Changes look good, unit tests pass and my concerns have been addressed. We can make follow-up changes as needed if the 10000 limit ends up being too high.

Merging.

PR #1530, opened 2023-12-07
KFSPTS-30381 remove routing for GADI doc type

db483:


nkimble (approved):
Routing looks appropriate. Merging to develop for record of this customization. Workflow will also need to be ingested.

PR #1531, opened 2023-12-12
KFSPTS-30410 fix DV null pointer exception when changing account with different FO

db483:


nkimble (approved):
Changes look good. Unit tests have passed. Merging to develop.

PR #1533, opened 2023-12-14
KFSPTS-30485 fix labor scrubber null pointer exception

db483:


cah292 (approved):

cah292:
One other note: The messageBuilderService NPE bug appears to be present in the latest financials code as well. We should report this back to KualiCo if they're not aware of it.

PR #1534, opened 2023-12-14
KFSPTS-28701 Upgrade to the 1/25/2023 Version of Financials

ajd299:


ajd299:
retest this please

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPurapServiceImpl.java
```diff
@@ -253,11 +253,18 @@ public KualiDecimal getApoLimit(final PurchasingDocument purchasingDocument) {
...
         if (ObjectUtils.isNull(purchaseOrderTotalLimit)) {
-            String defaultLimitParameterName = shouldUseFederalApoLimit(purchasingDocument)
-                    ? CUPurapParameterConstants.AUTOMATIC_FEDERAL_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT
-                    : PurapParameterConstants.AUTOMATIC_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT;
-            
-            final String defaultLimit = parameterService.getParameterValueAsString(RequisitionDocument.class, defaultLimitParameterName);
+            String defaultLimit = "";
+
+            if (shouldUseFederalApoLimit(purchasingDocument)) {
+                defaultLimit = parameterService.getParameterValueAsString(RequisitionDocument.class, CUPurapParameterConstants.AUTOMATIC_FEDERAL_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT);
```
cah292, line 259: Since the related base `AUTOMATIC_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT` parameter is being renamed to `APO_LIMIT` with a `PurchaseOrder` component, we may want to consider renaming this CU-specific parameter to follow a similar convention. It might be worth asking the functionals if they think such a change to our customization makes sense.

src/main/java/edu/cornell/kfs/module/purap/document/CuPurchaseOrderAmendmentDocument.java
```diff
@@ -121,7 +121,7 @@ public Set<Person> getAllPriorApprovers() throws WorkflowException {
     // KFSUPGRADE-339
     protected boolean isContractManagementReviewRequired() {
         final ParameterService parameterService = SpringContext.getBean(ParameterService.class);
-        final KualiDecimal automaticPurchaseOrderDefaultLimit = new KualiDecimal(parameterService.getParameterValueAsString(RequisitionDocument.class, PurapParameterConstants.AUTOMATIC_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT));
+        final KualiDecimal automaticPurchaseOrderDefaultLimit = new KualiDecimal(parameterService.getParameterValueAsString(RequisitionDocument.class, PurapParameterConstants.APO_LIMIT));
```
cah292, line 124: Should this be modified to use the namespace-component-parameter variant instead of the classname-parameter variant? I thought this parameter was updated to use the "PurchaseOrder" component.

src/test/java/edu/cornell/kfs/module/purap/document/service/impl/CuPurapServiceImplTest.java
```diff
@@ -241,7 +241,13 @@ private ParameterService buildMockParameterService() {
         Parameter costSourceApoOverrideParameter = buildParameterForCostSourceApoOverride();
         
         when(parameterService.getParameterValueAsString(
-                RequisitionDocument.class, PurapParameterConstants.AUTOMATIC_PURCHASE_ORDER_DEFAULT_LIMIT_AMOUNT))
+                RequisitionDocument.class, PurapParameterConstants.APO_LIMIT))
```
cah292, line 244: Once my other comments related to `APO_LIMIT` have been addressed, please double-check this method's mock setup for `APO_LIMIT`.

cah292 (changes requested):
I've finished my initial review. The changes look good overall, but there are some items that may need further revisions.

There's also one item I noticed that affects a file not currently updated by this PR. I'll put the details in a separate comment.

cah292 (changes requested):
In addition to my other comments, I noticed one other item that may need attention:

In the `CuDisbursementVoucherDocumentPreRules` class, please check if revisions are needed to the `checkWireTransferTabState()` method. The base class appears to have made some updates to that method.

ajd299:
@cah292 Please see my latest commits

cah292:
retest this please

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPurapServiceImpl.java
cah292: I saw Cathy's comment on the user story, where it sounds like she would prefer to keep the parameter name as-is if there are no objections from other functionals. Given that, I'm fine with leaving this custom parameter name as-is.

cah292 (commented):
There's one minor item that you may want to double-check, but other than that the code changes look good.

pom.xml
```diff
@@ -537,10 +537,6 @@
                     <groupId>io.netty</groupId>
                     <artifactId>netty-common</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>io.netty</groupId>
-                    <artifactId>netty-transport</artifactId>
```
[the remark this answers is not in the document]
ajd299: When I checked, there were indeed conflicts between minor versions of netty-transport. Restoring the exclusion seems to fix this.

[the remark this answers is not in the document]
cah292: Thanks for checking on the netty JAR versions.

At some point, we should upgrade to the newer netty JARs, given the details on the FINP-9194 ticket. Since we seem to be transitively pulling in those JARs from the `secretsmanager` dependency, we should investigate upgrading that JAR in the near future, so that we'll automatically pull in the appropriate netty JAR versions.

I'll create a follow-up user story to look into upgrading to a newer `secretsmanager` version, to avoid scope creep for this PR's current changes.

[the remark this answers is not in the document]
cah292: I've created KFSPTS-30649 to handle additional cleanup of the `secretsmanager` and `netty` dependencies. We'll address those dependencies as part of that ticket instead, so that we can keep the 01/25 patch moving forward.

cah292 (approved):
Changes look good, unit tests pass, my concerns have been addressed and the related PRs are also ready. Merging.

PR #1535, opened 2023-12-19
KFSPTS-20442 Add create-acct-doc duplicate file checks

cah292:

Also, I'm waiting for functional feedback on some of the new messages introduced by these changes. Please don't merge until after I hear back about any needed wording changes.

This PR allows the Create Accounting Documents Job to prevent the processing of files that were already handled in a previous run; such functionality is controllable by a new parameter. I believe I've implemented the requirements noted in the user story, but please let me know if I missed anything. (Adding the new table to the Purge Job will be handled in a follow-up user story, as noted in the ticket.)

I also fixed a pre-existing bug that was preventing certain error information from being displayed properly in the reports. (I needed that functionality for the duplicate file error reporting.) This partially fixes the KFSPTS-30149 issue, but that ticket still needs more follow-up work.

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
@@ -3,6 +3,7 @@ package edu.cornell.kfs.fp.batch.service.impl;
 import java.sql.Timestamp;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
@@ -13,6 +14,7 @@ import java.util.stream.Stream;
 
 import javax.mail.internet.InternetAddress;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -24,10 +26,12 @@ import org.kuali.kfs.krad.document.Document;
 import org.kuali.kfs.krad.exception.ValidationException;
 import org.kuali.kfs.krad.service.BusinessObjectService;
 import org.kuali.kfs.krad.service.DocumentService;
+import org.kuali.kfs.krad.service.SequenceAccessorService;
 import org.kuali.kfs.krad.util.ErrorMessage;
 import org.kuali.kfs.krad.util.GlobalVariables;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.sys.KFSConstants;
+import org.kuali.kfs.sys.KFSPropertyConstants;
 import org.kuali.kfs.sys.batch.BatchInputFileType;
 import org.kuali.kfs.sys.batch.service.BatchInputFileService;
 import org.kuali.kfs.sys.context.SpringContext;
@@ -36,6 +40,8 @@ import org.kuali.kfs.sys.service.FileStorageService;
 import org.springframework.util.AutoPopulatingList;
 
 import edu.cornell.kfs.fp.CuFPConstants;
+import edu.cornell.kfs.fp.CuFPConstants.CreateAccountingDocumentConstants;
+import edu.cornell.kfs.fp.CuFPConstants.CreateAccountingDocumentConstants.FileEntryFieldLengths;
 import edu.cornell.kfs.fp.CuFPKeyConstants;
 import edu.cornell.kfs.fp.CuFPParameterConstants;
 import edu.cornell.kfs.fp.batch.CreateAccountingDocumentReportItem;
@@ -52,9 +58,7 @@ import edu.cornell.kfs.sys.CUKFSConstants;
 import edu.cornell.kfs.sys.util.LoadFileUtils;
 
 public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocumentService {
-	private static final Logger LOG = LogManager.getLogger(CreateAccountingDocumentServiceImpl.class);
-
-	private static final int MAX_OVERVIEW_LENGTH_FOR_FILE_ENTRY = 250;
+    private static final Logger LOG = LogManager.getLogger(CreateAccountingDocumentServiceImpl.class);
 
     protected BatchInputFileService batchInputFileService;
     protected BatchInputFileType accountingDocumentBatchInputFileType;
@@ -66,6 +70,7 @@ public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocu
     protected CreateAccountingDocumentValidationService createAccountingDocumentValidationService;
     protected BusinessObjectService businessObjectService;
     protected DateTimeService dateTimeService;
+    protected SequenceAccessorService sequenceAccessorService;
 
     @Override
     public boolean createAccountingDocumentsFromXml() {
@@ -95,7 +100,11 @@ public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocu
             if (ObjectUtils.isNotNull(fileEntry)) {
                 reportItem.setDuplicateFile(true);
                 if (shouldExcludePreviouslyProcessedFiles()) {
-                    throw new DuplicateCreateAccountingDocumentFileException(fileEntry);
+                    LOG.error("processAccountingDocumentFromXml: File {} was already processed by a previous run "
+                            + "of the job; it will be excluded from the current run.  File details: {}",
+                            fileEntry.getFileName(), fileEntry);
+                    throw new DuplicateCreateAccountingDocumentFileException(
+                            "File " + fileEntry.getFileName() + " was already processed by a prior run");
                 } else {
                     LOG.warn("processAccountingDocumentFromXml: The file {} was already processed by a prior run, "
                             + "but the job has been configured to reprocess it anyway.", fileEntry.getFileName());
@@ -105,7 +114,9 @@ public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocu
             AccountingXmlDocumentListWrapper accountingXmlDocuments = unmarshalAccountingDocumentFile(fileName);
             reportItem.setReportEmailAddress(accountingXmlDocuments.getReportEmail());
             reportItem.setFileOverview(accountingXmlDocuments.getOverview());
-            createOrUpdateFileEntry(accountingXmlDocuments, fileEntry, fileName);
+            if (!reportItem.isDuplicateFile()) {
+                createFileEntry(accountingXmlDocuments, fileName);
+            }
             
             if (createAccountingDocumentValidationService.isValidXmlFileHeaderData(accountingXmlDocuments, reportItem)) {
                 int documentCount = accountingXmlDocuments.getDocuments().size();
@@ -131,13 +142,10 @@ public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocu
         } catch (DuplicateCreateAccountingDocumentFileException e) {
             reportItem.setXmlSuccessfullyLoaded(false);
             reportItem.setNonBusinessRuleFailure(true);
-            CreateAccountingDocumentFileEntry fileEntry = e.getExistingEntry();
-            LOG.error("processAccountingDocumentFromXml: File {} was already processed by a previous run of the job; "
-                    + "it will be excluded from the current run.  File details: {}",
-                    fileEntry.getFileName(), fileEntry);
             String duplicateFileMessage = configurationService.getPropertyValueAsString(
                     CuFPKeyConstants.REPORT_CREATE_ACCOUNTING_DOCUMENT_DUPLICATE_FILE_ERROR);
             reportItem.setReportItemMessage(duplicateFileMessage);
+            LOG.error("processAccountingDocumentFromXml: Error processing accounting document XML file", e);
         } catch (Exception e) {
             reportItem.setXmlSuccessfullyLoaded(false);
             reportItem.setNonBusinessRuleFailure(true);
@@ -168,7 +176,10 @@ public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocu
 
     protected CreateAccountingDocumentFileEntry findExistingFileEntry(String fileName) {
         String cleanedFileName = getCleanedFileName(fileName);
-        return businessObjectService.findBySinglePrimaryKey(CreateAccountingDocumentFileEntry.class, cleanedFileName);
+        Map<String, String> criteria = Map.of(KFSPropertyConstants.FILE_NAME, cleanedFileName);
+        Collection<CreateAccountingDocumentFileEntry> fileEntries = businessObjectService.findMatching(
+                CreateAccountingDocumentFileEntry.class, criteria);
+        return CollectionUtils.isNotEmpty(fileEntries) ? fileEntries.iterator().next() : null;
     }
 
     protected String getCleanedFileName(String fileName) {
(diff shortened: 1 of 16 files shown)


src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -236,14 +323,28 @@ protected void removeDoneFileQuietly(String dataFileName) {
...
         if (reportItem.isXmlSuccessfullyLoaded()) {
-            toAddress = reportItem.getReportEmailAddress();
+            primaryToAddress = reportItem.getReportEmailAddress();
+        } else {
+            primaryToAddress = fromAddress;
+        }
+        
+        List<String> toAddresses;
+        if (!reportItem.isXmlSuccessfullyLoaded() && reportItem.isDuplicateFile()
+                && shouldExcludePreviouslyProcessedFiles()) {
```
jhulslander, line 336: Should this boolean expression just be reportItem.isDuplicateFile()?  We only send email in production, and we will always prevent duplicate in prod.  Also, if duplicates are are found, XML won't fully be loaded, correct?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -258,6 +359,29 @@ protected String getCreateAccountingDocumentReportEmailAddress() {
                 CuFPParameterConstants.CreateAccountingDocumentService.CREATE_ACCT_DOC_REPORT_EMAIL_ADDRESS);
     }
     
+    protected String getDuplicateFileReportEmailAddress() {
```
jhulslander, line 362: I think we want DUPLICATE_FILE_REPORT_EMAIL_ADDRESS to have a list of addresses.  We want the developers to know because there may be something technically going awry, and the functionals (FIS / Cathy) would also want to know

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -258,6 +359,29 @@ protected String getCreateAccountingDocumentReportEmailAddress() {
...
+            String reportEmail = accountingXmlDocuments.getReportEmail();
+            if (StringUtils.isNotBlank(reportEmail)) {
+                InternetAddress reportInternetAddress = new InternetAddress(reportEmail);
+                reportInternetAddress.validate();
+                return Optional.of(reportEmail);
+            } else {
+                return Optional.empty();
+            }
+        } catch (Exception e) {
+            LOG.error("getReportEmailFromFileIfPossible, Could not obtain and validate report email from file", e);
```
jhulslander, line 380: I am thinking about the situation where the XML file is fine, and fully loaded the KFS documents, but the report email address is empty or not valid.  I think we need to alert the developers that we received a file without a report email address.

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
```diff
@@ -1327,4 +1327,15 @@
 		<field-descriptor name="subAccountNumber" column="SUB_ACCT_NBR" jdbc-type="VARCHAR" />
 	</class-descriptor>
 
+    <class-descriptor class="edu.cornell.kfs.fp.businessobject.CreateAccountingDocumentFileEntry" table="CU_FP_CRTE_ACCT_DOC_FIL_T">
+        <field-descriptor name="fileName" column="FIL_NM" jdbc-type="VARCHAR" primarykey="true" index="true" />
```
jhulslander, line 1331: sorry I didn't mention it before, but it may be nice to include sequence ID field as the primary key, and make FIL_NM a unique constraint.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -118,14 +155,64 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
...
+        CreateAccountingDocumentFileEntry fileEntry;
+        if (ObjectUtils.isNull(existingEntry)) {
+            fileEntry = new CreateAccountingDocumentFileEntry();
+            fileEntry.setFileName(getCleanedFileName(fileName));
+        } else {
+            fileEntry = existingEntry;
+        }
+        fileEntry.setFileCreatedDate(new Timestamp(accountingXmlDocuments.getCreateDate().getTime()));
+        fileEntry.setFileProcessedDate(dateTimeService.getCurrentTimestamp());
+        fileEntry.setReportEmailAddress(accountingXmlDocuments.getReportEmail());
```
jhulslander, line 209: similar to the length check on the overview, I think we want a length check on the email and file name

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -118,14 +155,64 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
...
+                CuFPParameterConstants.CreateAccountingDocumentService.DUPLICATE_FILE_CHECK_IND);
+    }
+
+    private AccountingXmlDocumentListWrapper unmarshalAccountingDocumentFile(String fileName) {
+        byte[] fileData = LoadFileUtils.safelyLoadFileBytes(fileName);
+        return (AccountingXmlDocumentListWrapper) batchInputFileService.parse(
+                accountingDocumentBatchInputFileType, fileData);
+    }
+
+    private void createOrUpdateFileEntry(AccountingXmlDocumentListWrapper accountingXmlDocuments,
```
jhulslander, line 198: I find this convention a bit confusing, and perhaps error prone with potential future updates.  If an XML was previously loaded, we don't want to update the values for existingEntry.  This naming convention makes it sound like that is possible.  Can this be refactored to help with that?

jhulslander (changes requested):
Changes are looking good, but I found a few items I think need to be looked at

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -39,11 +46,16 @@
...
 import edu.cornell.kfs.fp.batch.xml.AccountingXmlDocumentListWrapper;
+import edu.cornell.kfs.fp.businessobject.CreateAccountingDocumentFileEntry;
+import edu.cornell.kfs.fp.exception.DuplicateCreateAccountingDocumentFileException;
+import edu.cornell.kfs.sys.CUKFSConstants;
 import edu.cornell.kfs.sys.util.LoadFileUtils;
 
 public class CreateAccountingDocumentServiceImpl implements CreateAccountingDocumentService {
 	private static final Logger LOG = LogManager.getLogger(CreateAccountingDocumentServiceImpl.class);
 
+	private static final int MAX_OVERVIEW_LENGTH_FOR_FILE_ENTRY = 250;
```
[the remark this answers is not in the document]
cah292: I'll add max length checks on the other fields accordingly.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -75,10 +91,21 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
         String headerValidationErrorMessage = KFSConstants.EMPTY_STRING;
         try {
             LOG.info("processAccountingDocumentFromXml: Started processing accounting document XML file: " + fileName);
-            byte[] fileData = LoadFileUtils.safelyLoadFileBytes(fileName);
-            AccountingXmlDocumentListWrapper accountingXmlDocuments = (AccountingXmlDocumentListWrapper) batchInputFileService.parse(accountingDocumentBatchInputFileType, fileData);
+            CreateAccountingDocumentFileEntry fileEntry = findExistingFileEntry(fileName);
+            if (ObjectUtils.isNotNull(fileEntry)) {
+                reportItem.setDuplicateFile(true);
```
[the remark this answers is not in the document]
cah292: There's some logging down in the "catch" block, but I'll see if I can move most of that logging over into this block instead.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -101,6 +128,16 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
             }
             LOG.info("processAccountingDocumentFromXml: Finished processing accounting document XML file: " + fileName);
             
+        } catch (DuplicateCreateAccountingDocumentFileException e) {
+            reportItem.setXmlSuccessfullyLoaded(false);
+            reportItem.setNonBusinessRuleFailure(true);
+            CreateAccountingDocumentFileEntry fileEntry = e.getExistingEntry();
+            LOG.error("processAccountingDocumentFromXml: File {} was already processed by a previous run of the job; "
+                    + "it will be excluded from the current run.  File details: {}",
+                    fileEntry.getFileName(), fileEntry);
```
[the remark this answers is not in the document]
cah292: Adding the fileEntry to the log statement will cause its toString() content to be added in the appropriate placeholder. I purposely overrode the BO's toString() method so that it would print the BO's properties, in case we needed that information in the logs to assist with troubleshooting.

If you think we shouldn't add that info to the logs, please let me know.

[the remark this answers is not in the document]
jhulslander: That makes sense, and I agree, thank you

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: I think we need at least the `isDuplicateFile()` and `shouldExcludePreviouslyProcessedFiles()` checks here. If the parameter is configured to allow duplicate processing, the report item will be marked as a duplicate but KFS will process it anyway. I don't think we want the extra duplicate-file recipients to be notified if KFS has been configured to allow such duplication. If you think otherwise, please let me know.

I believe I could at least simplify this to remove the `!reportItem.isXmlSuccessfullyLoaded()` check.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: The enclosing `getReportEmailFromFileIfPossible()` method is only meant to be used in cases where the file was rejected due to the duplicate-file check. We already have existing code in `CreateAccountingDocumentValidationServiceImpl` that will validate the correctness of the report email, if the file was not rejected by the duplication check. An email validation error at that phase will cause a header-level validation failure in the file, and will cause the error report to be sent to the email specified by `CREATE_ACCT_DOC_REPORT_EMAIL_ADDRESS`.

If you think we need to adjust the existing code so that such email error reports are also sent to a developer EGA, please let me know.

src/main/resources/edu/cornell/kfs/fp/cu-ojb-fp.xml
cah292: I'll add the suggested columns and constraints.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: I'll add the extra length checks accordingly.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: I can modify this to only support creating entries and not updating them. I wasn't sure if we needed to allow for updating existing entries when duplication is permitted, but if you're fine with the file entries only reflecting the first processed file of a given name (as opposed to the latest file), then I'm fine with making this a create-only method.

cah292:
I think I've updated the areas that needed further changes. This is ready for review again.

I have a question out to the functionals as to whether any extra emails need to be included as duplicate-file-error recipients. Once I get feedback on that and make any necessary follow-up changes, this can be merged once it has been reviewed and approved.

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
jhulslander: I don't think we want to adjust previously saved records, they should be a historic record of a processed file

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
jhulslander: that should be fine, thanks

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -118,14 +163,66 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
...
+                accountingDocumentBatchInputFileType, fileData);
+    }
+
+    private void createFileEntry(AccountingXmlDocumentListWrapper accountingXmlDocuments, String fileName) {
+        String cleanedFileName = getCleanedFileName(fileName);
+        Long nextFileId = sequenceAccessorService.getNextAvailableSequenceNumber(
+                CreateAccountingDocumentConstants.FILE_ID_SEQUENCE_NAME);
+        CreateAccountingDocumentFileEntry fileEntry = new CreateAccountingDocumentFileEntry();
+        fileEntry.setFileId(nextFileId.toString());
+        fileEntry.setFileName(StringUtils.left(cleanedFileName, FileEntryFieldLengths.FILE_NAME));
```
jhulslander, line 215: I think the file size clean should be included in getCleanedFileName() so the search will find entries of file names that had to be trimmed for length

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -118,14 +163,66 @@ protected void processAccountingDocumentFromXml(String fileName, CreateAccountin
...
+
+    private AccountingXmlDocumentListWrapper unmarshalAccountingDocumentFile(String fileName) {
+        byte[] fileData = LoadFileUtils.safelyLoadFileBytes(fileName);
+        return (AccountingXmlDocumentListWrapper) batchInputFileService.parse(
+                accountingDocumentBatchInputFileType, fileData);
+    }
+
+    private void createFileEntry(AccountingXmlDocumentListWrapper accountingXmlDocuments, String fileName) {
+        String cleanedFileName = getCleanedFileName(fileName);
+        Long nextFileId = sequenceAccessorService.getNextAvailableSequenceNumber(
```
jhulslander, line 211: Is this needed?  I thought BusinessObjectService / OJB would set this if not set.

jhulslander (commented):
Changes are looking good, i found a few more things to consider

jhulslander:
retest this please

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: I'll refactor the cleaning and truncating of the filename accordingly.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
cah292: I'll see if I can set this up through OJB metadata instead of the SequenceAccessorService.

jhulslander: I don't think you need to set the fileId value, won't OJB set it atomically from the sequence?

cah292: Yes, I meant to say that I'm changing this to use just the OJB XML for this setup, and that I'll remove the manual ID-populating code (at least on the non-unit-test side). Sorry for the confusion.

cah292:
I've committed some more changes to address the latest feedback. (A small change has also been committed to the SQL PR.) This is ready for review again.

Also, I've received the needed functional feedback on my questions, so this is now ready to be merged once both PRs have been approved.

jhulslander (approved):
code changes look good, as do SQL changes, unit tests pass, I shall merge

PR #1536, opened 2023-12-21
KFSPTS-29567 fix failing unit test after utest database refresh

jhulslander:


cah292 (approved):
Changes look good and unit tests passed. Merging.

PR #1538, opened 2023-12-21
KFSPTS-30149 Fix create-acct-doc report formatting

cah292:
This PR makes adjustments to how some of the error reports are formatted:

* If a document within the file fails for certain non-business-rule reasons (such as a not-authorized exception when creating the KFS doc), the report will now use the regular layout instead of one of the total-file-failure layout. The problematic docs will be noted in the appropriate sections.
* If a file fails for certain header validation reasons (such as an invalid report email address), the related total-file-failure layout will now attempt to avoid outputting the error message twice. The #1535 changes fixed a bug that was preventing some file-level error messages from being printed to the reports, but this had the side effect of potentially printing some of the messages twice. (The upstream code prepares a longer version of the error message that contains the original message; that's why this PR is using a `containsIgnoreCase()` check in the appropriate spot.)

Please let me know if you think adjustments to the upstream code are also needed, to try and prevent the configuration from getting into these unexpected situations.


jhulslander (approved):
I think this code change is OK as is.  I think it will overly complicate this PR to try and prevent the duplication of error messages  before we get to the report wrier.
I shall merge this PR

PR #1539, opened 2024-01-04
KFSPTS-30631 Fix handling of bad Create-Acct-Doc create dates

cah292:
The #1535 Create-Acct-Doc changes incorrectly assumed that an exception would be thrown by the XML parser if the file's create date could not be converted properly; it's actually being set to null on the POJO in such cases. This PR adjusts those prior changes so that if such a null create date is encountered, it will be logged as a warning and then it will be ignored, thus avoiding unexpected NullPointerExceptions when preparing the file entry BOs. (The Create Accounting Documents Job does not need that particular date field for its processing, so it's safe to ignore invalid values  there.)


jhulslander (commented):
Changes look good, I will merge one the unit tests pass

jhulslander (approved):
changes look fine, unit tests pass, I shall merge

PR #1542, opened 2024-01-15
KFSPTS-30711 Fix general ledger entries for stopped checks

db483:


src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
```diff
@@ -175,9 +175,14 @@ private void populatePaymentGeneralLedgerPendingEntry(
             glPendingTransaction.setAccountNumber(paymentAccountDetail.getAccountNbr());
             glPendingTransaction.setSubAccountNumber(paymentAccountDetail.getSubAccountNbr());
             glPendingTransaction.setChartOfAccountsCode(paymentAccountDetail.getFinChartCode());
-
-            glPendingTransaction.setFinancialDocumentTypeCode(paymentAccountDetail.getPaymentDetail().getFinancialDocumentTypeCode());
-
+            
+            if (StringUtils.equals(paymentGroup.getDisbursementType().getCode(), PdpConstants.DisbursementTypeCodes.ACH)) {
+                glPendingTransaction.setFinancialDocumentTypeCode(state.documentTypeForDisbursementType(PdpConstants.DisbursementTypeCodes.ACH));
```
jhulslander, line 180: The previous set functions were
glPendingTransaction.setFinancialDocumentTypeCode(achFdocTypeCode);
or
 glPendingTransaction.setFinancialDocumentTypeCode(checkFdocTypeCod);

I am not sure if achFdocTypeCode or checkFdocTypeCod  exist any longer, but you could use the PDP constant.
Why do you need to reference the state object?

jhulslander (commented):
I had a question, and I think the PR description needs to be updated so the commit message is meaningful

src/main/java/edu/cornell/kfs/pdp/service/impl/CuPendingTransactionServiceImpl.java
db483: Good question! KualiCo added the GeneratePdpGlpeState object with the 1/4 upgrade and this has methods that will get the doc type for the given disbursement type. Different maps are used for the different cases like stop, stale cancelled, reissued, etc. This method was updated to follow base code changes.

jhulslander: Thanks, I shall merge

jhulslander (approved):
my question was answered, the unit tests pass, I shall merge

PR #1543, opened 2024-01-17
KFSPTS-30681 add logging for check, ach, and bundled ach file names

jhulslander:
do NOT merge this PR, it is just needed for analysis at this point

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
@@ -307,7 +307,7 @@ public class CuExtractPaymentServiceImpl extends ExtractPaymentServiceImpl {
     protected void writeExtractCheckFile(
             final PaymentStatus extractedStatus, final PaymentProcess p, String filename,
             final Integer processId) {
-        LOG.info("writeExtractCheckFile, writing check file name {}", filename);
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -306,6 +307,7 @@ protected void writeExtractBundledAchFile(
     protected void writeExtractCheckFile(
             final PaymentStatus extractedStatus, final PaymentProcess p, String filename,
             final Integer processId) {
+        LOG.info("writeExtractCheckFile, writing check file name {}", filename);
```
nkimble, line 310: I would add the processId to this log message

jhulslander:
I added the process ID to the check log entry.  unfortunately the ACH log functions don't have the process ID.

cah292 (commented):
Do you need to add extra logging to `Iso20022FormatExtractor` as well?

jhulslander:
@cah292 what did you have in mind?

cah292:
@jhulslander In the `Iso20022FormatExtractor`, you could add some logging to the `extractAchs()` and `extractChecks()` methods, specifically in the sections that iterate over the `PaymentProcess` collections. Those objects' primary keys are the Process IDs, so you could log those there. You may also want to log the generated filenames after the `determineFilename()` invocations (or within that method itself).

Alternatively, most (if not all) of this info is being printed to the logs at the debug level, so you could just configure `Iso20022FormatExtractor` for debug-level logging.

PR #1554, opened 2024-02-06
KFSPTS-30931 update PMW to gracefully fail when payment works endpoints have exceptions

jhulslander:
The payment works integration is currently working again, payment works reverted a change they implemented that caused our problem.

The change in this PR should allow the PMW batch jobs to fail gracefully when an API error happens again.  if an API error happens again, this change should make it so a batch container restart won't be required to run a job again.  The functionals and Sandy agreed that we don't want to purse a more complicated solution at this time.


db483 (approved):
Code changes look good.

PR #1562, opened 2024-02-20
KFSPTS-28702 Upgrade to 01/29/2023 version of financials (Revised)

cah292:
This is a modified version of the #1559 changes that attempts to simplify part of the file overlay process. If necessary, we can discard this PR and stick with the original #1559 PR instead.

There were several files in this upgrade (including a few large ones) that required minor but crucial changes related to the Person masking functionality. To minimize the amount of overlays that we need to add to the codebase, I've been experimenting with an alternative process that will extract the appropriate base financials files and then apply patch files to them. This greatly reduces the amount of base financials files in our codebase since it only requires our project to store appropriate .patch files, rather than having to include the financials files directly.

Note that for large or complex file customizations, we can (and should) still use our strategy of copying financials files into our project and modifying them there. This new patching feature is primarily meant for relatively minor changes.

The following commit shows the differences between  the #1559 changes and this PR's changes: https://github.com/CU-CommunityApps/cu-kfs/commit/6f4b1067ec102a5e558f53abf843362ee1cca206

If you think this type of patching is too complex to introduce at this time, please let me know. (We can also discuss this further at future team meetings.) If we decide against this patching strategy, I'll need to make a few follow-up changes to the #1559 PR, since I discovered and fixed a couple minor upgrade issues while preparing the .patch files.


cah292:
Upon further consideration, I think it makes more sense to set up this PR so that it merges into the #1559 PR's branch instead of into "develop" directly. That should make it easier to distinguish between the regular 01/29/2023 financials patch changes and the proposed enhancements to it.

I'll close this PR and open a new one accordingly when ready.

dp65:
Regarding your Note, "Jenkins was having difficulty finding the Unix patch command". If this was new Jenkins, we could probably have installed that for you as part of our AMI build. We add in other unix tools that are not part of the base OS, so this would be just one more. However, using the "git apply" is probably better because it only uses tools that are already installed and available, as long as that's not too burdensome to you.

PR #1564, opened 2024-02-21
KFSPTS-28702 Enhance upgrade to 01/29/2023 version of financials

cah292:
This PR contains some optional enhancements to the 01/29/2023 financials upgrade. It will merge the changes into the branch used by the #1559 PR, rather than merging into the "develop" branch directly. If we decide that these enhancements should not move forward, then we can simply close this PR without affecting the main 01/29/2023 upgrade PR.

Our 01/29/2023 upgrade involved making some minor but crucial overlays to various code and config files, especially with regards to the masking of Person data fields. Our current process is limited to manually copying the files into our cu-kfs project and modifying them there, even if only 1 or 2 lines need modification. To minimize the number of base files included directly in our project, this PR introduces an alternative process that relies upon patch files to automatically extract the files within the JARs/WARs, apply the patches to them, and include them in the WAR overlay.

This PR also takes the 01/29-related files that were the best candidates for the patching setup, and switches them over to the new patching method instead. Most of them were overlays; one was an extension that was attempting to avoid an overlay (but has now been moved to an auto-patched overlay here).

Note that for larger or more complex changes, we can and should still use our existing manual-copy-and-edit overlay process. This patching feature is meant for situations where a base financials file (especially a large one) must be overlayed but only needs minor changes.

Since these changes required modifying the "maven-dependency-plugin" configuration, I also adjusted the existing portion of that to clarify its current intent. It was previously used for unpacking Cynergy or KFS overlays of Rice files, but now it's only being used to work around a potential Data Dictionary issue; details are in the POM's comments. If we decide to close this PR without merging it, then we should create a separate user story to at least do the "maven-dependency-plugin" cleanup at a future date.

Also, I had difficulty getting the "maven-patch-plugin" to work with Jenkins, which is why I switched this feature over to using the "git apply" command (invoked via the "maven-antrun-plugin").

Please let me know if you have questions about how this setup works. I can also discuss it at one of our team meetings if that would help.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,15 +52,18 @@
         <validation-api.version>2.0.1.Final</validation-api.version>
         <xapool.version>1.5.0-patch7</xapool.version>
         <maven.compiler.version>3.10.1</maven.compiler.version>
+        <maven.patch.version>1.2</maven.patch.version>
         <project.java.version>1.11</project.java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <unpack.includes>META-INF/**,org/kuali/kfs/kim/impl/**/*.xml</unpack.includes>
-        <unpack.meta.excludes>META-INF/maven/**,META-INF/MANIFEST.MF,META-INF/*weaving.xml</unpack.meta.excludes>
-        <unpack.excludes>${unpack.meta.excludes}</unpack.excludes>
-        <unpack.overlays.directory>${project.build.directory}/unpacked-overlays</unpack.overlays.directory>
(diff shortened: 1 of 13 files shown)


cah292:
NOTE: Since this PR's target branch was recently rebased, I went ahead and rebased its source branch as well. There were no merge conflicts, but I did end up performing an interactive rebase to avoid re-committing the target branch's older versions of its 01/29 upgrade changes. If you notice any unexpected issues as a result of this rebasing, please let me know.

cah292:
retest this please

cah292:
I think I accidentally messed up the commit history on this branch when I rebased it. (I had to take different steps than usual when rebasing it since its target branch was also rebased.) I'll create a new branch for this instead.

PR #1566, opened 2024-02-23
KFSPTS-31070 Upgrade to latest Spring 5.3.x version

cah292:
There are PRs for this in cu-kfs, cornell_customizations, cynergy-server-customizations and cynergy-standalone. Please make sure all four are good to go before merging.

I also had a quick question about the scope of the changes. I recently noticed that the cu-kfs WAR is using two different versions of the Log4J dependencies. I currently have the 01/29/2023 financials upgrade set up to fix that, but I'm wondering if it would make sense to bundle that fix with this dependency-updating PR. Please let me know your thoughts on that. (If we do decide to bundle in the Log4J version fix, it might make sense to update the Cynergy-side Log4J versions as well.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
         <powermock.version>2.0.7</powermock.version>
         <pw-iso20022.version>SRU2021-9.2.7</pw-iso20022.version>
         <squiggly-filter-jackson.version>1.3.18</squiggly-filter-jackson.version>
-        <spring.version>5.3.27</spring.version>
+        <spring.version>5.3.32</spring.version>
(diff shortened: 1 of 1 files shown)


db483:
@cah292 The changes look good and I agree that we should update the spring version.
Regarding the log4j version fix, I would suggest that we have a separate user story for that, so that we keep the changes on a separate commit. We can move both the log4j and the spring update user stories through dev and test at the same time so that the functionals only have to do one regression testing. Thank you.

ajd299 (approved):
Changes look good, merging

PR #1569, opened 2024-02-29
KFSPTS-30687: Include data file name in report file name for Accounting Document Batch Job.

nkimble:


nkimble:
Our customization has report writer attributes that hold data that is not consistent with the standard file path, name, and extension values. For example, fullFilePath contains the fully qualified directory path along with a file’s name and extension. Because of this quirk, the code falls a little short on being self-documenting and that is why I have included a few more comments than usual to help clarify this new coding.

src/main/java/edu/cornell/kfs/sys/service/ReportWriterService.java
```diff
@@ -66,4 +66,6 @@ public interface ReportWriterService extends org.kuali.kfs.sys.service.ReportWri
 	public void setTitle(String title);
 
 	public void destroy();
+
+	public void initalize(String dataFileNamePrefix);
```
cah292, line 70: Minor typo in method name.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -40,6 +41,9 @@ public class ReportWriterTextServiceImpl extends org.kuali.kfs.sys.service.impl.
 		implements ReportWriterService {
 	private static final Logger LOG = LogManager.getLogger();
 
+	private static String REPORT_FILE_NAME_INFIX = "_report_";
+	private static int NUM_CHARS_ALLOWED_FOR_FULL_FILE_PATH = 512;
```
cah292, line 45: If these are meant to be constants, then please declare them as `final`.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +70,36 @@ public void initialize() {
 		this.writeHeader(title);
 	}
 
+	@Override
+    public void initalize(String dataFileNamePrefix) {
```
cah292, line 74: Some of the formatting is a bit off. If the newly-added methods are only meant to contain spaces, please check to make sure their first and last lines have been converted from tabs to spaces accordingly.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +70,36 @@ public void initialize() {
...
+        if (StringUtils.isNotBlank(dataFileNamePrefix)
+                && fullFilePathLimitWouldNotBeExceeded(dataFileNamePrefix)) {
+            LOG.info("initalize, with XML data file name prefix: {}", dataFileNamePrefix);
+            setFileNamePrefix(dataFileNamePrefix + REPORT_FILE_NAME_INFIX);
```
cah292, line 78: This setter appears to overwrite the filename prefix permanently. In the rare event that a data filename does not meet the requirements for being used as a report filename prefix, you may want to add some way of restoring the Spring-configured prefix. Otherwise, a file that's ineligible for this prefixing could end up using the prefix from a previously processed file that was eligible.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +70,36 @@ public void initialize() {
...
+        }
+        initialize();
+    }
+
+	private boolean fullFilePathLimitWouldNotBeExceeded(String dataFileNamePrefix) {
+        //initialized to force use of default report file prefix when processing fails for any reason
+        int numCharsCalculated = NUM_CHARS_ALLOWED_FOR_FULL_FILE_PATH + 1;
```
cah292, line 85: Since you're immediately overwriting this preinitialized value afterwards without using it, you could remove this line and initialize the variable as part of the length calculation logic.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +70,36 @@ public void initialize() {
...
+
+        numCharsCalculated = filePath.length() + File.separator.length()
```
cah292, line 87: Do any of the string segments need null checking? `StringUtils.length()` provides automatic null handling if you need it.

cah292 (changes requested):
Changes look good overall, though I did notice some areas that could be cleaned up or simplified further.

nkimble (commented):

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +75,79 @@ public void initialize() {
...
+	        onlyDataFileName = StringUtils.substringAfterLast(onlyDataFileName, File.separator);
+	    }
+	    return onlyDataFileName;
+	}
+	
+	private boolean fullFilePathLimitWouldNotBeExceeded(String dataFileNamePrefix) {
+	    //initialized to force use of default report file prefix when processing fails for any reason
+	    int numCharsCalculated = NUM_CHARS_ALLOWED_FOR_FULL_FILE_PATH + 1;
+
+	    numCharsCalculated = StringUtils.length(filePath) + File.separator.length()
```
cah292, line 120: As noted in the prior review, the static-value initialization is not needed because you're immediately replacing it with the filename-derived value.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +75,79 @@ public void initialize() {
...
+	    }
+	}
+	
+	private void configureFileNamePrefixForThisReport(String dataFileNamePrefix) {
+	    resetFileNamePrefixToSpringValue = true;
+	    springFileNamePrefixBackupCopy = fileNamePrefix;
+	    setFileNamePrefix(dataFileNamePrefix + REPORT_FILE_NAME_INFIX);
+	}
+	
+	private void restoreFileNamePrefixToSpringConfiguredPrefix() {
```
cah292, line 142: Should this also be invoked by the 1-arg `initialize()` method, in the event that an error occurs while generating the report but before its `destroy()` method could be called? I don't think our various report-generating areas in the code are forcibly calling `destroy()` if a mid-report-generation error occurs. (I'm working on a report-related user story where I'll be forcibly invoking a report writer's `destroy()` method via a `finally` block, but fixing our code's existing report writer usage is beyond the scope of that user story.)

cah292 (changes requested):
I noticed a few more areas that may need adjustments, but overall the changes look good.

nkimble:
This is ready to be reviewed again. Unit tests are running.

nkimble:
Unit tests have completed.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +75,91 @@ public void initialize() {
...
+	        onlyDataFileName = removeFileExtension(fullyQualifiedDataFileName);
+	        
+	        if (StringUtils.isNotBlank(onlyDataFileName)) {
+	            onlyDataFileName = StringUtils.substringAfterLast(onlyDataFileName, File.separator);
+	        }
+	    }
+	    return onlyDataFileName;
+	}
+	
+	private String removeFileExtension(String fullyQualifiedDataFileName) {
```
cah292, line 121: Could this method be replaced with a call to `StringUtils.substringBeforeLast()` (or have such a call be the body of this method)? That would make the file extension removal much simpler.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
cah292: I was asking whether the 1-arg `initialize()` method should call this prefix-restoring method, not the other way around. Sorry if that wasn't clear in my earlier comment.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
```diff
@@ -66,6 +75,91 @@ public void initialize() {
...
+	    int numCharsCalculated = StringUtils.length(filePath) + File.separator.length() 
+	            + StringUtils.length(dataFileNamePrefix) + StringUtils.length(REPORT_FILE_NAME_INFIX) 
+	                + StringUtils.length(dateTimeService.toDateTimeStringForFilename(dateTimeService.getCurrentDate())) 
+	                    + StringUtils.length(fileNameSuffix);
+
+	    if (numCharsCalculated < NUM_CHARS_ALLOWED_FOR_FULL_FILE_PATH) {
+	        LOG.debug("fullFilePathLimitWouldNotBeExceeded: report file name prefix {} ok for path limit.", dataFileNamePrefix);
+	        return true;
+	    } else {
+	        LOG.warn("fullFilePathLimitWouldNotBeExceeded: Using data file XML name as prefix for report file prefix "
```
cah292, line 141: Since the new feature has been updated to allow for more than just XML files, please update this message to avoid mentioning XML.

cah292 (changes requested):
There are a few spots that might need a little more cleanup, but otherwise the changes look good.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
nkimble: Adjusted.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
nkimble: Fixed.

src/main/java/edu/cornell/kfs/sys/service/impl/ReportWriterTextServiceImpl.java
cah292: After discussing things further with Nancy, we agreed that the current setup for this method is fine as-is. If we need further adjustments to this method or its invocations in the future, we will handle those in a follow-up ticket.

nkimble (commented):
My local validation is completed. This is ready to be reviewed again.

cah292 (approved):

PR #1573, opened 2024-03-06
KFSPTS-30649 Upgrade Maven Artifact secretsmanager Version

ajd299:


cah292 (commented):
Changes look good and unit tests pass. However, I did have a quick question:

Do we need to explicitly declare other Jackson dependencies in the POM (such as `jackson-core`)? Or is it enough to just declare the ones noted in this PR and to just pull in the rest transitively?

ajd299:
For com.fasterxml.jackson I have tried to add explicit dependencies for the packages we are using. Please let me know if I am missing any or added any that do not need to be included

cah292 (approved):
Changes look good and unit tests pass. The version handling seems to be working as expected with the PR's current dependency setup. We can make further revisions later if issues arise.

Merging.

PR #1574, opened 2024-03-08
KFSPTS-6225 Remove CapAsset Struts class overlay

cah292:
This PR removes the `CapitalAssetInformationActionBase` overlay and relocates the custom processing to a `CapitalAssetInformationDetail` overlay instead. We need some way to handle the primary key setup on our CapAsset-related extended attribute (which is what the old overlay accomplished), but doing so through subclassing is not practical because `CapitalAssetInformationActionBase` is an abstract superclass with numerous descendants, and because that superclass constructs `CapitalAssetInformationDetail` instances explicitly. As far as I can tell, overlaying `CapitalAssetInformationDetail` and putting the custom logic into its pre-persistence methods seems like the least invasive way of simplifying our customization. That class is also much easier to maintain than the other class.

If you have any other suggestions on how to simplify this without using any overlays, please let me know.


src/main/java/org/kuali/kfs/fp/businessobject/CapitalAssetInformationDetail.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private void populatePrimaryKeyOnExtendedAttribute() {
+        CapitalAssetInformationDetailExtendedAttribute extension =
+                (CapitalAssetInformationDetailExtendedAttribute) getExtension();
+        extension.setDocumentNumber(getDocumentNumber());
+        extension.setCapitalAssetLineNumber(getCapitalAssetLineNumber());
+        extension.setItemLineNumber(getItemLineNumber());
+    }
+
+    // ==== End CU Customization ====
```
ajd299, line 192: Could this code go before the getters and setters or is our practice to put the customizations at the end of the file?

ajd299 (commented):
This approach seems appropriate, since there are a lot less lines of code to maintain.

src/main/java/org/kuali/kfs/fp/businessobject/CapitalAssetInformationDetail.java
cah292: I don't think the location within this particular file matters too much, but I'll shift them up farther for consistency with similar classes that override these methods.

cah292:
I've relocated the custom code so that it's above the getters and setters. This is ready for review again.

ajd299:
Okay I will merge once the unit tests pass

ajd299 (approved):

PR #1580, opened 2024-03-21
Update License test

db483:
DO NOT MERGE


db483:
PR only used to check unitests after utest database refresh.

PR #1581, opened 2024-03-25
KFSPTS-31229 Remove empl info from Affiliate entries

cah292:
We've received a request to stop treating the "Affiliate" affiliation type as an employee-ID-supporting one. This PR updates the KIM Feed Batch Job to tread "Affiliate" affiliations the same as the other non-employee affiliation types (Student, Alumni, etc.).

Note that there's no associated SQL PR for this. I would suggest using the KIM Feed process to correct the existing KFS data once these changes are in place. We can ask Chris Grippin to generate appropriate KIM Feed delta rows for "Affiliate" users as needed.


db483 (approved):
Changes look good.

PR #1586, opened 2024-04-15
KFSPTS-30809 Fix clearing of DV Wire Transfer tab

cah292:
An initial review can be performed now if desired, but please don't merge until after I get functional feedback on whether the scope of the changes should be expanded further.

DV documents have a pre-rules question that allows for auto-clearing the Wire Transfer tab if the payment method has been set to something other than Wire Transfer. To keep this feature working with our Wire Transfer extended attribute, this PR updates the relevant pre-rules sections so that the extension fields will also be cleared out.


cah292:
Cathy has confirmed that this fix's current scope is correct. This PR can be merged once it has been reviewed and approved.

db483 (approved):
Changes look good.

PR #1588, opened 2024-04-18
KFSPTS-30233 Concur SAE Line Level Validation Error in Subject

ajd299:


src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurReportEmailServiceImpl.java
```diff
@@ -46,12 +47,42 @@ protected String buildEmailSubject(ConcurEmailableReportData reportData) {
...
+
+        for (String errorMessage : concurBatchReportLineValidationErrorItem.getItemErrorResults()) {
+            if (errorMessage.contains("The line has the Pseudo (XXXX) payment code")) {
+                continue;
+            } else if (errorMessage.contains("The line detected as cash advance request. Should have been processed by first batch job step")){
+                continue;
+            } else if (errorMessage.contains("Cash Advance with key") && errorMessage.contains("did not have a corresponding entry from the Request Extract")) {
+                continue;
+            } else if (errorMessage.contains("Line was not processed because a Cash Advance under the same Report ID was missing a related Request Extract entry")) {
+                continue;
```
ajd299, line 77: Based on the requirement of only show the subject under certain conditions, I chose to go this route. Another solution could be to not add the error or add a flag based on the error. Let me know if you would suggest a different solution.

cah292:
retest this please

src/main/java/edu/cornell/kfs/concur/batch/service/impl/ConcurReportEmailServiceImpl.java
cah292: I would not recommend this text-based analysis for determining the file error header logic, especially if the messages end up needing rewording in the future. I think a flag-based approach would be a much cleaner option, perhaps naming it something like `reportableAsLineLevelValidationError`. One of the main methods to be reworked is `ConcurStandardAccountingExtractCollectorBatchBuilder.reportUnprocessedLine()`, where you may need to add a third argument for the flag (and adjust the report item DTO's constructors accordingly). Since that method is also being used as a `BiConsumer` method reference, the relevant areas may need adjustment to use a three-arg reference.  (You may have to add a temporary 3-arg functional interface; Commons Lang3 does have `TriConsumer` interface but I think KFS is using a version that hasn't introduced that yet.)

cah292 (commented):
Please see my prior comment concerning the new code block. I think a flag-based solution would be a better approach to dealing with the email subject setup.

ajd299:
Thank you for your suggestions. Please let me know what you think of the refactored code

src/main/java/edu/cornell/kfs/concur/batch/report/ConcurBatchReportLineValidationErrorItem.java
```diff
@@ -224,4 +233,12 @@ public void addItemErrorResult(String itemErrorResult) {
         this.itemErrorResults.add(itemErrorResult);
     }
 
+    public boolean getReportableAsLineLevelValidationError() {
```
cah292, line 236: Since this is a getter for a boolean property, you may want to use "is" instead of "get" as the prefix.

cah292 (commented):
There's one minor item that may or may not need adjustments, but otherwise the revised changes look good.

ajd299:
Please review again

cah292 (approved):

PR #1589, opened 2024-04-19
KFSPTS-28426 Allow SEC view-doc overrides based on route log

cah292:
There are PRs for this in both cu-kfs and nonprod-sql. Please make sure both are good to go before merging. Also, please coordinate with the team on whether these changes should be merged before or after the merging of the 2023-01-29 financials upgrade.

If the KIM permissions and Access Security restrictions are configured in a certain way, there are situations where users can view a document for which they have an active Action Request, but then lose that access after taking action. To allow such users to continue to view documents that they had previously received during routing, this PR customizes the Access Security processing to look for both prior and current requests.

Note that, as requested, the relaxed restrictions are only applied to a specific set of documents (controlled by a new parameter). At the time of this writing, ST and YEST documents are the only ones that will be using this in its initial setup.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sec/CuSecParameterConstants.java
+++ b/src/main/java/edu/cornell/kfs/sec/CuSecParameterConstants.java
@@ -1,0 +1,8 @@
+package edu.cornell.kfs.sec;
+
+public final class CuSecParameterConstants {
+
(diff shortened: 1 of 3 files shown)


cah292:
retest this please

db483 (approved):
Changes look good.

PR #1590, opened 2024-04-24
KFSPTS-31295 Restore Customer Profile

ajd299:


ajd299:
At this point the Customer Profile is working (tagged locally). However is moving the overlay file out of the org/kuali... directory the same thing as deleting it?

PR #1591, opened 2024-04-24
KFSPTS-31509 Improve Person masking behavior

cah292:
The 01/29 upgrade initially set up the custom name/email/phone masking to behave similarly to KualiCo's customized address masking. However, part of that workaround was causing browser session issues when entering NetIDs into certain form fields.

This PR removes the problematic configuration that was interfering with the browser sessions. As a result, the name/email/phone masking has been adjusted to behave similarly to the pre-01/29 releases, where the Person info is generally masked on most screens (even to users with unmasking permissions), though the Person Inquiry and/or Person Document have some easing of those restrictions.

Note that KualiCo's address masking (and our related custom alt-address masking) are still working as-is, for consistency with base code behavior. Also note that the `CuPersonInquirableImpl` changes are similar to what we currently have in Production.


db483 (approved):
Changes look good.

PR #1594, opened 2024-04-26
KFSPTS-31521 Add more Person field conversion rules

cah292:
This PR adds more maintenance XML conversion rules for clearing out Person fields on various BOs. Some of KualiCo's prior changes caused the maintenance XML to no longer include the Person classname information on Person fields, thus making some of our existing Person conversion rules ineffective on newer KFS documents.


db483 (approved):
Changes look good.

PR #1596, opened 2024-04-29
KFSPTS-31545 Fix handling of Person references on BO Notes

cah292:
The KFSPTS-31521 changes fixed the maintenance XML conversion for older pre-01-29-patch documents, but then the functionals discovered that the maintenance docs were encountering errors upon taking workflow action. Further analysis revealed that the OJB list proxying on the PersonExtension's affiliations list was messing up newly-saved maintenance XML, at least on the Person references within BO Notes. To prevent future list-proxy-related and Note-related issues, this PR changes the problematic list to a non-proxy one, and also adds a maintenance XML conversion rule to remove the Person references on the BO Notes, similar to the Person removal rules introduced by KFSPTS-31521.


db483 (approved):
Changes look good.

PR #1597, opened 2024-04-30
KFSPTS-31462 Backport FINP-10589 case sensitivity fix

cah292:
This PR backports the FINP-10589 KualiCo fix, which corrects a case sensitivity issue when searching for users by principal name. Since we already had a CU override of the affected service, and since the needed code changes were very small, the backport was added to the CU subclass instead of in an overlay.

Note that a small modification was made to the backported fix, as noted in the code comments, due to a possible misconfiguration from KualiCo. I will create a follow-up user story to report the issue back to KualiCo.


db483 (approved):
Changes look good.

PR #1602, opened 2024-05-02
KFSPTS-31428 removing customization-fix no longer needed

jhulslander:


cah292 (approved):
Overlay removal looks good and unit tests pass. Merging.

PR #1605, opened 2024-05-14
KFSPTS-30887 add purchasing assistant owner netid

db483:


src/main/java/edu/cornell/kfs/module/purap/document/IWantDocument.java
```diff
@@ -100,6 +100,8 @@ public class IWantDocument extends FinancialSystemTransactionalDocumentBase impl
     private boolean completed;
     private String processorNetId;
     private String processorName;
+    private String purchasingAssistantNetId;
+    private String purchasingAssistantName;
```
jhulslander, line 104: should this be transient since it isn't persisted?

jhulslander (commented):
I had one question on the code,
When opening the I Want document, and the netID field was previously set, should purchasingAssistantName be populated also?  I think IWantDocumentAction.loadDocument() is where we would want to do that.

src/main/java/edu/cornell/kfs/module/purap/document/IWantDocument.java
db483: updated

jhulslander (approved):

PR #1608, opened 2024-05-21
KFSPTS-29345: Upgrade to the 03/05/2023 version of financials.

nkimble:


nkimble:
Please pay attention to the changes associated with FINP-8973 which "Replaced java.util.Calendar with java.time in financials\kfs-core\src\main\java\org\kuali\kfs\fp". I have two concerns that may need to be addressed with these changes and/or follow-up user stories:

1. Watch out for possible issue in file "edu.cornell.kfs.fp.document.service.impl.CuDisbursementVoucherDefaultDueDateServiceImpl.java" which is our customization left in place and NOT changed for this upgrade. FIle "edu.cornell.kfs.fp.document.CuDisbursementVoucherDocument.java" line 602 replaced base code call that was changed by FINP-8973.
 
2. Should we create a follow-up user story to deal with replacing java.util.Calendar with java.time in a similar fashion as FINP-8973 so that our customizaions in edu/cornell/kfs/fp  match that FINP's changes? I think these files would be affected:
    edu.cornell.kfs.fp.businessobject.USBankRecordFieldUtils.java
    edu.cornell.kfs.fp.dataaccess.impl.ProcurementCardDocumentDaoOjb.java
    edu.cornell.kfs.fp.document.CuProcurementCardDocument.java
    edu.cornell.kfs.fp.service.impl.RecurringDisbursementVoucherDocumentServiceImpl.java
    edu.cornell.kfs.fp.service.impl.ScheduledAccountingLineServiceImpl.java
    edu.cornell.kfs.fp.CuFPConstants

src/main/java/org/kuali/kfs/fp/document/web/struts/DisbursementVoucherAction.java
```diff
@@ -57,6 +57,7 @@
 import org.kuali.kfs.sys.businessobject.PaymentMethod;
 import org.kuali.kfs.sys.businessobject.WireCharge;
```
cah292, line 58: In addition to removing a few imports (such as this one), it looks like the upgrade removed the `retrieveWireChargeMessage()` method entirely. Please include that change in this file.

cah292 (changes requested):
I'm still reviewing the code, but I wanted to proceed with pointing out one minor item that may need to be addressed.

src/main/java/org/kuali/kfs/fp/document/web/struts/DisbursementVoucherAction.java
nkimble: With this next set of changes, I will be trying to make this customized overlay file more closely align to the corresponding base code file.  There are a number of places where "final" is missing from our version as well as "this" existing in our version.  Question: Do you want me to make those items reflect base code OR should I mark those items as a CU Customization?  IMHO one or the other should be done to make future upgrade compares easier.

cah292: With regards to the various minor differences (like the usage of "final" or "this"), they should NOT be marked as customizations; it would be best to make them line up with base code. If you think such rework expands the scope of the upgrade changes too much, then I would suggest creating a follow-up user story (or multiple such stories) to perform this cleanup. Thanks for checking.

nkimble: The latest set of changes for this file more closely aligns it to base code so that our customization is obvious when performing future upgrades.

cah292:
Regarding the date-time handling, I agree that we should create separate user stories to convert our custom date-related code over to using `java.time` instead. (This would involve both Java Calendar and Joda DateTime conversion.) I don't think the conversion is urgent at this point, except for code that overrides/overlays base code.

As for `CuDisbursementVoucherDefaultDueDateServiceImpl`, it looks like you would only need to change the 3 lines in `findDefaultDueDate()` accordingly. It would be good to clean that up now if it's not too much trouble, but if you'd rather defer the cleanup, please make sure it's mentioned in the appropriate follow-up ticket.

src/main/java/com/rsmart/kuali/kfs/cr/document/service/impl/GlTransactionServiceImpl.java
```diff
@@ -168,11 +168,11 @@ private void generateGlPendingTransaction(PaymentGroup paymentGroup, String fina
...
-                OffsetDefinition offsetDefinition = SpringContext.getBean(OffsetDefinitionService.class).getByPrimaryId(glPendingTransaction.getUniversityFiscalYear(), glPendingTransaction.getChartOfAccountsCode(), paymentAccountDetail.getPaymentDetail().getFinancialDocumentTypeCode(), glPendingTransaction.getFinancialBalanceTypeCode());
+                OffsetDefinition offsetDefinition = SpringContext.getBean(OffsetDefinitionService.class).getActiveByPrimaryId(glPendingTransaction.getUniversityFiscalYear(), glPendingTransaction.getChartOfAccountsCode(), paymentAccountDetail.getPaymentDetail().getFinancialDocumentTypeCode(), glPendingTransaction.getFinancialBalanceTypeCode()).orElse(null);
                 
                 glPendingTransaction.setAccountNumber(accountNbr);
                 glPendingTransaction.setChartOfAccountsCode(finCoaCd);
-                glPendingTransaction.setFinancialObjectCode(offsetDefinition != null ? offsetDefinition.getFinancialObjectCode() : finObjectCode);
+                glPendingTransaction.setFinancialObjectCode(ObjectUtils.isNotNull(finObjectCode) ? offsetDefinition.getFinancialObjectCode() : finObjectCode);
```
cah292, line 175: I think this null check should examine `offsetDefinition`, not `finObjectCode`.

src/main/java/org/kuali/kfs/module/purap/document/web/struts/RequisitionAction.java
cah292, line 1: You may want to verify with the functionals if they're using the REQS document's Organization routing (or are planning to in the near future). If they are, then you may want to override or overlay this action class's `approve()` method, because there's a slight typo in the requeue/refresh message that we fixed as part of the original backport. (KualiCo later fixed that typo for FINP-9395, so our original custom fix could be rebranded as a backport if necessary.)

cah292 (changes requested):
Changes look good overall, though I did find a few items that may need cleanup. I'll also add a separate comment concerning a file that's currently not affected by this PR.

cah292:
Concerning the extra follow-up comment that I had hinted at before:

I was originally going to point out some changes to `ArConstants.java` that did not get added to our overlay, but now I realize that those changes were ultimately made in the subsequent 2023-03-08 patch, not in the 2023-03-05 patch. Our overlay of that particular class should be fine for now; sorry for the confusion.

src/main/java/com/rsmart/kuali/kfs/cr/document/service/impl/GlTransactionServiceImpl.java
nkimble: fixing

src/main/java/org/kuali/kfs/module/purap/document/web/struts/RequisitionAction.java
nkimble: The functional users have confirmed that they WILL be using the REQS document's Organization routing. I am going to rebrand this overlay backport to be FINP-9395 fix released by KualiCo on 2023-04-05

src/main/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImpl.java
```diff
@@ -20,9 +20,8 @@ public class CuDisbursementVoucherDefaultDueDateServiceImpl implements CuDisburs
 
     @Override
     public Date findDefaultDueDate() {
-        Calendar calendar = dateTimeService.getCurrentCalendar();
-        calendar.add(Calendar.DAY_OF_MONTH, findNumberOfDaysInTheFuture());
-        return new Date(calendar.getTimeInMillis());
+        LocalDate calculatedDefaultDvDueDate = dateTimeService.getLocalDateNow().plusDays(findNumberOfDaysInTheFuture());
+        return dateTimeService.getSqlDate(calculatedDefaultDvDueDate);
```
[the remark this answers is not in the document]
cah292: If you want `getLocalDateNow()` to return a pre-defined date but you want `getSqlDate()` to use the service's actual code, you could set up the dateTimeService as a "spied" object. First create the dateTimeService using an actual instance, then use `Mockito.spy()` to create a proxy that will delegate to the underlying object by default. You can then use one of the `Mockito.doXXX()` methods to override the behavior of individual service methods. (In most cases, don't use `Mockito.when()` to configure a "spied" object, because doing it that way will cause the setup process to invoke the actual method on the wrapped instance.)

For an example of how to set this up, see the `ConcurRequestV4ServiceImplTest` class. It creates a "spied" dateTimeService instance using our custom `TestDateTimeServiceImpl` class for the wrapped object.

nkimble:
retest this please

nkimble:
JIRAs KFSPTS-31897 and KFSPTS-31898 created to address outstanding items in followup user stories. This is ready to be reviewed again.

db483 (approved):
The remaining concern will be addressed on a separate user story.

db483 (approved):

PR #1611, opened 2024-05-30
KFSPTS-31842: Fix stacktrace for users without primary departments on their person record not being able to return carts from eShop to KFS.

nkimble:


jhulslander:
This code change looks reasonable.  I was able to confirm the version of FinancialSystemUserServiceImpl from base code was the correct version, and I agree with this fix.
I will be doing developer validation in sandbox as we can't do eshop locally.

jhulslander (approved):
Code changes look fine, I shall merge

PR #1615, opened 2024-06-10
KFSPTS-31931 Backport FINP-9525 doc requeue fix

cah292:
This PR backports KualiCo's FINP-9525 fix, which prevents a requeue of all pending Account-route-node documents when inactivating a Person that has no Fiscal Officer delegations. It also includes a fix for preventing such mass requeues when performing minor changes on a Role document; however, the latter part of the fix doesn't seem to be working in certain cases. Nevertheless, the functionals are fine with this backported FINP moving forward as-is, and we'll report the latter issue to KualiCo and/or use a separate user story for fixing it as needed.


ajd299 (approved):
Changes look good, merging

PR #1618, opened 2024-06-24
KFSPTS-31189: Create blacklist parameter and service that disallows attachments on KFS documents with the identified file extensions.

nkimble:


src/main/java/org/kuali/kfs/kns/web/struts/action/KualiDocumentActionBase.java
```diff
@@ -173,6 +177,10 @@ public class KualiDocumentActionBase extends KualiAction {
     private BusinessObjectMetaDataService businessObjectMetaDataService;
     private AdHocRoutingService adHocRoutingService;
     private WorkflowDocumentService workflowDocumentService;
+    
+    //CU customization: Added blacklist attachment processing
+    private static final String ERROR_UPLOADFILE_EXTENSION = "error.uploadFile.extension";
```
db483, line 182: Should this constant be moved to CUKFSKeyConstants?

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -738,3 +738,5 @@ errors.reject.invoice.po.vendor.mismatch=Vendor from Kuali PO does not match wit
 
 error.cam.building.room.combination.invalid=Invalid Room #%s for Building Code %s
 error.cam.building.code.invalid=Invalid Building Code %s
+
+error.uploadFile.extension=File '{0}' is not allowed as an attachment; please select a different file such as an Excel, Word or PDF document.
```
db483, line 742: Was this message provided by the functionals? If not, it would be good to confirm it with Cathy. Thank you.

db483 (changes requested):
Changes look good. I just had a couple of suggestions for improvement. Thank you.

src/main/java/org/kuali/kfs/kns/web/struts/action/KualiDocumentActionBase.java
nkimble: Moved constant.

src/main/resources/CU-ApplicationResources.properties
nkimble: Cathy an I spoke about what the wording should be and she provided the specifics that are listed here.

db483 (approved):
Changes look good and unit tests passed. I will merge.

PR #1620, opened 2024-06-25
KFSPTS-30797 PDP Extract Unique Filename and Logging

ajd299:


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -98,12 +98,14 @@ public void extractAchPayments() {
...
     
         String filename = getOutputFile(achFilePrefix, processDate);
-        LOG.debug("MOD: extractAchPayments() filename = " + filename);
 
         /** 
         * MOD: This is the only section in the method that is changed.  This mod calls a new method that bundles 
         * ACHs into single disbursements if the flag to do so is turned on.
+        * Modified the SimpleDateFormat to have milliseconds to prevent files being overwritten
         */
+        sdf = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss-SSS", Locale.US);
```
cah292, line 107: I don't think this particular date format needs updating, because this line of code should be unreachable when ISO 20022 formatting is enabled.  (The `if` block above with the `shouldUseIso20022Format()` check should force the method to return early.) Also, the downstream code might be ignoring or replacing this format anyway.

I believe the correct way to modify the filename for the legacy Check/ACH files is to change the date formatter in the `getOutputFile()` method. That method is in base code but can be overridden by this subclass. (Note that we're still generating legacy KFS-proprietary Check files in addition to ISO Check files, so such an override will be needed until we turn off the KFS-proprietary Check file handling in a future sprint.)

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -375,6 +375,7 @@ public void extractChecks(
                     message.setCstmrCdtTrfInitn(customerCreditTransferInitiation);
 
                     final String filename = determineFilename(directoryName, extractTypeContext);
+                    LOG.info("extractChecks writing file " + filename);
```
cah292, line 378: There's an equivalent spot in the `extractAchs()` method where another logging statement should be inserted. Alternatively, you could move the custom Check/ACH filename logging to the `writeMessageToFile()` method.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,7 +1816,7 @@ private String determineFilename(
         }
 
         final Date disbursementDate = extractTypeContext.getDisbursementDate();
-        final SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US);
+        final SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd_HHmmssSSS", Locale.US);
```
cah292, line 1819: Adjusting this format to include milliseconds won't be enough on its own to guarantee filename uniqueness. Based on the `extractAchs()` and `extractChecks()` methods, the date that's used in the filename is being defined BEFORE iterating over the file creation loops, meaning this filename-building method will likely be reusing the same source date for a given set of Check/ACH files.

To get the filename uniqueness to work as expected, you may need to make one or more of the following changes:

* Redefine the source date for each iteration of the Check/ACH `for` loops. (Not sure whether this would interfere with the data being sent to the bank, though.)
* Modify the `determineFilename()` method so that it also examines the filesystem to see if the given file already exists. If it does, then make adjustments to the filename accordingly, such as by tweaking the millisecond portion or adding/adjusting a custom suffix.

cah292 (changes requested):
I noticed a few things that will need further updates, especially with regards to the filename uniqueness feature.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,16 +1815,29 @@ private String determineFilename(
...
 
         final Date disbursementDate = extractTypeContext.getDisbursementDate();
-        final SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US);
+        final String filename = buildUniqueFilename(directoryName, finalFormattedCheckFilePrefix, disbursementDate, null);
+
+        LOG.debug("determineFilename(...) - : filename={}", filename);
+        return filename;
+    }
+
+    private String buildUniqueFilename(String directoryName, String finalFormattedCheckFilePrefix, Date disbursementDate, Integer uniquenessSuffix) {
```
cah292, line 1824: Since this file is an overlay of base code, please add appropriate comments to point out this customization to it.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,16 +1815,29 @@ private String determineFilename(
...
                 directoryName
                 + File.separator
                 + finalFormattedCheckFilePrefix
                 + "_"
                 + sdf.format(disbursementDate)
+                + (uniquenessSuffix != null ? "-" + uniquenessSuffix : "")
                 + ".xml";
 
-        LOG.debug("determineFilename(...) - : filename={}", filename);
+        if ((new File(filename).exists())) {
```
cah292, line 1836: There appears to be one extra unnecessary set of parentheses on this line.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,16 +1815,29 @@ private String determineFilename(
...
+            LOG.info("buildUniqueFilename filename={} already exists, adding uniqueness suffix", filename);
+            return buildUniqueFilename(directoryName, finalFormattedCheckFilePrefix, disbursementDate, ++uniquenessSuffix);
```
cah292, line 1838: Your initial call to this method sets `uniquenessSuffix` to null. Will that cause the "++" part on this line to encounter a null pointer?

You may need to make some more adjustments to keep this null-safe.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1895,6 +1908,7 @@ private static void writeMessageToFile(
              * of changing/removing the namespace prefixes, rather than performing String.replaceAll() substitution.
              * Also added setup of a CU-specific EscapeHandler to customize the XML-escaping process.
              */
+            LOG.info("writeMessageToFile writing file " + filename);
```
cah292, line 1911: I would suggest modifying this log statement to use a placeholder for inserting the filename.

cah292 (changes requested):
Not sure of this was ready for review again, but I did notice a few minor follow-up items that may need to be addressed.

ajd299:
Please let me know if I have addressed all of the feedback thus far.

cah292:
retest this please

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,16 +1815,31 @@ private String determineFilename(
...
                 + "_"
                 + sdf.format(disbursementDate)
+                + (uniquenessSuffix != null ? "-" + uniquenessSuffix : "")
                 + ".xml";
 
-        LOG.debug("determineFilename(...) - : filename={}", filename);
+        if (new File(filename).exists()) {
+            LOG.info("buildUniqueFilename filename={} already exists, adding uniqueness suffix", filename);
+            int suffixCounter = uniquenessSuffix == null ? 0 : ++uniquenessSuffix;
+            return buildUniqueFilename(directoryName, finalFormattedCheckFilePrefix, disbursementDate, suffixCounter);
```
cah292, line 1840: This back-and-forth shifting between primitive and wrapper integers is a little confusing. You may want to consider using just a primitive integer instead, set it 0 or 1 as the initial value, and exclude it from the filename if it's currently at the starter value. It might also be helpful to rename the variable to "fileIndex" or something along those lines.

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
cah292: The placeholder suggestion for this log statement has not been implemented yet.

cah292 (changes requested):
There are a couple more areas that might need revisions, but otherwise the changes look good.

ajd299:
Please see my latest changes

src/main/java/org/kuali/kfs/pdp/batch/service/impl/Iso20022FormatExtractor.java
```diff
@@ -1815,16 +1815,30 @@ private String determineFilename(
...
+    private String buildUniqueFilename(String directoryName, String finalFormattedCheckFilePrefix, Date disbursementDate, int fileIndex) {
+        final SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd_HHmmssSSS", Locale.US);
+
         final String filename =
                 directoryName
                 + File.separator
                 + finalFormattedCheckFilePrefix
                 + "_"
                 + sdf.format(disbursementDate)
+                + (fileIndex == 0 ? "" : ("-" + fileIndex ))
```
cah292, line 1834: Unnecessary whitespace prior to the closing parentheses.

cah292 (changes requested):
There's one very minor formatting item that would be good to clean up, but otherwise the changes look good.

cah292 (approved):

PR #1622, opened 2024-06-26
KFSPTS-31189: Create blacklist parameter and service that disallow attachments on KFS documents based on the file extensions listed in the parameter.

nkimble:
The only difference between this version of the code and the previous version is the user error message in file CU-ApplicationResources.properties


cah292 (approved):

PR #1627, opened 2024-07-15
KFSPTS-31370: Purge all Initiated KFS docs based upon new system parameter KREW_DOC_HDR_T_INITIATED_NUMBER_OF_DAYS_OLD.

nkimble:


nkimble:
Related SQL PR that must be merged and applied with this code is located here: https://github.com/CU-CommunityApps/nonprod-sql/pull/1415

src/main/java/edu/cornell/kfs/sys/batch/service/impl/DocumentRouteHeaderValueLookupCriteriaPurgeServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.ojb.broker.query.Criteria;
+import org.kuali.kfs.kew.api.KewApiConstants;
+
+import edu.cornell.kfs.sys.batch.service.TableLookupCriteriaPurgeService;
+
+public class DocumentRouteHeaderValueLookupCriteriaPurgeServiceImpl extends TableLookupCriteriaPurgeServiceImpl
+        implements TableLookupCriteriaPurgeService {
+
+    private static final String DOCUMENT_ROUTE_STATUS = "docRouteStatus";
+    private static final String CREATE_DATE = "createDate";
```
jhulslander, line 14: the String "createDate" is used in a number of places in our code, but we don't have a constant. Should this be moved into one of our constant classes, and a follow up user story to update the classes to use a constant?

jhulslander (changes requested):
code changes look good, one minor idea to consider

src/main/java/edu/cornell/kfs/sys/batch/service/impl/DocumentRouteHeaderValueLookupCriteriaPurgeServiceImpl.java
nkimble: Base code KRADPropertyConstants defines the createDate String as a constant Replacing with that base code value instead of creating a local customization for it.
DOCUMENT_ROUTE_STATUS is not defined anywhere in the code base so leaving local to this class.

jhulslander (commented):
This code is ready to merge when we are ready

nkimble:
Retest this please

jhulslander (approved):
The unit tests pass, and the SQL is ready too, I shall merge

PR #1628, opened 2024-07-22
KFSPTS-32208 Add fixes for IWNT return-to-SSC feature

cah292:

The changes from #1619 added the "Return to SSC" feature to the IWNT document, but it was later discovered that some additional changes were needed. This PR and its related SQL PR implement and combine the following fixes/changes to this feature:

* KFSPTS-32208: Fix the "Return to SSC" behavior on existing saved/enroute IWNT documents.
* KFSPTS-32219: When returning the IWNT to the SSC, require specifying a reason for doing so, which will be added as both a document note and a route log annotation.
* KFSPTS-32158: Update IWNT fields/nodes with the "Purchasing Contract Assistant" naming to use the "Procurement Contract Assistant" naming instead.

The return-to-previous-node behavior has been updated to be more similar to the Recall and/or Disapprove actions, especially with regards to the confirmation prompt. One of our overlays needed adjustments to allow the prompt-handling feature to be used by the IWNT action class.

This code PR also makes some miscellaneous improvements, such as adding the return-to-previous-node method at the interface level and making the post-button-click behavior similar to that of other document actions.


cah292:
retest this please

db483 (approved):
Code changes look good.

PR #1629, opened 2024-07-26
KFSPTS-31897 fix unit test

db483:


src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -118,11 +103,18 @@ protected ParameterService buildMockParameterService(String numberOfDays) {
         return service;
     }
     
-    protected DateTimeService buildMockDateTimeService(Calendar currentCalendar) {
+    protected DateTimeService buildMockDateTimeService() {
         DateTimeService service = Mockito.mock(DateTimeService.class);
-        Mockito.when(service.getCurrentCalendar()).thenReturn(currentCalendar);
+        Mockito.when(service.getLocalDateNow()).thenReturn(java.time.LocalDate.now(Clock.systemDefaultZone()));
+        when(service.getSqlDate(any(LocalDate.class))).then(this::getSqlDate);
```
cah292, line 109: I would suggest updating this method to create an instance of `TestDateTimeServiceImpl` and then use `Mockito.spy()` to create a spied version of it. That way, you can let the actual class handle most of the operations, and you can use one of the `Mockito.doXXX()` methods to override the behavior of specific service methods. This will allow the unit test to override the current date-time as before (though the test should be updated to use a `java.time` class instead of `Calendar`).

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -1,9 +1,13 @@
 package edu.cornell.kfs.fp.document.service.impl;
 
+import static org.mockito.ArgumentMatchers.any;
 import static org.junit.Assert.*;
+import static org.mockito.Mockito.when;
```
cah292, line 5: Other areas of this class are already using `Mockito.when()` instead of the static import variant. You may want to adjust this test so that the import can be removed.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -39,51 +45,30 @@ public void tearDown() throws Exception {
 
     @Test
```
cah292, line 46: Given how relatively small and simple this test class is, it might be a good idea to go ahead and convert it over to using the JUnit 5 annotations.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -39,51 +45,30 @@ public void tearDown() throws Exception {
...
     public void testFindDefaultDueDateToday() {
-        prepCuDisbursementVoucherDefaultDueDateServiceImpl("1", Calendar.getInstance());
+        prepCuDisbursementVoucherDefaultDueDateServiceImpl("1");
         Calendar expectedCalendar = Calendar.getInstance();
```
cah292, line 49: This should be converted from `Calendar` to an appropriate `java.time` class.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -39,51 +45,30 @@ public void tearDown() throws Exception {
...
         expectedCalendar.add(Calendar.DAY_OF_MONTH, 1);
 
-// KFSPTS-31897 - fix failing unit test
-//        assertServiceDateEqualsExpectedDate(new Date(expectedCalendar.getTimeInMillis()));
+        assertServiceDateEqualsExpectedDate(new Date(expectedCalendar.getTimeInMillis()));
     }
     
-    @Test
-    public void testFindDefaultDueDateLeapYear() {
```
cah292, line 51: If it's practical to implement the suggested `DateTimeService` improvements, then these removed test methods should be added back in (and updated for `java.time` compatibility).

cah292 (changes requested):
I have some suggestions on how to preserve the various non-current-date test cases, along with a few other improvements.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -92,37 +100,57 @@ public void testFindDefaultDueDateBadParamEmptyNull() {
...
+        java.time.LocalDate p = invocation.getArgument(0);
+        return java.sql.Date.valueOf(p);
+    }
+
+    private DateTimeService buildSpiedDateTimeService(TestDateTimeServiceImpl actualDateTimeService, int year,
+            Month month, int dayOfMonth) {
+        TestDateTimeServiceImpl dateTimeService = Mockito.spy(actualDateTimeService);
+        Mockito.doAnswer(invocation -> getMockLocalDateNow(year, month, dayOfMonth)).when(dateTimeService)
+                .getLocalDateNow();
+        Mockito.doAnswer(invocation -> getSqlDate(invocation)).when(dateTimeService).getSqlDate(any(LocalDate.class));
```
cah292, line 138: Does the `getSqlDate()` method need to be mocked? From what I can tell, the underlying service should be able to handle those method calls successfully.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -79,11 +85,13 @@ public void testFindDefaultDueDateBadParamString() {
         }
         assertTrue("This should have generated an IllegalStateException", caughtError);
```
cah292, line 86: If this class gets converted over to using the JUnit 5 assertion methods, you may want to consider updating the various error-throwing cases to use `assertThrows` assertions instead.

cah292 (changes requested):
I had some more suggestions for cleaning up these unit tests further.

src/test/java/edu/cornell/kfs/fp/document/service/impl/CuDisbursementVoucherDefaultDueDateServiceImplTest.java
```diff
@@ -1,128 +1,110 @@
...
-        dateFormat = null;
     }
 
     @Test
-    public void testFindDefaultDueDateToday() {
-        prepCuDisbursementVoucherDefaultDueDateServiceImpl("1", Calendar.getInstance());
-        Calendar expectedCalendar = Calendar.getInstance();
-        expectedCalendar.add(Calendar.DAY_OF_MONTH, 1);
+    public void testFindDefaultDueDateToday() throws Exception {
+        prepCuDisbursementVoucherDefaultDueDateServiceImpl("1", LocalDate.of(LocalDate.now().getYear(), LocalDate.now().getMonth(), LocalDate.now().getDayOfMonth()));
```
cah292, line 39: Could this spot just use `LocalDate.now()` directly, rather than picking apart the current year/month/day and manually adding them to a new instance?

cah292 (changes requested):
I had one more minor code cleanup suggestion, but otherwise the latest revisions look good.

cah292 (approved):

PR #1631, opened 2024-08-05
KFSPTS-32287 Backport FINP-10040 AR Cash Control Import

ajd299:


jhulslander:
I am still reviewing these changes, I think the code comments noted above will be helpful.
I am not sure about just being able to cherry pick this change into sandbox due to the action class override, but I need to look more closely.  If Kris is ready for sandbox to be refreshed, we won't need to look into that.

ajd299:
- When you say sandbox do you mean KFS-AR environment? 
- Please see my latest commit with the customization marker comments.

src/main/java/org/kuali/kfs/module/ar/ArPropertyConstants.java
```diff
@@ -257,6 +257,15 @@ public static class CashControlDocumentFields {
 
     public static class CashControlDetailFields {
         public static final String CASH_CONTROL_DETAILS_TAB = "document.cashControlDetails";
+
+        /* CU Customization: Backport FINP-10040 These constants can be removed 2023-09-20 financials patch. */
```
jhulslander, line 261: the FederalFormReportFields sub class in this file has gotten out of sync with base code.  I wonder if it would be good to catch that up now.
@cah292 do you have an opinion on this?

src/main/java/org/kuali/kfs/module/ar/document/web/struts/CashControlForm.java
```diff
@@ -0,0 +1,141 @@
...
+
+    public boolean isCashPaymentMediumSelected() {
+        return cashPaymentMediumSelected;
+    }
+
+    public void setCashPaymentMediumSelected(final boolean cashPaymentMediumSelected) {
+        this.cashPaymentMediumSelected = cashPaymentMediumSelected;
+    }
+
+    public FormFile getDetailImportFile() {
```
jhulslander, line 127: A note should be added here on these functions and the new variable detailImportFile that they are part of the back port

src/main/java/org/kuali/kfs/module/ar/document/web/struts/CashControlAction.java
```diff
@@ -0,0 +1,443 @@
...
+    /**
+     * Import items to the document from a spreadsheet.
+     *
+     * @param mapping  An ActionMapping
+     * @param form     An ActionForm
+     * @param request  The HttpServletRequest
+     * @param response The HttpServletResponse
+     * @return An ActionForward
+     */
+    public ActionForward importDetails(
```
jhulslander, line 368: The functions added  should also have a comment that they are part of the backport

src/main/webapp/WEB-INF/tags/module/ar/cashControlDetails.tag
```diff
@@ -0,0 +1,150 @@
...
+              description="The DataDictionary entry containing attributes for cash control detail fields."%>
+<%@ attribute name="readOnly" required="true"
+              description="If document is in read only mode"%>
+<%@ attribute name="editDetails" required="true"
+              description="If document details are in edit mode"%>
+<%@ attribute name="editPaymentAppDoc" required="true"
+              description="If payment application document number should be a link"%>
+
+<kul:tab tabTitle="Cash Control Details" defaultOpen="true" tabErrorKey="${KFSConstants.CASH_CONTROL_DETAILS_ERRORS}"
+         helpUrl="${KualiForm.detailsImportInstructionsUrl}" helpLabel="Import Templates">
```
jhulslander, line 50: Can you confirm the help URL works as expected

src/main/webapp/WEB-INF/tags/module/ar/cashControlDetails.tag
```diff
@@ -0,0 +1,150 @@
...
+              type="java.util.Map"
+              description="The DataDictionary entry containing attributes for cash control detail fields."%>
+<%@ attribute name="readOnly" required="true"
+              description="If document is in read only mode"%>
+<%@ attribute name="editDetails" required="true"
+              description="If document details are in edit mode"%>
+<%@ attribute name="editPaymentAppDoc" required="true"
+              description="If payment application document number should be a link"%>
+
+<kul:tab tabTitle="Cash Control Details" defaultOpen="true" tabErrorKey="${KFSConstants.CASH_CONTROL_DETAILS_ERRORS}"
```
jhulslander, line 49: I think we want further override comments in here too

jhulslander (changes requested):
I think this code change looks good for the most part, I think we just want to add in the "Section" comments about specific function over rides in the new files too.
I am not sure about ARPropertyConstants that have gotten out of sync with base code.

src/main/java/org/kuali/kfs/module/ar/ArPropertyConstants.java
cah292: @jhulslander The reason why the `FederalFormReportFields` nested class is out-of-sync is because we previously backported the FINP-7147 changes into this file. The sections regarding the FINP-7147 backport should be kept as-is until we upgrade to the related KualiCo patch version (2023-06-28).

src/main/webapp/WEB-INF/tags/module/ar/cashControlDetails.tag
ajd299: I think it is working now after I added a new parameter component definition

ajd299:
Please see my latest commits and the commit on the SQL Pull Request. Note that for certain whole files since we are bringing in I put a comment to remove the whole file based on the backport.

ajd299:
To validate Locally I use branch KFSPTS-32287-ar which is based on the ar-environment branch and then spin up a database from the ar database.

PR #1635, opened 2024-08-15
KFSPTS-30386 Move Accounting Document XML filetype to new class

cah292:
This is a newer version of the #1633 changes. The code is almost the same, but with the following key differences:

* The XSD's `DocumentWrapper` element definition has been updated to use `xsd:all` instead of `xsd:sequence`, thus allowing the expected sub-elements to appear in any order.
* A new unit test case has been added to verify that the relaxed ordering works as expected.


ajd299 (approved):
Code is concise and well written per our team standards, merging code to develop

PR #1637, opened 2024-08-19
KFSPTS-29125 allow canceling of DVs when accounts are expired

jhulslander:


cah292 (commented):
I'm wondering whether or not setting override codes on the accounting lines is the right way to go with this. If the operations are performing behind-the-scenes saves on finalized documents but are failing due to expired-account business rule errors, then perhaps it would make sense to adjust the doc-saving operations so that they bypass the business rules validation. Certain PURAP documents already do this via the `PurapService.saveDocumentNoValidation()` method, so maybe we need an FP module equivalent to handle recurring and spawned DVs. (Based on that service, calling the 2-arg `DocumentService.saveDocument()` method and passing in a `DocumentSystemSaveEvent` instance should allow for saving the document without checking the business rules.)

I see that you're also overriding expired accounts when a workflow cancel is being performed. I didn't think that would be necessary on an actual cancel operation, but if it is then the save-event workaround above won't work for that case.

Thoughts?

jhulslander:

cah292 (approved):

PR #1638, opened 2024-08-19
KFSPTS-32455 IWantDocument Show Contract Tab When Document is Enroute

ajd299:


src/main/java/edu/cornell/kfs/module/purap/document/authorization/IWantDocumentAuthorizer.java
```diff
@@ -147,8 +147,13 @@ private boolean isDocumentAtConfirmationStep(Document document) {
     
     public boolean canViewContractTab(Document document, Person person) {
         WorkflowDocument workflowDocument = document.getDocumentHeader().getWorkflowDocument();
-        return (isInOrgHierarchyOrProcurementAssistantNode(document) && (isCurrentUserOrgReviewer(person) 
-                || isCurrentUserProcurementContractAssistant(person))) || workflowDocument.isFinal();
+
+        if (workflowDocument.isEnroute() || workflowDocument.isFinal()) {
```
jhulslander, line 151: Based on the description of the JIra
I think this should include a permission check for " SSC or Purchasing Assistant reviewers" the status should be not workflowDocument.isFinal()

jhulslander (changes requested):
I think the logic could use some adjustment based on the Jira

ajd299:
Based on the feedback by Cathy:

- The viewing of the contract tab has been modified to only be restricted when document is Saved/Initiated
- The editing of the netid field (editProcurementAssistantNetId) will be addressed by KFSPTS-32444
- The editing of the contract indicator is currently being addressed by canEditContractIndicator

jhulslander (approved):

PR #1640, opened 2024-08-21
KFSPTS-32444 Fix editing of IWNT Proc Assistant NetID

cah292:
This PR adjusts the editability of the IWNT Procurement Assistant NetID, so that it's only editable at a specific route node by a specific role's members.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/authorization/IWantDocumentPresentationController.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/authorization/IWantDocumentPresentationController.java
@@ -13,6 +13,7 @@ import org.kuali.kfs.sys.service.FinancialSystemWorkflowHelperService;
 import org.kuali.kfs.sys.service.impl.KfsParameterConstants;
 
 import edu.cornell.kfs.module.purap.CUPurapConstants;
+import edu.cornell.kfs.module.purap.CUPurapConstants.IWantRouteNodes;
 import edu.cornell.kfs.module.purap.CUPurapParameterConstants;
 import edu.cornell.kfs.module.purap.document.IWantDocument;
 import edu.cornell.kfs.sys.CUKFSConstants;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/authorization/IWantDocumentPresentationController.java
```diff
@@ -148,13 +149,15 @@ public Set<String> getEditModes(Document document) {
...
         editModes.add(CUPurapConstants.IWNT_DOC_DISPLAY_NOTE_OPTIONS);
         
         if (isContractFunctionalityEnabled()) {
             editModes.add(CUPurapConstants.IWNT_DOC_DISPLAY_CONTRACT_TAB);
             editModes.add(CUPurapConstants.IWNT_DOC_EDIT_CONTRACT_INDICATOR);
-
+            if (workflowDocument.isEnroute() && workflowDocument.getCurrentNodeNames().contains(
+                    IWantRouteNodes.PROCUREMENT_CONTRACT_ASSISTANT)) {
+                editModes.add(CUPurapConstants.I_WANT_DOC_EDIT_PROCUREMENT_ASSISTANT_NET_ID);
+            }
```
ajd299, line 160: Could this code go in the canEditProcurementAssistantNetId method?

cah292: I realize that the related authorizer class already contains many of the route-node-based and workflow-status-based checks for the other edit modes. However, I was trying to set this part up to put most of the non-person-specific checks into the presentation controller, which I think is how KFS generally expects edit modes and action authorization to be handled.

I could alternatively put these checks into the authorizer class for consistency with our other IWNT setup. If you have a strong opinion about which approach would be better, please let me know.

ajd299: I am fine doing it this way if it aligns with our best practices.

ajd299 (approved):
Code looks good, merging

PR #1641, opened 2024-08-26
KFSPTS-32580 Enable notes on Person and Role docs

cah292:
This PR adds the Notes and Attachments tab to the Person Document and the Role Document. Note that, because both of these documents set the "readOnly" flag at the request level, it needs to be temporarily set to false when rendering the Notes and Attachments tab; otherwise the add-line section might not be fully editable. Also, the Person Document needed a minor tweak to allow for adding attachments.


db483 (approved):
Changes look good.

PR #1643, opened 2024-09-05
KFSPTS-32020 Fix batch jobs getting stuck when encountering errors

cah292:
This PR fixes the issue that is causing our batch jobs to get stuck when they throw exceptions. It appears that KualiCo messed up the `BatchStepExecutor` class's log-related refactoring for FINP-9025, so I overlayed it to rearrange the error handling and have it create the .error file(s) at the correct times.

### DIFF
--- a/src/main/java/org/kuali/kfs/sys/context/BatchStepExecutor.java
+++ b/src/main/java/org/kuali/kfs/sys/context/BatchStepExecutor.java
@@ -1,0 +1,243 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2023 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.sys.context;
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/kuali/kfs/sys/context/BatchStepExecutor.java
```diff
@@ -0,0 +1,243 @@
...
+        batchStepFile.setStepIndex(stepIndex);
+
+        try (CloseableThreadContext.Instance ignored = setupNDCLogging()) {
+            runWithLogging(stepRunDate);
+            /*
+             * CU Customization: Copied the contents of the "catch" block below this line into the runWithLogging()
+             * method instead, and replaced the original contents below with warning-only error handling. 
+             */
+        } catch (final Exception throwable) {
+            LOG.warn(
```
jhulslander, line 105: should these be error log messages rather than warnings? Also, should the full error be logged, rather than just the message?

src/main/java/org/kuali/kfs/sys/context/BatchStepExecutor.java
```diff
@@ -0,0 +1,243 @@
...
+                LOG.info("Step returned false");
+                batchContainerDirectory.writeBatchStepErrorResultFile(batchStepFile);
+            }
+            /*
+             * CU Customization: Copied the run() method's original "catch" block contents to this spot instead.
+             * That way, if the batch step throws an exception, KFS will create the .error file BEFORE running
+             * the code that checks for .success or .error files.
+             */
+        } catch (final Exception throwable) {
+            LOG.warn(
```
jhulslander, line 136: similar thought about it being an error, and logging the full error

jhulslander (changes requested):
I added a few thoughts to consider

src/main/java/org/kuali/kfs/sys/context/BatchStepExecutor.java
cah292: I'm not sure why, but KualiCo had set this part up so that the error would be logged as a warning and would only show partial details here. I was trying to be consistent with how the `BatchStepExecutor` was currently logging things.

Also, the downstream code in `Job.runStep()` should already be logging the exception's full stack trace at error level.

jhulslander (approved):
OK, lets proceed as you you coded

PR #1644, opened 2024-09-09
KFSPTS-32353 Backport FINP-10787 Award doc fix

cah292:
This PR fixes an Award-doc-related issue by backporting the FINP-10787 code changes. It only involved overlaying 1 file and adding a few new lines to it.

NOTE: The related user story originally requested backporting FINP-10710, but upon further analysis, it was determined that FINP-10787 should be backported instead.


db483 (approved):
Backport looks good.

PR #1645, opened 2024-09-10
KFSPTS-31658 Clean up config properties

cah292:
This PR removes some unneeded properties from our KFS `institutional-config.properties` file. Some pertained to a Rice auto-XML-ingestion feature that no longer exists in KFS, others pertained to Quartz jobs that were later converted into KFS batch jobs, and one pertained to a thread dump feature that was removed by FINP-7441.

Also, for consistency with our other application resource property files, the two tax-related `message.batchUpload.*` properties have been moved into the appropriate tax module file instead.

There are some unneeded `data.xml.*` properties in the KFS config property template in kuali-puppet as well. If you think a related kuali-puppet PR should be created as part of this cleanup work, please let me know.


db483 (approved):
Clean up looks good.

PR #1647, opened 2024-09-11
KFSPTS-29488 Remove Mellon references from Check Recon

cah292:

This PR removes the Mellon-specific portions of the Check Recon processing. The only non-unit-test portion that I could identify was the legacy date parsing that was only used for handling Mellon files. However, a large amount of rework was needed on the unit test side to remove the Mellon test files and to adjust things to reference the new bank. Some miscellaneous test cleanup was performed as well, such as updating the expected bank code and adjusting one of the tests to use a more convenient means of adding/deleting test directories. (The lattermost change was not made to the stale-checks test because it's still using JUnit 4 code; perhaps we could use a separate user story to bring that test up-to-date with JUnit 5.)


jhulslander (approved):
These changes look good, and so does the SQL, I shall merge

PR #1652, opened 2024-09-27
Potential fix for ACH Account Extract errors

cah292:
This PR adjusts the transaction handling for the Payee ACH Account Extract Job, in an attempt to fix some intermittent transaction-rollback-related issues that are starting to become more frequent. I'm unable to replicate the issue and I haven't found the specific cause of the issue, so I'm not certain if these adjustments will fix it.

The data-updating code has now been moved into areas that run in their own separate transactions (rather than in a shared transaction), and the `GlobalVariables` handling has been modified to use the `doInNewGlobalVariables()` helper method instead. Hopefully this will help minimize or eliminate the intermittent transaction rollback errors.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
@@ -13,26 +13,21 @@ import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.kuali.kfs.core.api.datetime.DateTimeService;
 import org.kuali.kfs.coreservice.framework.parameter.ParameterService;
-import org.kuali.kfs.krad.service.BusinessObjectService;
-import org.kuali.kfs.krad.util.GlobalVariables;
-import org.kuali.kfs.krad.util.KRADPropertyConstants;
+import org.kuali.kfs.kim.api.identity.PersonService;
+import org.kuali.kfs.kim.impl.identity.Person;
 import org.kuali.kfs.krad.util.ObjectUtils;
 import org.kuali.kfs.pdp.PdpConstants.PayeeIdTypeCodes;
-import org.kuali.kfs.pdp.businessobject.PayeeACHAccount;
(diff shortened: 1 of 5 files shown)


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
```diff
@@ -366,29 +345,18 @@ protected String processACHBatchDetail(PayeeACHAccountExtractDetail achDetail) {
...
-        processingError = addOrUpdateACHAccountIfNecessary(payee, achDetail, PayeeIdTypeCodes.EMPLOYEE, payee.getEmployeeId());
+        processingError = payeeACHAccountDocumentService.addOrUpdateACHAccountIfNecessary(
+                payee, achDetail, PayeeIdTypeCodes.EMPLOYEE, payee.getEmployeeId());
         
         return processingError;
     }
 
-    protected String addOrUpdateACHAccountIfNecessary(
-            Person payee, PayeeACHAccountExtractDetail achDetail, String payeeType, String payeeIdNumber) {
-        GlobalVariables.getMessageMap().clearErrorMessages();
```
jhulslander, line 381: would it be worth clearing error messages in the new version of this function, in case any down steam processing expected this function to clear any previous errors?
I know the new version makes the achService call in a new global variables, but maybe it would be good to clear errors for other processing?

jhulslander (changes requested):
I had one question, also, we should update the PR description with a KFSPTS number before we squash merge this code to develop

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/PayeeACHAccountExtractServiceImpl.java
cah292: The old code was making this `clearErrorMessages()` call prior to each attempted ACH account insertion or update, so I don't see what other downstream processing would need to have the error messages retained beyond the insert/update attempt. Also, there are only a small number of areas in base code that are calling `clearErrorMessages()` directly (likely because they haven't been refactored to use the do-in-new-variables approach), so I would advise against including this direct call unless you know of a specific use case.

jhulslander (approved):
Thanks, I am not sure I expressed my concern correctly, but you did answer it indirectly.
I am fine moving forward with this.  I will add KFSPTS-32797 to the commit message and merge this into develop

PR #1653, opened 2024-10-01
KFSPTS-27162 Default Role Member Active From Date

ajd299:


cah292:
@ajd299 I realize that Jay is still reviewing this (and I'm sorry if this duplicates anything that Jay has already discovered), but I noticed something that I wanted to bring to your attention:

The user story is asking for the new Active From Date handling on KSR documents, but your current code changes are introducing such handling on the Person document instead. Further changes are needed to get this working for KSR documents as well. (KSR docs rely upon the Role and/or Responsibility APIs, so I'm not sure if they support setting Active From Date or if you'll need additional handling to set those dates.)

Based on the user story, it sounds like users are manually adding the Active From Date on Person documents. If you think automatic setup of such dates should be expanded to the Person doc, please check with the functionals first to verify if they truly want that added behavior.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
```diff
@@ -482,8 +486,20 @@ public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEven
         if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
             setIfGroupsEditable();
             setIfRolesEditable();
+
+            for (final PersonDocumentRole role : roles) {
+                for (final KimDocumentRoleMember rolePrncpl : role.getRolePrncpls()) {
+                    if (ObjectUtils.isNull(rolePrncpl.getActiveFromDate())) {
```
jhulslander, line 492: I am not sure the best approach just yet, but I think there might be an attribute of rolePrncpl that might work for us.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
```diff
@@ -482,8 +486,20 @@ public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEven
...
+                        Timestamp timestampNow = new Timestamp(new java.util.Date().getTime());
```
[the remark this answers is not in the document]
ajd299: okay thank you

ajd299:
@jhulslander @cah292 Please see my latest commits

- Do we need to confirm with functionals that it is correct to default the RoleMember ActiveFromDate everywhere regardless of where it is created (PersonDoc, KSRDoc, RoleDoc)? I assume that is desired.

jhulslander:
@ajd299 ,
I read the requirements as asking for the active from date to be set from all three like you did.  It sounds like Chad read it differently so it would be good to confirm the requirements

cah292:
Based on the user story's description, it sounds like the functionals are already handling the Action From Date manually on Person and Role docs, but they want the KSR documents to auto-handle such setup since the functionals can't edit the Active From Date from those documents. If you think the automatic Active From Date handling should be extended to the Person and Role documents as well, please check with the functionals to verify if that's what they want.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementRoleDocument.java
```diff
@@ -261,6 +261,11 @@ public void addMember(final KimDocumentRoleMember member) {
         final String roleMemberId = nextSeq.toString();
         member.setRoleMemberId(roleMemberId);
         setupMemberRspActions(member);
+
+        if (ObjectUtils.isNull(member.getActiveFromDate())) {
+            member.setActiveFromDate(getDateTimeService().getCurrentTimestamp());
```
cah292, line 266: If the functionals want Role documents to auto-initialize the Active From Date when necessary, I would suggest moving such initialization to the `doRouteStatusChange()` method override so that it's performed when the document finishes routing, not when the member is first added to the document.

src/main/java/org/kuali/kfs/kim/impl/role/RoleServiceImpl.java
```diff
@@ -1769,6 +1770,9 @@ public RoleMember assignPrincipalToRole(
         newRoleMember.setMemberId(principalId);
         newRoleMember.setType(MemberType.PRINCIPAL);
 
+        DateTimeService dateTimeService = SpringContext.getBean(DateTimeService.class);
```
cah292, line 1773: `RoleServiceImpl` already has a `dateTimeService` field, so you don't need to manually retrieve the service.

cah292 (commented):
I had a few more suggestions on top of what I had mentioned in my previous comment.

ajd299:
Okay i will clarify with the functionals

ajd299:
@cah292 @jhulslander Please see my latest commits

src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
```diff
@@ -482,8 +484,19 @@ public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEven
         if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
             setIfGroupsEditable();
             setIfRolesEditable();
+
+            for (final PersonDocumentRole role : roles) {
```
cah292, line 488: Since this is a modification to an overlayed file, I would suggest adding a comment to mark this spot as a customized one.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementRoleDocument.java
```diff
@@ -338,7 +338,15 @@ protected void setAttrDefnIdForDelMemberQualifier(
     @Override
     public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEvent) {
         super.doRouteStatusChange(statusChangeEvent);
+        
         if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
+
+            for (final KimDocumentRoleMember member : getMembers()) {
```
cah292, line 344: Another spot that should be marked as a customization.

src/main/java/org/kuali/kfs/kim/impl/role/RoleServiceImpl.java
```diff
@@ -1769,6 +1770,8 @@ public RoleMember assignPrincipalToRole(
         newRoleMember.setMemberId(principalId);
         newRoleMember.setType(MemberType.PRINCIPAL);
 
+        newRoleMember.setActiveFromDateValue(dateTimeService.getCurrentTimestamp());
```
cah292, line 1773: Another spot that should be marked as a customization.

src/main/java/org/kuali/kfs/kim/impl/role/RoleServiceImpl.java
```diff
@@ -52,6 +52,7 @@
 import org.kuali.kfs.kim.impl.type.KimType;
 import org.kuali.kfs.kim.util.KimCommonUtilsInternal;
 import org.kuali.kfs.krad.service.KRADServiceLocator;
+import org.kuali.kfs.sys.context.SpringContext;
```
cah292, line 55: Is this import still needed?

cah292 (commented):
I had a few minor suggestions for improvement, but otherwise the revisions look good.

cah292 (commented):
I'm wondering whether the larger customized blocks should also have an end-of-customization comment. I'm fine either way since the larger blocks have a small-ish `for` loop as a wrapper. I'll defer to Jay as to whether he thinks extra comments are needed on those blocks.

Apart from that, the changes look good to me.

ajd299:
@jhulslander @cah292 Please see my latest commit

cah292 (commented):
Changes look good to me. If Jay also approves of the changes, then this PR should be ready to merge.

src/main/java/org/kuali/kfs/kim/document/IdentityManagementPersonDocument.java
```diff
@@ -482,6 +484,17 @@ public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEven
         if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
             setIfGroupsEditable();
             setIfRolesEditable();
+
+            // ==== CU Customization KFSPTS-27162 Set default for activeFromDate field ===
+            for (final PersonDocumentRole role : roles) {
+                for (final KimDocumentRoleMember rolePrncpl : role.getRolePrncpls()) {
+                    if (ObjectUtils.isNull(rolePrncpl.getActiveFromDate())) {
+                        rolePrncpl.setActiveFromDate(getDateTimeService().getCurrentTimestamp());
```
jhulslander, line 492: With this solution this will update all the role active from dates to "today" that don't previously have an active from date.  Should we check with the functionals on this?
I think we only want to set the active from date on new records

jhulslander (commented):
I have one question that we should check with the functionals on.

ajd299:
I modified the code to only set the activeFromDate field when a new RoleMember is created.

jhulslander (approved):
This seems reasonable, I shall merge

PR #1654, opened 2024-10-02
KFSPTS-24998 Add role inquiry links to KSR docs

cah292:
This PR adds icons/links to KSR documents that allow for opening the associated Roles' inquiries in a new tab/window. Because Role inquiries take a long time to render or potentially time out if they have a large number of members (due to the inquiries displaying a non-paged members list), I configured the KSR docs to render only the open-in-new-tab links, rather than additionally rendering the usual open-in-modal links. That's why the proposed changes do NOT use the `kul:inquiry` tag.

The changes have been configured to be similar to other areas of KFS that render the open-in-new-tab links directly.

### DIFF
--- a/src/main/webapp/WEB-INF/tags/ksr/securityRequestRole.tag
+++ b/src/main/webapp/WEB-INF/tags/ksr/securityRequestRole.tag
@@ -9,6 +9,7 @@
 <c:set var="securityRequestRole" value="${KualiForm.document.securityRequestRoles[securityRequestRoleIndex]}" />
 <c:set var="roleTitle" value="${securityRequestRole.roleInfo.id} : ${securityRequestRole.roleInfo.namespaceCode} - ${securityRequestRole.roleInfo.name}"/>
 <c:set var="roleInquiryUrl" value="${ConfigProperties.application.url}/inquiry.do?methodToCall=start&businessObjectClassName=org.kuali.kfs.kim.impl.role.Role&id=${securityRequestRole.roleInfo.id}&mode=standalone"/>
+<c:set var="roleLinkTitle" value="Open details for ${securityRequestRole.roleInfo.namespaceCode} - ${securityRequestRole.roleInfo.name} in new tab"/>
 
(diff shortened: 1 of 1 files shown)


src/main/webapp/WEB-INF/tags/ksr/securityRequestRole.tag
```diff
@@ -8,12 +8,18 @@
...
+<c:set var="roleInquiryUrl" value="${ConfigProperties.application.url}/inquiry.do?methodToCall=start&businessObjectClassName=org.kuali.kfs.kim.impl.role.Role&id=${securityRequestRole.roleInfo.id}&mode=standalone"/>
 
 <h3>&nbsp;</h3>
 <table width="100%" border="0" cellpadding="0" cellspacing="0" class="datatable">
   <tr>
     <th colspan="2" width="100%">
-      <b><c:out value="${roleTitle}"/></b>
+      <b>
+        <c:out value="${roleTitle}"/>
+        <a href="<c:out value='${roleInquiryUrl}'/>" target="_blank" title="Open in new tab" class="new-window" onclick="event.stopPropagation();">
```
jhulslander, line 19: in the interest of being nice to screen readers, maybe the anchor title should include the role name, something like "Open details for role ROLE-NAME in a new tab"
Also, a space between the c:out and the anchor might be good

jhulslander (changes requested):
this seems reasonable, I had a few items to consider

src/main/webapp/WEB-INF/tags/ksr/securityRequestRole.tag
cah292: I don't think the other KFS new-tab-link anchors include extra info about what's being linked to, but I agree that it would be helpful to include that. I'll see if I can adjust the title accordingly.

I'll also add a non-breaking space between the printed title and the link.

cah292:
I've committed some changes to address the items that were brought up. This is ready for review again.

jhulslander (approved):

PR #1655, opened 2024-10-07
KFSPTS-32170 Remove rest of REQS CapAsset tab customization

cah292:
We used to have a customization that modified the Fiscal Officer's field-editing permissions on REQS documents, which was largely removed by KFSPTS-19994 but was later discovered to still potentially have a few lingering portions left in our code (as per KFSPTS-32026). This PR removes the rest of the customization's remaining portions that were affecting the REQS UI.

NOTE: Our `Requisition.jsp` overlay still contained carriage returns, so I went ahead and removed those from the file as well; that's why it looks as though the whole file was edited. Aside from the whitespace changes, the file was primarily updated to revert to the base code logic for handling the full-entry-mode checks.


jhulslander (approved):

PR #1657, opened 2024-10-11
KFSPTS-32374 Add batch job to copy legacy acct attachments

cah292:
There are PRs for this in both cu-kfs and nonprod-sql. Please make sure both are good to go before merging.

This PR Introduces a temporary KFS batch job that will retrieve account attachments from the DFA Legacy system and attach them to the corresponding KFS accounts. It relies upon a remote attachment download endpoint and a pre-populated KFS table containing the attachment mappings, both of which were already created by earlier user stories.

Since KualiCo is moving away from JAX-RS, I implemented the web client handling using Spring's `WebClient` instead. (I would prefer not to use Spring's `RestTemplate` since it's an older API that may eventually get phased out, and I can't use Spring's `RestClient` because we're not on Spring 6.x yet.)

The note text for copied attachments may need a few adjustments, since its line breaks are being displayed as regular spaces on the KFS inquiry screens. I tried using CSS to preserve the line breaks, but doing so also preserves a large number of other unwanted line breaks that the jsp/tag files are inserting within the same HTML tag. Please let me know if you have suggestions on some alternative formatting (comma separations, semicolon separators, etc.).

Note that are still a few comma-related or whitespace-related areas that I need to clean up as well.


src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,202 @@
...
+                CuCOAKeyConstants.LEGACY_ACCOUNT_ATTACHMENT_NOTE_TEXT_TEMPLATE);
+        final String noteText = String.format(Locale.US, noteTemplate, legacyAccountAttachment.getLegacyAccountCode(),
+                legacyAccountAttachment.getAddedBy(), legacyAccountAttachment.getFileDescription());
+        return StringUtils.left(noteText, noteTextMaxLength);
+    }
+
+    private int getParameterValueAsInteger(final String parameterName) {
+        final String parameterValue = parameterService.getParameterValueAsString(
+                CopyLegacyAccountAttachmentsStep.class, parameterName);
+        return Integer.parseInt(parameterValue);
```
jhulslander, line 151: Would it be helpful to log the parameter values the first time they are retrieved for the job

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,202 @@
...
+            LOG.info("copyLegacyAccountAttachmentsToKfs, There are no legacy attachments available for copying");
+            return true;
+        }
+
+        LOG.info("copyLegacyAccountAttachmentsToKfs, Copying {} legacy account attachments into KFS",
+                attachmentsToCopy.size());
+        final List<LegacyAccountAttachment> copiedAttachments = new ArrayList<>(attachmentsToCopy.size());
+        final List<LegacyAccountAttachment> copyFailures = new ArrayList<>();
+        
+        for (final LegacyAccountAttachment attachmentToCopy : attachmentsToCopy) {
```
jhulslander, line 70: Do you think downloading an attachment, adding a BO note to the account and updating the status of the LegacyAccountAttachment should be in a transactional block?
That would avoid a condition where an account has an attachment added but we don't update the LegacyAccountAttachment status.  That would also prevent subsequent errors preventing updates from saving.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/DownloadLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.web.reactive.function.BodyExtractors;
+import org.springframework.web.reactive.function.client.ClientResponse;
+import org.springframework.web.reactive.function.client.WebClient;
+
+import edu.cornell.kfs.coa.batch.CuCoaBatchConstants;
+import edu.cornell.kfs.coa.batch.businessobject.LegacyAccountAttachment;
+import edu.cornell.kfs.coa.batch.service.DownloadLegacyAccountAttachmentsService;
+import edu.cornell.kfs.sys.service.WebServiceCredentialService;
+
+public class DownloadLegacyAccountAttachmentsServiceImpl implements DownloadLegacyAccountAttachmentsService {
```
jhulslander, line 25: I am wondering if re-using AccountingXmlDocumentDownloadAttachmentServiceImpl.downloadByteArray() might simplify this service.

src/main/java/edu/cornell/kfs/sys/dataaccess/DtoPropertyHandler.java
```diff
@@ -0,0 +1,47 @@
+package edu.cornell.kfs.sys.dataaccess;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.function.BiConsumer;
+
+import org.apache.commons.lang3.function.FailableBiFunction;
+
+public final class DtoPropertyHandler<T, V> {
```
jhulslander, line 9: I commented earlier about what this new class buys us

src/main/java/edu/cornell/kfs/sys/util/CuAttachmentUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.net.URLConnection;
+import java.util.Locale;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.kuali.kfs.sys.KFSConstants;
+
+public final class CuAttachmentUtils {
```
jhulslander, line 11: There is a find mime type function in AccountingXmlDocumentDownloadAttachmentServiceImpl

src/main/resources/edu/cornell/kfs/coa/cu-coa-resources.properties
```diff
@@ -1,2 +1,6 @@
 error.document.accountMaintenance.icrTypeCodeCannotBeEmpty=When {0} is {1}, at least one ICR Type Code is required.
 error.document.accountMaintenance.icrSeriesIdentifierCannotBeEmpty=When {0} is {1}, at least one ICR Series Identifier is required.
+
+legacy.account.attachment.note.text.template=Legacy Account Number: %1$s\
```
jhulslander, line 4: I don't see where the functionals agreed with my proposed note message for these attachments, we should confirm that.

jhulslander (changes requested):
I found a few items to consider

src/main/java/edu/cornell/kfs/coa/batch/businessobject/LegacyAccountAttachment.java
```diff
@@ -0,0 +1,105 @@
+package edu.cornell.kfs.coa.batch.businessobject;
+
+import org.kuali.kfs.core.api.util.Truth;
+import org.kuali.kfs.krad.bo.TransientBusinessObjectBase;
+
+public class LegacyAccountAttachment extends TransientBusinessObjectBase {
```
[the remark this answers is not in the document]
cah292: Will override toString() accordingly.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
```diff
@@ -0,0 +1,116 @@
...
+        return legacyAccountAttachment;
+    }
+
+    @Override
+    public void markLegacyAccountAttachmentsAsCopied(final List<LegacyAccountAttachment> legacyAccountAttachments) {
+        final CuSqlQuery sqlQuery = buildQueryForMarkingAccountAttachmentsAsCopied(legacyAccountAttachments);
+        final int numUpdatedRows = executeUpdate(sqlQuery);
+        LOG.info("markLegacyAccountAttachmentsAsCopied, {} attachments were marked as copied", numUpdatedRows);
+        if (numUpdatedRows != legacyAccountAttachments.size()) {
+            LOG.warn("markLegacyAccountAttachmentsAsCopied, {} attachments were meant to be marked as copied, "
```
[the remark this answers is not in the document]
cah292: The query is using a single `UPDATE` statement to modify the account attachment entries in the given batch. If something unexpected happens to the data that's expected to exist, or if something odd happens with the attachment-handling logic, this log statement will help with alerting us about the discrepancy.

If you think the risk of mismatched row counts is low enough to not be worth reporting, please let me know.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,202 @@
...
+        LOG.info("====================");
+
+        return copiedAttachments.size() > 0;
+    }
+
+    private boolean copyLegacyAccountAttachmentToKfs(final LegacyAccountAttachment legacyAccountAttachment) {
+        try {
+            downloadLegacyAccountAttachmentsService.downloadAndProcessLegacyAccountAttachment(
+                    legacyAccountAttachment, this::saveAttachmentToKfs);
+            LOG.info("copyLegacyAccountAttachmentToKfs, Copied attachment '{}' targeting account '{}-{}'",
```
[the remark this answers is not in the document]
cah292: I'll adjust this log statement for clarity.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: I'm not sure if logging these values would do much for this job, given that only a couple are needed for the job and the value of one of them is likely to be reflected in the retrieved-batch-size log entries. If you think they would still add significant value, I could add them.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: The job is currently running in one large transaction, but I'll look into adjusting things so that each account attachment is handled in a new transaction that's not tied to the large one.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/DownloadLegacyAccountAttachmentsServiceImpl.java
cah292: That `downloadByteArray()` method was implemented using the JAX-RS client libraries. Because KualiCo is replacing JAX-RS with Spring MVC, I was trying to avoid using JAX-RS here.

I could use Spring's `RestTemplate` instead, but I think Spring is encouraging people to use `WebClient` or `RestClient` instead (where the latter requires Spring 6.x).

src/main/java/edu/cornell/kfs/sys/dataaccess/DtoPropertyHandler.java
cah292: Will update or remove this class accordingly as part of the refactoring work.

src/main/java/edu/cornell/kfs/sys/util/CuAttachmentUtils.java
cah292: As noted in the file's comments, I specifically copied that method's logic into this file. Should I create a future user story to merge our MIME type handling into a single unified location?

src/main/resources/edu/cornell/kfs/coa/cu-coa-resources.properties
cah292: I'll check with Cathy when she returns this week.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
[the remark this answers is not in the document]
jhulslander: with other re-work in progress, does this go away?
I am not saying we should get rid of the check, I was just curious how that would happen

src/main/java/edu/cornell/kfs/coa/batch/service/impl/DownloadLegacyAccountAttachmentsServiceImpl.java
cah292: Upon further analysis, I think I can reuse AccountingXmlDocumentDownloadAttachmentServiceImpl to handle the downloading if I change its primary method to accept more than just `Document` objects. I'll rework things to use this service instead.

Also, we already have KFSPTS-30451 to convert the existing service over to using Spring MVC.

cah292:
I've committed some changes based on the initial feedback. I'm still waiting for the functionals to verify what note text they want, but other than that the rest of the changes are ready for review again.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
```diff
@@ -0,0 +1,88 @@
...
+        return attachment;
+    }
+
+    @Override
+    public void markLegacyAccountAttachmentAsCopied(final LegacyAccountAttachment legacyAccountAttachment) {
+        final Long id = legacyAccountAttachment.getId();
+        final CuSqlQuery sqlQuery = new CuSqlChunk()
+                .append("UPDATE KFS.TEMP_ACCT_ATTACH_FOR_COPYING")
+                .append(" SET COPIED_IND = ").appendAsParameter(KRADConstants.YES_INDICATOR_VALUE)
+                .append(" WHERE COPYING_ACCT_ATTACH_ID = ").appendAsParameter(Types.BIGINT, id)
```
jhulslander, line 69: should this update the LATEST_ERR_MSG to an empty string?

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,232 @@
...
+        final List<LegacyAccountAttachment> attachmentsToCopy = copyLegacyAccountAttachmentsDao
+                .getLegacyAccountAttachmentsToCopy(batchSize, maxRetries);
+        if (attachmentsToCopy.size() == 0) {
+            LOG.info("copyLegacyAccountAttachmentsToKfs, There are no legacy attachments available for copying");
+            return true;
+        }
+
+        LOG.info("copyLegacyAccountAttachmentsToKfs, Copying {} legacy account attachments into KFS",
+                attachmentsToCopy.size());
+        final CopyLegacyAccountAttachmentsService proxiedService = SpringContext.getBean(
```
jhulslander, line 74: Why do we need a new instance of this service from SpringContext?  Can't we use the current instance of the service?

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+        return account;
+    }
+
+    private Attachment downloadAttachment(final LegacyAccountAttachment legacyAccountAttachment,
+            final Account account) throws URISyntaxException {
+        final AccountingXmlDocumentBackupLink backupLink = new AccountingXmlDocumentBackupLink();
+        backupLink.setCredentialGroupCode(CuCoaBatchConstants.DFA_ATTACHMENTS_GROUP_CODE);
+        backupLink.setFileName(legacyAccountAttachment.getFileName());
+        backupLink.setDescription(legacyAccountAttachment.getFileDescription());
```
jhulslander, line 138: Should this be the formatted note text discussed on the jira?

jhulslander (changes requested):
Changes look good, I had a question and a few ideas for improvement

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public void recordCopyingErrorForLegacyAccountAttachment(final LegacyAccountAttachment legacyAccountAttachment,
+            final String errorMessage) {
+        final Long id = legacyAccountAttachment.getId();
+        final String errorMessageToStore = StringUtils.left(errorMessage, MAX_ERROR_MESSAGE_LENGTH);
+        final CuSqlQuery sqlQuery = new CuSqlChunk()
+                .append("UPDATE KFS.TEMP_ACCT_ATTACH_FOR_COPYING")
+                .append(" SET RETRY_COUNT = RETRY_COUNT + 1,")
+                .append(" LATEST_ERR_MSG = ").appendAsParameter(errorMessageToStore)
```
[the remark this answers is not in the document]
cah292: I'm fine with having the error message cleared out when the row successfully copies, so I'll adjust the SQL accordingly.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
cah292: I'm fine with having the error message cleared out when the row successfully copies, so I'll adjust the SQL accordingly.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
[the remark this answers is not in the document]
cah292: The line that defines the `errorMessageToStore` variable already handles truncating the error message to fit the column.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: Based on your prior review, I updated the code so that each legacy account attachment will be processed in its own transaction. Since I placed the transaction-annotated method on the same service class, the method has to be invoked on the transaction-proxied service so that the new transaction will be created and committed/rolled back accordingly.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: The attachment-downloading service doesn't seem to be actively using this description, so I don't think it matters much in this case. Nevertheless, I could look into restructuring this to use the intended note text.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: I'll add a method that wraps the `SpringContext` call and clarifies its intent.

cah292:
I've committed some changes to rework the appropriate areas of concern. This is ready for review again.

nkimble (commented):
I have reviewed both the code and the SQL for this user story. I did have a few items to consider for possible improvements.

I did have a question regarding the attachments that have reached the 5 attempt processing limit. What mechanism is intended to be used to find and address those items when that threshold is reached on any of them?

I am also having trouble seeing how to have this batch job execute in a manner similar to the way SY-99-05 (Historical Purge Job) was run (i.e. repeated executions in specific time window). I know I must be missing something and will continue to review tomorrow.

Specifically, I am trying to see how the following can be applied from the historical purge job:  
Prod control was able to setup Tidal to run repeatedly in a set time window and to stop running with EITHER a force failure if the job ran faster than a specific amount of time OR be configured to execute N number of times when that job iteration value could be accurately determined. Either of these approaches is needed to prevent a large number of empty log files from being created when there is no data to process during the repeated executions.

src/main/java/edu/cornell/kfs/coa/batch/dataaccess/impl/CopyLegacyAccountAttachmentsDaoJdbc.java
cah292: The queries will set `COPIED_IND` to `'Y'` when download succeeds, or will increment the `RETRY_COUNT` in the event of a failure (or a repeated failure). Those two columns should be enough to help identify the processing state of a particular attachment.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,237 @@
...
+            final AccountingXmlDocumentBackupLink backupLink = createBackupLinkForAttachmentDownload(
+                    legacyAccountAttachment);
+            final Attachment attachment = accountingXmlDocumentDownloadAttachmentService
+                    .createAttachmentFromBackupLink(account, backupLink);
+            saveAttachmentToKfs(attachment, account, backupLink.getDescription());
+            LOG.info("copyLegacyAccountAttachmentToKfs, Copied attachment with ID '{}' targeting account '{}-{}'",
+                    legacyAccountAttachment.getId(), legacyAccountAttachment.getKfsChartCode(),
+                    legacyAccountAttachment.getKfsAccountNumber());
+            copyLegacyAccountAttachmentsDao.markLegacyAccountAttachmentAsCopied(legacyAccountAttachment);
+            return true;
```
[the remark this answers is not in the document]
cah292: I'll look into adjusting the log statements accordingly.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,237 @@
...
+
+    private CopyLegacyAccountAttachmentsService getThisServiceAsProxyToProcessAttachmentsInNewTransactions() {
+        return SpringContext.getBean(CopyLegacyAccountAttachmentsService.class);
+    }
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @Override
+    public boolean copyLegacyAccountAttachmentToKfs(final LegacyAccountAttachment legacyAccountAttachment) {
+        try {
+            LOG.debug("copyLegacyAccountAttachmentToKfs, Copying attachment: {}", legacyAccountAttachment);
```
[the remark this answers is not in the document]
cah292: I'm concerned about the logs containing too much "noise" if we print the entire attachment structure at this spot. (Also, the `catch` block is already set up to log the entire DTO contents when an exception occurs.) I could add an INFO message that logs the ID, Chart and Account (like in the post-success message), if that helps.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,237 @@
...
+            LOG.info("copyLegacyAccountAttachmentToKfs, Copied attachment with ID '{}' targeting account '{}-{}'",
+                    legacyAccountAttachment.getId(), legacyAccountAttachment.getKfsChartCode(),
+                    legacyAccountAttachment.getKfsAccountNumber());
+            copyLegacyAccountAttachmentsDao.markLegacyAccountAttachmentAsCopied(legacyAccountAttachment);
+            return true;
+        } catch (final Exception e) {
+            LOG.error("copyLegacyAccountAttachmentToKfs, Failed to copy attachment: {}", legacyAccountAttachment, e);
+            copyLegacyAccountAttachmentsDao.recordCopyingErrorForLegacyAccountAttachment(
+                    legacyAccountAttachment, e.getMessage());
+            return false;
```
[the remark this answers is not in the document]
cah292: Given the discovery that you made on KFSPTS-33040, I agree that it would be best for this method to re-throw the exception (or a wrapped exception) so that the rollback will happen accordingly. I'll adjust this part to do so, and I'll rework the logic so that the error-recording step will occur in bulk within the parent transaction (or a new standalone sub-transaction).

cah292:
Regarding the processing limit question: If an attachment fails to download too many times, then we'll need developer and/or DevOps analysis to determine if the attachment truly exists or if some network-related error prevented the download. If we determine that the row should indeed be processable, then we can make SQL-based corrections to fix any fields that need tweaking and/or adjust the upper retry count limit.

As for configuring the job for repeated executions, we still need further testing to determine how long it takes to complete for the proposed batch size (currently set to 300). A test run or two in Sandbox or DEV should help us determine the run duration and the approximate number of runs needed for full completion.

cah292:
I've committed some more changes to address the latest feedback. This is ready for review again.

nkimble:
My concerns have been addressed. Letting primary reviewer @jhulslander have opportunity to evaluate these latest code changes and perform the merge.

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    private String createNoteText(final LegacyAccountAttachment legacyAccountAttachment) {
+        final int noteTextMaxLength = dataDictionaryService.getAttributeMaxLength(Note.class,
+                KRADConstants.NOTE_TEXT_PROPERTY_NAME);
+        final String noteTemplate = configurationService.getPropertyValueAsString(
+                CuCOAKeyConstants.LEGACY_ACCOUNT_ATTACHMENT_NOTE_TEXT_TEMPLATE);
+        final String noteText = MessageFormat.format(noteTemplate, legacyAccountAttachment.getLegacyAccountCode(),
+                legacyAccountAttachment.getAddedBy(), legacyAccountAttachment.getFileDescription());
+        return StringUtils.left(noteText, noteTextMaxLength);
```
jhulslander, line 180: I think it is very unlikely the note text will be too long.  KFS Notes support 800 characters and the longes file description inn temp_acct_attach_for_copying is 75 characters.
However, if the length is too long, I think it would be best to error and let the functionals edit the file description.

jhulslander (changes requested):
Changes look good, I just found one thing to consider

src/main/java/edu/cornell/kfs/coa/batch/service/impl/CopyLegacyAccountAttachmentsServiceImpl.java
cah292: The file description is coming from a generated table that is not displayed in the KFS UI; there's no place for the functionals to "edit" the description unless they run SQL (or we run SQL for them).

I could change this to error out in too-long situations, but then we'd need manual SQL intervention to fix the file description. Thoughts?

jhulslander: Yes, I agree with that

cah292:
As requested, I've modified the note text handling to treat exceeds-max-length circumstances as errors. Note that the upstream code already logs more details about attachment entries that failed, so I didn't include that information in the new exception's text.

jhulslander (approved):
My concerns have been addressed, and the other reviews approved, I think this is ready to be merged

PR #1658, opened 2024-10-16
KFSPTS-32798 Filter unwanted chars from Concur SAE files

cah292:
This PR modifies the Concur SAE processing so that special characters or in-cell quotes will be filtered out prior to creating the PDP and Collector files. If characters do end up getting removed, the affected lines and column numbers will be recorded on the DTOs and will also be written to the report file.

For simplicity in checking for invalid characters, tabs in the SAE data will be silently converted to spaces. Please let me know if you think if the tabs should be left as-is or if you think they should be removed altogether. (I'll send a note to the functionals to confirm.)


src/main/java/edu/cornell/kfs/concur/batch/ConcurSaeDelimitedFlatFileSpecification.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        final ConcurStandardAccountingExtractLineBase saeLine =
+                (ConcurStandardAccountingExtractLineBase) parseIntoObject;
+        saeLine.setLineNumber(lineNumber);
+
+        final String[] cleanedSegments = new String[lineSegments.length];
+        int index = -1;
+        for (final String lineSegment : lineSegments) {
+            index++;
+            if (StringUtils.isEmpty(lineSegment)) {
```
db483, line 38: Would it be cleaner to change this to use the !StringUtils.isEmpty(lineSegment) to perform the logic bellow and then after, set the  cleanedSegments[index] = cleanedSegment; ? 

Also, would it be better to extract the code for cleaning the special characters into a separate method that is then called inside the for loop?

db483 (changes requested):
Changes look good but I had a couple of things to consider. Thank you.

src/main/java/edu/cornell/kfs/concur/batch/ConcurSaeDelimitedFlatFileSpecification.java
```diff
@@ -0,0 +1,56 @@
...
+
+import edu.cornell.kfs.concur.batch.businessobject.ConcurStandardAccountingExtractLineBase;
+import edu.cornell.kfs.sys.batch.CuDelimitedFlatFileSpecification;
+
+public class ConcurSaeDelimitedFlatFileSpecification extends CuDelimitedFlatFileSpecification {
+
+    private static final Logger LOG = LogManager.getLogger();
+
+    private static final Pattern TABS_PATTERN = Pattern.compile("[\\t]");
+    private static final Pattern REMOVABLE_CHARS_PATTERN = Pattern.compile("[^\\x20\\x21\\x23-\\x7E]");
```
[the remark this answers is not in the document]
cah292: This regex uses a `[^...]` character group, where the `^` at the start of the group negates it. This means that the pattern will match any characters that are NOT in the group.

I explicitly excluded `\x22` (double quotes) from the negated group, so the regex should affect double quotes as well.

src/main/java/edu/cornell/kfs/concur/batch/ConcurSaeDelimitedFlatFileSpecification.java
cah292: I'll look into refactoring this for clarity and to add helper methods as needed.

cah292:
I've committed some changes to address the concerns that were brought up. I also improved the reporting for tabs that get replaced by spaces.

One other thing: The functionals are fine with how we're proposing to remove/replace the various characters and on how the changes are being reported, so this PR can be merged once it's been approved.

src/main/java/edu/cornell/kfs/concur/batch/ConcurSaeDelimitedFlatFileSpecification.java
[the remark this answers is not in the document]
db483: That makes sense, thank you for explaining.

src/main/java/edu/cornell/kfs/concur/batch/ConcurSaeDelimitedFlatFileSpecification.java
```diff
@@ -35,22 +34,29 @@ public void parseLineIntoObject(
...
-            }
-            String cleanedSegment = TABS_PATTERN.matcher(lineSegment).replaceAll(KFSConstants.BLANK_SPACE);
-            final Matcher removableCharsMatcher = REMOVABLE_CHARS_PATTERN.matcher(cleanedSegment);
-            if (removableCharsMatcher.find()) {
-                LOG.warn("parseLineIntoObject, Found special characters or extra quotes on SAE line {} at column {}",
-                        lineNumber, index + 1);
-                saeLine.addColumnNumberContainingSpecialCharacters(index + 1);
-                removableCharsMatcher.reset();
-                cleanedSegment = removableCharsMatcher.replaceAll(KFSConstants.EMPTY_STRING);
+            String cleanedSegment;
```
db483, line 37: The updated code looks good but I think most of the code inside the for look could go into a separate method called something along the lines 'cleanSegment' or 'cleanText'? Thank you

cah292: I've moved more of the logic into another method as requested.

cah292:
I've committed some more changes. This is ready for review again.

db483 (approved):
Changes look good.

PR #1659, opened 2024-10-24
KFSPTS-29347: Upgrade to the 04/19/2023 version of financials.

nkimble:


cah292 (changes requested):
I'm still reviewing the code changes, but here are some items I've noticed thus far that may need attention. I'll also add a separate comment regarding one other needed change whose file has not yet been updated by this PR.

cah292 (changes requested):
In addition to the items noted above, I found something else that should be addressed:

In `Asset.xml`, we're currently adding some custom search fields to the Asset Lookup. Because this patch is configuring Assets to use the new lookup framework, we need to adjust our bean/field overrides so that the custom search fields will also be added to the converted Asset Lookup.

nkimble (commented):
Requested changes to PaymentRequestStatus.java had a ripple effect requiring adjustments to other classes with dependencies on those modifications.

src/main/resources/edu/cornell/kfs/vnd/businessobject/datadictionary/VendorDetail.xml
```diff
@@ -410,7 +383,11 @@
 	</bean>
 
 	<!-- Business Object Lookup Definition -->
-   <bean parent="DataDictionaryBeanOverride">
+    
+    <!-- KualiCo release 2024-04-19 of FINP-9046 replaced Cornell's similar customization but base code
+         did not include the "Default Payment Method" drop down without inquiry on Vendor Lookup. 
+         Adding that back in here.-->
+    <bean parent="DataDictionaryBeanOverride">
```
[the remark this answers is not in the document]
nkimble: I think what you are describing is what we want to happen (...have the inserted item be last in the list...) so this issue was not obvious. Thanks for catching it. I think I have fixed the problem and have changed the compare element appropriately.

nkimble (commented):
More review fixes.

cah292 (changes requested):
I'm still reviewing this, but I've found a few more areas needing attention.

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
```diff
@@ -73,7 +73,7 @@ public boolean confirmInvoiceNoMatchOverride(final AccountsPayableDocument accou
 	protected boolean checkWireTransferTabState(final VendorCreditMemoDocument cmDocument) {
```
cah292, line 73: Could this method be replaced with a call to the equivalent method on `PaymentSourcePreRulesService`?

I realize that the other service's method seems to be tailoring the question message for DV usage, despite base code also using that method for handling the equivalent PREQ logic. To keep the bulk of the logic consolidated in one place, I would suggest creating a `CuPaymentSourcePreRulesServiceImpl` subclass that adjusts the question message on a per-document basis, using a different config property for the base message if necessary. (I can look into preparing replication steps for the DV-tailored-message issue so that it can be reported to KualiCo.)

Thoughts?

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
```diff
@@ -97,37 +97,37 @@ protected boolean checkWireTransferTabState(final VendorCreditMemoDocument cmDoc
         return tabStatesOK;
     }
 
-    protected boolean hasWireTransferValues(final CreditMemoWireTransfer cmWireTransfer) {
+    protected boolean hasWireTransferValues(final PaymentSourceWireTransfer wireTransfer) {
```
cah292, line 100: Another method that could be replaced with `PaymentSourcePreRulesService` usage.

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
```diff
@@ -97,37 +97,37 @@ protected boolean checkWireTransferTabState(final VendorCreditMemoDocument cmDoc
...
-    	cmWireTransfer.setCmBankRoutingNumber(null);
-    	cmWireTransfer.setCmBankCityName(null);
-    	cmWireTransfer.setCmBankStateCode(null);
-    	cmWireTransfer.setCmBankCountryCode(null);
-    	cmWireTransfer.setCmPayeeAccountNumber(null);
-    	cmWireTransfer.setCmAttentionLineText(null);
-    	cmWireTransfer.setCmCurrencyTypeName(null);
-    	cmWireTransfer.setCmAdditionalWireText(null);
-    	cmWireTransfer.setCmPayeeAccountName(null);
+    protected void clearWireTransferValues(final PaymentSourceWireTransfer wireTransfer) {
```
cah292, line 119: Another method that could be replaced with `PaymentSourcePreRulesService` usage.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -59,76 +57,39 @@ protected CuDisbursementVoucherTaxService getCuDisbursementVoucherTaxService(){
...
-        hasValues |= StringUtils.isNotBlank(wireExtension.getDisbVchrCorrespondentBankName());
-        hasValues |= StringUtils.isNotBlank(wireExtension.getDisbVchrCorrespondentBankAddress());
-        hasValues |= StringUtils.isNotBlank(wireExtension.getDisbVchrCorrespondentBankRoutingNumber());
-        hasValues |= StringUtils.isNotBlank(wireExtension.getDisbVchrCorrespondentBankAccountNumber());
-        hasValues |= StringUtils.isNotBlank(wireExtension.getDisbVchrCorrespondentBankSwiftCode());
+    protected boolean hasWireTransferValues(final PaymentSourceWireTransfer wireTransfer) {
+        boolean hasValues = super.hasWireTransferValues(wireTransfer);
+        final PaymentSourceWireTransferExtendedAttribute wireExtension =
+                (PaymentSourceWireTransferExtendedAttribute) wireTransfer.getExtension();
+        hasValues |= StringUtils.isNotBlank(wireExtension.getBankStreetAddress());
```
cah292, line 65: If you do end up creating a custom `PaymentSourcePreRulesService` implementation, it would probably be best to move this extended attribute handling into the equivalent method on that service.

src/main/java/edu/cornell/kfs/fp/document/validation/impl/CuDisbursementVoucherDocumentPreRules.java
```diff
@@ -59,76 +57,39 @@ protected CuDisbursementVoucherTaxService getCuDisbursementVoucherTaxService(){
...
-        wireExtension.setDisbVchrCorrespondentBankName(null);
-        wireExtension.setDisbVchrCorrespondentBankAddress(null);
-        wireExtension.setDisbVchrCorrespondentBankRoutingNumber(null);
-        wireExtension.setDisbVchrCorrespondentBankAccountNumber(null);
-        wireExtension.setDisbVchrCorrespondentBankSwiftCode(null);
+    protected void clearWireTransferValues(final PaymentSourceWireTransfer wireTransfer) {
+        super.clearWireTransferValues(wireTransfer);
+        final PaymentSourceWireTransferExtendedAttribute wireExtension =
+                (PaymentSourceWireTransferExtendedAttribute) wireTransfer.getExtension();
+        wireExtension.setBankStreetAddress(null);
```
cah292, line 83: Another area that could be simplified by moving the logic into a new `PaymentSourcePreRulesService` implementation.

cah292 (changes requested):
I had a few more suggestions for improvement, especially with regards to the pre-rules handling on the wire-transfer-related documents.

cah292 (changes requested):
I found a few other items that may need further work.

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
cah292: Just as a quick follow-up, I've created KFSPTS-33228 to get the DV-message-on-PREQ issue reported back to KualiCo.

nkimble:
The changes for commit bef5801ac2494effcd0990196541f9a567682c36 only addresses adding Cornell extended attributes to the Asset Lookup page in the new frameworks. Searching from the Asset Lookup page is still stack tracing as it was before these changes were made. Refer to item 27 on the Technical Notes page for details on how to reproduce this error.

src/main/java/edu/cornell/kfs/module/purap/CUPurapWorkflowConstants.java
```diff
@@ -41,7 +41,7 @@ public enum NodeDetailEnum implements NodeDetails {
 	            ACCOUNT_REVIEW("Account", PaymentRequestStatuses.APPDOC_AWAITING_FISCAL_REVIEW, PaymentRequestStatuses.APPDOC_CANCELLED_POST_AP_APPROVE, true), 
 	            ORG_REVIEW("AccountingOrganizationHierarchy", PaymentRequestStatuses.APPDOC_AWAITING_ORG_REVIEW, PaymentRequestStatuses.APPDOC_CANCELLED_POST_AP_APPROVE, false), 
 	            // KFSPTS-1891
-	            PAYMENT_METHOD_REVIEW("PaymentMethodReviewer", CUPurapConstants.PAYMENT_METHODL_REVIEW, PaymentRequestStatuses.APPDOC_CANCELLED_POST_AP_APPROVE, true), 
+	            PAYMENT_METHOD_REVIEW("PaymentMethod", CUPurapConstants.PAYMENT_METHODL_REVIEW, PaymentRequestStatuses.APPDOC_CANCELLED_POST_AP_APPROVE, true),
```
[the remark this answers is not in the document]
nkimble: Fixing now.

src/main/java/org/kuali/kfs/vnd/businessobject/VendorDetail.java
```diff
@@ -733,8 +752,8 @@ public boolean isEqualForRouting(final Object toCompare) {
...
                     ((VendorDetailExtension)getExtension()).getHealthOffSiteLicenseExpirationDate(),((VendorDetailExtension)detail.getExtension()).getHealthOffSiteLicenseExpirationDate()).append(
                     ((VendorDetailExtension)getExtension()).getInsuranceNotes(),((VendorDetailExtension)detail.getExtension()).getInsuranceNotes()).append(
-                    ((VendorDetailExtension)getExtension()).getMerchantNotes(),((VendorDetailExtension)detail.getExtension()).getMerchantNotes()).append( 
-                    ((VendorDetailExtension)getExtension()).getDefaultB2BPaymentMethodCode(),((VendorDetailExtension)detail.getExtension()).getDefaultB2BPaymentMethodCode()).append( // end KFSUPGRADE-779
+                    ((VendorDetailExtension)getExtension()).getMerchantNotes(),((VendorDetailExtension)detail.getExtension()).getMerchantNotes()).append( // end KFSUPGRADE-779
+                    getDefaultPaymentMethodCode(), detail.getDefaultPaymentMethodCode()).append( // end KFSUPGRADE-779
```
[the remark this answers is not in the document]
nkimble: Fixed

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
```diff
@@ -124,37 +104,25 @@ public boolean answerSplitNodeQuestion(final String nodeName) throws Unsupported
                 "VendorIsEmployeeOrNonResidentAlien".equals(nodeName)) {
             return isVendorEmployeeOrNonresident();
         }
-        // KFSPTS-1891
-        if (nodeName.equals(CUPurapWorkflowConstants.TREASURY_MANAGER)) {
```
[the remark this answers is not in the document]
nkimble: Verified this method would now match base code so removing from our overlay.

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
```diff
@@ -124,37 +104,25 @@ public boolean answerSplitNodeQuestion(final String nodeName) throws Unsupported
...
-                 setPaymentMethodCode(
-                         ((VendorDetailExtension)po.getVendorDetail().getExtension()).getDefaultB2BPaymentMethodCode() );
-             }
+            final PurchaseOrderDocument po, final HashMap<String,
+            ExpiredOrClosedAccountEntry> expiredOrClosedAccountList) {
+        super.populatePaymentRequestFromPurchaseOrder(po, expiredOrClosedAccountList);
+
+        if (ObjectUtils.isNotNull(po.getVendorDetail()) 
+                && StringUtils.isNotBlank(((VendorDetail)po.getVendorDetail()).getDefaultPaymentMethodCode())) {
+             setPaymentMethodCode(((VendorDetail)po.getVendorDetail()).getDefaultPaymentMethodCode());
```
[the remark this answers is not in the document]
nkimble: Removing.

src/main/java/edu/cornell/kfs/module/purap/document/CuVendorCreditMemoDocument.java
```diff
@@ -83,17 +82,11 @@ public void customizeExplicitGeneralLedgerPendingEntry( final GeneralLedgerPendi
     
     @Override
     public boolean answerSplitNodeQuestion(final String nodeName) throws UnsupportedOperationException {
-    	   if (nodeName.equals(PurapWorkflowConstants.REQUIRES_IMAGE_ATTACHMENT)) return requiresAccountsPayableReviewRouting();
-           // KFSPTS-1891, KFSPTS-2851
-           if (nodeName.equals(CUPurapWorkflowConstants.TREASURY_MANAGER))
-               return isWireOrForeignDraft();
-           throw new UnsupportedOperationException("Cannot answer split question for this node you call \""+nodeName+"\"");
+        if (nodeName.equals(PurapWorkflowConstants.REQUIRES_IMAGE_ATTACHMENT)) 
+            return requiresAccountsPayableReviewRouting();
```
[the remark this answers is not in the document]
nkimble: Fixed.

src/main/java/org/kuali/kfs/sys/businessobject/PaymentSourceWireTransfer.java
```diff
@@ -0,0 +1,278 @@
...
+import org.kuali.kfs.krad.bo.BusinessObject;
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+import org.kuali.kfs.krad.datadictionary.AttributeSecurity;
+import org.kuali.kfs.krad.util.ObjectUtils;
+
+import edu.cornell.kfs.sys.businessobject.PaymentSourceWireTransferExtendedAttribute;
+
+/**
+ * Cornell Customizations: 
+ *     Overrode toString method in KualiCo 2024-04-19 version of this class
```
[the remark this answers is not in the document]
nkimble: fixed

nkimble (commented):
more review changes.

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
```diff
@@ -205,6 +205,7 @@
     <vendorUrlAddress>www.avisbudgetgroup.com</vendorUrlAddress>
     <vendorFirstLastNameIndicator>false</vendorFirstLastNameIndicator>
     <taxableIndicator>false</taxableIndicator>
+    <defaultPaymentMethodCode>P</defaultPaymentMethodCode>
```
cah292, line 208: This appears to be on the pre-conversion side of the file. Should it have been omitted?

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
```diff
@@ -375,7 +376,6 @@
       <vendorHeaderGeneratedIdentifier>10385</vendorHeaderGeneratedIdentifier>
       <vendorDetailAssignedIdentifier>1</vendorDetailAssignedIdentifier>
       <einvoiceVendorIndicator>N</einvoiceVendorIndicator>
-      <defaultB2BPaymentMethodCode>P</defaultB2BPaymentMethodCode>
```
cah292, line 378: This appears to have been removed from the pre-conversion side of the file. Should it be added back in?

cah292 (changes requested):
I've gotten through the rest of the initial pass on reviewing the code changes. I still need to re-review the areas that were updated today, but in the meantime I found a few more items needing attention. I also discovered something that may require additional SQL for resolution; I'll post some details on the review sub-task due to its sensitive nature.

nkimble:
@cah292 How do we need to adjust the validation spreadsheet to ensure this new LABOR CuScrubberValidatorImpl customization (existing merged with new base code change) is adequately tested?

cah292:
@nkimble Regarding the validation spreadsheet instructions for the Labor Scrubber changes, we need the functionals to confirm something like the following:

Verify that when the Labor Scrubber detects an LD fringe entry that needs to be moved to another account (such as the related "reports to" account), its sub-account will get cleared out (or replaced with dashes) and the related error or warning message will display the sub-account that was removed from the entry.

cah292:
@nkimble Regarding the Asset Lookup errors, the underlying problem seems to be affecting lookups in our current code; it's not specific to this upgrade. There appears to be a bug in the lookup handling, where a new lookup that returns values to another new lookup will sometimes use the values from the nested lookup's criteria, not the values from the returned row. That's why the Asset Lookup is encountering errors after returning results from the nested Building lookup, since the latter has an Active flag but the former does not. (Parameter objects also lack an Active field, but I was only able to replicate the search-criteria-returning bug and not the logged stack trace; perhaps the Asset's non-DB-mapped "active" field has something to do with making KFS use the field incorrectly?)

Because the bug is likely in the React-based code, we might not be able to fix it ourselves. I wasn't able to replicate the problem in the KualiCo Monsters environment, so one of the future KualiCo patches should already fix it.

src/main/resources/edu/cornell/kfs/module/purap/document/datadictionary/CuPaymentRequestDocument.xml
```diff
@@ -19,56 +17,5 @@
   <bean id="PaymentRequestDocument-bankCode" parent="PaymentRequestDocument-bankCode-parentBean">
       <property name="required" value="false"/>
   </bean>
-  
-  <!-- workflow attributes for routing -->
```
db483, line 23: We should add paymentMerthodCode back to the search fields. Base code did not add the payment method to the search fields but we used to have it in our customization and it makes PREQs searches easier. I think I removed this when I was cleaning up the code but I now realise we still need it. Thank you.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -761,12 +761,35 @@ protected PaymentRequestDocument initializePaymentRequestDocument(final Electron
...
+
+        LOG.info("{} accounts has been found as Expired or Closed", expiredOrClosedAccountList::size);
+
+        preqDoc.populatePaymentRequestFromPurchaseOrder(orderHolder.getPurchaseOrderDocument(),expiredOrClosedAccountList);
+        
         //Copied from PaymentRequestServiceImpl.populatePaymentRequest()
         //set bank code to default bank code in the system parameter
         //KFSPTS-1891
         boolean hasPaymentMethodCode = false;
         if ( preqDoc instanceof PaymentRequestDocument ) {
```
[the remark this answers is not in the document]
nkimble: Adjusting the code, cleaning up imports, and cleaning up service classes no longer needed.
Does this need an explicit validation note in the testing spreadsheet?

src/main/java/edu/cornell/kfs/module/ld/batch/service/impl/CuScrubberValidatorImpl.java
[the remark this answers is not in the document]
nkimble: I think I have addressed this issue in commit c9d22a088328fcdc9f1bae219b3cc4f9072b58b6

src/main/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocument.java
```diff
@@ -109,10 +109,14 @@ public void templateVendor(final VendorDetail vendor, final VendorAddress vendor
 
         dvPayeeDetail.setDisbursementVoucherPayeeTypeCode(KFSConstants.PaymentPayeeTypes.VENDOR);
         dvPayeeDetail.setDisbVchrPayeeIdNumber(vendor.getVendorNumber());
+        
+        //Cornell Customization -- start
```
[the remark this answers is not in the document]
cah292: On the testing spreadsheet, you might want to mention the DV Payee Type field as a spot to be validated. At least one of the extended attribute fields impacts the display of the information in the Payee Type field (under the "Payment Information" tab).

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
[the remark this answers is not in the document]
cah292: The spreadsheet mentions PREQ validation already, but you may want it to explicitly mention the eInvoice Extract and Route jobs as areas needing regression testing.

src/main/resources/edu/cornell/kfs/module/cam/businessobject/datadictionary/Asset.xml
```diff
@@ -82,27 +85,28 @@
       <property name="beanName" value="Asset-lookupDefinition" />
       <property name="fieldOverrides">
           <list>
+              <!-- NEW FRAMEWORK -->
```
cah292, line 88: In this release, Asset lookups appear to have been configured for both the old and new frameworks in base code. Instead of repurposing the old-lookup overrides for the new framework, I would suggest creating new field override beans to put these changes in the new framework.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
```diff
@@ -810,6 +772,11 @@ protected PaymentRequestDocument initializePaymentRequestDocument(final Electron
         LOG.info("{} accounts has been found as Expired or Closed", expiredOrClosedAccountList::size);
 
         preqDoc.populatePaymentRequestFromPurchaseOrder(orderHolder.getPurchaseOrderDocument(),expiredOrClosedAccountList);
+        
+        // Starting with KualiCo release 2023-04-19:
+        // Use base code service method to initialize payment method and bank. 
+        paymentRequestService.initializePaymentMethodAndBank(preqDoc);
```
cah292, line 778: Is this comment needed?

cah292 (changes requested):
I found a few more areas that might need cleanup.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuPurapGeneralLedgerServiceImpl.java
```diff
@@ -72,136 +72,144 @@ protected int getNextAvailableSequence(final String documentNumber) {
         return count + 1;
     }
     
-    protected boolean generateEntriesPaymentRequest(
```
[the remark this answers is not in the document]
nkimble: The reason it was fully commented out was to get the code to compile. Multiple issues popped up between base code and our customization. Thanks for catching that it was still commented out.

I have had to jump through some hoops to retain our customization and its not pretty. I took the approach I did because of where in the if-clause our customization needs to occur. If you think the extra comments don't add value let me know.

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
nkimble: Without this change and the one below that removes _defaultB2BPaymentMethodCode as P_ , unit test CuMaintainableXMLConversionServiceImplTest.testConversionOfVendors fails.

Is there a better way to address this unit test failure than making these two coding changes?

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
nkimble: See comment above.

src/main/resources/edu/cornell/kfs/module/purap/document/datadictionary/CuPaymentRequestDocument.xml
nkimble: paymentMethodCode search field put back in place. Also fixed formatting in the file.

cah292 (changes requested):
I noticed a few minor spots that could be cleaned up.

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
cah292: After testing locally, I found the problem in the test case, where the expected-results side had one of the payment method elements in the wrong spot. Here's how to fix the issue:

1. Undo the `<defaultPaymentMethodCode>` addition on this line.
2. Undo the `<defaultB2BPaymentMethodCode>` removal on or near line 378.
3. Copy and remove the `<defaultPaymentMethodCode>` addition on line 633, then paste it immediately after the closing `</extension>` tag on line 779. Do NOT put any whitespace or newlines between the closing `</extension>` tag and the pasted content.

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
cah292: See my other comment for details on what needs to be done here.

cah292:
Quick follow-up on the Asset lookup errors and the underlying new-lookup bug:

I've created KFSPTS-33276 to keep track of this bug and placed it in the backlog. We may not be able to fix the problematic React code, but if we wanted to try forcibly altering the click handlers on the "return value" links as a workaround, I've placed a proof-of-concept file overlay for doing so in that JIRA ticket. (It only goes as far as adjusting the click handlers; it doesn't try altering criteria field values yet.)

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuElectronicInvoiceHelperServiceImpl.java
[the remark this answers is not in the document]
nkimble: Adding to the validation steps for FINP-9046 in the testing spreadsheet.

src/main/java/edu/cornell/kfs/module/purap/service/impl/CuPdpExtractService.java
```diff
@@ -12,27 +12,25 @@
 import org.kuali.kfs.pdp.businessobject.PaymentDetail;
 import org.kuali.kfs.pdp.businessobject.PaymentGroup;
 import org.kuali.kfs.pdp.businessobject.PaymentNoteText;
+import org.apache.commons.codec.binary.StringUtils;
```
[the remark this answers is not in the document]
nkimble: Fixed.

nkimble (commented):
more code cleanup

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
nkimble: Thank you! Fixed as described.

src/test/resources/edu/cornell/kfs/krad/service/impl/VendorTest.xml
nkimble: Fixed.

nkimble (commented):
fix test case coding

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
```diff
@@ -14,17 +11,15 @@
...
-import edu.cornell.kfs.module.purap.CUPurapConstants;
-import edu.cornell.kfs.module.purap.businessobject.CreditMemoWireTransfer;
-import edu.cornell.kfs.module.purap.document.CuVendorCreditMemoDocument;
 import edu.cornell.kfs.pdp.service.CuCheckStubService;
-import edu.cornell.kfs.sys.CUKFSKeyConstants;
-
 
 public class CuCreditMemoDocumentPreRules extends CreditMemoDocumentPreRules {
     
+    private PaymentSourcePreRulesServiceImpl paymentSourcePreRulesService;
```
cah292, line 22: Unless you explicitly need to reference the Impl class directly, I would suggest adjusting this variable to reference the interface instead.

src/main/resources/CU-ApplicationResources.properties
```diff
@@ -365,8 +364,7 @@ error.document.creditmemo.paymentmethodcode.mustmatchpreq=When creating a credit
 
  # Prevent Cardholder Self Reconciliation - KITT-909
  warning.document.procurementcardholderdetail.cardholder.last.active=Cardholder is the only active member of the selected PCard Reconciler workgroup.
-question.preq.clearUnneededTab=The {0} for this payment request is {1}.  The {2} tab should not contain data if the {0} is not {3}.  Would you like to clear the tab?  Press YES to clear the tab and continue routing.  Press NO to return to the document for further editing.
-question.cm.clearUnneededTab=The {0} for this Credit Memo is {1}.  The {2} tab should not contain data if the {0} is not {3}.  Would you like to clear the tab?  Press YES to clear the tab and continue routing.  Press NO to return to the document for further editing.
+# nkk4: use for rewording::::   question.cm.clearUnneededTab=The {0} for this Credit Memo is {1}.  The {2} tab should not contain data if the {0} is not {3}.  Would you like to clear the tab?  Press YES to clear the tab and continue routing.  Press NO to return to the document for further editing.
```
cah292, line 367: Can this commented-out property be removed?

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
cah292: Can you please check whether this method is still being used at all? I think `PaymentSourcePreRulesServiceImpl` is invoking its own variant of it, rather than invoking the one on the document PreRules instance.

src/main/java/edu/cornell/kfs/module/purap/document/validation/impl/CuCreditMemoDocumentPreRules.java
cah292: Another method for which the `PaymentSourcePreRuleServiceImpl` variant might be getting used instead.

src/main/java/edu/cornell/kfs/module/ld/batch/service/impl/CuScrubberValidatorImpl.java
```diff
@@ -126,15 +100,20 @@ protected void clearSubAccountOnModifiedFringeTransaction(
             }
             laborWorkingEntry.setSubAccount(null);
             laborWorkingEntry.setSubAccountNumber(KFSConstants.getDashSubAccountNumber());
```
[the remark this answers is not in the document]
cah292: The overridden `checkAccountFringeIndicator()` method is invoking this custom `clearSubAccountOnModifiedFringeTransaction()` method, and it looks like the former already contains code to set the sub-account to null and the sub-account number to dashes. Given that, you might be able to remove the equivalent lines located in the latter method.

nkimble (commented):
Removing the LD scrubber validator customization since it now matches base code

src/main/java/edu/cornell/kfs/module/ld/batch/service/impl/CuScrubberValidatorImpl.java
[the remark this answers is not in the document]
cah292: Can you please double-check your latest removal of this class? I thought the custom `appendSubAccountWarningToMessage()` method was still needed so that the message would display the removed sub-account. (It looks like base code is printing the all-dashes sub-account for the modified entry, but I don't think it's printing the sub-account that was originally on the entry.)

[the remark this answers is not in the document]
nkimble: Fixed.

src/main/resources/edu/cornell/kfs/module/cam/businessobject/datadictionary/Asset.xml
nkimble: Fixed.

cah292 (changes requested):
I noticed a few more items that should be cleaned up on the Asset lookup overrides.

src/main/java/edu/cornell/kfs/sys/businessobject/PaymentSourceWireTransferExtendedAttribute.java
```diff
@@ -0,0 +1,119 @@
...
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectExtension;
+
+public class PaymentSourceWireTransferExtendedAttribute extends PersistableBusinessObjectBase implements PersistableBusinessObjectExtension {
+    
+    private String documentNumber;
+    private String payeeAccountTypeCode;
+    private String bankStreetAddress;
+    private String bankProvince;
+    private String bankSWIFTCode;
```
cah292, line 12: The old equivalent of this extended attribute had "...SWIFTCode" for the regular bank's property but had "...SwiftCode" for the correspondent bank's property, and that discrepancy was carried over to this new extended attribute. If it's not too much work, it would be good to update those properties to be consistent about the "SWIFT" capitalization.

cah292 (commented):
I did another quick spot-check of this PR's changes. Aside from the two known issues concerning the on-payment-method-changed JavaScript function and the sensitive wire transfer data, I only noticed one other obvious issue that might be good to clean up.

src/main/java/edu/cornell/kfs/sys/businessobject/PaymentSourceWireTransferExtendedAttribute.java
nkimble: fixed

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
```diff
@@ -109,7 +109,7 @@ public static class ConfidentialAttachmentTypeCodes {
...
+    public static final String DV_WIRETRANSFER_TAB_ERRORS_EXTENSION_FIELDS = "document.dvWireTransfer.extension.disbVchrBankStreetAddress,document.dvWireTransfer.extension.disbVchrBankProvince,document.dvWireTransfer.extension.disbVchrBankIBAN,document.dvWireTransfer.extension.disbVchrCorrespondentBankName,document.dvWireTransfer.extension.disbVchrBankSwiftCode,document.dvWireTransfer.extension.disbVchrCorrespondentBankAddress,document.dvWireTransfer.extension.disbVchrSortOrTransitCode,document.dvWireTransfer.extension.disbVchrCorrespondentBankSwiftCode,document.dvWireTransfer.extension.disbVchrCorrespondentBankRoutingNumber,document.dvWireTransfer.extension.disbVchrCorrespondentBankAccountNumber";
```
cah292, line 112: As a result of the wire transfer changes, do the properties listed in this constant and/or the `DV_WIRETRANSFER_TAB_ERRORS` constant need further updates?

src/main/webapp/WEB-INF/tags/module/purap/preqWireTransfer.tag
```diff
@@ -131,12 +131,12 @@
                     </td>
                 </tr>
                 <tr>
-                    <th scope=row class="bord-l-b"><div align="right"><kul:htmlAttributeLabel attributeEntry="${wireTransAttributes.preqBankSWIFTCode}"/></div></th>
+                    <th scope=row class="bord-l-b"><div align="right"><kul:htmlAttributeLabel attributeEntry="${wireTransAttributes.preqBankSwiftCode}"/></div></th>
```
cah292, line 134: Should the references to `preqBankSwiftCode` be replaced with `bankSwiftCode` instead?

You may also want to check if there are any other property references that need the `preq` prefix removed.

cah292 (changes requested):
I noticed a couple follow-up items while reviewing the changes to the Swift Code conventions.

src/main/webapp/WEB-INF/tags/module/purap/preqWireTransfer.tag
db483: @nkimble is this the comment that prompted the latest changes? I think this tag was just supposed to be removed.

nkimble: Lets talk off-line. There are some things going on with these tags.

PR #1660, opened 2024-11-04
KFSPTS-33161 remove code that updates active from date on existing records

db483:


src/main/java/org/kuali/kfs/kim/service/impl/UiDocumentServiceImpl.java
```diff
@@ -771,8 +771,8 @@ protected List<RoleMember> populateRoleMembers(
                         if (roleMember.getActiveFromDate() != null) {
                             roleMemberImpl.setActiveFromDateValue(
                                     new java.sql.Timestamp(roleMember.getActiveFromDate().getTime()));
-                        } else {
-                            // CU Customization KFSPTS-27162 Set default for activeFromDate field
+                        } else if(!roleMember.isEdit()){
```
cah292, line 774: Missing whitespace before and after the parenthesized condition.

src/main/java/org/kuali/kfs/kim/service/impl/UiDocumentServiceImpl.java
```diff
@@ -771,8 +771,8 @@ protected List<RoleMember> populateRoleMembers(
...
+                            // CU Customization If this is a new role and activeFromDate is null, set default for activeFromDate field
                             roleMemberImpl.setActiveFromDateValue(dateTimeService.getCurrentTimestamp());
```
cah292, line 776: This code appears to be affecting new role members that are not necessarily on a new role. Please update this comment for clarity.

cah292 (changes requested):
There are a couple minor areas that should be adjusted, but otherwise the changes look good.

Just to confirm, is any SQL needed to fix existing records?

src/main/java/org/kuali/kfs/kim/service/impl/UiDocumentServiceImpl.java
db483: updated

src/main/java/org/kuali/kfs/kim/service/impl/UiDocumentServiceImpl.java
db483: updated

cah292 (commented):
Changes look good, will merge once we get the unit tests to pass.

cah292:
retest this please

cah292 (approved):

PR #1662, opened 2024-11-11
KFSPTS-32805 Show DV payment info on IWNT doc

cah292:
This PR adds a new tab to the IWNT doc that shows a limited set of payment information from the associated DV document (if any).


db483 (approved):
Changes look good.

PR #1664, opened 2024-11-13
KFSPTS-32805 Show DV payment info on IWNT doc

cah292:
This is an updated version of #1662 that has been updated to retrieve the payment details correctly. Note that, for consistency with similar methods on the modified Payment Detail service, the new service method has been set up to return an iterator instead of a collection/list.


db483 (approved):
Changes look good.

PR #1666, opened 2024-11-19
KFSPTS-33051 Tax Sprintax 1042-S Report Process CSV

ajd299:


src/main/java/edu/cornell/kfs/tax/dataaccess/impl/SprintaxRowPrintProcessor.java
cah292, line 1: The logic for writing the Sprintax transaction file may need to be included in the class that writes the Sprintax biographic file. The mapping between detail rows and Sprintax rows might not necessarily be 1-1, and there's business logic that may impact whether the data from a given Transaction Detail object should be written to the file at all.

cah292 (changes requested):
I still have more to review, but here are my initial thoughts and suggestions for improvement. There's some simplification that could potentially be performed in this PR, as a starting point for longer-term refactoring of the tax processing code, unless you think it's not practical to do so at this time.

ajd299:
@cah292 I've tried to address all of your feedback items. There are some questions I have.

- Where is the logic in the current 1042-S process that dictates when the biographic records are supposed to be written?
- For the DTO, could we utilize edu.cornell.kfs.tax.businessobject.TransactionDetail? Or is there performance concerns with using OJB?

cah292:
Since the Sprintax work was reassigned to me, and since I'm taking a different implementation approach than the one introduced on the initial feature branch, I'm closing this PR to avoid confusion with the others. I will soon open up a new PR containing the changes from the first part of the revised Sprintax implementation efforts.

PR #1672, opened 2024-12-17
KFSPTS-27383 Search For Payment Lookup Remove (Stale) for ACH/Wire Payments

ajd299:


src/main/java/org/kuali/kfs/pdp/businessobject/PaymentGroup.java
```diff
@@ -335,9 +335,12 @@ public String getPaymentStatusCodeWithHistory() {
             paymentStatusWithHistory += " (Reissued)";
         }
 
-        // check for stale payments, if one payment detail is stale then they all are
-        // CU customization
-        if (!CRConstants.CLEARED.equalsIgnoreCase(paymentStatus.getCode())) {
+        // CU Customization: KFSPTS-27383 do not add Stale to ACH or WIRE
+        final boolean isAchOrWire = StringUtils.equals(disbursementType.getCode(), PdpConstants.DisbursementTypeCodes.ACH) ||
```
nkimble, line 339: Is it possible that disbursementType could ever be undefined (meaning null) so that we would get an exception thrown when attempting the getCode call?

nkimble (changes requested):
Coding looks good. 
I did see one thing we many want to have an additional check for if you think that data value could ever be null.

jhulslander (approved):
This seems reasonable to me, I shall merge

jhulslander:
the unit tests have now passed, I shall merge

PR #1463, opened 2023-06-19
KFSPTS-28699 upgrade to the 10/19/2022 version of financials

db483:

PR #1673, opened 2024-12-18
KFSPTS-32809: Print Boss only processing ISO 20022 check file format. Removing KFS proprietary format for checks.

nkimble:
This PR has dependent SQL in nonprod-sql PR: https://github.com/CU-CommunityApps/nonprod-sql/pull/1463
Make sure both are acceptable and merged together.

### DIFF
--- a/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
@@ -38,7 +38,6 @@ import org.kuali.kfs.sys.KFSConstants;
 import com.rsmart.kuali.kfs.pdp.service.AchBundlerHelperService;
 
 import edu.cornell.kfs.fp.document.CuDisbursementVoucherConstants;
-import edu.cornell.kfs.pdp.CUPdpParameterConstants;
 import edu.cornell.kfs.pdp.batch.service.CuPayeeAddressService;
 import edu.cornell.kfs.sys.CUKFSParameterKeyConstants;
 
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
```diff
@@ -112,9 +111,7 @@ public void extractAchPayments() {
     }
 
     /**
-     * Overridden to call a CU-specific version of the parameter-checking method to determine
-     * whether or not to use the ISO 20022 format, as well as to allow for generating
-     * the proprietary XML when the ISO 20022 format is enabled.
+     * Overridden and Cornell customized to not generate proprietary check format.
      */
     @Override
     public void extractChecks() {
```
jhulslander, line 117: maybe we should  remove this customized function.
In base code, when shouldUseIso20022Format() is false, extractChecksToProprietaryFormat() is called. Since we are Iso20022 format, maybe we don't need this customized function?

@cah292  since you are closest to this, do you have an opinion?

jhulslander (commented):
Changes look good, I found one other function that we may want to remove, but I will defer to @cah292 's opinion on the matter

src/main/java/edu/cornell/kfs/pdp/batch/service/impl/CuExtractPaymentServiceImpl.java
cah292: Yes, I agree that the `extractChecks()` override and most of the related copied private methods (like `extractChecksToProprietaryFormat()`) can be removed as well. We may still need a base-code-equivalent version of `shouldUseIso20022Format()` if we want to keep the ACH-related customization intact for now. (We'll eventually need to remove the ACH-side proprietary and/or FastTrack customizations, but that could be done at a future date if needed.)

Also, just to be clear, we're still sending the check files to PrintBoss, but they're now processing the ISO 20022 check files instead of the KFS proprietary check files. This PR's description should be updated to reflect that.

nkimble:
PR description for both the code and SQL has been updated as requested. 
Requested code changes have been made.

jhulslander (approved):
changes looks reasonable as does the SQL< I shall merge

PR #1677, opened 2025-01-08
KFSPTS-29347: Upgrade to the 04/19/2023 version of financials.

nkimble:


nkimble:
@cah292 The code changes in this PR should match the ones in the v7 feature branch. The unit tests are currently running and this PR is ready for review.

cah292 (approved):
The changes on the older PR looked good when I had reviewed them before. After doing some quick comparisons via the "git clone" and "git diff" commands, it looks like the rebased PR's upgrade changes are lining up as expected with those from the old PR. Also, it appears that the unit tests have passed on the new PR.

The code, SQL and report changes for the 04/19/2023 patch are now ready to merge to develop.

PR #1678, opened 2025-01-09
verify unit tests run after patch sql applied

nkimble:


nkimble:
verification of utest database was successful. closing pr.

PR #1679, opened 2025-01-10
KFSPTS-33004 add class loader logs in support of OJB error

db483:


jhulslander (changes requested):
I think we want to add TargetAccounting Line.
Also, should we add similar logging in concurEventNotificationV2ProcessingJob, procurementCardSummaryFeedJob and crGLTransactionJob?  We had this error happen in those jobs too.

db483:
Good question Jay. I added the logging to the Create Accounting Document job because this is the one that fails most frequently. I will remove this logging after we get an instance of this exception and we can determine if it is caused by a class loader problem.

PR #1681, opened 2025-01-14
KFSPTS-33051 Implement Part 1 of Sprintax 1042-S handling

cah292:
I had originally planned to have the Sprintax changes committed via the #1674 PR, but its scope was too large for one user story and needed to be split up further. This PR introduces the first part of the divided changes, along with further cleanup and the creation of relevant unit tests.

Here are the major items introduced by this PR, many of which won't be fully utilized until the subsequent user stories have been implemented:

1. Added the basic skeleton and initial coding for the new Sprintax batch job. The job will at least determine which tax year to process and then generate appropriate transaction DB rows via the old feature, but further processing and file generation will be deferred to the follow-up tickets. (The creation of the DB rows is using the old code for now, so that we can focus on preparing the Sprintax file generation part with the new code.)
2. Added the XML DTOs and KFS file type for handling new versions of the Tax Output Definition objects. The existing process already had DTOs with similar structure, but the new ones have been adjusted to make the upcoming output process simpler.
3. Added a Tax Parameter Service that wraps the KFS Parameter Service, converts certain multi-value parameters into the Map-based or Collection-based structures needed by the tax processing, and performs temporary caching of those Map/Collection objects. This provides a cleaner and less error-prone way of generating the appropriate parameter-derived Maps/Collections.
4. Made some minor adjustments to the old process, such as allowing for 1042-S transaction DB row generation without processing the rows into files, as well as adding or downgrading various log statements.
5. Added unit tests to validate the new services and output file type. Note that the tests rely upon some unit test infrastructure that was introduced by earlier non-tax-related user stories; some new test infrastructure was also added that may be useful for future tax and non-tax unit tests. If you have questions about how some of that test infrastructure works, please let me know.


src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+                        + "tax year later than 1970");
+            }
+            return parsedYear;
+        } catch (final NumberFormatException e) {
+            throw new IllegalStateException("The dates-to-process parameter contained a malformed tax year");
+        }
+    }
+
+    private TaxBatchConfig buildTaxBatchConfigForExplicitStartAndEndDates(final java.util.Date processingStartDate,
+            final String taxType, final Collection<String> taxRangeSettings) {
```
ajd299, line 115: Will the taxType always be 1042-S? Does it make sense to build this process to allow for multiple tax types?

src/test/java/edu/cornell/kfs/tax/batch/xml/TaxOutputDefinitionV2FileTypeTest.java
```diff
@@ -0,0 +1,226 @@
...
+                Pair.of("good-file-single-section-multiple-fields.xml", LocalTestCase.SINGLE_SECTION_MULTIPLE_FIELDS),
+                Pair.of("good-file-multiple-sections-multiple-fields.xml",
+                        LocalTestCase.MULTIPLE_SECTIONS_MULTIPLE_FIELDS)
+        )
+                .map(pair -> Arguments.of(pair.getLeft(), pair.getRight().toNamedAnnotationFixturePayload()));
+    }
+
+
+
+    @ParameterizedTest
```
ajd299, line 160: Are the extra newlines by convention. I am fine with them if it helps separate the code visually

src/test/java/edu/cornell/kfs/tax/fixture/TaxOutputDefinition.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import edu.cornell.kfs.sys.util.FixtureUtils;
+import edu.cornell.kfs.tax.batch.xml.TaxOutputDefinitionV2;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface TaxOutputDefinition {
```
ajd299, line 13: Do you think it would be appropriate to rename this class/file to TaxOutputDefinitionFixture?

src/test/java/edu/cornell/kfs/tax/fixture/TaxOutputField.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+import edu.cornell.kfs.sys.CUKFSConstants;
+import edu.cornell.kfs.tax.batch.CUTaxBatchConstants.TaxOutputFieldType;
+import edu.cornell.kfs.tax.batch.xml.TaxOutputFieldV2;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface TaxOutputField {
```
ajd299, line 16: Do you think it would be appropriate to rename this class to TaxOutputFieldFixture?

src/test/java/edu/cornell/kfs/tax/fixture/TaxOutputSection.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import edu.cornell.kfs.sys.util.FixtureUtils;
+import edu.cornell.kfs.tax.batch.xml.TaxOutputSectionV2;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface TaxOutputSection {
```
ajd299, line 13: Do you think it would be appropriate to rename this section TaxOutputSectionFixture?

src/test/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImplTest.java
```diff
@@ -0,0 +1,241 @@
...
+    public static TaxProcessingDao buildMockTaxProcessingDao() {
+        return Mockito.mock(TaxProcessingDao.class);
+    }
+
+    @SpringXmlTestBeanFactoryMethod
+    public static DateTimeService buildSpiedTestDateTimeService() throws Exception {
+        final TestDateTimeServiceImpl dateTimeService = new TestDateTimeServiceImpl();
+        dateTimeService.afterPropertiesSet();
+        return new CuMockBuilder<>(dateTimeService)
+                .withReturn(service -> service.getLocalDateNow(), TEST_DATE_2025_01_10)
```
ajd299, line 122: This could be replaced with a lambda method reference 
DateTimeServiceImpl::getLocalDateNow

src/test/resources/edu/cornell/kfs/tax/batch/output-definition-test/bad-file-static-field-with-derived-key.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<taxOutputDefinition fieldSeparator=",">
+    <section name="Example_Section">
+        <field name="recipient_id" length="25" type="STATIC" key="recipientId"/>
```
ajd299, line 4: In general, is the length field needed since we are generating a CSV that will have variable lengths?

src/main/java/edu/cornell/kfs/tax/service/impl/TaxParameterServiceImpl.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Override
+    public Collection<String> getParameterValuesAsString(final String componentCode, final String parameterName) {
+        validateParams(componentCode, parameterName);
+        return parameterService.getParameterValuesAsString(CUTaxConstants.TAX_NAMESPACE, componentCode, parameterName);
+    }
+
+    @Override
+    @Cacheable(cacheNames = Parameter.CACHE_NAME, key = "'{CU_getParameterValuesSetAsString}'+#p0+','+#p1")
+    public Set<String> getParameterValuesSetAsString(final String componentCode, final String parameterName) {
```
ajd299, line 47: Is the Cacheable required or necessary in this clas? Does it provide needed performance benefits?

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        return buildTaxBatchConfig(processingStartDate, CUTaxConstants.TAX_TYPE_1042S,
+                CUTaxConstants.TAX_1042S_PARM_DETAIL);
+    }
+
+    private TaxBatchConfig buildTaxBatchConfig(final java.util.Date processingStartDate,
+            final String taxType, final String taxParameterComponent) {
+        final Collection<String> taxRangeSettings = taxParameterService.getParameterValuesAsString(
+                taxParameterComponent, taxType + TaxCommonParameterNames.DATES_TO_PROCESS_PARAMETER_SUFFIX);
+        final TaxBatchConfig taxConfig;
+        if (taxRangeSettings.size() == 1) {
```
ajd299, line 64: I would suggest adding some spacing (newlines)... Maybe between these 2 lines

ajd299 (changes requested):
Please see my suggestions. I am fine with any changes you want to defer to avoid rebasing.

src/main/java/edu/cornell/kfs/tax/batch/TaxOutputDefinitionV2FileType.java
```diff
@@ -0,0 +1,20 @@
+package edu.cornell.kfs.tax.batch;
+
+import edu.cornell.kfs.sys.batch.CuXmlBatchInputFileTypeBase;
+import edu.cornell.kfs.tax.CUTaxConstants;
+import edu.cornell.kfs.tax.CUTaxConstants.CUTaxKeyConstants;
+import edu.cornell.kfs.tax.batch.xml.TaxOutputDefinitionV2;
+
+public class TaxOutputDefinitionV2FileType extends CuXmlBatchInputFileTypeBase<TaxOutputDefinitionV2> {
+
```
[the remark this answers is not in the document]
cah292: Yes, I was planning to add the actual non-test XML files in a follow-up user story. I wanted to make some adjustments to the XML structure (including some simplifications), which is why I created separate JAXB POJOs and a separate XSD instead of reusing the old ones.

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+    public void performTaxProcessingFor1042S(final java.util.Date processingStartDate) {
+        try {
+            LOG.info("performTaxProcessingFor1042S, Starting 1042-S tax processing...");
+            final TaxBatchConfig config = buildTaxBatchConfigFor1042S(processingStartDate);
+            LOG.info("performTaxProcessingFor1042S, Finding transactions to process...");
+            createTransactionDetailRowsFor1042SUsingLegacyProcess(config);
+            LOG.info("performTaxProcessingFor1042S, Generating 1042-S files...");
+            LOG.warn("performTaxProcessingFor1042S, 1042-S tax file generation has not been fully implemented yet");
+            taxFileGenerationService.generateFiles(config);
+            LOG.info("performTaxProcessingFor1042S, Finished 1042-S tax processing");
```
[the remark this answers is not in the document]
cah292: I'll add more whitespace to make the contents clearer.

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
```diff
@@ -0,0 +1,151 @@
...
+            final String taxType, final String taxParameterComponent) {
+        final Collection<String> taxRangeSettings = taxParameterService.getParameterValuesAsString(
+                taxParameterComponent, taxType + TaxCommonParameterNames.DATES_TO_PROCESS_PARAMETER_SUFFIX);
+        final TaxBatchConfig taxConfig;
+        if (taxRangeSettings.size() == 1) {
+            final String taxRangeSetting = taxRangeSettings.iterator().next();
+            taxConfig = buildTaxBatchConfigForSingleRangeSetting(processingStartDate, taxType, taxRangeSetting);
+        } else if (taxRangeSettings.size() == 2) {
+            taxConfig = buildTaxBatchConfigForExplicitStartAndEndDates(processingStartDate, taxType, taxRangeSettings);
+        } else if (taxRangeSettings.size() == 0) {
```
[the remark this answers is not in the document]
cah292: Changing to use `isEmpty()` call.

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
cah292: At some point, we may want to refactor the 1099 code for simplicity and easier maintainability. If we do eventually proceed with such refactoring, it might make sense to add its basic setup portions into this service instead of creating a separate one.

src/test/java/edu/cornell/kfs/tax/batch/xml/TaxOutputDefinitionV2FileTypeTest.java
cah292: I was trying to add some extra visual separation in certain classes (especially the unit test classes) without resorting to comments. Not sure if we have a specific convention, but I wanted some kind of extra separation to help make it easier for other developers to read the unit test code.

src/test/java/edu/cornell/kfs/tax/fixture/TaxOutputDefinition.java
cah292: I was originally trying to avoid clutter in the unit tests, but I agree that a "Fixture" suffix would help clarify that the annotation represents a fixture. I'll make adjustments accordingly.

src/test/java/edu/cornell/kfs/tax/fixture/TaxOutputField.java
cah292: Will rename accordingly.

src/test/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImplTest.java
cah292: Updating to use method reference.

src/test/resources/edu/cornell/kfs/tax/batch/output-definition-test/bad-file-static-field-with-derived-key.xml
cah292: Even for tax files that don't enforce fixed field lengths, they usually still expect a maximum length for certain fields. The spec spreadsheet for Sprintax also contains some fields that expect specific lengths or maximum lengths (Chapter 3 Status Code, the various dollar amount fields, etc.).

Would it make more sense to rename this attribute as a max length (such as "maxlength" or "maxLength" or "max-length")? Or do you think the current "length" name is clear enough?

src/main/java/edu/cornell/kfs/tax/service/impl/TaxParameterServiceImpl.java
cah292: There are various inclusion/exclusion checks that need to be performed on all or most of the relevant Transaction Detail rows, where many of those checks involve comparisons against parameter values. Because the checks are much easier to perform when the parameters have been parsed into Sets/Maps/etc., and because it's inefficient to recreate those collections for each data row iteration, I wanted to add a way to temporarily cache the generated Sets/Maps/etc. without necessarily storing them in a temporary helper object. Putting the temporary storage at the Spring Cache level (backed by Redis in our case) seemed like the best way to accomplish that.

cah292:
I've committed some changes to address the initial feedback. If there's something that needs more work or more clarification, please let me know.

src/test/resources/edu/cornell/kfs/tax/batch/output-definition-test/bad-file-static-field-with-derived-key.xml
ajd299: Okay, thank you for the explanation, I'm good with leaving it "length"

ajd299:
I was able to run all the tax unit tests successfully.

I was able to run the skeleton batch job locally.

Merging to develop

ajd299 (approved):
merging to develop

PR #1682, opened 2025-01-15
KFSPTS-33649 only add custom actions if user can edit

db483:
After the 4/19 upgrade functionals noticed that the custom links on the Asset lookup results show up for users that don't have them displayed in the previous version. After looking into the current production code it looks like the Actions column is fully ignored in the old lookup if the user does not have edit permission for Assets. Similar logic was implemented in the AssetActionsProvider.


jhulslander (approved):

PR #1683, opened 2025-01-15
KFSPTS-33694 implement PaymentSource for CuVendorCreditMemoDocument

db483:


nkimble (commented):
These changes look good. 

Before merging them, I did have one question.
@db483 Asking whether we need to be concerned with the following area as well: 

KualiCo introduced a new constant (checkAttachmentPropertyPath) local to the org.kuali.kfs.sys.document.validation.impl.PaymentSourceWireTransferValidation class.

I am still looking but do you think the error messages will be presented correctly for the PREQs and our customized recurring DVs that do have attachments with this base code define constant?    

I saw your comment on the JIRA that the CM's should be ok with this change but would require they receive functional user validation confirmation.

db483:
It doesn't look like that would work for PREQs but I think it would be good to test first before opening a jira for this issue.
I think the issue is separate from the one with the CM's routing to exception.

nkimble (approved):
My concern currently would not manifest itself for this document type (CM) because the changes in this PR always return "false", indicating that there are no attachments on the credit memo. Based on that logic, these changes look good to address the present issue being reported by the functional users and are Approved to move forward.

My concern will be further investigated with a future user story created to address it should that issue be found to manifest.

PR #1690, opened 2025-01-23
KFSPTS-33766: Fix CuCreditMemoDocumentPreRules stack trace when using a po or vendor to create the credit memo after the 2023-04-19 KualiCo upgrade changes.

nkimble:


cah292 (approved):

PR #1692, opened 2025-01-27
KFSPTS-33823: Fix vendor credit memo routing so it uses role qualifiers correctly after the 2023-04-19 KualiCo upgrade changes.

nkimble:


src/main/resources/edu/cornell/kfs/module/purap/document/datadictionary/CuVendorCreditMemoDocument.xml
```diff
@@ -20,7 +20,15 @@
    
     <!-- workflow attributes -->
     <bean id="VendorCreditMemoDocument-workflowAttributes" parent="VendorCreditMemoDocument-workflowAttributes-parentBean">
-    	<property name="searchingTypeDefinitions">
+    	<property name="routingTypeDefinitions">
+            <map>
```
cah292, line 24: You may want to specify the `merge="true"` attribute here, so that you only need to include the CU-specific routing definitions below (unless there's something else overriding things that requires explicitly specifying the base entries below).

cah292 (commented):
I had one minor suggestion to make the Spring XML a little cleaner, but otherwise the changes look good. If you think we should move this forward now and possibly defer the cleanup to a later date, please let me know.

cah292 (approved):

PR #1694, opened 2025-01-30
KFSPTS-33004: Sub-task KFSPTS-33849: Add more logging for procurementCardSummaryFeedJob ojb error support. Remove duplicate ProcurementCardSummaryEntry ojb definition.

nkimble:


src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
```diff
@@ -196,18 +196,22 @@ protected void loadDataInDB(List<ProcurementCardSummaryEntry> entriesToLoad) {
 			try {
```
cah292, line 196: In the `if` block that's above this `try` block, I had noted during our team discussion that the block was copying the version number but not the object ID. Should it be updated to copy the object ID as well?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
```diff
@@ -196,18 +196,22 @@ protected void loadDataInDB(List<ProcurementCardSummaryEntry> entriesToLoad) {
...
+    private void logClassLoaderDebugInfo(ProcurementCardSummaryEntry entry, ProcurementCardSummaryEntry retrievedEntry) {
         Class repoClass = org.apache.ojb.broker.metadata.ClassDescriptor.class;
         LOG.info("Repository Class Loader: " + repoClass.getClassLoader());
 
         Class pcardSummaryEntryClass = edu.cornell.kfs.fp.businessobject.ProcurementCardSummaryEntry.class;
         LOG.info("ProcurementCardSummaryEntry Class Loader: " + pcardSummaryEntryClass.getClassLoader());
+        
+        LOG.info("ProcurementCardSummaryEntry Class Loader: entry attributes = {}", entry.toString());
+        
+        LOG.info("ProcurementCardSummaryEntry Class Loader: retrievedEntry attributes = {}", retrievedEntry.toString());
```
cah292, line 214: `retrievedEntry` might be null at this point, so I would suggest adding some null checking here.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
cah292, line 1: General Question:

Since this is a follow-up PR to a reopened user story whose prior changes were not reverted, should the PR title only mention the sub-task instead of the parent? Or would it make sense to split out the extra logging into its own separate user story?

cah292 (changes requested):
Changes look good overall, though there are a few spots that might need minor revisions.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
nkimble: I am not sure if that is going to help. And, after further analysis, I am thinking that the entire "if" block you are referencing may need to be removed with the version number not being copied either. My reasoning for both of those ideas: 

- The physical table CU_FP_PCARD_SUMMARY_T (as well as the OJB data definition for ProcurementCardSummaryEntry) does NOT have a column defined for either object ID nor version number even though those data values are listed as attributes on base class PersistableBusinessObjectBase which ProcurementCardSummaryEntry extends. What would copying those values achieve in this case?

- This is an intermittent problem. Could the intermittent nature of this failure be caused by the "if" logic actual being executed? Meaning: Data exists in table CU_FP_PCARD_SUMMARY_T where one of the table data rows will match to a record in the new incoming data file. The existing data in the table is removed/deleted/dropped BUT the routine still has access to that information due to the transactional nature of the processing. So, the processing updates the new data with the old entry's version number which causes the OJB error. The next time the job runs, the next data file does not contain any data that matches what is in the table with the job completing successfully.

Thoughts?

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
nkimble: My reasoning for using both of the JIRA numbers is that if we use and tie everything to the first/reopened user story, we will be able to find all of the changes that pertained to this OJB issue by searching just on that one ticket number. Including the sub-task numbers adds in the identifying each specific change. 
IMHO, breaking it up into multiple user stories and just linking them in JIRA runs a higher risk of not removing all of this code at a later date.

cah292: Thanks for the explanation. Given that, I'm fine with both JIRA numbers being listed in the commit message and with the changes being kept in one user story.

src/main/java/com/rsmart/kuali/kfs/cr/batch/GlTransactionStep.java
```diff
@@ -297,11 +297,15 @@ public boolean execute(String jobName, Date jobRunDate) throws InterruptedExcept
     }
     
     private void logClassLoaderDebugInfo() {
+        Class classHelperClass = org.apache.ojb.broker.util.ClassHelper.class;
+        LOG.info("logClassLoaderDebugInfo:: OJB Broker ClassHelper.ClassLoader.name: {}", 
+                (ObjectUtils.isNull(classHelperClass) || ObjectUtils.isNull(classHelperClass.getClassLoader()) ? "ClassHelper or ClassLoader IS NULL" : classHelperClass.getClassLoader().getName()));
```
cah292, line 302: Just to clarify, the class loader that should be logged here (and in the similar methods) is the one from `ClassHelper.getClassLoader()`, not the one from `ClassHelper.class.getClassLoader()`.

Also, for consistency with the other logged class loaders, I would suggest logging its `toString()` value (implicitly or explicitly) instead of logging the `getName()` result.

cah292 (changes requested):
I noticed something in the latest revision that needs some minor adjustments.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/CreateAccountingDocumentServiceImpl.java
```diff
@@ -282,17 +282,19 @@ protected void createAndRouteAccountingDocumentFromXml(AccountingXmlDocumentEntr
     }
     
     private void logClassLoaderDebugInfo() {
+        LOG.info("logClassLoaderDebugInfo:: OJB Broker ClassHelper.ClassLoader.name: {}", org.apache.ojb.broker.util.ClassHelper.getClassLoader());
```
cah292, line 285: The log message labels are inconsistent between the various modified methods. You may want to update them for consistency, especially since your latest changes are logging the ClassLoader and not its "name()" call.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/ProcurementCardSummaryFeedServiceImpl.java
```diff
@@ -185,29 +186,28 @@ protected void loadDataInDB(List<ProcurementCardSummaryEntry> entriesToLoad) {
...
+        LOG.info("logClassLoaderDebugInfo:: OJB Broker ClassHelper.ClassLoader.name: {}", org.apache.ojb.broker.util.ClassHelper.getClassLoader());
+        
         Class repoClass = org.apache.ojb.broker.metadata.ClassDescriptor.class;
-        LOG.info("Repository Class Loader: " + repoClass.getClassLoader());
-
+        LOG.info("logClassLoaderDebugInfo:: Repository Class Loader: {}", repoClass.getClassLoader());
+        
         Class pcardSummaryEntryClass = edu.cornell.kfs.fp.businessobject.ProcurementCardSummaryEntry.class;
-        LOG.info("ProcurementCardSummaryEntry Class Loader: " + pcardSummaryEntryClass.getClassLoader());
+        LOG.info("logClassLoaderDebugInfo:: ProcurementCardSummaryEntry Class Loader:{} ", pcardSummaryEntryClass.getClassLoader());
```
cah292, line 206: For consistency with your other log messages, I would suggest adding a space between the colon and the placeholder.

cah292 (commented):
There are some minor message-format-related items that might need cleanup, but otherwise the changes look good.

cah292 (approved):
Changes look good, will merge when the unit tests pass.

PR #1696, opened 2025-01-31
KFSPTS-33579 Implement Part 2 of Sprintax 1042-S handling

cah292:
This PR implements the 2nd phase of the Sprintax 1042-S work, which adds tax object metadata handling to assist with reading/writing tax-related data. The related user story was originally going to include the code that performs the actual reads/writes; however, it ended up being too much scope to include both the metadata part and the read/write part, so the latter has been deferred to the Part 3 work.

Even though the old Tax Processing Job performed reading/writing of most of the tax data via JDBC, it analyzed the objects' related OJB metadata to determine what table and column names were associated with the various business objects. This potentially made certain things easier (like keeping up with column/field name changes from KualiCo), but it was implemented in a way that was difficult to read and maintain. This PR introduces a different style of OJB-metadata-based derivation that should be easier to maintain and also takes advantage of more modern Java features (especially lambdas).

The new process uses annotated DTO classes and fields to discern how the various business objects should be mapped to the DTOs.  (The Part 3 work will implement the DB-level interactions between the DTOs and the business objects.) Such DTO classes can then be passed to a new metadata lookup service that will derive the appropriate mapping metadata, using both the DTO annotations and OJB's metadata.

Note that, in addition to introducing a fair amount of unit test code related to OJB testing and XML filtering, this PR relies upon several existing unit test utilities that were added by prior user stories. If you have questions about how the new or existing unit test code works, please let me know.


cah292:
After further discussion with the developer team, I will be making more adjustments to simplify the metadata handling, since the current solution is still too complex for long-term maintenance. I'm closing this PR in the meantime, since I will likely introduce the adjusted changes on a different feature branch.

PR #1697, opened 2025-02-05
KFSPTS-32561 Fix boolean value handling for VN-99-03

cah292:
This PR fixes the VN-99-03 batch job so that, rather than expecting the CSV file's boolean values to be written as "yes"/"no" (case-insensitive), it can now handle KualiCo's other accepted boolean string values as well (0, 1, N, Y, etc.).

Some (but not all) of the associated unit test data has been updated, so that the tests can validate both the yes/no setup and the 1/0 setup.


jhulslander (approved):
These changes seem reasonable, I shall merge

PR #1698, opened 2025-02-07
KFSPTS-33579 Implement Part 2 of Sprintax 1042-S handling

cah292:
This PR contains the Part 2 changes for implementing Sprintax, with a focus on handling metadata for table/column mappings. It is a simpler version of the rejected #1696 changes and has a closer resemblance to the code from the abandoned #1674 PR, but has been adjusted to allow greater flexibility when choosing a metadata mapping strategy. (A hard-coded mapping implementation is included here, as requested; a few OJB-related implementation classes are still present but I can remove them if they're not needed anymore.)

To help demonstrate how this metadata will be used by the new code, I've included some of the SQL-related code that I was originally planning to include in Part 3. The remainder of the SQL handling will be implemented in the follow-up PR(s).

Depending on the changes in the subsequent PRs, I might end up removing the `DocumentHeaderLite` DTO. That's why I didn't add a mapper class for it yet.


db483 (approved):
This implementation is better. Any further improvements can be done at a later time.

PR #1699, opened 2025-02-11
KFSPTS-33478 Update account search such that the reports to chart field does not have a default value

jhulslander:
Kuali base code has the following inheritance
Account-reportsToCOACode -> Organization-reportsToChartOfAccountsCode -> ChartAttribute-SimpleDropdown

The definition of ChartAttribute-SimpleDropdown includes setting the default value.  This PR creates a new version of ChartAttribute-SimpleDropdown that does not have a default value.

This problem doesn't exist in Monsters.  I was not able to find a misconfiguration in our code as compared to base-code.  I am not sure why we see this problem and I don't know what is causing this.  However, this PR is a solution to this bug.  However, if you think there is a better way to deal with this, please let me know.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/Account.xml
+++ b/src/main/resources/edu/cornell/kfs/coa/businessobject/datadictionary/Account.xml
@@ -332,6 +332,14 @@
     <bean id="invoiceTypeValuesFinder"
           class="edu.cornell.kfs.module.cg.businessobject.options.InvoiceTypeValuesFinder"
           p:keyValuesService-ref="keyValuesService"/>
+          
+    <bean id="ChartAttribute-SimpleDropdown-no-default" abstract="true" parent="ChartAttribute-TextControl" p:control-ref="ChartSimpleDropdownControl" />
+    
+    <bean id="Account-reportsToCOACode"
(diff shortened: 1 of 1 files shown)


cah292 (changes requested):
It looks like KualiCo fixed this with FINP-9877 in the 2023-06-14 financials release. Since we're currently working on upgrading to the 2023-06-28 version of financials, we'll automatically receive this fix with our next upgrade.

If the functionals think this is important enough to fix now (instead of waiting for the 2023-06-28 upgrade), then please adjust the fix to be closer to base code.

For KualiCo's fix, the child of the "Account-reportsToCOACode" bean (on line 1327 in "Account.xml") was updated to the following XML. You might be able to use one of the Field Override helper beans to perform the substitution.

```xml
                <bean parent="Account-reportsToCOACode"
                      p:disableLookup="true">
                    <property name="defaultValueFinder">
                        <null/>
                    </property>
                </bean>
```

jhulslander:
Thanks for noticing the base code Jira.  I have let Cathy know about that and asked her if she wants to wait for the current upgrade.
I was investigating the the solution you noted, but I had an XML syntax error with my first few attempts, so I dropped that approach.  Now I know how to properly write that.

PR #1700, opened 2025-02-12
KFSPTS-34001 Fix permission handling for Vendor report files

cah292:
Base code doesn't explicitly specify "reports/vnd" as a mapped batch file directory. This PR adds such a mapping, so that the related KIM permissions will properly allow authorized users to download files from the "reports/vnd" directory.


jhulslander (approved):
Changes look good, I will merge when the unit tests pass

PR #1703, opened 2025-02-20
KFSPTS-34074 allow older PREQ and CM documents in canceled or disapproved status to be opened

jhulslander:
The post XML ingestion SQL from the 2023-04-19 KualiCo patch did not update canceled or disapproved PREQ documents.  This code change allows those PREQ documents to be opened.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
+++ b/src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
@@ -95,8 +95,6 @@ import java.util.Set;
 public class PaymentRequestDocument extends AccountsPayableDocumentBase implements PaymentSource {
 
     private static final Logger LOG = LogManager.getLogger();
-    
-    private static final String TREASURY_MANAGER_ROUTE_NODE = "TreasuryManager";
 
     protected Date invoiceDate;
     protected String invoiceNumber;
@@ -1240,14 +1238,6 @@ public class PaymentRequestDocument extends AccountsPayableDocumentBase implemen
         }
         if (nodeName.equals(PurapWorkflowConstants.IS_DOCUMENT_AUTO_APPROVED)) {
             return isAutoApprovedIndicator();
-        }
-        if (nodeName.equals(TREASURY_MANAGER_ROUTE_NODE)) {
-            /*
-             * CU Customization KFSPTS-34074
-             * This allows canceled and disapproved PREQ documents that were done before 2/9/2025,
-             * which is when the 2023-04-19 version of Kuali Financials was installed into cu-kfs.  
-             */
-            return false;
         }
         throw new UnsupportedOperationException("Cannot answer split question for this node you call \"" + nodeName +
                 "\"");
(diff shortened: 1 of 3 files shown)


src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
cah292, line 1: Would it better to place the change in our `CuPaymentRequestDocument` subclass instead of placing it in this base class overlay?

Also, I just remembered that Credit Memos had their route nodes modified for the 04/19 patch as well, so the fix should be expanded to include those documents. I'm very sorry for not bringing that up earlier. (57802873 is one example of a problematic CM document in PROD.)

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
```diff
@@ -95,6 +95,8 @@
 public class PaymentRequestDocument extends AccountsPayableDocumentBase implements PaymentSource {
 
     private static final Logger LOG = LogManager.getLogger();
+    
+    private static final String TREASURY_MANAGER_ROUTE_NODE = "TreasuryManager";
```
cah292, line 99: I would suggest using `CUPurapWorkflowConstants.TREASURY_MANAGER` instead.

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
```diff
@@ -1239,6 +1241,14 @@ public boolean answerSplitNodeQuestion(final String nodeName) throws Unsupported
         if (nodeName.equals(PurapWorkflowConstants.IS_DOCUMENT_AUTO_APPROVED)) {
             return isAutoApprovedIndicator();
         }
+        if (nodeName.equals(TREASURY_MANAGER_ROUTE_NODE)) {
+            /*
+             * CU Customization KFSPTS-34074
+             * This allows canceled and disapproved PREQ documents that were done before 2/9/2025,
```
cah292, line 1247: This may need rewording to say what specifically the change allows, or to alternatively replace "allows" with "fixes".

cah292 (changes requested):
I noticed a few minor items that may need to be addressed. Also, I believe one other document needs to be included within the scope of this fix; sorry for not thinking of it earlier.

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
jhulslander: ok

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
jhulslander: ok

cah292 (approved):
Changes look good and my concerns have been addressed. Will merge when the unit tests pass.

jhulslander:

PR #1704, opened 2025-02-25
KFSPTS-34015: Changes needed to fix GL entry issue on PRNC docs after the 2023-04-19 KualiCo patch was applied.

nkimble:


src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
```diff
@@ -208,28 +226,6 @@ public void customizeExplicitGeneralLedgerPendingEntry(
         }
     }
     
-    public void doActionTaken(final ActionTakenEvent event) {
```
cah292, line 211: Just to confirm, was this method's removal needed for the first part of the fix (the part that Daniela had prepared a fix for)? Or was it needed for the double-extract part? Or should it be left in?

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
```diff
@@ -149,6 +147,26 @@ public void populateDocumentForRouting() {
...
+    /* Cornell Customization : KFSPTS-34015
+     * We needed local fixes due to changes introduced by the KualiCo 2023-04-19 Upgrade patch changes.
+     * 
+     * In that patch, method org.kuali.kfs.module.purap.document.PaymentRequestDocument.doRouteStatusChange 
+     * introduced the following service call
+     *       getPdpExtractService().extractPaymentRequestDocument(this);
+     * which saves the paymentRequestDocument to record the extracted and paid timestamps resulting in 
+     * cash offset GL entries being duplicated.
+     * This GL duplicate entry issue was because method populateDocumentForRouting is invoked after a 
+     * docuemnt is saved. This check was added to the method so that populateDocumentForRouting prevented
```
cah292, line 159: Minor typo in comment: `docuemnt` --> `document`

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
```diff
@@ -748,13 +751,16 @@ public void doRouteStatusChange(final DocumentRouteStatusChange statusChangeEven
         if (isAutoApprovedIndicator()) {
             updateAndSaveAppDocStatus(PaymentRequestStatuses.APPDOC_AUTO_APPROVED);
             // DOCUMENT PROCESSED .. //KFSCNTRB-1207 - UMD - Muddu -- end
-        }
-        if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
+        } else if (getDocumentHeader().getWorkflowDocument().isProcessed()) {
+          // Cornell Customization : KFSPTS-34015: Made above line "else if" to be logic prior to 202304-19 changes.
```
cah292, line 755: Is the "else if" change above still needed? I thought we determined that we didn't need it to fix the problem.

cah292 (commented):
There are a few items that might need minor adjustments, but otherwise the changes look good.

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
nkimble: fixed

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
nkimble: Please compare this method to the method of the same name in overlay file PaymentRequestDocument. I believe that it is exactly the same code. The difference IMHO are just in the way that the service calls are being defined/made and the manner in which the constants are being stored/retrieved.

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
nkimble: This is what Cathy executed in sandbox. Please refer to feature branch link in JIRA sub-task for comparison. 
Logic would not be exactly the same as what she executed if that change was made.

src/main/java/edu/cornell/kfs/module/purap/document/CuPaymentRequestDocument.java
cah292: Thanks for the clarification. I agree that the logic and effects are the same, so I'm fine with this method being removed.

src/main/java/org/kuali/kfs/module/purap/document/PaymentRequestDocument.java
cah292: In that case, I'm okay with this change remaining in place for now, and we can do a follow-up fix later if it causes auto-approve-related issues.

cah292 (approved):
Changes look good, unit tests pass and my concerns have been addressed. We can make further changes in a follow-up PR if necessary.

Merging.

PR #1705, opened 2025-02-25
KFSPTS-33582 Implement Part 3 of Sprintax 1042-S handling

cah292:
This PR implements the next section of the Sprintax 1042-S handling, with a main focus on the following areas:

* Added/modified related DAO and criteria/query classes to help with querying and processing the relevant tax data. This section relies heavily upon the metadata code that was introduced by the Part 2 PR. (A few significant changes were made to the metadata handling as well.)
* Added helper classes for processing the TransactionDetail objects and printing related tax files. This PR only implements the classes needed for printing the plain transactions to a tab-delimited text file (like in the old code); the next phase will implement the classes that print the actual Sprintax 1042-S files.
* Made a few other miscellaneous changes, such as adding (or partially adding) some objects that won't be fully utilized until Phase 4 is implemented.

Note that the new tax query API does not provide a full suite of the possible query criteria; more can be added in future user stories as needed. Also, the unit test cases are not exhaustive, so please let me know if you think further scenarios are needed.


src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
```diff
@@ -0,0 +1,309 @@
...
+                        parentDetail.getVendorDetailAssignedIdentifier(), vendorId.getLeft(), vendorId.getRight());
+                return new VendorQueryResults(vendorDetail, null);
+            } else {
+                return new VendorQueryResults(vendorDetail, parentDetail);
+            }
+        }
+    }
+
+
+
```
ajd299, line 194: I am okay with the multiple newlines if that was your intention? There is other places in this file where this situation occurs.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
```diff
@@ -0,0 +1,309 @@
...
+    private static final Logger LOG = LogManager.getLogger();
+
+    private TaxTableMetadataLookupService taxTableMetadataLookupService;
+    private EncryptionService encryptionService;
+
+    @Override
+    public <U> TaxStatistics processTransactionDetails(final TaxBatchConfig config,
+            final TransactionDetailRowMapperFactory<U> rowMapperFactory,
+            final TransactionDetailHandler<U> handler) throws SQLException {
```
ajd299, line 62: Is the throws SQLException needed? I'm okay with leaving it if it is needed once implementation happens.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
```diff
@@ -0,0 +1,309 @@
...
+    @Override
+    public VendorQueryResults getVendor(final Integer vendorHeaderId, final Integer vendorDetailId)
+            throws SQLException {
+        final Pair<Integer, Integer> vendorId = Pair.of(vendorHeaderId, vendorDetailId);
+        final TaxDtoDbMetadata vendorMetadata = taxTableMetadataLookupService
+                .getDatabaseMappingMetadataForDto(VendorField.class);
+        final CuSqlQuery query = createVendorQuery(vendorId, vendorMetadata);
+
+        return queryForResults(query,
+                resultSet -> extractVendorQueryResults(resultSet, vendorId, vendorMetadata));
```
ajd299, line 130: Is this a factory design pattern? Wouldn't it be simpiler to get the results on one line and then call the extract method on the next line?

src/main/java/edu/cornell/kfs/tax/batch/dto/SprintaxInfo1042S.java
```diff
@@ -1,11 +1,114 @@
...
+import org.kuali.kfs.krad.bo.BusinessObject;
+
+import edu.cornell.kfs.tax.batch.dataaccess.TaxDtoFieldEnum;
+
 public class SprintaxInfo1042S {
 
+    private String rowId;
+    private String taxId;
+    private String payeeId;
+    private Integer vendorHeaderId;
```
ajd299, line 15: I'm confused to what this class is representing by looking at the name. Since it is holding the Vendor information should it be named Sprintax1042SVendorInfo? Do I have the purpose correct?

src/main/java/edu/cornell/kfs/tax/batch/service/impl/TaxFileRowWriterBase.java
```diff
@@ -0,0 +1,117 @@
...
+    protected String performDefaultFormatting(final String sectionName, final Object value) {
+        return value != null ? value.toString() : KFSConstants.EMPTY_STRING;
+    }
+
+    @Override
+    public void writeHeaderRow(final String sectionName) throws IOException {
+        Validate.notBlank(sectionName, "sectionName cannot be blank");
+
+        final TaxOutputSectionV2 section = getSection(sectionName);
+        Validate.validState(section.isHasHeaderRow(),
```
ajd299, line 57: Could this be hasHeaderRow() instead of isHasHeaderRow()? For Sprintax won't all of the files have header rows?

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
```diff
@@ -24,7 +24,8 @@ public class TaxProcessingV2ServiceImpl implements TaxProcessingV2Service {
     private static final Logger LOG = LogManager.getLogger();
 
     private TaxProcessingDao legacyTaxProcessingDao;
-    private TaxFileGenerationService taxFileGenerationService;
+    private TaxFileGenerationService taxFileGenerationServiceFor1042S;
+    private TaxFileGenerationService taxFileGenerationServiceForTransactionPrinting;
     private TaxParameterService taxParameterService;
```
ajd299, line 29: Are you re-using a class to do 2 different things? Is the first one generating files?

src/main/resources/edu/cornell/kfs/tax/batch/DefaultTransactionRowOutputDefinitionV2.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<taxOutputDefinition fieldSeparator="&#x9;" includeQuotes="false">
+    <section name="Plain_Transaction_Detail_Row" hasHeaderRow="true" useExactFieldLengths="false">
```
ajd299, line 3: What is the "Plain" implying? Are all the rows DERIVED? Could that be defaulted instead of explicitly specified by each field element?

ajd299 (changes requested):
In addition to the questions within the initial review. I have the following higher level questions. It may be appropriate to have a meeting to go over the code in more detail.

- Is there anyway to make the code simpler (maybe by removing some templated types) without duplicating code? 
- Since KualiCo is using final in their code we are using final whenever possible? In my opinion makes the code bloated and goes against our clean code methodology.
- Is it correct that the code was designed in anticipation of eventually using portions of the code to generate the 1099 files as well?

cah292:
@ajd299 Regarding your high-level questions:

* The templated types are needed to provide flexibility in querying or outputting varying objects. However, I could look into simplifying or removing certain portions, such as replacing the PR's new query criteria API with simple SQL instead.
* Now that base financials code is using the `final` keyword as much as possible, I've been trying to follow that convention when writing or updating our cu-kfs code. If you think this interferes with cleanliness or readability of our code, then we could discuss it further with the team.
* Yes, I've been structuring this new code so that we can eventually use it for rewriting the 1099 process in the future.

src/main/java/edu/cornell/kfs/sys/util/CuSqlQueryPlatformAwareDaoBaseJdbc.java
```diff
@@ -48,6 +50,42 @@ protected int executeUpdate(final CuSqlQuery sqlQuery, boolean logSQLOnError) {
...
+            return getJdbcTemplate().query(statementHandler, statementHandler, resultSetExtractor);
+        } catch (RuntimeException e) {
+            if (logSQLOnError || LOG.isDebugEnabled()) {
+                logSQL(sqlQuery);
+            }
+            LOG.error("queryForResults, Unexpected error encountered while running query!", e);
+            throw e;
+        }
+    }
+
```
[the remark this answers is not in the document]
cah292: I added these new methods on one of our base classes that is not specific to the Tax module, in case we ever need to implement this type of querying in other areas of our code. Because of that, and because even the PR's changes are using this code to query more than just the `TransactionDetail` objects, it needs to use the templated type.

src/main/java/edu/cornell/kfs/sys/util/CuSqlQueryPreparedStatementCreatorAndSetter.java
```diff
@@ -0,0 +1,52 @@
...
+ * Convenience CuSqlQuery-to-JDBC-query class that can be used as both a PreparedStatementCreator
+ * and a PreparedStatementSetter by a JdbcTemplate. Furthermore, it also allows specifying the desired
+ * type and concurrency for the generated ResultSet.
+ */
+public class CuSqlQueryPreparedStatementCreatorAndSetter extends ArgumentTypePreparedStatementSetter
+        implements PreparedStatementCreator, SqlProvider {
+
+    private final CuSqlQuery query;
+    private final int resultSetType;
+    private final int resultSetConcurrency;
```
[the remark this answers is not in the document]
cah292: Many of the queries that retrieve `TransactionDetail` objects need to use ResultSets that have `CONCUR_UPDATABLE` concurrency; that way, they can make minor changes to the objects as they iterate over them. The old 1099 and 1042-S processes also have this behavior.

src/main/java/edu/cornell/kfs/tax/batch/TaxBatchConfig.java
```diff
@@ -2,8 +2,16 @@
...
+import edu.cornell.kfs.tax.util.TaxUtils;
+
 public final class TaxBatchConfig {
 
+    public enum Mode {
+        CREATE_TAX_FILES,
+        PRINT_TRANSACTION_ROWS;
+    }
+
+    private final Mode mode;
```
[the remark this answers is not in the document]
cah292: This enum and field were meant to help pass around a setting that controls the criteria setup for the `TransactionDetail` retrieval query, while also keeping it encapsulated within one of the objects that's already being passed to the relevant data-querying method. The `PRINT_TRANSACTION_ROWS` setting is meant for the phase where the plain transactions are printed to a flat text file, which takes place after printing the 1042-S files (which often consolidate several transactions into a single file line instead).

I'll see if I can rename the enum and/or its constants to make the intent clearer. Sorry for the confusion.

src/main/java/edu/cornell/kfs/tax/batch/CUTaxBatchConstants.java
```diff
@@ -523,6 +523,12 @@ private DerivedFieldNames() {
         }
     }
 
+    public static final class TaxFileSections {
+        public static final String SPRINTAX_BIOGRAPHIC_ROW_1042S = "Sprintax_Bio_Row_1042S";
+        public static final String SPRINTAX_PAYMENT_ROW_1042S = "Sprintax_Payment_Row_1042S";
+        public static final String PLAIN_TRANSACTION_DETAIL_ROW = "Plain_Transaction_Detail_Row";
+    }
+
```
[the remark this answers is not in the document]
cah292: These sections are currently in separate files. As a result of how things are being restructured, though, there's only going to be 1 section per output definition file for the 1042-S and plain-transaction outputs. When/if we rewrite 1099 to use the new process, we may end up having multiple sections in 1 output definition file.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailMapperForSprintax.java
```diff
@@ -0,0 +1,30 @@
...
+     * This method will be fully implemented in a future user story. Only some very minimal coding
+     * has been added thus far for demonstration purposes.
+     */
+    @Override
+    protected void prepareCurrentRowForUpdate(final SprintaxInfo1042S sprintaxInfo) throws SQLException {
+        updateString(TransactionDetailField.vendorName, sprintaxInfo.getVendorNameForOutput());
+        updateString(TransactionDetailField.form1042SBox, CUTaxConstants.TAX_1042S_UNKNOWN_BOX_KEY);
+    }
+
+}
```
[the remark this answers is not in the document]
cah292: When/if we update 1099 to use this new code, I anticipate creating a new subclass of `TransactionDetailMapperBase` for doing so (such as `TransactionDetailMapperFor1099`).

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
```diff
@@ -0,0 +1,309 @@
...
+                encryptionService, vendorMetadata, resultSet);
+        if (!vendorMapper.moveToNextRow()) {
+            return new VendorQueryResults(null, null);
+        }
+
+        final VendorDetailLite vendorDetail = vendorMapper.readCurrentRow();
+        final VendorDetailLite parentDetail;
+        if (!vendorMapper.moveToNextRow()) {
+            if (vendorId.getRight().equals(vendorDetail.getVendorDetailAssignedIdentifier())) {
+                parentDetail = null;
```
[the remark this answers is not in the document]
cah292: Good catch, I'll update this accordingly to remove the unneeded null assignment here.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
cah292: Yes, I intended to have 3-line newline sections in various areas to help with some of the code grouping.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
cah292: Given the exception handling of the JDBC Template, I might be able to remove the `throws SQLException` part here. I'll look into removing it.

src/main/java/edu/cornell/kfs/tax/batch/dataaccess/impl/TransactionDetailProcessorDaoJdbcImpl.java
cah292: This lambda implements the JDBC-Template-related `ResultSetExtractor` interface, which is used by query operations where the callback code manually iterates over the `ResultSet` and returns an object representing the final results. (The `extractVendorQueryResults()` method performs this type of manual iteration over the `ResultSet`.) Using a `ResultSetExtractor` seemed like the best way to support queries that need an updatable `ResultSet`, and I decided to use such extractors for the DAO's other queries as well.

src/main/java/edu/cornell/kfs/tax/batch/dto/SprintaxInfo1042S.java
cah292: This class will eventually hold the information that's needed for generating the file lines in the Sprintax biographic and payment files. Many of these fields come from my earlier all-in-one Sprintax PR that was later abandoned, so most of the code that actually uses these fields won't be in place until Part 4 is implemented.

I'll remove the unused fields for now (and possibly rename the class), and will add things back in for the Part 4 work.

src/main/java/edu/cornell/kfs/tax/batch/service/impl/TaxFileRowWriterBase.java
cah292: This is a boolean property, so its getter method needs to start with "is" or "get" to conform to the JavaBean conventions. I agree that all of the current use cases will have this property set to true; however, I'm not certain whether that will be the case on the 1099 side.

In the meantime, though, I could remove this property, and then we can add it back in later if necessary.

src/main/java/edu/cornell/kfs/tax/service/impl/TaxProcessingV2ServiceImpl.java
cah292: These are two different service implementations with their own distinct tasks. One is responsible for creating the 1042-S Sprintax files, while the other is responsible for printing the transaction rows to a plain text file. The functionals sometimes take the latter file and use it to construct another text file for batch uploads of Transaction Override objects.

src/main/resources/edu/cornell/kfs/tax/batch/DefaultTransactionRowOutputDefinitionV2.xml
cah292: The "Plain" is implying that the `TransactionDetail` contents are being printed as-is, aside from minor formatting changes and/or optional data masking. This particular file will always have the "DERIVED" type for each field; however, when printing the Sprintax files, I anticipate that certain the fields will always have empty or hard-coded values, and such fields will use the "STATIC" type instead. I would prefer not to default the field type unless you think it introduces too much clutter.

cah292:
Upon further consideration, I believe I can update the DB-row-mapping-related objects to use reflection for reading/writing the DTOs from/to the database. It won't run quite as fast, but it will allow me to consolidate several mapping classes into one or two, which should help simplify and cut down on the amount of code needed. I'm currently working on implementing such simplification.

cah292:
I've committed some changes to simplify the new code further. Below are some of the highlights:

* Consolidated the row-mapping classes into a single implementation class. Additional properties were added to the related field enums so that some of the specialized processing could still be performed.
* Consolidated the file-row-writing classes into a single implementation class. Additional attributes were added to the Tax Output Definition JAXB POJOs to support this change, and the "DERIVED" field types have been split up into more specific types.
* Reduced the usage of generic type attributes on the new classes/interfaces where feasible.

Unit tests are still running, so I may need to make minor revisions if the tests fail for some reason. Apart from that, the changes should be ready for review again.

Please let me know if there are any remaining items from your previous review that I still need to address further.

ajd299 (commented):
- I was able run the batch job and unit tests locally with success. 
- The simplification revisions were effective I believe.
- One question: Do the static interfaces within a class need to be static? (FieldReader, FieldValueFormatter, Criteria)

cah292:
Regarding the nested/inner interfaces I was using:

Based on the following StackOverflow article, inner/nested interfaces are already implicitly static. (Inner classes can be non-static if desired, but that doesn't seem to be the case with inner interfaces.)

https://stackoverflow.com/questions/55650365/can-an-interface-nested-inside-a-class-be-non-static

Thus, I could exclude the `static` keyword in their declarations if desired, but I would prefer to leave them in place for clarity.

ajd299 (approved):
Okay I will approve and merge

PR #1706, opened 2025-02-26
  KFSPTS-33035 CreateAccountingDocument Attachment Download Handle Ad…

ajd299:
…ditional Exception

### DIFF
--- a/src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
@@ -9,6 +9,7 @@ import java.net.URLConnection;
 import java.util.Collection;
 import java.util.Locale;
 
+import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Invocation;
 import javax.ws.rs.client.Invocation.Builder;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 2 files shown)


src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -69,18 +70,24 @@ public Attachment createAttachmentFromBackupLink(GloballyUnique parentObject, Ac
...
                 LOG.error("createAttachmentFromBackupLink, Unable to download attachment due to illegal argument exception: " + accountingXmlDocumentBackupLink.getLinkUrl(), iae);
                 throw new ValidationException(UNABLE_TO_DOWNLOAD_ATTACHMENT_MESSAGE + accountingXmlDocumentBackupLink.getLinkUrl());
             }
+        } catch (ProcessingException processingException) {
+            return handleFileException(processingException, accountingXmlDocumentBackupLink);
         }
     }
 
+    protected Attachment handleFileException(Exception e, AccountingXmlDocumentBackupLink backupDocLink) throws ValidationException {
+        LOG.error("createAttachmentFromBackupLink, Unable to download attachment: " + backupDocLink.getLinkUrl(), e);
```
jhulslander, line 87: This log message should reference the function name handleFileException

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
```diff
@@ -69,18 +70,24 @@ public Attachment createAttachmentFromBackupLink(GloballyUnique parentObject, Ac
...
+        throw new ValidationException(UNABLE_TO_DOWNLOAD_ATTACHMENT_MESSAGE + backupDocLink.getLinkUrl());
```
jhulslander, line 88: I think we need to adjust this some more.  One of the goals of this Jira is to help the person receiving this report know that they need to contact the IT people who support their system so they can fix the SSL certificate error, or what whatever might be wrong.

jhulslander (changes requested):
I think this accomplishes the goal of not failing when SSL errors happen.  
We want to give more detailed messages on the report so the person knows they need to contact their IT staff to fix whatever is awry.

src/main/java/edu/cornell/kfs/fp/batch/service/impl/AccountingXmlDocumentDownloadAttachmentServiceImpl.java
ajd299: Please see my latest commit

ajd299:
Please take a look at the formatted error message from my latest commit.

jhulslander (approved):
This looks reasonable to me, I shall merge when the unit tests pass

PR #1707, opened 2025-02-28
KFSPTS-33241 add cache prefix for environment

jhulslander:


src/main/java/org/kuali/kfs/sys/cache/CacheConfiguration.java
```diff
@@ -205,4 +212,10 @@ public RedisCacheManager cacheManager(
 
         return redisCacheManager;
     }
+    
+    protected String getEnvironmentPrefix() {
```
cah292, line 216: Instead of adding this separate method to grab the config property from the Configuration Service, you could try injecting it as a parameter in the `cacheManager()` method. That method (along with others in this class) should have examples of how to use the `@Value` annotation for injecting property values as method parameters.

src/main/java/edu/cornell/kfs/sys/CUKFSConstants.java
```diff
@@ -211,6 +211,7 @@ public static final class WebappPaths {
     }
 
     public static final String CU_ALLOW_LOCAL_BATCH_EXECUTION_KEY = "cu.allow.local.batch.execution";
+    public static final String CU_CACHE_ENVIRONEMNT_PREFIX = "cu.cache.environment.prefix";
```
cah292, line 214: If you still need this constant, then please fix the typo in its name.

cah292 (commented):
Here are my thoughts on this proof-of-concept work for the caching improvements. Aside from a few minor adjustments, I think this solution looks good.

cah292:
Closing this PR since it was meant as a proof of concept and its code is not ready for merging. I will be opening a follow-up PR momentarily that contains a revised version of this feature.

PR #1708, opened 2025-02-28
KFSPTS-32561: Sub-task KFSPTS-34103: Adjust columns in VN-99-03 report file

cah292:
This PR makes some additional VN-99-03 batch job enhancements to accompany the #1697 changes:

* Adds a "Vendor Type" column to the batch job's generated report file. As requested by the functionals, the column is only being added to the report file, not to the CSV files being sent to/from Workday.
* Right-justifies the report's existing "Vendor ID" column.
* Made a few other misc. changes, such as fixing the cleanup of failed report generation attempts and improving the setup of the related unit tests.


ajd299 (approved):
Code looks good, merging

PR #1693, opened 2025-01-30
KFSPTS-32827 I Want Document Warning Message No Attachment

ajd299:

PR #1713, opened 2025-03-13
KFSPTS-34259 Fix IWNT routing when in 4-step mode

cah292:
This PR revises the #1693 IWNT no-attachments prompt feature to fix a few issues:

* Updated code to skip the `promptBeforeValidation()` logic when a pre-route IWNT is in 4-step mode, due to the IWNT document performing equivalent prompts at the end of Step 3 if necessary.
* Updated code to avoid redirecting to the IWNT-submit-success page if the document did not actually get submitted successfully.

The changes above fix a bug where 4-step IWNT docs that lack attachments don't actually get submitted for routing, because the document routing was being stopped by the attachment question prompt but the docs were still redirecting to the success page anyway.

Note that this PR is merging into an emergency fix branch that was created from the master branch, so the unit tests might fail because the 2023-06-28 financials patch is not on the master branch yet.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
+++ b/src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
@@ -884,7 +884,7 @@ public class IWantDocumentAction extends FinancialSystemTransactionalDocumentAct
 
         ActionForward actionForward = super.route(mapping, form, request, response);
         
-        if (CUPurapConstants.IWantDocumentSteps.ROUTING_STEP.equalsIgnoreCase(step)) {
+        if (CUPurapConstants.IWantDocumentSteps.ROUTING_STEP.equalsIgnoreCase(step) && documentRoutingSucceeded()) {
             iWantDocForm.setStep(CUPurapConstants.IWantDocumentSteps.REGULAR);
             iWantDocument.setStep(CUPurapConstants.IWantDocumentSteps.REGULAR);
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
```diff
@@ -911,6 +911,31 @@ private void saveUserOption(String principalId, String userOptionName, String us
...
+    /**
+     * Overridden to skip the missing-attachments prompt when submitting an IWNT document
+     * via the 4-step process, because that process already performs an equivalent prompt
+     * after Step 3 if necessary.
+     * 
+     * If we ever update the IWNT document to include additional prompts, then we may need
+     * to revisit this workaround or revisit the Step-3-prompting logic.
+     */
+    @Override
+    public ActionForward promptBeforeValidation(
```
jhulslander, line 929: I think this approach makes sense here.  I suggest adding a note to the IWantDocumentPreRules.doPrompts() noting if additional prompts are added, this function will need to be thought about too

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
```diff
@@ -911,6 +911,31 @@ private void saveUserOption(String principalId, String userOptionName, String us
         }
     }
 
+    // This logic is based upon the routing-succeeded check from the AccountsPayableActionBase.route() method.
+    private boolean documentRoutingSucceeded() {
```
jhulslander, line 915: this function name makes me think it will be looking at a document workflow status, or something along those lines. Should this function be renamed to convey it is just looking for a "successfully routed" message?
Is the comment about AccountsPayableActionBase helpful?

jhulslander (commented):
I found a few minor things to consider

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
cah292: Ok, I'll add a note there accordingly.

src/main/java/edu/cornell/kfs/module/purap/document/web/struts/IWantDocumentAction.java
cah292: I'll rename the method for clarity. I also agree that the comment is likely unnecessary, so I'll remove that too.

cah292:
I've committed some more changes to address the concerns that were brought up. This is ready for review again.

jhulslander (commented):
changes look good, we can merge when the unit tests pass

jhulslander (approved):
Unit tests now pass, I shall merge

PR #1714, opened 2025-03-13
KFSPTS-34259 Fix IWNT routing when in 4-step mode

cah292:
This PR has the exact same code changes as those from the #1713 PR, except that this PR is applying them to the develop branch instead of an emergency fix branch. That way, we can finish migrating the fix into the regular develop and master branches.


jhulslander (approved):
I compared this PR with https://github.com/CU-CommunityApps/cu-kfs/pull/1713 and they seem to be the same change, so that is good.
The unit tests pass, so I shall merge

PR #1720, opened 2025-03-26
Kfspts 34140 jdh34

jhulslander:


src/main/java/edu/cornell/kfs/sys/batch/service/impl/CuCacheServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+public class CuCacheServiceImpl extends CacheServiceImpl {
+    private static final Logger LOG = LogManager.getLogger();
+
+    protected CacheManager cacheManager;
+
+    @Override
+    public void clearSystemCaches() {
+        LOG.debug("clearSystemCaches, cache manager class {}", cacheManager.getClass());
+        for (final String cacheName : cacheManager.getCacheNames()) {
+            LOG.debug("clearSystemCaches, attempting to clear cache for {}", cacheName);
```
cah292, line 17: After looking at the spring-data-redis code further, you may want to also log the prefix info from the cache configuration object(s), since the prefixes might not necessarily be shown in the regular cache-name-returning methods.

Here are some examples of how to retrieve a cache's configuration object and the configured prefix. Note that you may have to typecast the cache manager as a `RedisCacheManager` (or typecast the returned cache as a `RedisCache`).

```java
cacheManager.getCacheConfigurationFor(cacheName).getKeyPrefix()
cacheManager.getCacheConfigurationFor(cacheName).getKeyPrefixFor(cacheName)
cacheManager.getCache(cacheName).getCacheConfiguration().getKeyPrefix()
cacheManager.getCache(cacheName).getCacheConfiguration().getKeyPrefixFor(cacheName)
```

cah292: @jhulslander,

Sorry, I must have been looking at the wrong versions of the files for the first two calls. However, I believe the last two calls from earlier should work, if you typecast the `cacheManager.getCache()` result to a `RedisCache`:

```java
((RedisCache) cacheManager.getCache(cacheName)).getCacheConfiguration().getKeyPrefix()
((RedisCache) cacheManager.getCache(cacheName)).getCacheConfiguration().getKeyPrefixFor(cacheName)
```

jhulslander:
@cah292 I think found the problem
in sandbox transactional in kfs-config.properties environment is kfs-sandbox
in sandbox batch, in build.xml environment is sandbox
so the calculated prefix as seen in sandbox batch is "kfs_sandbox_"

I think the best option is to add cu.redis.cache.environment.prefix as new value we add to the kfs-config.properties file in our server environments

cah292:
@jhulslander Thanks for finding the discrepancy! I agree that overriding `cu.redis.cache.environment.prefix` in the puppet config files would be the best approach, given the current setup for keeping the regular `environment` property value distinct between runtime and batch.

If you want me to adjust my PR's `cu.redis.cache.environment.prefix` setup to avoid referencing the `environment` property, please let me know.

jhulslander:
I think it would be best to force cu.redis.cache.environment.prefix to come from our config file, so we don't accidentally forget to set it and have a mis-match between transactional and batch.  I will work on getting update into puppet

dp65:
After rereading both Chad and Jay's comments, I guess I'm agreeing with you both. Sorry for not catching on to that sooner, Chad.

jhulslander:
this PR was just for testing / debugging, we don't need this any longer

PR #1721, opened 2025-03-31
KFSPTS-34443 Fix issues with unexpected drop-down values

cah292:
This PR fixes two drop-down-related issues that seem to have resulted from the 06/28/2023 financials upgrade changes:

1. Fixed an issue where the IWNT document's Yes/No drop-downs were instead displaying Y/N/(Blank) as selectable values. There appears to be another values finder class in the CAM module with the same name as the one used for IWNT Yes/No fields; something must have changed with the 06/28/2023 patch to cause the CAM version to be treated as the latest override. Renaming the IWNT's related values finder class appears to resolve the problem.
2. Fixed an issue where our add-blank-entry customization for newer-style lookup drop-downs was not properly adjusted for the 06/28/2023 upgrade. It used to be located in a JAX-RS resource class that was removed by the 06/28/2023 patch, but our upgrade efforts moved it into a `KeyValuesBase` method that might not actually affect the lookups as intended. I was able to fix the problem by moving the customization into an overlay of a Spring-MVC-related helper service.


src/main/java/org/kuali/kfs/sys/rest/resource/businessobject/BusinessObjectsControllersHelperService.java
```diff
@@ -0,0 +1,423 @@
...
+                }
+                final List<HierarchicalData> values = valuesFinder.getHierarchicalControlValues();
+                valuesMap.put(singleAttributeName, values);
+            } else {
+                final KeyValuesFinder valuesFinder = control.getValuesFinder();
+                if (valuesFinder == null) {
+                    continue;
+                }
+                // CU Customization: keyValues list now comes from the helper method below.
+                final List<KeyValue> keyValues = getKeyValuesForLookup(valuesFinder);
```
jhulslander, line 223: would it make more sense if this block was included in the if statement above and the log was changed to valuesFinder != null

src/main/java/org/kuali/kfs/sys/rest/resource/businessobject/BusinessObjectsControllersHelperService.java
```diff
@@ -0,0 +1,423 @@
...
+     * CU Customization: Added this method and the one right below it,
+     * to forcibly add a blank key-value entry if the values finder does not return one.
+     */
+    private List<KeyValue> getKeyValuesForLookup(final KeyValuesFinder valuesFinder) {
+        final List<KeyValue> keyValues = valuesFinder.getKeyValues();
+        if (hasEntryForBlankKey(keyValues)) {
+            return keyValues;
+        } else {
+            final KeyValue blankKeyValue = new ConcreteKeyValue(StringUtils.EMPTY, StringUtils.EMPTY);
+            return Stream.concat(Stream.of(blankKeyValue), keyValues.stream())
```
jhulslander, line 242: Do we need to do anything to force the empty key value pair to be the first item in the returned list?

jhulslander (changes requested):
I had a question and found something to possibly improve this

src/main/java/org/kuali/kfs/sys/rest/resource/businessobject/BusinessObjectsControllersHelperService.java
cah292: In my opinion, such restructuring would result in unnecessary deviation from how KualiCo had coded this base method. The subsequent put-in-map line needs to be skipped if the values finder is null, meaning there's more than just this one customized line that needs adjustments if we modify the if-null-then-continue block above. Or are you referring to some other type of refactoring here?

jhulslander: ah, I think I misunderstood somthing, lets leave it as you coded it

src/main/java/org/kuali/kfs/sys/rest/resource/businessobject/BusinessObjectsControllersHelperService.java
cah292: This particular setup should already return the empty key/value pair as the first one. `Stream.of()` and `Collection.stream()` both return sequential streams, so their concatenation via `Stream.concat()` should also be a sequential stream.

jhulslander (approved):
my concerns have been answered, I shall merge

PR #1724, opened 2025-04-11
KFSPTS-34577 create api authentication framework

jhulslander:
This is NOT ready for review. 
I used vsCode and cline to vibe code my idea for framework .... two birds with one stone ... experiment with vibe coding and come up with an idea for the api authentication framework

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/service/impl/APIAuthenticatorServiceImpl.java
+++ b/src/main/java/edu/cornell/kfs/sys/service/impl/APIAuthenticatorServiceImpl.java
@@ -1,0 +1,51 @@
+package edu.cornell.kfs.sys.service.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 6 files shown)


src/main/java/edu/cornell/kfs/sys/service/impl/APIAuthenticatorServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.kuali.kfs.krad.service.BusinessObjectService;
+
+import edu.cornell.kfs.sys.businessobject.APIAuthenticator;
+import edu.cornell.kfs.sys.businessobject.APIEndpointDescriptor;
+import edu.cornell.kfs.sys.service.APIAuthenticatorService;
+
+public class APIAuthenticatorServiceImpl implements APIAuthenticatorService {
```
jhulslander, line 12: Meh, this service logic isn't right, but the rest of code isn't bad

PR #1729, opened 2025-04-15
KFSPTS-34127 Fix quote handling in Concur SAE files

cah292:
We sometimes encounter SAE files whose cells are unquoted yet may contain quotes within the value. Rather than letting the parser fail, we want some way of detecting and mitigating such scenarios.

This PR introduces some changes to how the Concur SAE Job parses the quotes within input files. An extra pre-cleaning process has been added that will look for the quotes mentioned above, and then double-up the quotes to allow the parser to succeed. Afterwards, our existing `ConcurSaeDelimitedFlatFileSpecification` class will strip off the internal quotes and keep track of which cells were cleaned up, like it currently does in our Production code.

There may still be some cases where improper quoting can fail the file (such as unmatched quotes at the start/end of a cell), but it should generally be okay to let those remain as error cases, since it would be difficult to discern whether the cell was meant to be fully quoted in such situations.


ajd299 (approved):
Code looks good, merging

PR #1731, opened 2025-04-16
KFSPTS-34622: Convert PurgeTablesStep joda DateTime to java.time.LocalDateTime

nkimble:


nkimble:
File cu-spring-sys.xml is touched by these code changes as well as the Kuali Co 08/30/2023 patch. Moving this change set out to next spring so that it can applied after that KualiCo patch goes into the develop branch.

PR #1732, opened 2025-04-16
KFSPTS-34628: Remove all pieces of decommissioned batch job CreateClosedAccountCsvJob instead of updating joda to java time libraries.

nkimble:


nkimble:
There is a dependent SQL PR: https://github.com/CU-CommunityApps/nonprod-sql/pull/1507

Please make sure both are good to go before merging.

jhulslander (approved):
Changes look good.  I shall merge
I think we need to add sub tasks for devops to remove the closed account folders as we progress this code change to the different environments.
Should we also have sub tasks to confirm with prod control that the job has totally been removed from Tidal, including their steps to copy the files to workday?

nkimble:
Three sub-tasks added with work grouped as follows:
1.  DEV and SANDBOX
2. TEST
3. PROD and EMERG

I Have add two sub-tasks for prod control for this batch job (test first then prod).

My preference is to have another user story for the file copy that they used to do because that is a separate batch job with a link to this issue. I don't want the decommission of CO-01-03 to be hard to find in the future and it might be if it were buried in a sub-task on an issue with a different title.

PR #1508, opened 2023-10-10
KFSPTS-18193 update create accounting document job, payee address id, subobject, subaccount and virus checking

jhulslander:
This user story has a collection of fixes to create accounting document job.  While working on this, I noticed Jenkins wasn't running all the unit tests.  That got fixed by the change to pom.xml. I had to make CreateAccountingDocumentServiceImplTestBase abstract otherwise Jenkins gave an error on that file.  A few other changes to unit tests had to be added to get them to run again

A SQL foreign key error happens when the sub account or sub object have spaces in the XML.  If invalid characters are entered on either field, those are handled correctly.  I opted to fix that by trimming the strings when marshaing/unmarshalling the XML.  To do so I created TrimmedStringXmlAdapter.  I added a unit test to make sure the various iterations of input are handled as expected.  If you think there is a better way of fixing this, let me know.

PR #1737, opened 2025-04-23
KFSPTS-34668 PaymentApplication GLPE Generation CustomerInvoiceDetail…

ajd299:
….ObjectCode PersistanceBrokerException Workaround

### DIFF
--- a/src/main/java/org/kuali/kfs/module/ar/document/PaymentApplicationDocument.java
+++ b/src/main/java/org/kuali/kfs/module/ar/document/PaymentApplicationDocument.java
@@ -1,0 +1,1616 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2023 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


cah292:
I'm still looking into this, but if the error is being caused by an OJB proxy that maps to a null/nonexistent object, it would be helpful to figure out which absent or inaccurate data value is causing the refresh-reference-object calls to fail to reload the object. If you haven't already, could you please check to see whether there are problems with the Invoice Detail's postingYear, chartOfAccountsCode or financialObjectCode field values?

I'll continue reviewing this tomorrow unless someone else wants to pick this up first.

ajd299:
@cah292 In KFS-TEST, AR_INV_DTL_T.FDOC_POST_YR is null. 

- In a local temporary database I will try to update it to 2025 via SQL and see if that fixes the issue.

ajd299:
After updating AR_INV_DTL_T.FDOC_POST_YR the error isn't fixed.

queries for reference:
`select * 
from AR_INV_DTL_T
where FDOC_NBR='58345281';

update AR_INV_DTL_T
set FDOC_POST_YR='2025'
where FDOC_NBR='58345281';

select *
from CA_OBJECT_CODE_T
where FIN_OBJECT_CD in ('4480', '1300')
and UNIV_FISCAL_YR=2025
and FIN_COA_CD='IT'`

ajd299:
I have opened PR to update the year.

https://github.com/CU-CommunityApps/nonprod-sql/pull/1508/files

db483:
@ajd299 The above SQL will fix the existing PaymentApplication but I think you will also need a code fix to make sure the fdoc post year is properly populated in the code? Thank you

PR #1738, opened 2025-04-23
KFSPTS-34600: REQUIRED SECOND COMMIT FOR THIS ISSUE: Fix for UTC time still being locale time in file header. Added more unit tests due to original one not catching this.

nkimble:


src/test/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerGenerateSupplierXmlServiceImplTest.java
```diff
@@ -126,6 +126,46 @@ void testDateFormatters() {
...
     }
     
+    @Test
+    void testDateFormatterLocaleFileName() {
+        Clock myClock = Clock.systemDefaultZone();
+        LocalDateTime dayAndTimeNow = LocalDateTime.now(myClock);
+        
+        String manualFileNameDateString = new String();
+        int myNanoSeconds = dayAndTimeNow.getNano() / 1000000;
+        manualFileNameDateString = manualFileNameDateString + dayAndTimeNow.getYear()
```
db483, line 136: Perhaps these two new tests could be simplified by using a String formatter and a constant date for testing?

src/main/java/edu/cornell/kfs/module/purap/batch/service/impl/JaggaerGenerateSupplierXmlServiceImpl.java
```diff
@@ -329,7 +333,7 @@ protected List<SupplierSyncMessage> buildSupplierSyncMessageList(List<Supplier>
     private Header buildHeader() {
         Header header = new Header();
         header.setMessageId(UUID.randomUUID().toString());
-        header.setTimestamp(dateTimeService.getLocalDateTimeNow().format(DATE_TIME_ZONE_UTC_FORMATTER_yyyy_MM_dd_T_HH_mm_ss_SSS_Z));
+        header.setTimestamp(LocalDateTime.now(utcClock).format(DATE_TIME_ZONE_UTC_FORMATTER_yyyy_MM_dd_T_HH_mm_ss_SSS_Z));
```
db483, line 336: is this change needed? The old code was header.setTimestamp(DATE_FORMATTER_FOR_HEADER_DATE.print(dateTimeService.getCurrentDate().getTime()));

Is this going to change the timestamp in the header when it shouldn't?

db483: I think we should look at old files in DEV/TEST to see what the header date was. Unfortunately I am getting a 502 when trying to get in either DEV or TEST this morning.

db483 (changes requested):
Please see my thoughts on these changes.

PR #1744, opened 2025-04-30
KFSPTS-33874 remove aws secrets validation job

db483:


pom.xml
```diff
@@ -41,6 +40,7 @@
         <jsp-api.version>2.3.1</jsp-api.version>
         <jsr311-api.version>1.1.1</jsr311-api.version>
         <junit.version>5.9.3</junit.version>
+        <lettuce.version>6.2.4.RELEASE</lettuce.version>
```
cah292, line 43: What was the reason for updating the "lettuce" and "spring-data-redis" dependencies as part of this PR? I don't see how they relate to the removal of the AWS Secrets handling.

cah292 (commented):
There are some POM-related changes that I wanted to get some clarification on, but other than that the PR looks good.

pom.xml
db483: I added them due to these failures: https://new-jenkins.kuali.cucloud.net/job/unit-test-develop-pr/edu.cornell.kfs$kfs-war/2570/testReport/junit/edu.cornell.kfs.coa.businessobject.lookup/AccountGlobalSearchLookupableHelperServiceImplIntegTest/testValidateSearchParameters/

Perhaps there is a better solution? Thank you

cah292: It looks like we have several dependency-level and WAR-overlay-level netty JAR exclusions, possibly to prevent version conflicts with those that were being pulled in by the secretsmanager dependency. Perhaps removing some or all of those exclusions will fix the problem; just make sure that such adjustments don't result in duplicate dependencies when building the WAR.

db483: Thank you for your feedback, I made some further changes. Please let me know if this looks better.

cah292: Thanks for the updates. I confirmed locally that your feature branch can be built into a WAR containing the appropriate netty JARs. These changes should be ready to move forward now.

cah292 (approved):

PR #1746, opened 2025-05-06
KFSPTS-33899 upgrade to the 09/27/2023 version of financials

db483:


cah292 (changes requested):
I'm still reviewing the code changes, but I did find one item thus far that might need attention.

src/main/java/org/kuali/kfs/coa/document/validation/impl/AccountGlobalRule.java
```diff
@@ -135,296 +137,286 @@ public void setupConvenienceObjects() {
...
-		success &= checkRemoveExpirationDate();
-		success &= checkRemoveContinuationChartAndAccount();
-		success &= checkRemoveIncomeStreamChartAndAccount();
+        success &= checkEmptyValues();
+        success &= checkGeneralRules(document);
+        success &= checkContractsAndGrants();
 
-		success &= checkEmptyValues();
-		success &= checkGeneralRules(document);
-		success &= checkCloseAccounts();
```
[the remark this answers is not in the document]
db483: updated

src/main/resources/edu/cornell/kfs/module/cam/document/workflow/EquipmentLoanOrReturnDocument.xml
cah292, line 1: I thought we weren't adding doc type XML to our codebase unless it contained CU-specific customizations. Should this XML file only be included in the XML ingestion sub-tasks instead?

Also, just to confirm what I think you already mentioned in the cu-kfs-upgrades reports: Are we leaving our workflow as-is for the documents INVW (Customer Invoice Writeoff) and PAAT (Payee ACH Account)?

src/main/resources/edu/cornell/kfs/sys/workflow/RoutingFieldConfigMaintenanceDocument.xml
cah292, line 1: Another workflow XML file that might need removal if there are no CU-specific changes. Not sure why Git thought the file was being renamed.

cah292 (changes requested):
I found a couple items that might need quick cleanup, but overall the changes look good.

I'll add a separate comment for one other issue I found.

cah292 (changes requested):
There's one other thing that might need to be addressed:

Please double-check the CU-specific field override beans in `ACHPayee.xml`. Additional overrides might be needed for the new/converted variant of the lookup.

src/main/resources/edu/cornell/kfs/module/cam/document/workflow/EquipmentLoanOrReturnDocument.xml
db483: Good question Chad. I wasn't sure what is the best way to go about the new doc type xml. I can remove them and attach them to the "Ingest xml jiras".

That is correct, we are leaving the  INVW (Customer Invoice Writeoff) and PAAT (Payee ACH Account) as is.

src/main/resources/edu/cornell/kfs/sys/workflow/RoutingFieldConfigMaintenanceDocument.xml
db483: removed

db483:
Thank you for noting the ACHPayee.xml, I am looking.

db483:
I made the changes to the ACHPayee set up but I am still running into an issue where the principalName field is empty in the new lookup results and even if I enter a value in the search field it is not take into account. To test the new lookup you need to go directly to: http://localhost:8080/kfs/webapp/lookup/ACHPayee
When debugging I can see that the new search service is not being called.
I haven't yet figured out what is causing this issue. I will continue to look into it Monday morning. If you happen to spot the problem please let me know. Thank you!

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
```diff
@@ -1,17 +1,21 @@
...
-  <bean id="ACHPayee" parent="ACHPayee-parentBean">
-    <property name="businessObjectClass" value="edu.cornell.kfs.pdp.businessobject.CuACHPayee" />
-    <property name="attributes">
-      <list merge="true">
-        <ref bean="ACHPayee-principalName" />
-      </list>
-    </property>
-    <property name="relationships">
+  
+  <bean id="CuACHPayee" parent="CuACHPayee-parentBean"/>
```
cah292, line 9: I think the custom lookup field is failing because the PR changed this bean file to no longer override the "ACHPayee" data dictionary bean. Please adjust this to override the "ACHPayee" bean as before.

Because our code is using a custom `CuACHPayee` subclass, you may need to use a URL path like the following to access the converted lookup: http://localhost:8080/kfs/webapp/lookup/CuACHPayee

Please also check if the converted lookup is accessible from the Payee ACH Account Extract lookup (by clicking the lookup icon next to the "Payee Number" field). If that icon still opens the old lookup, that should be fine for now, and we can re-test the icon when we upgrade to the 2023-12-06 patch (which refactors the Payee ACH Account Extract lookup as per FINP-10429).

If the data dictionary expects a bean named "CuACHPayee" to be present, then add a bean alias tag that references the overridden "ACHPayee" bean.

Please let me know if you continue to run into issues with this lookup refactoring.

cah292 (changes requested):
@db483 I believe I've found the problem with the ACH Payee lookup conversion. If you continue to run into issues after making the suggested changes, please let me know.

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
db483: Thank you @cah292, I initially tried to override ACHPayee but I am getting the following error:
`is java.lang.NullPointerException] with root cause
java.lang.NullPointerException
	at org.kuali.kfs.sys.rest.resource.businessobject.BusinessObjectsControllersHelperService.getLookup(BusinessObjectsControllersHelperService.java:121)
	at org.kuali.kfs.sys.rest.resource.businessobject.BusinessObjectsLookupController.getLookup(BusinessObjectsLookupController.java:80)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
`

I committed those changes. Maybe I have a mistake in the way I override the bean? Thank you

cah292: @db483 I think your latest edit accidentally removed the `parent="ACHPayee-parentBean"` attribute from the overridden "ACHPayee" bean. Adding that attribute back in should (hopefully) fix the problem.

db483: Thank you @cah292 , I had tried that version too but I am getting the same error.

cah292: @db483 Based on where it's failing, it appears that our overlay of `BusinessObjectsControllersHelperService.getLookup()` is not able to find the needed `BusinessObjectEntry` from the data dictionary. I would suggest debugging the overlayed method to determine what `businessObjectName` argument value is being passed in; that should help with determining next steps.

db483: The businessObjectName is CuACHPayee. There is no mistake in the way I set up the override in the data dictionary. The lookup will work if the http://localhost:8080/kfs/webapp/lookup/CuACHPayee is used. If we want this url to use the http://localhost:8080/kfs/webapp/lookup/ACHPayee form then I think we need to adjust the way principal name is added to the ACHPayee BO, perhaps by using an extended attribute.

cah292: In that case, I agree that adjusting our ACHPayee customization to use an extended attribute would be the best way to move forward. Also, since ACHPayee is a transient business object, such a change might not need database updates.

db483: Thank you @cah292 . Since ACHPayee is a transient bo and it does not extend PersistableBusinessObjectBase it cannot use an extended attribute. I decided to overlay the ACHPayee file to implement this change.

db483: @cah292 This is ready for review again. Thank you

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
```diff
@@ -33,6 +55,20 @@
     <property name="fieldOverrides">
       <list>
         <!-- Add "principalName" to the ACHPayee lookup input fields. -->
+
+        <bean parent="FieldOverrideForListElementInsert">
+          <property name="propertyName" value="displayAttributeDefinitions" />
```
cah292, line 60: These new field override beans need adjustments so that they line up with the XML comments. ("formAttributeDefinitions" represents the search criteria fields; "displayAttributeDefinitions" represents the search results fields. Also, "ACHPayee-entityId" is in the search criteria, whereas "DisbursementPayee-payeeIdNumber" is in the search results.)

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
```diff
@@ -0,0 +1,61 @@
...
+import org.kuali.kfs.krad.util.GlobalVariables;
+import org.kuali.kfs.pdp.PdpConstants;
+import org.kuali.kfs.pdp.PdpKeyConstants;
+import org.kuali.kfs.pdp.businessobject.ACHPayee;
+import org.kuali.kfs.pdp.businessobject.service.ACHPayeeSearchService;
+import org.kuali.kfs.sys.KFSPropertyConstants;
+import org.kuali.kfs.vnd.businessobject.VendorDetail;
+import org.springframework.util.MultiValueMap;
+
+public class CuACHPayeeSearchService extends ACHPayeeSearchService{
```
cah292, line 27: Missing space prior to the curly brace.

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
```diff
@@ -0,0 +1,61 @@
...
+    
+    public CuACHPayeeSearchService(DataDictionaryService dataDictionaryService,
+            DisbursementVoucherPayeeService disbursementVoucherPayeeService, Lookupable vendorLookupable,
+            PersonService personService) {
+        super(dataDictionaryService, disbursementVoucherPayeeService, vendorLookupable, personService);
+    }
+
+    protected Map<String, String> getPersonFieldValues(final Map<String, String> fieldValues) {
```
cah292, line 35: Should the `@Override` annotation be added to this class's methods accordingly?

src/main/java/org/kuali/kfs/pdp/businessobject/service/ACHPayeeSearchService.java
```diff
@@ -0,0 +1,347 @@
...
+            final VendorDetail vendorDetail = (VendorDetail) vendor;
+            final DisbursementPayee payee = getPayeeFromVendor(vendorDetail);
+            payeeList.add(payee);
+        }
+
+        return payeeList;
+    }
+
+    //CU customization change from private to protected
+    protected ACHPayee getPayeeFromVendor(final VendorDetail vendorDetail) {
```
cah292, line 227: Do you still need to override the visibility of this method? Can it be reverted back to the visibility from base code?

cah292 (changes requested):
I found a few more areas with the ACHPayee fix that need cleanup. Other than that, the changes look good.

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/PayeeACHAccount.xml
```diff
@@ -16,7 +16,7 @@
       <list merge="true">
         <bean parent="RelationshipDefinition">
           <property name="objectAttributeName" value="achPayee"/>
-          <property name="targetClass" value="edu.cornell.kfs.pdp.businessobject.CuACHPayee" /><!-- CU Customization: Changed target class. -->
+          <property name="targetClass" value="org.kuali.kfs.pdp.businessobject.ACHPayee" />
```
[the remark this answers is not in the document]
db483: Good question. My understanding is that this is adding additional support attributes. Is there a better way to do this? Or do you think the additional support attributes not needed?

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
```diff
@@ -21,9 +22,30 @@
             </list>
           </property>
         </bean>
+          <bean parent="RelationshipDefinition"
+                      p:objectAttributeName="achPerson">
```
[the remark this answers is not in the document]
db483: update

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
db483: update

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
db483: updated

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
db483: updated

src/main/java/org/kuali/kfs/pdp/businessobject/service/ACHPayeeSearchService.java
db483: updated

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/PayeeACHAccount.xml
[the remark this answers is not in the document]
cah292: Would one of the DD Field Override helper beans work here? Not sure what effect that would have on the Relationship Definition functionality, though.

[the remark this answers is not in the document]
db483: The current implementation works as expected so I would prefer keeping it as is.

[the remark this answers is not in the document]
cah292: In that case, would it be feasible to remove the `merge="true"` setting from the wrapper list, so that the "relationships" list is fully overwritten? (It might be helpful to add a comment as to why a no-merge approach is being used.) If you'd rather keep this as-is for now and investigate adjusting the list in a future user story, please let me know.

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
```diff
@@ -58,8 +58,43 @@
             </list>
           </property>
         </bean>
+        
+        <!-- Add "principalName" to the ACHPayee lookup results. -->
```
cah292, line 62: What I meant before was that the XML comments were making distinctions between input field overrides and result field overrides; sorry if that wasn't clear. Some reordering or additional comments are needed to get things to line up; alternatively, you could just remove the input field and result field comments if you think the beans are self-explanatory enough.

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
```diff
@@ -58,8 +58,43 @@
             </list>
           </property>
         </bean>
+        
+        <!-- Add "principalName" to the ACHPayee lookup results. -->
+        <bean parent="FieldOverrideForListElementInsert">
+          <property name="propertyName" value="formAttributeDefinitions" />
+          <property name="propertyNameForElementCompare" value="name" />
+          <property name="element">
+            <ref bean="DisbursementPayee-payeeIdNumber"/>
```
cah292, line 67: I believe the "formAttributeDefinitions" comparison should check against the `ACHPayee-entityId` bean instead. Likewise, the "displayAttributeDefinitions" comparison should check against the `DisbursementPayee-payeeIdNumber` bean instead. Sorry if that suggestion from my previous review wasn't clear.

Note that if the comparison element is not found in the list, I think KFS will stick the insertion beans at the end of the list instead. That explains why KFS can still launch even though these comparison elements cannot be found.

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        // add entity
+        personFieldValues.put(KIMPropertyConstants.Person.ENTITY_ID, fieldValues.get(KIMPropertyConstants.Person.ENTITY_ID));
+        // add principal name
+        personFieldValues.put(KIMPropertyConstants.Principal.PRINCIPAL_NAME, fieldValues.get(KIMPropertyConstants.Principal.PRINCIPAL_NAME));
+        return personFieldValues;
+    }
+    
+    protected DisbursementPayee getPayeeFromPerson(final Person personDetail, final Map<String,String> fieldValues) {
```
cah292, line 55: This method should also have an `@Override` annotation added.

cah292 (changes requested):
In addition to the one remaining list-overwrite suggestion, I noticed a few items from the latest revisions that could be cleaned up.

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
db483: Good catch! Updating

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/ACHPayee.xml
db483: updating

src/main/java/edu/cornell/kfs/pdp/businessobject/service/CuACHPayeeSearchService.java
db483: updated

src/main/resources/edu/cornell/kfs/pdp/businessobject/datadictionary/PayeeACHAccount.xml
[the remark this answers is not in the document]
db483: Updating

cah292 (commented):
Code changes look good and my concerns have been addressed.

Do you want me to wait until Tuesday morning to merge the changes, given that we also need to run the SQL in DEV? Or do you want me to proceed with the merge today anyway?

db483:
@cah292 Thank you for checking! I can appl the SQL to the DEV database tomorrow morning before the DEV deploy if you merge the changes.

cah292:
@db483 Thanks! I'll get the changes merged momentarily.

cah292 (approved):
Changes look good, unit tests pass and the related PRs are also ready. Merging.

PR #1748, opened 2025-05-09
KFSPTS-34622: Convert PurgeTablesStep joda.time and java.util.Date to java.time.LocalDate

nkimble:


src/main/java/edu/cornell/kfs/sys/batch/PurgeTablesStep.java
```diff
@@ -2,19 +2,29 @@
...
 import java.util.Date;
 
+import org.kuali.kfs.core.api.datetime.DateTimeService;
 import org.kuali.kfs.sys.batch.AbstractStep;
 
 import edu.cornell.kfs.sys.batch.service.TablesPurgeService;
 
 public class PurgeTablesStep extends AbstractStep {
     
+    private DateTimeService dateTimeService;
```
db483, line 12: Is the dateTimeService needed here? Or could this just use the one from AbstractStep?

db483 (changes requested):
Changes look good but I had one thing to clarify. Thank you.

src/main/java/edu/cornell/kfs/sys/batch/PurgeTablesStep.java
nkimble: Good catch. Changing.

db483 (approved):
Changes look good and my concern has been addressed.

PR #1749, opened 2025-05-13
KFSPTS-34870 Fix Org/Account maintenance XML conversion

cah292:
We discovered during FYE2025 Dress Rehearsal testing that the 08/30/2023 patch's removal of the Org field `organizationInFinancialProcessingIndicator` was preventing certain maintenance documents (especially Account documents) from opening successfully, due to improper maintenance XML conversion. We had already included the relevant FINP-10001 maintenance XML conversion changes as part of the 08/30/2023 upgrade; however, it appears that KualiCo's new conversion rules are not sufficient for converting nested Organization objects within another BO (Account, Award, etc.). This PR adds another set of conversion rules to properly convert those nested objects.

NOTE: For the refactoring of the unit tests, the contents of `AccountTest.xml` have been transferred over to `LegacyAccountTest.xml`, and the former file has been updated with a different Account document that needs the new conversion rules in place.

### DIFF
--- a/src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
+++ b/src/main/resources/edu/cornell/kfs/krad/config/MaintainableXMLUpgradeRules.xml
@@ -673,6 +673,17 @@
             </pattern>
         </pattern>
         <pattern>
+            <class>organization</class>
+            <pattern>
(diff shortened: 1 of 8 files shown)


db483 (approved):
Changes look good.

PR #1751, opened 2025-05-21
KFSPTS-33900 upgrade to the 10/25/2023 version of financials

db483:


jhulslander:
The follow up changes that Chad made seem reasonable to me, this is ready to be merged

jhulslander (approved):
I have reviewed Chad's edit and they look fine to me.  if needed I will dismiss Chad's review so this can be merged

PR #1753, opened 2025-05-21
KFSPTS-34685 Partial Backport FINP-10379 CustomerInvoiceDetail Posting Year

ajd299:


jhulslander:
@ajd299 I found this KualiCo PR that fixes this problem: https://github.com/KualiCo/financials/pull/6364
I think we should back port this fix if possible

jhulslander (changes requested):
see my note about https://github.com/KualiCo/financials/pull/6364

jhulslander:
This looks right, but I need to review this some more in the morning.  I did notice it looked like there some formatting discrepenscies.  I can confirm that in the morning with fresh eyes.  
Just to confirm, you started with the 2023-09-27 version of ContractsGrantsInvoiceDocumentServiceImpl in base code, and then applied the changes on the FINP PR noted above?

ajd299:
Yes, the only practical differences between the 2023-09-27 version should be:

1. The first parameter is the document instead of the docNumber (and also reformatted the parameters)
2. The postingYear is set

To avoid doing an overlay I created new method in our extension file.

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
```diff
@@ -43,6 +46,145 @@
...
+                        sequenceNumber
+                );
+                awardAccountingLines.add(cide);
+            }
+        }
+        return awardAccountingLines;
+    }
+
+
+    protected CustomerInvoiceDetail createSourceAccountingLineBackport(
```
jhulslander, line 149: why did you need to rename createSourceAccountingLine to createSourceAccountingLineBackport?
I think you can use the same function name as in basecode, and include an override annotation

jhulslander (commented):
I think this looks reasonable, but I have a question about renaming a function.  Also, the commit message should also  mention this is a partial back port

src/main/java/edu/cornell/kfs/module/ar/document/service/impl/CuContractsGrantsInvoiceDocumentServiceImpl.java
jhulslander: I think this aspect is OK, I am fine with this

jhulslander (changes requested):
I think the KualiCo PR includes changes. to the function createSourceAccountingLinesByContractControlAccount.  Should the be included here?  Also, please adjust the PR commit message to note this is a partial back port

ajd299:
@jhulslander Please see the latest commit that updates the already extended createSourceAccountingLinesByContractControlAccount to use the Backport method

jhulslander:
This looks reasonable, I can merge once this unit tests pass

jhulslander (approved):
the unit tests passing, I shall merge

PR #1754, opened 2025-05-22
KFSPTS-35018 Add CU-specific version of FINP-10238 PREQ fix

cah292:
This PR adds a CU-specific implementation of the proposed FINP-10238 PREQ enhancement, which updates the PREQ initialization screen to auto-populate the Invoice Received Date from the Invoice Date if the former is blank. Note that KualiCo has not implemented FINP-10238 yet, which is why we have to create our own implementation of it.

In addition to the "onblur" handling proposed by FINP-10238, this CU-specific implementation adds such handling to the "onchange" event as well, so that the field will update properly when the user enters a date via the date-picker widget instead of direct input. Furthermore, our version will perform a simple regex check on the Invoice Date before using it to populate the Invoice Received Date, to help reduce the risk of auto-populating malformed dates.

Note that the "webScriptFiles" bean property had to be configured as a full list override instead of a merge (though the sub-list part does use merging). This is because the parent bean specifies the "commonWebScriptFiles" list for this property, and Spring isn't allowing direct merges into that list, possibly because it's a factory bean. (We needed a similar workaround in our script file overrides for the Requisition Document.)


db483 (approved):
Changes look good.

PR #1756, opened 2025-05-23
KFSPTS-35018 Add CU-specific version of FINP-10238 PREQ fix

cah292:
This is almost the same as the #1754 code changes, except that the auto-completion function's regex has been modified to allow for both 2-digit and 4-digit years.

### DIFF
--- a/src/main/webapp/scripts/module/purap/preqInitUtils.js
+++ b/src/main/webapp/scripts/module/purap/preqInitUtils.js
@@ -1,10 +1,10 @@
 function populateInvoiceReceivedDateIfNecessary(invoiceDateField) {
-    const datePattern = /^\d{1,2}\D\d{1,2}\D\d{2}(\d{2})?$/;
-    if (invoiceDateField && invoiceDateField.value && datePattern.test(invoiceDateField.value)) {
+    const invoiceDate = invoiceDateField && invoiceDateField.value;
+    if (invoiceDate && invoiceDate.trim().length > 0) {
         const invoiceReceivedDateField = document.getElementById("document.invoiceReceivedDate");
         const invoiceReceivedDate = invoiceReceivedDateField && invoiceReceivedDateField.value;
         if (invoiceReceivedDateField && (!invoiceReceivedDate || invoiceReceivedDate.trim().length == 0)) {
-            invoiceReceivedDateField.value = invoiceDateField.value;
+            invoiceReceivedDateField.value = invoiceDate;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/webapp/scripts/module/purap/preqInitUtils.js
```diff
@@ -0,0 +1,10 @@
+function populateInvoiceReceivedDateIfNecessary(invoiceDateField) {
+    const datePattern = /^\d{1,2}\/\d{1,2}\/\d{2}(\d{2})?$/;
```
db483, line 2: I am wondering if it might be better to skip the regular expression test and always set the invoiceReceived date to whatever the value is in invoiceDate? 
In KFS-TEST I noticed that the Invoice Date filed will also accept a 2-2-23 or 2-2-2023 value. This will then get formatted to dd/mm/yyyy format on the created PREQ.

cah292: I had put the regex in place to reduce the likelihood of accidentally copying malformed Invoice Dates into the Invoice Received Date field. Nevertheless, I agree that this pattern should be more flexible.

I could update the regex to allow any non-digit character as the separator (instead of a slash or other specific character), which I think would address the issue. Please let me know if you think that will be enough or if you would prefer the regex logic to be removed altogether.

db483: I like your suggested approach but my concern is that there is no feedback for the user if an invalid value is entered, so it can be confusing as to why sometimes the value is not copied. If an invalid value is entered in invoice date and copied to invoice received date that will be caught on "continue" and a detailed error message displayed. Would it make sense to check with the functional user if any checks should be done on the invoice date value?

cah292: I've posted a question to Eric Banfield on the related user story.

cah292: While we're waiting for Eric's response, I went ahead and committed the regex changes that we would need if we decided to keep the regex check in place. If Eric decides against the regex check, then I'll perform a follow-up commit to remove the regex part.

cah292: As per Eric's response, he's fine with the Invoice Date always being copied to the Invoice Received Date regardless of formatting, so I'll remove this regex-based logic accordingly.

cah292:
I've removed the regex-based check for the copying logic, as per the request from the functionals on the main user story. This is ready to be reviewed again.

PR #1760, opened 2025-06-06
KFSPTS-34586 CapAsset API Add Campus Field

ajd299:


src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -184,18 +187,44 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.ASSET_TAG_PAR
...
+    @GET
+    @Path("/campusCodes")
+    public Response getCampusCodes(@Context HttpHeaders headers) {
+        try {
+            List<Map<String, String>> campusCodes = new ArrayList<>();
+            
+            Map<String, Object> fieldValues = new HashMap<>();
+            fieldValues.put(CuCamsConstants.CapAssetApi.ACTIVE, true);
+            Collection<org.kuali.kfs.sys.businessobject.Campus> campuses = 
+                SpringContext.getBean(BusinessObjectService.class).findMatching(
```
cah292, line 200: Instead of embedding the `SpringContext.getBean()` call here, you may want to separate it out into a separate method, like what this class does with its other service references.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -184,18 +187,44 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.ASSET_TAG_PAR
...
+                    org.kuali.kfs.sys.businessobject.Campus.class, fieldValues);
```
cah292, line 201: Is there a particular reason why the fully qualified `Campus` classname is being used throughout this file? Can an import be used instead, or is there already a conflicting classname in use?

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -184,18 +187,44 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.ASSET_TAG_PAR
...
+            
+            for (org.kuali.kfs.sys.businessobject.Campus campus : campuses) {
+                Map<String, String> campusMap = new HashMap<>();
+                campusMap.put("value", campus.getCode());
+                campusMap.put("label", campus.getCode());
```
cah292, line 206: You may want to check with the functionals on whether the label should include both the campus code and the campus name, instead of containing merely the campus code.

cah292 (changes requested):
I noticed a few areas that may need cleanup, but overall the changes look good.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
ajd299: I have an open question to the functionals on this one

ajd299:
@cah292 Please see the need code and comments on this and related user story

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
```diff
@@ -184,18 +189,45 @@ public Response updateAsset(@PathParam(CuCamsConstants.CapAssetApi.ASSET_TAG_PAR
...
+            Map<String, Object> fieldValues = new HashMap<>();
+            fieldValues.put(CuCamsConstants.CapAssetApi.ACTIVE, true);
+            Collection<Campus> campuses = getBusinessObjectService().findMatching(Campus.class, fieldValues);
+            
+            for (Campus campus : campuses) {
+                Map<String, String> campusMap = new HashMap<>();
+                campusMap.put("value", campus.getCode());
+
+                String campusLabel = String.format("%s - %s", campus.getCode(), campus.getName());
+                campusMap.put("label", campusLabel);
```
cah292, line 208: You may want to use constants for both the map keys and the format string.

cah292 (commented):
I had a minor improvement suggestion, but otherwise the changes look good. If you think the improvement is not needed and that the code should be left as-is, please let me know so that I can merge it.

src/main/java/edu/cornell/kfs/module/cam/rest/resource/CuCapAssetInventoryApiResource.java
ajd299: Please see my latest changes

PR #1764, opened 2025-06-11
KFSPTS-34289 Add MACHINE Person affiliation type

cah292:

The only change on the code side was adding a new affiliation code constant, for consistency with the other affiliations. Nothing is actively using the new constant at this time.


jhulslander (approved):
My concerns have been discussed, we can move forward with this

PR #1769, opened 2025-06-17
KFSPTS-32702 Fix eInvoice unbounded query issue

cah292:
There are PRs for the related user story in cu-kfs, einvoice-api-server and einvoice-website. Please make sure all of them are good to go before merging.

The primary changes are on the eInvoice application/website side; however, I did discover an issue where if there are no invitation records in the Postgres database, then opening the Invitations screen/tab could cause the related KFS API call to search for all KFS Vendors (thus potentially causing the API call to time out or encounter some other error). This specific PR fixes the bug by making the API call return an empty array if no vendor numbers were passed in.

If you think this fix should be deferred to another user story or should be implemented on the eInvoice application side instead, please let me know.


ajd299 (commented):
Good to go along with the other changes. Note, I was able to locally hit this code.

PR #1770, opened 2025-06-24
KFSPTS-34457 Fix Asset Location Global on-save issue

cah292:
Asset Location Global documents, like other global maintenance documents, persist the maintenance BO (and child objects) to the corresponding database tables upon saving, instead of merely storing the BO in the XML content prior to finalization. However, the Asset Location Global Detail child objects have database constraints on the Building and Room references, which can interfere with saving docs containing invalid Building/Room data because save-only maint doc actions aren't aborted for regular business rule failures. This PR fixes the problem by using KualiCo's existing `isPersistable()` Global BO method to include the appropriate validation, thus forcing the save operation to abort if the current data would result in a DB-level constraint violation.

Note that it was more convenient to overlay the `AssetLocationGlobal` BO instead of creating a custom subclass, since the latter option might require adding XML conversion rules and other overrides that may ultimately make the subclass strategy more complex than intended.

### DIFF
--- a/src/main/java/org/kuali/kfs/module/cam/businessobject/AssetLocationGlobal.java
+++ b/src/main/java/org/kuali/kfs/module/cam/businessobject/AssetLocationGlobal.java
@@ -1,0 +1,252 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
+ *
+ * Copyright 2005-2023 Kuali, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package org.kuali.kfs.module.cam.businessobject;
+
+import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/kuali/kfs/module/cam/businessobject/AssetLocationGlobal.java
```diff
@@ -0,0 +1,252 @@
...
+                    location.setAssetLocationStateCode(null);
+                    location.setAssetLocationCountryCode(null);
+                    location.setAssetLocationZipCode(null);
+                }
+            }
+        }
+    }
+
+    /*
+     * CU Customization: Modified isPersistable() implementation to prevent the document from being saved
```
jhulslander, line 184: should we include the JIra in this comment?

src/main/java/org/kuali/kfs/module/cam/businessobject/AssetLocationGlobal.java
```diff
@@ -0,0 +1,252 @@
...
+                GlobalVariables.getMessageMap().removeFromErrorPath(errorPath);
+                index++;
+            }
+        }
+        return canPersist;
+    }
+
+    private boolean detailHasValidOrIncompleteForeignKeyValues(final AssetLocationGlobalDetail detail) {
+        boolean valid = true;
+        final boolean hasCampusCode = StringUtils.isNotBlank(detail.getCampusCode());
```
jhulslander, line 209: do we need to check  if there is just a campus code and the campus code is wrong?
What if there is a building code, but no campus code?
what is there is a room number but no building code?

jhulslander (changes requested):
I had a couple of questions

src/main/java/org/kuali/kfs/module/cam/businessobject/AssetLocationGlobal.java
cah292: I was only encountering the unexpected exception when one of the database foreign key constraints was not being satisfied. The two problematic constraints in this case were for the Building (using Campus Code and Building Code) and the Room (using Campus Code, Building Code and Room Number); this document's BOs don't appear to have a foreign key constraint for just the Campus. Also, as long as at least one of the fields involved in the foreign key constraint is null/blank, the constraint won't be validated by the database when saving the document.

jhulslander (approved):
My concerns have been addressed, we can move forward with this.

PR #1773, opened 2025-07-02
KFSPTS-35347 backport FINP-11623 update jackson core to 2.18.1

jhulslander:
This PR updates our jackson libraries to 2.18.1.
I added three defined dependencies in our POM file as they were coming in from base code, and wanted to get the 2.18.1 version of those jackson libraries too.
I added the jackson exclususion to avoid pull in the older version from base code.


cah292 (approved):
Changes look good, unit tests pass, and I confirmed locally that the built app has the correct library versions. Merging.

PR #1774, opened 2025-07-02
KFSPTS-22330 Remove Concur Event Notification Job v1

ajd299:


ajd299:
My understanding is that we want to wait until KFSPTS-34583 to remove the Concur Account Validation Web Service (API Resource). 

- Therefore do we need to wait to remove the edu.cornell.kfs.concur.businessobjects.ConcurEventNotification class and it's database table CU_CNCR_EVNT_NTFCTN_T?

- Do we need to keep the ConcurManageRefreshTokenForm? 

- Is it okay if a non-devops person commits to kuali puppet as long as a DevOps person reviews it?

dp65:
@ajd299 asked: Is it okay if a non-devops person commits to kuali puppet as long as a DevOps person reviews it?

Yes. Follow the usual procedures (create a branch with KFSPTS-NNNNN in the name, make the change, and create the PR). If you want to test it prior to creating the PR, we can rebuild an environment using the new kuali-puppet branch, but that will need DevOps involvement.

ajd299:
@cah292 At this time I'm not sure if the test failures are related to my most recent commit. Please review my latest changes based on your feedback

src/main/java/edu/cornell/kfs/concur/ConcurParameterConstants.java
```diff
@@ -67,6 +64,5 @@ public class ConcurParameterConstants {
     public static final String CONCUR_EVENT_V2_PROCESSING_REPORT_DETAIL_ITEM = "concur.event.v2.processing.report.detail.item";
     public static final String CONCUR_EVENT_V2_PROCESSING_REPORT_TITLE = "concur.event.v2.processing.report.title";
     public static final String CONCUR_EVENT_V2_PROCESSING_REPORT_FILE_NAME = "concur.event.v2.processing.report.file.name";
-    public static final String CONCUR_EVENT_V2_PROCESSING_REPORT_EMAIL_SUBJECT = "concur.event.v2.processing.report.email.subject";
```
cah292, line 70: Why was this constant and its related resource property removed? They appear to be related to the Event Notification V2 process, not the V1 process. was it because they seemed to be unused?

cah292 (commented):
I had a quick follow-up question about one of the removed constants, but otherwise the code changes look good. As for the unit/integration test that was failing, it was due to a non-PROD Jaggaer availability problem; it was not a problem with these code changes.

src/main/java/edu/cornell/kfs/concur/ConcurParameterConstants.java
cah292: Thanks for the explanation. Given that the constant and its related property appear to be unused, I'm okay with them being removed in this PR.

cah292 (approved):
Code and SQL changes are ready now, and DevOps has already merged the kuali-puppet PR. Merging.

PR #1782, opened 2025-07-21
KFSPTS-31717 update KIM Feed V2 logging details

jhulslander:
I updated KimFeedEdwDaoJdbc to extend CuSqlQueryPlatformAwareDaoBaseJdbc to make use of the logging functionality.  If you think KimFeedEdwDaoJdbc should continue to extend PlatformAwareDaoBaseJdbc, I could move the logging functionality to a static utility class.  Let me know if you think that would be a better approach.


db483:
The code changes look good. I agree that moving the logging functionality to a static utility class might make more sense.

jhulslander:
I have made some changes, this is ready for review again

db483 (approved):

PR #1783, opened 2025-07-24
KFSPTS-35567 Fix opening new lookups from Struts pages

cah292:
KualiCo's 12/20/2023 patch makes several changes to how Struts pages navigate to/from converted lookup screens. However, KualiCo configured those changes to assume that the KFS application URL ends with "/fin". This PR overlays the relevant file so that the appropriate section of the URL is being derived from the "app.context.name" property instead, which is set to "kfs" in the cu-kfs institutional config property file by default.

I was initially encountering some blank-screen issues after fixing the problem above, but I have been unable to replicate them today. It might have been just a temporary problem triggered by one of my earlier versions of this fix.

### DIFF
--- a/src/main/webapp/WEB-INF/tags/kr/modernLookupSupport.tag
+++ b/src/main/webapp/WEB-INF/tags/kr/modernLookupSupport.tag
@@ -1,0 +1,335 @@
+<%--
+
+    The Kuali Financial System, a comprehensive financial management system for higher education.
+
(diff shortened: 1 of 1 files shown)


jhulslander (approved):

dp65:
That sounds like a KualiCo bug - using a hard-coded value rather than the appropriate config variable. If so, is it being reported back to them?

PR #1784, opened 2025-07-28
KFSPTS-35513 upgrade to the 02/28/2024 version of financials

db483:


pom.xml
```diff
@@ -46,9 +46,9 @@
         <jotm-core.version>2.1.10-kuali-1</jotm-core.version>
         <jsp-api.version>2.3.3</jsp-api.version>
         <jsr311-api.version>1.1.1</jsr311-api.version>
-        <junit.version>5.10.1</junit.version>
+        <junit.version>5.10.2</junit.version>
         <ojdbc.version>19.18.0.0</ojdbc.version>
-        <opencsv.version>5.8</opencsv.version>
+        <opencsv.version>5.9</opencsv.version>
         <plexus-utils.version>3.0.24</plexus-utils.version>
         <mockito.version>4.9.0</mockito.version>
```
cah292, line 53: Since we removed the PowerMock dependency from the POM a few weeks ago, can we now upgrade Mockito to the same version that KualiCo uses?

src/main/java/org/kuali/kfs/vnd/businessobject/VendorDetail.java
```diff
@@ -553,8 +558,7 @@ public String getVendorCommoditiesAsString() {
             } else {
                 first = false;
             }
-            // CU customization KFSPTS-3487
-            sb.append(vcc.getCommodityCode().getCommodityDescription());
+            sb.append(vcc.getCommodityCode().getPurchasingCommodityCode());
```
cah292, line 561: Why was this customization removed? Should it be added back in?

db483: Good catch, updating

cah292 (changes requested):
I still have a little bit more to review on this, but I did find some additional items that may need cleanup. Also, I'll be adding a separate comment regarding some upgrade issues that are not being affected yet by this PR.

cah292 (changes requested):
In addition to the items from my prior comments, I found a few more things that may need fixing:

* `CuScrubberProcessImpl` is another file where some `new Integer()` calls should be replaced with `Integer.valueOf()` instead.
* The new/converted Vendor lookup might not work properly when using an Oracle database. The related queries in `VendorDaoOjb` are using plain SQL to gather the appropriate data, but the queries seem to be using MySQL-specific features that would need to be rewritten for Oracle compatibility. (For example, any `CONCAT()` functions with 3 or more arguments need to be rewritten to use either the `||` operator or nested 2-arg `CONCAT()` calls. Also, `GROUP_CONCAT()` appears to be a MySQL-specific function; I think `LISTAGG()` is the Oracle equivalent.) We may have to override or overlay the appropriate areas if we want to get the new/converted Vendor lookup to work as expected.
* At some point, we may want to revise our `CUVendorService` interface to only contain CU-specific methods and to extend from `VendorService` to inherit the rest. (Such a setup made it a bit harder to verify that our KFSUPGRADE-911 customization wasn't going to interfere with the FINP-10641 DV batch upload changes.) If such consolidation is beyond the scope of this PR, then we should create a separate user story to do so.
  * If we're not simplifying `CUVendorService` yet, then it might still be a good idea to add this upgrade's new `VendorService.getVendorAddresses()` method to it).

src/test/java/edu/cornell/kfs/fp/document/CuDisbursementVoucherDocumentTest.java
```diff
@@ -407,6 +407,11 @@ public boolean isVendorContractExpired(Document document, Integer integer, Vendo
         public VendorAddress getVendorDefaultAddress(Integer integer, Integer integer1, String s, String s1, boolean b) {
             return null;
         }
+        
+        public Collection<VendorAddress> getVendorAddresses(int vendorHeaderId, int vendorDetailId, String addressType){
```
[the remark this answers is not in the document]
db483: Adding

src/main/java/edu/cornell/kfs/module/ld/batch/service/impl/CuFileEnterpriseFeederHelperServiceImpl.java
```diff
@@ -47,32 +47,22 @@ public class CuFileEnterpriseFeederHelperServiceImpl extends FileEnterpriseFeede
...
-            throw new RuntimeException(e);
-        }
-        catch (final RuntimeException e) {
-            LOG.error("Error occured trying to parse the recon file.", e);
-            errorMessages.add(new Message("Error occured trying to parse the recon file.", Message.TYPE_FATAL));
-            reconciliationBlock = null;
-            statusAndErrors.setStatus(new FileReconBadLoadAbortedStatus());
-            throw e;
-        }
+        final ReconciliationBlock reconciliationBlock =
```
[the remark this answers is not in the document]
db483: Good question. Looking

src/main/resources/edu/cornell/kfs/vnd/businessobject/datadictionary/VendorDetail.xml
```diff
@@ -383,6 +383,18 @@
...
                </bean>
+               <bean parent="FieldOverrideForListElementInsert">
+                   <property name="propertyName" value="formAttributeDefinitions" />
+                   <property name="propertyNameForElementCompare" value="name" />
+                   <property name="element">
+                       <ref bean="VendorDetail-vendorHeader.vendorDebarredIndicator"/>
+                   </property>
+                   <property name="insertAfter">
+                       <list>
+                           <ref bean="VendorDetail-defaultPaymentMethodCode"/>
```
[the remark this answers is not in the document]
db483: updated

db483:
Still need work:
- Given the DateTimeServiceImpl simplifications in this upgrade, we might be able to remove this TestDateTimeServiceImpl class altogether. If you think doing so now is too much scope to be bundled with the upgrade, then I would suggest we create a follow-up user story to handle the removal.

- At some point, we may want to revise our CUVendorService interface to only contain CU-specific methods and to extend from VendorService to inherit the rest. (Such a setup made it a bit harder to verify that our KFSUPGRADE-911 customization wasn't going to interfere with the FINP-10641 DV batch upload changes.) If such consolidation is beyond the scope of this PR, then we should create a separate user story to do so.

- The new Vendor querying in VendorDaoOjb has at least two spots where the code is closing the Connection returned by the OJB Connection Manager. However, based on other areas of modified code in this upgrade (such as in ActionListDAOOjbImpl), such Connection objects are not supposed to be explicitly closed. It looks like KualiCo fixed that in FINP-10792 for the 2024-03-20 financials patch, so we might need at least a partial backport of that fix.

cah292 (commented):
While looking over the new `VendorDaoOjb` overlay, I noticed one other area that will need a backported fix:

KualiCo's Vendor pagination logic contains a bug that sets the first row's offset to the product of the row-skip count and the per-page size. It looks like KualiCo fixed this bug with FINP-11585 from the 2024-11-20 financials patch, so we may need to backport and adapt that fix as well.

db483:
Good find, adding changes from FINP-11585.

db483:
> While looking over the new `VendorDaoOjb` overlay, I noticed one other area that will need a backported fix:
> 
> KualiCo's Vendor pagination logic contains a bug that sets the first row's offset to the product of the row-skip count and the per-page size. It looks like KualiCo fixed this bug with FINP-11585 from the 2024-11-20 financials patch, so we may need to backport and adapt that fix as well.

Good find, I added the changes from FINP-11585.

src/test/java/edu/cornell/kfs/sys/service/impl/TestDateTimeServiceImpl.java
```diff
@@ -1,27 +1,11 @@
...
-import org.kuali.kfs.core.api.CoreConstants;
 import org.kuali.kfs.core.impl.datetime.DateTimeServiceImpl;
 
-import edu.cornell.kfs.sys.CUKFSConstants;
-
 /**
  * Micro-testable DateTimeService implementation that will only use
  * the default datetime formats.
  */
 public class TestDateTimeServiceImpl extends DateTimeServiceImpl {
```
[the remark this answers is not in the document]
db483: Removed

src/test/java/edu/cornell/kfs/sys/batch/service/impl/CuBatchInputFileServiceImplTest.java
```diff
@@ -166,7 +166,6 @@ private final void checkForDoneFile(File batchFile, boolean isExist) {
     private class MockDateTimeService extends DateTimeServiceImpl {
```
cah292, line 166: Since you're removing the `TestDateTimeServiceImpl` class, it would be good to remove this nested `MockDateTimeService` class as well.

cah292 (commented):
I found one unit-test-related change that would be good to put in place, but otherwise the upgrade's code changes look good.

src/test/java/edu/cornell/kfs/sys/batch/service/impl/CuBatchInputFileServiceImplTest.java
db483: Good point

db483: Updated

PR #1785, opened 2025-07-28
KFSPTS-35584 Fix Vendor maintenance XML conversion

cah292:
One of our previous financials patch upgrades resulted in the removal of a custom extended attribute for Vendor Supplier Diversity, thus requiring some changes to our maintenance XML conversion rules so that older Vendor documents could still be opened. However, that particular set of conversion rule changes was accidentally misconfigured, resulting in errors when users attempt to open or act upon older Vendor documents containing Supplier Diversity data. This PR fixes the issue by correcting the problematic conversion rules; it also adds a new unit test case to verify their correctness.


db483 (approved):
Changes look good.

PR #1788, opened 2025-07-30
KFSPTS-35643 fix spring configuration for disencumbranceEnterpriseFeederFileSetType to have enterpriseFeederFileSetType as parent

db483:


cah292 (commented):
I had a question about which bean should be specified as the parent, but otherwise the changes look good.

cah292 (approved):

PR #1789, opened 2025-07-31
KFSPTS-35626 Omit tenant message from email subjects

cah292:
The 12/20/2023 financials upgrade changed email messages so that the tenant name (or the "No tenant specified" message when tenant name isn't configured) is always included in the email subject line. This PR modifies the email service so that if the environment hasn't configured a tenant name, the "No tenant specified" message will be omitted from the email subject line.


db483 (approved):
Changes look good.

PR #1790, opened 2025-07-31
KFSPTS-35666 Backport FINP-12199 lookup fix

cah292:
The functionals discovered at least one error condition that can occur when maintenance documents interact with converted lookup screens. This PR backports KualiCo's FINP-12199 fix to address the issue, since those changes correct the "formKey" data handling that was triggering the problem.


jhulslander (commented):
changes look good, this can be merged when the unit tests complete

jhulslander (approved):
unit tests pass, this can be merged

PR #1797, opened 2025-08-14
KFSPTS-35826 Remove unneeded Award refresh customization

cah292:
We have a particular Award document customization that doesn't seem to be doing much, other than to suppress certain object refreshes from occurring. Not only have I not found any reason why those refreshes should be suppressed, it also appears that the suppressed refresh of the Award's Proposal reference is causing Award documents to route to EXCEPTION in the 02/28/2024 financials upgrade. (KualiCo added a certain Proposal field to the Award for FINP-10372, so that might have prevented the Award's Proposal reference from being auto-refreshed by other parts of the code.)

This PR fixes the problem by removing our customized `refreshAward()` method, thus causing our custom Award Maintainable to use the superclass's implementation of that method instead. If you think that a less impacting solution should be used instead (such as keeping the custom `refreshAward()` method but adding a statement to explicitly refresh the Proposal reference), please let me know.

### DIFF
--- a/src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
+++ b/src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
@@ -2,40 +2,13 @@ package edu.cornell.kfs.module.cg.document;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.kuali.kfs.krad.util.ObjectUtils;
-import org.kuali.kfs.module.cg.businessobject.Award;
 import org.kuali.kfs.module.cg.businessobject.AwardAccount;
-import org.kuali.kfs.module.cg.businessobject.Proposal;
 import org.kuali.kfs.module.cg.document.AwardMaintainableImpl;
-import org.kuali.kfs.sys.KFSPropertyConstants;
 
 import edu.cornell.kfs.module.cg.businessobject.AwardAccountExtendedAttribute;
 
 public class CuAwardMaintainableImpl extends AwardMaintainableImpl {
 
-    /**
-     * Load related objects from the database as needed.
-     *
-     * @param refreshFromLookup
-     */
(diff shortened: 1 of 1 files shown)


src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
```diff
@@ -2,40 +2,13 @@
...
 
-    /**
-     * Load related objects from the database as needed.
-     *
-     * @param refreshFromLookup
-     */
-    @Override
-    protected void refreshAward(final boolean refreshFromLookup) {       
-        final Award award = getAward();
-        final Proposal tempProposal = getAward().getProposal();
```
jhulslander, line 23: it looks like these two proposal steps are the only things not in the base code version of this function.  Should we do these two steps, then call the the super. refreshAward()?
I am not sure what these lines would accomplish, so I am OK moving forward with the total removal.  However, I wanted to ask and think about it.

jhulslander (commented):
I had a question on one part of this PR

src/main/java/edu/cornell/kfs/module/cg/document/CuAwardMaintainableImpl.java
cah292: Our old code was retrieving the Award, accessing its Proposal reference, and reassigning the Award's Proposal reference with that same value. The getter and setter for that Proposal reference don't contain extra processing, so our custom handling was effectively leaving the Proposal reference unchanged, rather than refreshing the reference like in base code. I'm not certain why it was set up this way, unless it was modeled after a similar customization elsewhere in our code (but didn't get adjusted properly when it was added to this class).

Given the above, I don't think we need to keep our override of this method.

jhulslander (approved):
Thanks for the explanation.  I am fine moving forward as is.  I shall merge

PR #1798, opened 2025-08-15
KFSPTS-35842 Fix radio button fields on converted lookups

cah292:
We have a customization for adjusting new/converted lookup drop-downs so that a blank key-value entry will be auto-added if necessary. However, the functionals recently discovered that the custom code is also adding such entries to radio button fields as well, resulting in unintended blank radio button options. This PR fixes the problem by adjusting our customization to only affect drop-down fields.


jhulslander (commented):
Changes look good, I can merge when the unit tests pass

jhulslander (approved):
this can be merged now

PR #1802, opened 2025-08-22
KFSPTS-35484: Adjust routing of all fiscal year end documents with parent FPYE to add AccountingOrganizationHierarchy and ObjectCode to the exisiting routings.

nkimble:


jhulslander (changes requested):
When I exported document type FPYE from prod the opening tag was
```<?xml version="1.0" encoding="UTF-8" standalone="yes"?>```
I think this file should have the opening tag updated.
Should the routingVersion be incremented here, since this has a new route path?

jhulslander (approved):
This can be merged now

PR #1803, opened 2025-08-28
KFSPTS-35912 Add support for non person contract manager email

db483:


jhulslander (approved):
code changes look good

PR #1804, opened 2025-08-28
KFSPTS-35695 Upgrade to the 04/24/2025 version of financials

cah292:
There are PRs for this upgrade in cu-kfs, cu-kfs-upgrades and nonprod-sql. Please make sure all three are good to go before merging.


db483 (changes requested):
The changes look good. I only found one minor thing that I am ok if we don't update.

cah292:

db483 (approved):
Changes look good.

PR #1805, opened 2025-09-04
KFSPTS-36010 Fix PMW Vendor return-from-lookup behavior

cah292:
One of the recent KualiCo patches changed the refactored multi-value lookups so that fields with inquiry/details links are the only ones whose values are returned to the caller. Such behavior was interfering with the PaymentWorks Vendor Global document, because the PaymentWorks Vendor ID field (the primary key) was not having an inquiry link generated for it, due to a Data Dictionary setting that we might not want to change.

To work around the issues above, this PR modifies the PaymentWorks Vendor lookup to show inquiry links in the ID field. This also required creating a custom Inquirable implementation to force an inquiry link for the ID field.


jhulslander (approved):
these changes look reasonable, unit tests work, I shall merge

PR #1808, opened 2025-09-15
KFSPTS-36120 fix masking on bank account number

db483:


src/main/webapp/WEB-INF/tags/sys/innerWireTransfer.tag
```diff
@@ -138,7 +139,8 @@
                 }"/>
                 <c:set var="mask"
                        value="${
-                    accountNumberReadOnly
+                    (maskNonProd and not KualiForm.document.documentHeader.workflowDocument.initiated)
+                    or accountNumberReadOnly
```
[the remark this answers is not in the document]
db483: Good question. After looking into this some more yesterday I decided when the document is in saved status the field should not be visible because that would allow access to existing documents in saved status. I also tested on a DV in intdev and the behaviour is the same, after the DV is saved and reloads the account number is masked.

src/main/webapp/WEB-INF/tags/sys/innerWireTransfer.tag
```diff
@@ -26,6 +26,7 @@
 <c:set var="wireTransAttributes" value="${DataDictionary.PaymentSourceWireTransfer.attributes}" />
 <%-- Cornell Customization --%>
 <c:set var="wireTransExtendedAttributes" value="${DataDictionary.PaymentSourceWireTransferExtendedAttribute.attributes}" />
+<c:set var="maskNonProd" value="${kfunc:isNonProductionEnvAndUnmaskingTurnedOff()? 'true' : 'false'}" />
```
cah292, line 29: You may want to add a space before the question mark.

Also, I would suggest adding appropriate customization comments to the areas you modified, especially the part that changes the "mask" derivation logic.

cah292 (changes requested):
There are one or two areas that could use some minor revisions, but otherwise the changes look good.

src/main/webapp/WEB-INF/tags/sys/innerWireTransfer.tag
db483: Thank you Chad, adding changes

nkimble:
@db483,  Changes look good and I agree with your approach based upon your explanations. I have also exercised these changes from a technical perspective in a non-prod environment and did not encounter any issues. 
I am just waiting on ChadH's requested adjustments before merging.

db483:
Good question Chad. We checked with Cathy and we don't use wire on CM. I will look into PREQ creation.

nkimble:
@cah292 Based on Daniela's findings, are you okay with moving forward with these changes to DEV?

nkimble (approved):

cah292 (approved):
I approve this as well; my concerns have been addressed.

PR #1809, opened 2025-09-16
KFSPTS-36151 fix SQL typo in VendorDaoImpl

jhulslander:


cah292 (changes requested):
The issue appears to be stemming from a typo accidentally introduced by our team (specifically, me). I had some suggestions for improving the documentation regarding the fix, but otherwise the changes look good.

PR #1852, opened 2025-12-09
KFSPTS-36784 Error Correction on BA & DI Customization Blanket Approval Parameter Fix

ajd299:


nkimble (approved):
These changes look good. They properly align the system parameter lookup calls in our customizations to the base code changes introduced by FINP-8704.

PR #1861, opened 2026-01-09
KFSPTS-36620 AI Payflow Phase 2 Create PREQ Document

ajd299:


src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -457,4 +470,139 @@ protected boolean isEligibleForAutoApproval(PaymentRequestDocument document, Kua
...
+        preqDoc.setSpecialHandlingInstructionLine3Text(preqDto.getSpecialHandlingLine3());
+
+        ShippingTitle shippingTitle = new ShippingTitle();
+        shippingTitle.setVendorShippingTitleCode(preqDto.getShippingPrice().toString());
+        shippingTitle.setVendorShippingTitleDescription(preqDto.getShippingDescription());
+
+        if (CollectionUtils.isNotEmpty(preqDto.getNotes())) {
+            for (PaymentRequestNoteDto noteDto : preqDto.getNotes()) {
+                LOG.info(preqDoc.getNoteType().getCode());
+                documentService.createNoteFromDocument(preqDoc, noteDto.getNoteText());
```
jhulslander: There is a note type preq documents, we want to set that

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -457,4 +470,139 @@ protected boolean isEligibleForAutoApproval(PaymentRequestDocument document, Kua
...
+        try {
+
+            PurchaseOrderDocument poDoc = purchaseOrderService.getCurrentPurchaseOrder(preqDto.getPoNumber());
+            if (ObjectUtils.isNull(poDoc)) {
+                throw new Exception("Purchase Order Document not found for PO #" + preqDto.getPoNumber());
+            }
+
+            CuPaymentRequestDocument preqDoc = generateNewPaymentRequestDocument(poDoc, preqDto);
+            documentService.saveDocument(preqDoc);
+            Document routedPreqDoc = documentService.routeDocument(preqDoc, CUPurapConstants.Payflow.PAYFLOW_DOCUMENT_ANNOTATION, null);
```
jhulslander: for this part of the work, we only want to save the preqs

src/main/java/edu/cornell/kfs/module/purap/rest/resource/PaymentRequestResource.java
```diff
@@ -67,17 +76,32 @@ public Response createPaymentRequestDocument(InputStream requestBodyStream) {
...
+
+                    LOG.info("createPaymentRequestDocument, PREQ Document #{} Created", preqDoc.getDocumentNumber());
+
+                    HashMap<String, Object> responseBody = buildResponseBody(preqDoc);
+
+                    return Response.ok(gson.toJson(results)).entity(responseBody).build();
+
+                } catch (Exception e) {
+                    LOG.error("createPaymentRequestDocument, Unexpected error occurred while creating PREQ Document", e);
+                    return Response.status(500).build();
```
jhulslander: We need to capture validation errors and give them to the caller, if present.
We do want to give a 500 error if there is some unknown problem though, and we can use the constants, like Status.INTERNAL_SERVER_ERROR

src/main/java/edu/cornell/kfs/module/purap/CUPurapConstants.java
```diff
@@ -576,5 +576,9 @@ public static final class IWantRoles {
     public static final String PURCHASE_ORDER_NUMBER_FORMAT_ERROR_RESPONSE_MESSAGE = "The " + PURCHASE_ORDER_NUMBER_URL_PARAMETER_NAME + " was not formatted correctly";
     public static final String PURCHASE_ORDER_NOT_FOUND_MESSAGE = "Purchase Order not found";
     public static final String VENDOR_NOT_FOUND_MESSAGE = "Vendor detail not found";
-    
+
+    public static final class Payflow {
+        public static final String PAYFLOW = "Payflow";
+        public static final String PAYFLOW_DOCUMENT_ANNOTATION = "Document Automatically Generated by Payflow"; // suggestion for better initial annotation?
```
jhulslander: I think we can defer the annotation to part 3 when we do the routing the document.
I think we want reference our restful endpoint / resource rather than the calling application though

jhulslander (changes requested):
Take a at the code I referenced, I think if model that approach, it should be able simplify some of this.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
ajd299: Okay, looking at the createNoteFromDocument method the noteType is set from the preqDoc.noteType.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -457,4 +472,131 @@ protected boolean isEligibleForAutoApproval(PaymentRequestDocument document, Kua
...
+
+        if (CollectionUtils.isNotEmpty(preqDto.getNotes())) {
+            for (PaymentRequestNoteDto noteDto : preqDto.getNotes()) {
+                documentService.createNoteFromDocument(preqDoc, noteDto.getNoteText());
+            }
+        }
+
+        for (PaymentRequestLineItemDto dtoItem : preqDto.getItems()) {
+            PaymentRequestItem paymentRequestItem = findPaymentRequestItem(dtoItem, preqDoc);
+            paymentRequestItem.setItemQuantity(dtoItem.getItemQuantity());
```
jhulslander: we should set the item quantity and item price from the DTO

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -457,4 +472,131 @@ protected boolean isEligibleForAutoApproval(PaymentRequestDocument document, Kua
...
+        preqDoc.setProcessingCampusCode(poDoc.getDeliveryCampusCode());
+
+        preqDoc.setVendorInvoiceAmount(preqDto.getInvoiceAmount());
+        preqDoc.setSpecialHandlingInstructionLine1Text(preqDto.getSpecialHandlingLine1());
+        preqDoc.setSpecialHandlingInstructionLine2Text(preqDto.getSpecialHandlingLine2());
+        preqDoc.setSpecialHandlingInstructionLine3Text(preqDto.getSpecialHandlingLine3());
+
+        if (CollectionUtils.isNotEmpty(preqDto.getNotes())) {
+            for (PaymentRequestNoteDto noteDto : preqDto.getNotes()) {
+                documentService.createNoteFromDocument(preqDoc, noteDto.getNoteText());
```
jhulslander: The notes aren't being saved.  I think we need to take into account the note type too.

jhulslander (changes requested):
This is looking better.  When I tested this it doesn't look like the notes nor items are working right though

jhulslander (approved):
I think this is reasonable for part 2.
We can figure out how to catch and return business errors when we try and route the document

PR #1867, opened 2026-01-28
KFSPTS-37234 fix Permission and Responsibility lookup

db483:


cah292 (changes requested):
There's one spot that needs some minor corrections, but otherwise the changes look good.

cah292 (approved):

PR #1869, opened 2026-01-29
KFSPTS-37234 fix Permission and Responsibility lookup

db483:


jhulslander (approved):
this seams reasonable

PR #1870, opened 2026-02-03
KFSPTS-36246 upgrade to the 11/20/2024 version of financials

db483:


db483:
Due to CEMI work we will not pursue this upgrade.

PR #1892, opened 2026-03-10
KFSPTS-37456 Restore Funding Source Code Field on PO

ajd299:


jhulslander (changes requested):
changes look good for the most part, I had a few questions/suggestions.  Should we confirm with the functionals that this value should NOT be able to controlled by functionals through the UI?

src/main/java/edu/cornell/kfs/module/purap/CUPurapConstants.java
```diff
@@ -215,6 +215,10 @@ public static final class POTransmissionMethods {
         public static final String MANUAL = "MANL";
         public static final String CONVERSION = "CNVS";
     }
+
+    public static final class PurapFundingSources {
+        public static final String FEDERAL_FUNDING_SOURCE = "FEDR";
```
jhulslander: should this be a parameter?  Would we have multiple federally funding source codes?

jhulslander (changes requested):

jhulslander (approved):
changes look good, unit tests pass, i shall merge

PR #1901, opened 2026-03-23
KFSPTS-37657 Fix Supplier Excel file structure

cah292:
This PR makes the following adjustments to the Supplier Extract Job:

* Updates the Excel-writing code to forcibly remove and reinsert the header rows on the modified sheets. This fixes an issue where each sheet's dimension/range info was not taking the pre-existing header rows into account (at least in streaming mode), which might be why we were getting discrepancies on some of the row counts during the merge for Huron.
  * Some adjustments were also made to generate a new spreadsheet from the modified template, without performing an initial copy-and-save of the template.
* Updates the spreadsheet zip/archive settings to disable ZIP64 mode. That mode allows for compressing content larger than 4GB, but Apache POI doesn't seem to be populating the ZIP64 headers properly, at least when in streaming mode. That seems to be why the subsequent merge process was reporting "Bad uncompressed size" errors.
* Adds a new unit test that validates the fixes for both of the above.

NOTE: To assist with validating the conditions above, I created branch `KFSPTS-37657-cah292-test` in `einvoice-api-server` that imports the JS-related spreadsheet library and exposes a simple web service for reading .xlsx files. See the `testxlsx.js` file on that branch for usage instructions.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/batch/service/impl/CemiExcelWriter.java
+++ b/src/main/java/edu/cornell/kfs/sys/batch/service/impl/CemiExcelWriter.java
@@ -2,9 +2,9 @@ package edu.cornell.kfs.sys.batch.service.impl;
 
 import java.io.Closeable;
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Map;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


db483 (changes requested):
Changes looks good. I had a couple of small suggestions for improvement.

db483 (approved):
Changes look good and the concerns raised in code review have been addressed.

PR #1902, opened 2026-03-25
KFSPTS-37577 Merge CEMI Supplier addresses/phones/emails

cah292:
This PR adjusts the CEMI Supplier address/phone/email handling to consolidate duplicate records where possible. The related DTOs have been modified to accept a list of Vendor addresses/phones/emails instead of individual BOs.

In general, the updated code will attempt to preserve the original order of the address/phone/email data.

Also note that there are several Vendors for which the new email-count-limit warning message will get printed to the logs. There are at least 11 such Vendors in the local-env abridged results.

UPDATE: In addition to some miscellaneous standardization and cleanup, I also consolidated the is-primary-address logic so that the new utils-related methods will handle it in all relevant places. Thus, this PR will also handle the needed changes for the KFSPTS-37573 improvement.

### DIFF
--- a/src/main/java/edu/cornell/kfs/sys/util/CemiUtils.java
+++ b/src/main/java/edu/cornell/kfs/sys/util/CemiUtils.java
@@ -75,20 +75,30 @@ public final class CemiUtils {
                 .toArray(String[]::new);
     }
 
-    public static List<String> createListWithElementsAndMinimumSize(final int minSize, final String... elements) {
+    public static List<String> createListPaddedToMinimumSizeIfNecessary(final int minSize, final String... elements) {
+        return createListPaddedToMinimumSizeIfNecessary(KFSConstants.EMPTY_STRING, minSize, elements);
+    }
+
+    @SafeVarargs
+    public static <T> List<T> createListPaddedToMinimumSizeIfNecessary(
+            final T emptyValue, final int minSize, final T... elements) {
         if (elements.length == 0) {
-            return createListOfEmptyStrings(minSize);
+            return createListOfEmptyValues(minSize, emptyValue);
         } else if (elements.length >= minSize) {
             return List.of(elements);
         } else {
-            final String[] minSizeArray = Arrays.copyOf(elements, minSize);
-            Arrays.fill(minSizeArray, elements.length, minSize, KFSConstants.EMPTY_STRING);
+            final T[] minSizeArray = Arrays.copyOf(elements, minSize);
+            Arrays.fill(minSizeArray, elements.length, minSize, emptyValue);
             return List.of(minSizeArray);
         }
     }
 
     public static List<String> createListOfEmptyStrings(final int size) {
-        return Collections.nCopies(size, KFSConstants.EMPTY_STRING);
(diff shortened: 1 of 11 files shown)


src/main/java/edu/cornell/kfs/sys/util/CemiUtils.java
```diff
@@ -47,4 +56,49 @@ public static int getDataColumnCount(final CemiSheetDefinition sheetDefinition)
...
+        return Stream.of(propertyValues)
+                .map(StringUtils::trimToEmpty)
+                .map(propertyValue -> propertyValue.toUpperCase(Locale.US))
+                .collect(Collectors.joining(CUKFSConstants.SEMICOLON));
+    }
+
+    public static <T> String[] getDistinctValuesFromMatchingSubLists(
+            final Map<String, List<String>> subLists, final List<T> dataObjects, final Function<T, String> keyGetter) {
+        return dataObjects.stream()
+                .map(keyGetter::apply)
```
db483, line 69: This is a long pipeline of chained lambdas. It is not immediately obvious what it is trying to accomplish. A developer unfamiliar with the codebase would need to reason through the type at each step, which will make the maintainability of this code trickier. I am wondering if it might be better to break the chain into two steps: generateKeyForGroupingDuplicates first and then getDistinctValuesFromMatchingSubLists?

src/main/java/edu/cornell/kfs/sys/util/CemiUtils.java
```diff
@@ -47,4 +56,49 @@ public static int getDataColumnCount(final CemiSheetDefinition sheetDefinition)
         return sheetDefinition.getFields().size();
     }
 
+    public static String generateKeyForGroupingDuplicates(final String... propertyValues) {
```
nkimble, line 59: I am not sure about others but the term "group" for me has the connotation of security which does not pertain for this functionality. Would "bundle" or some other term make more sense?

src/main/java/edu/cornell/kfs/vnd/batch/dto/CemiSupplierEmailSubEntry.java
```diff
@@ -1,59 +1,69 @@
...
-        
-        List<String> uses = CemiVendorConstants.ADDRESS_TENANTED_USES.get(addressTypeCode);
-        if (uses == null) {
-            return Collections.nCopies(4, CemiVendorConstants.EMPTY_STRING);
-        }
-        
-        List<String> result = new ArrayList<>(uses);
-        while (result.size() < 4) {
-            result.add(CemiVendorConstants.EMPTY_STRING);
+    private static List<String> determineUsesForTenanted(final List<VendorAddress> vendorAddresses,
```
nkimble, line 55: Should email be added to this method name to make the functionality clearer?

nkimble (changes requested):

nkimble (approved):
The code review has been completed.
A few maintainability type items were identified in the PR. Due to those issues not being seen as critical for overall functionality, this PR was merged with those recommendations being deferred. 
Linked user story KFSPTS-37914 was created to track and address those issues when additional coding is performed in this same area.

PR #1904, opened 2026-03-31
KFSPTS-37804 remove plexus-utils dependency

jhulslander:
We received a dendabut alert PR for a plexus update.  This library isn't used by base code.  We can remove it from our overlay.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,6 @@
         <junit.version>5.11.0</junit.version>
         <ojdbc.version>19.18.0.0</ojdbc.version>
         <opencsv.version>5.9</opencsv.version>
-        <plexus-utils.version>3.0.24</plexus-utils.version>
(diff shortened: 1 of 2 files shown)


cah292 (commented):
I agree that the test class's changes look good and that the unit tests pass. However, you may want to check whether `CuBatchFileLookupableHelperServiceImpl` is even being used by cu-kfs anymore. If it's dead code, then it might be better to remove the lookupable and its test class altogether.

cah292:
Actually, upon further analysis, I see that our Labor Ledger Batch File lookup is using the code, so it's active code after all. Sorry for the hassle.

cah292 (approved):
Changes look good and unit tests pass. Also, since the `Files.walk()` stream appears to be a sequential one, I'm okay with the `forEach()` operation being used to update the list.

Merging.

PR #1905, opened 2026-04-01
KFSPTS-37770 Add batch job to fix attachment file locations

cah292:
This PR creates a temporary batch job for performing further follow-up on some account attachment mapping corrections. KFSPTS-34442 remapped several attachments to the correct accounts but left the file contents in the same file system locations; however, it seems that the files need moving as well because of how the directory structure is derived from the object IDs. By using a batch job to perform the corrections based on the remapping data from the KFSPTS-34442 table, the files should get moved to the correct locations.

More testing is needed in a non-Production environment if possible.


cah292:
As per our team discussion, we will attempt to move the files via some DevOps scripts instead. (DevOps will create one script to copy the problematic files into the DEV environment and another script to move the files to the proper locations within KFS; only the latter script will be used when applying the fix to Production.) We will not be pursuing the batch-job-based solution further unless the DevOps scripts end up not being feasible.

PR #1906, opened 2026-04-03
KFSPTS-37760 Send default funding text for Jaggaer

ajd299:


jhulslander (approved):
changes look reasonable,  I shall merge

PR #1907, opened 2026-04-06
KFSPTS-35484 FPYE Document Add Nodes to Route Hierarchy

ajd299:
1) Export current version of FPYE document from production (https://kfs-prod.adminapps.cornell.edu/kfs/webapp/details/DocumentType/287408d4-1c24-4862-8eb8-c1cfb82dce47)

2) Reapply changes


nkimble (commented):
Changes look good. 
The only question I had was, would it make the routing easier to follow later on if we set this up such that the order of the role names in the routeNodes items list matched the order of the new routePath?

nkimble (approved):
Changes look good. Unit tests have passed. Approved to merge to develop.

PR #1913, opened 2026-04-20
KFSPTS-37605: Refactor all CEMI code to new package in Cornell overlay. Adjust data extract file location to be in new package. Add new namespace KFS-CEMI.

nkimble:


src/main/java/edu/cornell/kfs/cemi/sys/util/CemiCuSqlChunk.java
```diff
@@ -0,0 +1,14 @@
+package edu.cornell.kfs.cemi.sys.util;
+
+import edu.cornell.kfs.sys.util.CuSqlChunk;
+
+public final class CemiCuSqlChunk {
+    
+    public static CuSqlChunk asSqlBetweenCondition(final String columnName, final int sqlType,
```
cah292, line 7: I would suggest removing this `CemiCuSqlChunk` class and moving the `asSqlBetweenCondition()` method back to the `CuSqlChunk` class. `CuSqlChunk` is meant to be used by multiple modules, and though the new `asSqlBetweenCondition()` method was originally added for the CEMI work, it's not intended to be CEMI-specific.

src/main/java/edu/cornell/kfs/cemi/sys/CemiKFSConstants.java
```diff
@@ -0,0 +1,11 @@
+package edu.cornell.kfs.cemi.sys;
+
+import edu.cornell.kfs.sys.CUKFSConstants;
+
+public class CemiKFSConstants extends CUKFSConstants {
+    
+    public static final class FileExtensions {
+        public static final String XLSX = ".xlsx";
```
cah292: This constant doesn't have to be CEMI-specific, but if you think it should remain CEMI-specific, then I would suggest moving it to the `CemiBaseConstants` class instead.

src/main/java/edu/cornell/kfs/cemi/sys/CemiKFSKeyConstants.java
```diff
@@ -0,0 +1,9 @@
+package edu.cornell.kfs.cemi.sys;
+
+import edu.cornell.kfs.sys.CUKFSKeyConstants;
+
+public class CemiKFSKeyConstants extends CUKFSKeyConstants {
```
cah292: You may want to rename this to something like `CemiKeyConstants` instead, and there's no need for it to inherit from another constants class.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/CemiVendorDaoOjb.java
```diff
@@ -0,0 +1,11 @@
+package edu.cornell.kfs.cemi.vnd.dataaccess;
+
+import java.util.stream.Stream;
+
+import org.kuali.kfs.vnd.businessobject.VendorDetail;
+
+public interface CemiVendorDaoOjb {
+
+    Stream<VendorDetail> getVendorsForCemiSupplierExtractAsCloseableStream();
```
cah292: The new Vendor DAO setup is a little confusing. Would making the following changes keep things simpler? Thoughts?

1. Rename the `CemiVendorDaoJdbc` interface back to `CemiVendorDao` and add the `getVendorsForCemiSupplierExtractAsCloseableStream()` method to it.
2. Having `CemiVendorDaoOjb` as an interface instead of an impl is confusing. You may want to remove it or change its name to something else (such as `CemiVendorHelperDao`).
3. Inject the OJB variant of the CEMI Vendor DAO into a new a property on the JDBC DAO variant.
4. In the JDBC DAO, have its implementation of `getVendorsForCemiSupplierExtractAsCloseableStream()` invoke the method on the OJB DAO.

src/test/java/edu/cornell/kfs/cemi/vnd/CemiVendorTestConstants.java
```diff
@@ -0,0 +1,9 @@
+package edu.cornell.kfs.cemi.vnd;
+
+public final class CemiVendorTestConstants {
+
+    public static final class VendorSpringBeans {
+        public static final String CEMI_OUTPUT_DEFINITION_FILE_TYPE = "cemiOutputDefinitionFileType";
```
cah292, line 6: If you're moving this property to this new class, then you may want to remove the old version of it from `CuVendorTestConstants`.

cah292 (changes requested):
Here are my initial thoughts and suggestions regarding the refactored CEMI code.

cah292:
Regarding the failed unit test, I believe you'll need to update `cu-spring-vnd-cemi-supplier-file-test.xml` so that it imports the appropriate CEMI Spring files, now that the CEMI code that was formerly in the Vendor module is now in the CEMI module.

src/main/java/edu/cornell/kfs/cemi/sys/util/CemiCuSqlChunk.java
db483: This is a good point Chad, however, we may want to isolate the CEMI code. We want the CEMI code to take full advantage of the KFS framework but we do not want the CEMI code to impact KFS. This code is meant for data extraction only and it is part of the cu-kfs project only because we wanted to take advantage of the existing batch job functionality and ORM framework.

cah292: That works for me, thanks for explaining.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/CemiVendorDaoOjb.java
cah292: As per today's standup, we agreed on the following changes:

1. Keep the DAOs distinct; don't duplicate methods between them.
2. Rename `CemiVendorDaoJdbc` to `CemiVendorDao`.
3. Make sure the JDBC implementation is named `CemiVendorDaoJdbcImpl` and implements `CemiVendorDao`.
4. Rename `CemiVendorDaoOjb` to `CemiVendorOrmDao`.
5. Rename `CemiVendorDaoOjbImpl` to `CemiVendorOrmDaoOjbImpl` and have it implement `CemiVendorOrmDao`.

src/main/java/edu/cornell/kfs/cemi/sys/CemiKFSConstants.java
nkimble: Moving to CemiBaseConstants constants class.

src/main/java/edu/cornell/kfs/cemi/sys/CemiKFSKeyConstants.java
nkimble: Made requested changes.

src/test/java/edu/cornell/kfs/cemi/vnd/CemiVendorTestConstants.java
nkimble: Removed from CuVendorTestConstants.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/CemiVendorDaoOjb.java
nkimble: Changes made

nkimble (commented):
Also made the change to fix the unit test.

cah292 (commented):
Revised changes look good. I'm wondering whether the ORM/OJB DAO still needs to extend from the Vendor one or if it should be standalone from it, but if you'd rather defer that cleanup to a future date, that would be okay with me.

Waiting for Daniela to perform her follow-up review.

db483:
Updated changes look good. Please create a separate user story to look into whether the ORM/OJB DAO still needs to extend from the Vendor one. Thank you.

PR #1914, opened 2026-04-23
KFSPTS-37856 remove un-needed server references and convert to use central maven repo and github artifacts

jhulslander:
some new exclusions were added as they were libraries we don't need and there were problems being able to get them from Maven.
Please double check that the dependencies aren't needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,3 +1,5 @@
+
+
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
--- a/pom.xml
+++ b/pom.xml
@@ -304,23 +304,20 @@
     </build>
     <repositories>
         <repository>
-            <id>kualico.nexus.financials.releases</id>
-            <name>KualiCo Nexus Financials</name>
-            <url>https://nexus.kuali.co/nexus/content/repositories/financials-repos/</url>
-            <snapshots>
-                <enabled>true</enabled>
-                <updatePolicy>always</updatePolicy>
-            </snapshots>
-            <releases>
-                <enabled>true</enabled>
-            </releases>
+            <id>github</id>
+            <name>GitHub Packages</name>
+            <url>https://maven.pkg.github.com/kualico/financials</url>
+        </repository>
+        <repository>
+            <id>central</id>
+            <url>https://repo.maven.apache.org/maven2</url>
         </repository>
     </repositories>
     <pluginRepositories>
         <pluginRepository>
-            <id>kualico.nexus.financials.releases</id>
-            <name>KualiCo Nexus Financials</name>
-            <url>https://nexus.kuali.co/nexus/content/repositories/financials-repos/</url>
+            <id>github</id>
+            <name>GitHub Packages</name>
+            <url>https://maven.pkg.github.com/kualico/financials</url>
             <snapshots>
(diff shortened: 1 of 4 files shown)


src/test/java/edu/cornell/kfs/module/purap/jaggaer/supplier/xml/SupplierSyncMessageCornellTest.java
cah292, line 1: Did these various Jaggaer unit tests require refactoring? I don't see how their changes correlate to the POM updates. (Unless the two tests were executing in parallel and stepping on each other's directories?)

pom.xml
```diff
@@ -958,6 +864,15 @@
                     <artifactId>log4j</artifactId>
                     <groupId>log4j</groupId>
                 </exclusion>
+                <!-- CORBA libraries unavailable from defunct nexus.kuali.org repo; not needed for KFS -->
```
cah292, line 867: Did you try creating and routing some KFS edocs while running locally? Just wanted to confirm that the JOTM and XAPool functionality is still working with these exclusions in place.

pom.xml
```diff
@@ -1086,6 +1001,15 @@
                     <groupId>javax.xml.bind</groupId>
                     <artifactId>jaxb-api</artifactId>
                 </exclusion>
+                <!-- SDO plugin pulls in commonj.sdo from defunct maven.java.net repo; not needed for KFS -->
```
cah292, line 1004: Since these exclusions impact JAX-WS, you may want to confirm that the Account Verification Web Service is working locally.

cah292 (commented):
I confirmed that I was able to get KFS to build locally with this branch and the settings.xml file changes in place. (I ran out of time today to check whether KFS would launch, though.) I also updated the Confluence instructions to include Windows setup steps.

However, there were a few items I wanted to get clarity on, and it would be helpful to have another Mac developer test out this setup.

src/test/java/edu/cornell/kfs/module/purap/jaggaer/supplier/xml/SupplierSyncMessageCornellTest.java
jhulslander: Yes, when I was validating my changes locally the tests were stepping on each other regularly, so I thought it prudent to adjust the directories in the tests to avoid that.

pom.xml
jhulslander: the account verification webservice works locally

pom.xml
jhulslander: I sometimes run into a weird problem locally when creating  (that was happening before this change)
java.lang.NullPointerException
	at org.kuali.kfs.sys.service.impl.GeneralLedgerPendingEntryServiceImpl.populateOffsetGeneralLedgerPendingEntry(GeneralLedgerPendingEntryServiceImpl.java:434)

can you try a transactional document?

jhulslander: I was able to create a person doc and a few others

src/test/java/edu/cornell/kfs/module/purap/jaggaer/supplier/xml/SupplierSyncMessageCornellTest.java
cah292: Thanks for clarifying. Given that, I'm fine with these tests being updated.

pom.xml
cah292: I was able to successfully submit an IWNT document, though I didn't try other transactional doc types. Perhaps your error was resulting from some missing year/period/date info in KFS, but I'm not entirely certain.

Anyway, the creating and routing of KFS edocs is also working for me.

pom.xml
cah292: That web service also works for me locally.

PR #1915, opened 2026-04-23
KFSPTS-37839 Adjust PREQ auto-approve handling

cah292:
There are PRs for this change in both cu-kfs and nonprod-sql. Please make sure both PRs are good to go before merging.

This PR implements and combines two major changes:

* KFSPTS-37857: Adjusts the PREQ auto-approval action so that it does NOT check if receiving requirements are met and, consequently, does not attempt to change the annotation to a receiving-related one. The auto-approval action will use the regular "Total is below threshold" message instead.
* KFSPTS-37839: Adjusts the code that checks if a PREQ is eligible for auto-approval, to allow for specifying funds or sub-funds that are ineligible for auto-approval. One of the existing not-eligible messages has also been updated to only be written when necessary, to avoid reporting false positives in the logs.


cah292:
As per the parameter renaming from the nonprod-sql PR, I've also committed a code change to rename the parameters on the cu-kfs side.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -466,22 +471,69 @@ protected boolean isPRNCDocument(PaymentRequestDocument paymentRequestDocument)
...
+    private boolean paymentRequestSubFundGroupsAreEligibleForAutoApproval(
+            final PaymentRequestDocument document, final List<PurApAccountingLine> preqAccountingLines) {
+        final Collection<String> ineligibleSubFundGroups = parameterService.getParameterValuesAsString(
+                PaymentRequestDocument.class, CUPurapParameterConstants.SUB_FUND_GROUPS_PREVENTING_AUTO_APPROVE);
+        final boolean isEligible = preqAccountingLines.stream()
+                .map(PurApAccountingLine::getAccount)
+                .map(Account::getSubFundGroupCode)
+                .noneMatch(subFundGroupCode -> ineligibleSubFundGroups.contains(subFundGroupCode));
+        if (!isEligible) {
+            LOG.info(" -- PayReq [{}] skipped due to ineligible Sub-Fund Group.", document.getDocumentNumber());
```
jhulslander: per our norm, I think we want to include the function name.  I suggest the log mentioning that auto approval was prevented rather than saying "ineligible" also.  Would it be worth include the sub fund that prevented auto approval?

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
```diff
@@ -466,22 +471,69 @@ protected boolean isPRNCDocument(PaymentRequestDocument paymentRequestDocument)
...
+            final PaymentRequestDocument document, final List<PurApAccountingLine> preqAccountingLines) {
+        final Collection<String> ineligibleFundGroups = parameterService.getParameterValuesAsString(
+                PaymentRequestDocument.class, CUPurapParameterConstants.FUND_GROUPS_PREVENTING_AUTO_APPROVE);
+        final boolean isEligible = preqAccountingLines.stream()
+                .map(PurApAccountingLine::getAccount)
+                .map(Account::getSubFundGroup)
+                .map(SubFundGroup::getFundGroupCode)
+                .noneMatch(fundGroupCode -> ineligibleFundGroups.contains(fundGroupCode));
+        if (!isEligible) {
+            LOG.info(" -- PayReq [{}] skipped due to ineligible Fund Group.", document.getDocumentNumber());
```
jhulslander: similar thought about logging here as above

jhulslander (changes requested):
Changes look good, I just had a suggesting about clarifying  some of the log statements

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
cah292: The related delivered log statements in `PaymentRequestServiceImpl` don't include the function name as a prefix, so I was trying to have those logs follow that pattern for consistency. (The similar existing log entry in `CuPaymentRequetServiceImpl` does the same.) If you still think that the function prefix should be added, please let me know.

I'll work on rewording the log statement accordingly, and will see if I can update the logic so that the restricted fund/sub-fund can be logged as well.

src/main/java/edu/cornell/kfs/module/purap/document/service/impl/CuPaymentRequestServiceImpl.java
cah292: See my reply in the other comment.

cah292:
I've committed some changes based on the feedback thus far. Please let me know if you still want those prefixes added to the log messages.

jhulslander (approved):
This looks reasonable, and so do that SQL, I shall merge

PR #1918, opened 2026-05-01
KFSPTS-37414 Add sheet table storage

cah292:

This PR replaces the CSV-based sheet data storage with JDBC-based sheet data storage. The classes for the CSV-based storage are still present, but I'm not certain if they fully work with the changes being implemented here.

Notable areas of change:

* Added a "name" attribute to the Output Definition DTO, to help with grouping sheets based on extract file type.
* Added several new Field Definition types to support sensitive field encryption, limited support for other data types, and adding storage-only fields that are not included in the final Excel file. See the comments in `CemiSupplierExtractFileOutputDefinition.xml` for an explanation of the new types.
* Added new metadata classes to help with generating the SQL for the JDBC-based storage.
  * Table and column names will be uppercased, will have whitespace segments replaced with underscores, and will have all other non-word characters removed.
  * Table name format: `CU_CEMI_[Output Definition Name]_[Sheet Name]_T`
  * Column names generally follow the basic conversion outlined earlier, but with the following adjustments:
    * If a column name clashes with a reserved word, then a suffix of `_VAL` will be added. (Only one reserved word is defined for now; see the `RESERVED_WORDS` constant in the `CemiUtils` class.)
    * If a column name clashes with a prior column on the same sheet, then an indexed suffix (`_2`, `_3`, etc.) will be added.
* Added various new services and DAOs to support the new functionality.
  * Also included some services to perform caching of both the Output Definition objects and the new metadata objects. Those services include methods for flushing the caches, but they're not being called yet.
* `CemiOutputDefinitionServiceImpl` now handles the loading of Output Definitions. It expects the Output Definitions to be placed under the following folder structure, where the bracketed items below should be used as arguments in the service method that loads the definition:
  * `classpath:edu/cornell/kfs/cemi/[Module Name]/batch/[Bare Filename].xml`
* Added a helper DTO base class for storing a numeric index and the job execution date-time string. Together, these can be used to filter table rows based on extraction run, and to also maintain the insertion order of the rows so that they can be written to the Excel file in the same order.
  * The various Supplier DTOs and the Supplier Extract Job have been updated to populate and utilize these new fields.
* Updated the `WriteCemiSupplierFileTest` class so that, with only a few minor changes, you can use the unit test to auto-generate the CREATE TABLE SQL for all the sheets in a particular Output Definition. The file's comments contain more details on how to trigger that functionality.
  * Manual changes may still be needed, since the functionality uses a hard-coded max length for all of the VARCHAR2 columns.

NOTE: I was originally planning to have the new DAO use the `JdbcTemplate.queryForStream()` method to retrieve the sheet table contents, but doing so resulted in unexpected "closed connection" errors at runtime. I ultimately used a different strategy that works around the issue while still being memory-efficient, though it did require overriding one of the base File Appender methods to avoid using the Stream-based approach. (Some of the code from the failed approach, such as the `CemiDaoBaseJdbc` class, has not been cleaned up yet; such cleanup should be performed either here or in a follow-up user story.)

NOTE: Because the work for consolidating addresses/phones/emails was already performed in a prior PR, the results of running the table-based extract will NOT completely line up with the results from March's extract. If we need to get the CSV data from the March run into the new tables, that would likely be a manual effort, unless we want to add a feature for importing the CSV data into the tables.

### DIFF
--- a/src/main/java/edu/cornell/kfs/cemi/sys/CemiBaseConstants.java
+++ b/src/main/java/edu/cornell/kfs/cemi/sys/CemiBaseConstants.java
@@ -1,14 +1,38 @@
 package edu.cornell.kfs.cemi.sys;
 
+import java.sql.Types;
+
+import org.kuali.kfs.sys.KFSConstants;
+
 public final class CemiBaseConstants {
 
     public static final String CEMI_ENVIRONMENT_LANE_NAME = "kfs-cemi";
 
     public static final String CEMI_OUTPUT_DEFINITION_FILE_TYPE_IDENTIFIER = "cemiOutputDefinitionFileType";
 
+    public static final String CEMI_SCHEMA = "CEMI";
+    public static final String CEMI_SCHEMA_PREFIX = CEMI_SCHEMA + KFSConstants.DELIMITER;
+    public static final String VARCHAR2_TYPE = "VARCHAR2";
+    public static final String NUMBER_TYPE = "NUMBER";
+    public static final int DEFAULT_SHEET_COLUMN_SIZE = 200;
+    public static final int DEFAULT_SHEET_COLUMN_PRECISION = 10;
+    public static final int SHEET_TABLE_BATCH_SIZE = 200;
+
     public enum CemiFieldDefinitionType {
-        STATIC,
-        STRING;
+        STATIC(Types.VARCHAR, true),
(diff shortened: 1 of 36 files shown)


db483:
Changes look good overall but I am wondering if we could go with a simpler solution that takes advantage of the existing KFS framework and use ojb for storing the data.

PR #1921, opened 2026-05-08
KFSPTS-37585 Payflow Authentication Simplification

ajd299:


jhulslander (approved):
This looks right to me.
@cah292  Can you take a look at this too.  If you think it looks good, I think it is safe to merge

cah292 (commented):
I noticed that the removed filter is explicitly setting the "Access-Control-Allow-Methods" header to "GET". Do you need this setup in the new filter? If so, then add the following to the new Spring bean:

`p:accessControlAllowMethods="GET"`

Other than that, the changes look good to me.

jhulslander (changes requested):
Please see previously noted changes

jhulslander (approved):
this looks reasonable, I shall merge

PR #1924, opened 2026-05-27
KFSPTS-37938 Add CEMI Remit To Supplier Extract

cah292:

This implementation writes and reads the sheet data to/from the database without creating temporary CSV files. It also uses a "factory" class to create the sheet business object, rather than using a plain DTO.

Some hooks have been added to allow for data masking if we end up needing to add bank account info to the connection ID, but it's not in use yet since we only have up to 1 bank account per supplier. If that changes, then it should be fairly simple to adjust the code to add and mask the bank info.

Some (but not all) of the Supplier Extract sheet business objects have been added, to help with retrieving the Supplier Extract data that's needed for generating the Supplier Remit To data. Note that the Supplier Bank Account BO is not actively being used yet, so if needed I could exclude that one for now.


src/main/java/edu/cornell/kfs/cemi/sys/batch/businessobject/CemiIndexedBusinessObjectBase.java
```diff
@@ -0,0 +1,28 @@
+package edu.cornell.kfs.cemi.sys.batch.businessobject;
+
+import org.kuali.kfs.krad.bo.PersistableBusinessObjectBase;
+
+public abstract class CemiIndexedBusinessObjectBase extends PersistableBusinessObjectBase {
+
+    private static final long serialVersionUID = 2446925945540509482L;
+
+    private String jobRunDate;
+    private Long rowIndex;
```
nkimble: Per our offline discussion, this attribute will be renamed to jobRunRowIndex

src/main/java/edu/cornell/kfs/cemi/vnd/batch/businessobject/CemiSupplierBankAccountBo.java
```diff
@@ -0,0 +1,397 @@
+package edu.cornell.kfs.cemi.vnd.batch.businessobject;
+
+import edu.cornell.kfs.cemi.sys.batch.businessobject.CemiIndexedBusinessObjectBase;
+
+public class CemiSupplierBankAccountBo extends CemiIndexedBusinessObjectBase {
```
nkimble: You indicated that this class is not being actively used by these changes.
I would lean toward not introducing the class to the code base until it is truly being used.

src/main/java/edu/cornell/kfs/cemi/sys/util/CemiUtils.java
```diff
@@ -101,4 +117,26 @@ public static <T> List<T> createListOfEmptyValues(final int size, final T emptyV
...
+        try (
+            final InputStream inputStream = CuCoreUtilities.getResourceAsStream(
+                    CemiBaseConstants.CEMI_OUTPUT_DEFINITION_FILE_PATH_PREFIX + subPath);
+        ) {
+            final byte[] fileContents = IOUtils.toByteArray(inputStream);
+            return cemiOutputDefinitionFileType.parse(fileContents);
+        }
+    }
+
+    public static boolean isFormattedAsValidIdentifier(final String name) {
```
nkimble: I am not seeing where this method is being used. Which would also imply that the private constant is not being used.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/businessobject/CemiRemitToSupplierBo.java
```diff
@@ -0,0 +1,154 @@
+package edu.cornell.kfs.cemi.vnd.batch.businessobject;
+
+import edu.cornell.kfs.cemi.sys.batch.businessobject.CemiIndexedBusinessObjectBase;
+
+public class CemiRemitToSupplierBo extends CemiIndexedBusinessObjectBase {
+
```
nkimble, line 6: Asking: Should all of the business objects have an attribute that 1-1 corresponds to the database table fields?  Job run date and job run index attributes are not present.

src/main/java/edu/cornell/kfs/cemi/vnd/util/CemiVendorUtils.java
```diff
@@ -59,7 +59,7 @@ private static boolean isPurchaseOrderVendor(final String vendorTypeCode) {
         return StringUtils.equals(vendorTypeCode, VendorConstants.VendorTypes.PURCHASE_ORDER);
     }
 
-    private static boolean addressTypeIsActiveAndIsDefaultAndMatches(
+    public static boolean addressTypeIsActiveAndIsDefaultAndMatches(
```
nkimble, line 64: With visibility as private, any changes to this method would be scoped to where isPrimaryVendorAddress was used in the original extraction that method isPrimaryVendorAddress was created for.

Do the two extractions (the original one the routine was written for and this new one) really work in the same manner where both extractions can accept the same modification in the future to this private method since this will now be a public utility class routine that could be utilized by any vendor extraction?

src/main/resources/edu/cornell/kfs/cemi/vnd/cu-ojb-cemi-vnd.xml
```diff
@@ -0,0 +1,160 @@
...
+        <field-descriptor name="jobRunDate" column="EXTR_FILE_RUNDATE" jdbc-type="VARCHAR" primarykey="true" index="true"/>
+        <field-descriptor name="rowIndex" column="ROW_INDEX" jdbc-type="BIGINT" primarykey="true" index="true"/>
+        <field-descriptor name="supplierId" column="SUPPLIER_ID" jdbc-type="VARCHAR"/>
+        <field-descriptor name="supplierConnectionId" column="SUPPLIER_CONNECTION_ID" jdbc-type="VARCHAR"/>
+        <field-descriptor name="supplierConnectionName" column="SUPPLIER_CONNECTION_NAME" jdbc-type="VARCHAR"/>
+        <field-descriptor name="defaultPaymentType" column="DEFAULT_PAYMENT_TYPE" jdbc-type="VARCHAR"/>
+        <field-descriptor name="acceptedPaymentType1" column="ACCEPTED_PAYMENT_TYPE_1" jdbc-type="VARCHAR"/>
+        <field-descriptor name="acceptedPaymentType2" column="ACCEPTED_PAYMENT_TYPE_2" jdbc-type="VARCHAR"/>
+        <field-descriptor name="acceptedPaymentType3" column="ACCEPTED_PAYMENT_TYPE_3" jdbc-type="VARCHAR"/>
+        <field-descriptor name="settlementBankAccount" column="SETTLEMENT_BANK_ACCOUNT" jdbc-type="VARCHAR"/>
```
nkimble, line 100: Should the encryption field conversion be added?

nkimble (commented):
Initial review has been completed. Provided some items to consider.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/businessobject/CemiSupplierBankAccountBo.java
cah292: Ok, will remove that class.

src/main/java/edu/cornell/kfs/cemi/sys/util/CemiUtils.java
cah292: Will remove this method and its constant. It was left over from a feature that I ultimately abandoned.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/businessobject/CemiRemitToSupplierBo.java
cah292: These sheet business objects should already have 1-1 mappings between object field and database table field (as well as between object field and sheet column). The fields for the job run date and the row index are inherited from the `CemiIndexedBusinessObjectBase` parent class, so that we don't have to duplicate them on each sheet BO.

src/main/java/edu/cornell/kfs/cemi/vnd/util/CemiVendorUtils.java
cah292: I think it's safe for both extracts to have the same functionality for this particular case. This method uses KFS VendorAddress objects, not Workday sheet BOs, for its derivations, and it merely checks whether the VendorAddress is an active and default one with the specified address type.

src/main/resources/edu/cornell/kfs/cemi/vnd/cu-ojb-cemi-vnd.xml
cah292: This field (not currently populated) only contains the ID of the related bank account row from the Supplier extract. It does not contain the bank account number itself, and we're currently not placing any bank account number substrings into the ID.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiRemitToSupplierExtractServiceImpl.java
```diff
@@ -0,0 +1,243 @@
...
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @Override
+    public void populateListOfInScopeAddresses() {
+        LOG.info("populateListOfInScopeAddresses, Querying and storing the list of extractable addresses...");
+        cemiRemitToSupplierDao.queryAndStoreAddressIdsForRemitToSupplierExtract();
+    }
+
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    @Override
+    public void generateIntermediateRemitToSupplierExtractData(final LocalDateTime jobRunDate) {
```
db483, line 95: generateRemitToSupplierExtractData is only called here and that is all this method does. Do we need two separate methods?

db483 (approved):
I had one minor question but changes look good and I will merge it so we can move forward with the extract.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiRemitToSupplierExtractServiceImpl.java
cah292: Good point, I agree that those two methods could be consolidated. We could perform that cleanup in a future user story, given that this PR has already been merged.

PR #1925, opened 2026-06-09
KFSPTS-37939 Add Order From Supplier Extract

cah292:

This PR adds the code needed for generating the Order From Supplier file. Note that it relies on some address joining logic that is somewhat complex and could potentially be simplified later (unless it's needed now); see my related note in the nonprod-sql PR.

Since we were told to only create Order From rows when a particular supplier is expected to have 2 or more such connections, this extract will skip cases where a supplier would have had only one such connection. Furthermore, it forces punchout/CXML suppliers to have exactly 2 connections (unless their email data is missing or misconfigured).

A portion of the code that's similar to what was used in the Remit To job has been added to the main Vendor ORM DAO, so that we could potentially cut down on duplicate code when/if we refactor the Remit To job again.

As another improvement to help cut down on duplicate code, this PR introduces a new base class for creating a job's main extraction service. We could potentially use that as part of the future refactoring work on the batch jobs (and for simplifying the creation of new jobs).


src/main/java/edu/cornell/kfs/cemi/sys/batch/service/impl/CemiDataExtractServiceBase.java
```diff
@@ -0,0 +1,152 @@
...
+import edu.cornell.kfs.cemi.sys.CemiBaseConstants.FileExtensions;
+import edu.cornell.kfs.cemi.sys.batch.CemiOutputDefinitionFileType;
+import edu.cornell.kfs.cemi.sys.batch.service.CemiFileAppenderService;
+import edu.cornell.kfs.cemi.sys.batch.xml.CemiOutputDefinition;
+import edu.cornell.kfs.cemi.sys.util.CemiUtils;
+import edu.cornell.kfs.cemi.vnd.CemiVendorParameterConstants;
+import edu.cornell.kfs.core.api.util.CuCoreUtilities;
+import edu.cornell.kfs.sys.CUKFSConstants;
+
+public abstract class CemiDataExtractServiceBase {
```
nkimble, line 31: I am a little surprised that a new base class is being introduced at the same time analysis is being done to determine patterns in the code.
Code changes in this PR due to this are not an option at this because of data extraction file due dates. The analysis for this will have to be redone.

src/main/java/edu/cornell/kfs/cemi/sys/batch/service/impl/CemiDataExtractServiceBase.java
```diff
@@ -0,0 +1,152 @@
...
+    protected CemiOutputDefinitionFileType cemiOutputDefinitionFileType;
+    protected ParameterService parameterService;
+
+    protected CemiDataExtractServiceBase(final Environment environment) {
+        Validate.notNull(environment, "environment cannot be null");
+        this.environment = environment;
+    }
+
+    protected boolean isCemiEnvironment() {
+        return Strings.CI.equals(environment.getLane(), CemiBaseConstants.CEMI_ENVIRONMENT_LANE_NAME);
```
nkimble, line 49: Notes for the analysis not requesting change: new way of doing String comparisons that will need to be made consistent in other classes.

src/main/java/edu/cornell/kfs/cemi/sys/util/CemiUtils.java
```diff
@@ -50,7 +51,11 @@ public static String convertToBooleanValueForFileExtract(final boolean value) {
         return Boolean.toString(value)
                 .toUpperCase(Locale.US);
     }
-    
+
+    public static String convertToBooleanValueForEIBFileExtract(final boolean value) {
```
nkimble, line 55: I do not think that it is appropriate to have a conversion routine specific to a certain extract file in a common utilities package.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierBoFactory.java
```diff
@@ -0,0 +1,190 @@
...
+        return this;
+    }
+
+    public CemiOrderFromSupplierBo createCemiOrderFromSupplierBo() {
+        Validate.validState(supplier != null, "A supplier was not defined");
+        Validate.validState(supplierEmailRow != null, "A supplier email record was not defined");
+        Validate.validState(!isPunchoutConnection || isPunchoutSupplier,
+                "Entries marked as punchout connections must also be marked as having a punchout supplier");
+        Validate.validState(!isPunchoutConnection || isFirstRowForSupplier,
+                "Entries marked as punchout connections must also be marked as being the supplier's first connection");
```
nkimble, line 85: Notes for analysis not requesting changes: Need to determine better was to identify the actual data items causing the errors during our processing so that these messages don't become just noise in the logs.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierBoFactory.java
```diff
@@ -0,0 +1,190 @@
...
+            emailAddress = supplierEmailRow.getEmailAddress1();
+        } else if (Strings.CI.equals(emailFromKfsVendorAddress, supplierEmailRow.getEmailAddress2())) {
+            emailRowId = Integer.toString(1);
+            emailId = supplierEmailRow.getEmailId2();
+            emailAddress = supplierEmailRow.getEmailAddress2();
+        } else if (Strings.CI.equals(emailFromKfsVendorAddress, supplierEmailRow.getEmailAddress3())) {
+            emailRowId = Integer.toString(1);
+            emailId = supplierEmailRow.getEmailId3();
+            emailAddress = supplierEmailRow.getEmailAddress3();
+        } else {
```
nkimble, line 148: Is throwing an exception and failing the batch job really what we want to do here?

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
```diff
@@ -0,0 +1,243 @@
...
+                        + "eligible to be included in the Order From Supplier extract. This Supplier will be "
+                        + "excluded from the extract altogether.", supplier.getSupplierId());
+                addressesForOutput = List.of();
+            } else {
+                // Just duplicate the first address; the loop below will handle the punchout logic accordingly.
+                final Pair<String, CemiSupplierAddressBo> firstAddressWithUniqueEmail = addressesWithUniqueEmails.get(0);
+                addressesForOutput = List.of(firstAddressWithUniqueEmail, firstAddressWithUniqueEmail);
+            }
+        } else if (addressesWithUniqueEmails.size() <= 1) {
+            LOG.debug("createAndStoreOrderFromSupplierRows, Supplier {} only has one unique email across its "
```
nkimble, line 158: Asking: Would we need to know this supplier id for any reason (data fixing or data clean-up by the users) where this should be a warning message instead of debug?

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
```diff
@@ -0,0 +1,243 @@
...
+
+        int connectionRowId = 1;
+
+        for (final Pair<String, CemiSupplierAddressBo> emailAndAddressPair : addressesForOutput) {
+            final boolean isFirstRowForSupplier = (connectionRowId == 1);
+            final boolean isPunchoutConnection = isPunchoutSupplier && isFirstRowForSupplier;
+            final CemiOrderFromSupplierBo orderFromSupplierRow = new CemiOrderFromSupplierBoFactory()
+                    .withSupplier(supplier)
+                    .withSupplierEmailRow(emailRow)
+                    .withEmailFromKfsVendorAddress(emailAndAddressPair.getLeft())
```
nkimble, line 175: getLeft could return null. 
It looks like the factory deals with that possibility with the Strings.CI.equals conditionals. 
Is there anywhere else we need to be concerned about that value being null?

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
```diff
@@ -0,0 +1,243 @@
...
+        if (addressGroup == null) {
+            LOG.warn("getEmailAddress, UNEXPECTED: No KFS Vendor Addresses found for Supplier Address: {}", addressKey);
+            return KFSConstants.EMPTY_STRING;
+        }
+
+        return addressGroup.stream()
+                .map(VendorAddress::getVendorAddressEmailAddress)
+                .filter(StringUtils::isNotBlank)
+                .findFirst()
+                .orElse(KFSConstants.EMPTY_STRING);
```
nkimble, line 259: There is a lot going on in this single statement. 
Not suggesting changes now but we may need to unpack this in the future to pin-point failures or log data generating problems.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/impl/CemiOrderFromSupplierDaoJdbcImpl.java
```diff
@@ -0,0 +1,207 @@
...
+        return CemiUtils.generateConcatenatedKey(
+                vendorAddress.getVendorLine1Address(), vendorAddress.getVendorLine2Address(),
+                vendorAddress.getVendorCityName(), vendorAddress.getVendorStateCode(),
+                vendorAddress.getVendorZipCode(), vendorAddress.getVendorCountryCode());
+    }
+
+    private <T> void storeAsListOfAddressLinks(final Class<T> addressClass, final CuSqlQuery query,
+            final Iterator<T> addressIterator) {
+        int count = 0;
+        final List<T> cachedAddresses = new ArrayList<>(ADDRESS_BATCH_SIZE);
```
nkimble: I am not sure if others will but I find the use of "cache" in this local variable name confusing. I immediately think it is referring to the Redis cache.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/impl/CemiOrderFromSupplierDaoJdbcImpl.java
```diff
@@ -0,0 +1,207 @@
...
+                .append("SELECT COUNT(1) FROM KFS.CU_CEMI_MAPPING_SPLR_VNDR_EXTR_FILE_T EXT ")
+                .append("JOIN KFS.CU_CEMI_EXTR_ORD_FRM_SUPP_B2B_VNDR_T B2B ")
+                        .append("ON EXT.VNDR_HDR_GNRTD_ID = B2B.VNDR_HDR_GNRTD_ID ")
+                        .append("AND EXT.VNDR_DTL_ASND_ID = B2B.VNDR_DTL_ASND_ID ")
+                .append("WHERE EXT.WKDY_SPLR_ID = ").appendAsParameter(supplierId)
+                .append(" AND EXT.EXTR_FILE_RUNDATE = ").appendAsParameter(supplierJobRunDate)
+                .toQuery();
+
+        final int punchoutSupplierCount = queryForResults(query, resultSet -> {
+            Validate.validState(resultSet.next(), "Punchout supplier count query didn't return any count results");
```
nkimble, line 201: Are we going to be able to troubleshoot when this validateState method throw an unhandled exception?

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/impl/CemiRemitToSupplierOrmDaoOjbImpl.java
```diff
@@ -53,6 +53,7 @@ private boolean shouldUseLessDataDuringCemiDevelopment() {
                 CemiBaseConstants.CU_CEMI_DEVELOPMENT_USE_SMALLER_DATA_SET_KEY);
     }
  
+    // NOTE: We should eventually replace this method with the equivalent one from CemiVendorOrmDaoOjbImpl instead.
```
nkimble, line 56: This is Remit To Supplier being updated by these changes. 
Okay to leave in with these changes as it is only a comment.

src/main/java/edu/cornell/kfs/cemi/vnd/util/CemiVendorUtils.java
```diff
@@ -82,4 +88,45 @@ public static String generateAddressKey(final CemiSupplierAddressBo supplierAddr
...
+        final Comparator<VendorAddress> vendorAddressComparator = getVendorAddressPrecedenceComparator(
+                addressTypeWithHighestPrecedence);
+        for (final List<VendorAddress> subGroup : groupedVendorAddresses.values()) {
+            Collections.sort(subGroup, vendorAddressComparator);
+        }
+
+        return groupedVendorAddresses;
+    }
+
+    private static Comparator<VendorAddress> getVendorAddressPrecedenceComparator(
```
nkimble, line 111: Noting that this is an area that could prove challenging to troubleshoot if any bad data is encountered. There is a lot going on here that would need to be unpacked. Not requesting changes at this time due to deliverable dates.

src/main/resources/edu/cornell/kfs/cemi/cu-ojb-cemi.xml
```diff
@@ -1,5 +1,7 @@
 <descriptor-repository version="1.0">
```
nkimble, line 1: Should this file be included with these changes?

nkimble (changes requested):
There are a few items that may need to be cleaned up. 
I have noted were the comments are just notes for the pattern analysis or where changes are not feasible at this time due to extraction due dates.

src/main/java/edu/cornell/kfs/cemi/sys/batch/service/impl/CemiDataExtractServiceBase.java
cah292: Sorry about dropping this new base file in at the last minute. I wanted to get this in (or at least a discussion of it) prior to the completion of the data patterns analysis, given the amount of duplicated code in the extraction services.

src/main/java/edu/cornell/kfs/cemi/sys/util/CemiUtils.java
cah292: `CemiUtils` already has a `convertToBooleanValueForFileExtract()` method, but it's meant for handling boolean values in CloverDX workbooks. Since EIB workbooks seem to be using Y/N instead of TRUE/FALSE for booleans, I added this new boolean method here, so that it can be shared amongst the extracts that deal with EIB workbooks.

We could rename `convertToBooleanValueForFileExtract()` to `convertToBooleanValueForCloverDXFileExtract()` if you think that would help avoid confusion, though such a change might need to be deferred to a follow-up ticket.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierBoFactory.java
cah292: At this point in the processing, the upstream code from `CemiOrderFromSupplierDataBuilderDefaultImpl` should have already verified that the current record is either marked as a punchout/CXML connection or has an email that exists in the related Supplier Email record. That's why I configured this spot to treat a lack of email or a lack of punchout as an exception.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
cah292: Putting this statement at WARN level would add hundreds (if not thousands) of extra lines to the report logs, since I had trouble filtering out single-email results in the source queries and ended up doing most of that filtering at the code level instead. I was hoping the functional validation would help identify any suppliers that should have had Order From connections but didn't.

I could write out these skipped-supplier results to a separate report file if you think that would be helpful.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
cah292: The code that creates this particular `Pair` (in `getUniqueEmailsForAddresses()`) already ensures that the email on the left-hand side is non-null. I don't anticipate any areas needing extra null safety for this, unless you've found such a spot already.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/impl/CemiOrderFromSupplierDaoJdbcImpl.java
cah292: Ok, I'll rename it something like `batchAddresses` or `bulkAddresses` to avoid confusion.

src/main/java/edu/cornell/kfs/cemi/vnd/dataaccess/impl/CemiOrderFromSupplierDaoJdbcImpl.java
cah292: The upstream try/catch block in `CuSqlQueryPlatformAwareDaoBaseJdbc` should automatically handle logging and rethrowing any runtime exceptions that occur. As for the `resultSet.next()` call, the JDBC Template should automatically wrap/rethrow SQL exceptions as runtime exceptions accordingly.

src/main/resources/edu/cornell/kfs/cemi/cu-ojb-cemi.xml
cah292: Good catch, I'll undo the changes to this file since they're not needed for the extract.

src/main/java/edu/cornell/kfs/cemi/vnd/batch/service/impl/CemiOrderFromSupplierDataBuilderDefaultImpl.java
cah292: I added a simple feature for writing the skipped suppliers to a plain text file.

db483 (approved):
Thank you for getting this ready Chad! I agree with Nancy's concerns and we should address those in the follow up work on this code. Given that we are so close to the deadline, I will approve this so that you have the time to generate and deliver the extract.

As we think about this work going forward, given the tight deadlines and that we are currently interviewing contractors to help with the extracts, we will look at opportunities to make the code easy and quick to follow by a developer with no KFS experience (and possibly no experience with the latest Java features). The goal is for a new developer to be able to quickly get into the code base and make small adjustments to fix bugs and build new extracts rapidly following similar patterns as the existing ones. We'll work on this together next.

PR #1928, opened 2026-06-11
KFSPTS-38144 Payflow PREQ AP Approver Edit Address

ajd299:


src/main/java/edu/cornell/kfs/module/purap/document/authorization/CuPaymentRequestDocumentAuthorizer.java
```diff
@@ -0,0 +1,64 @@
...
+import org.kuali.kfs.module.purap.PaymentRequestStatuses;
+import org.kuali.kfs.module.purap.PurapAuthorizationConstants;
+import org.kuali.kfs.module.purap.document.PaymentRequestDocument;
+import org.kuali.kfs.module.purap.document.authorization.PurchasingAccountsPayableTransactionalDocumentAuthorizerBase;
+
+/**
+ * CU Customization: KFSPTS-38144 Custom authorizer for PaymentRequestDocument that allows AP Approver to edit
+ * vendor address fields when the document is in "Awaiting AP Review" status.
+ */
+public class CuPaymentRequestDocumentAuthorizer extends PurchasingAccountsPayableTransactionalDocumentAuthorizerBase {
```
cah292: I would suggest updating this subclass to extend from `PaymentRequestDocumentAuthorizer` instead, and modify your override of `getDocumentActions()` to **remove** the `CAN_EDIT_VENDOR_ADDRESS` action if it's currently present but both of the following apply:

1. The document is not in INITIATED or SAVED status.
2. `canEditVendorAddressInApReview()` returns false. (Make sure this method returns false if the document is not in ENROUTE status.)

src/main/webapp/WEB-INF/tags/module/purap/vendor.tag
```diff
@@ -39,8 +39,7 @@
 <c:set var="restrictFullEntry" value="${(not empty KualiForm.editingMode['restrictFullEntry'])}" />
 <c:set var="tabindexOverrideBase" value="30" />
 <%-- CU Customization: Update value to include FINP-11621 fix. --%>
-<c:set var="canEditVendorAddress" value="${not isPaymentRequest or not empty
-KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}" />
+<c:set var="canEditVendorAddress" value="${not empty KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}" />
```
cah292: This `canEditVendorAddress` variable should be changed back to its prior value. `vendor.tag` is used by multiple PURAP documents, so we don't want to mess up the edit-vendor-address logic on REQS, PO and CM docs.

You may want to add a separate `canEditVendorAddressForPREQ` variable, since that would help with some of my other suggested changes.

src/main/webapp/WEB-INF/tags/module/purap/vendor.tag
cah292: I would suggest adding `and not canEditVendorAddressForPREQ` to the part that calculates this value, since `readOnlyForPREQ` only seems to be used for controlling editability of the address fields.

src/main/webapp/WEB-INF/tags/module/purap/vendor.tag
```diff
@@ -116,10 +115,10 @@ KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}"
                 <td class="datacell" width="25%">
                     <kul:htmlControlAttribute 
                     	attributeEntry="${documentAttributes.vendorCityName}" 
-                    	property="document.vendorCityName" 
-                    	readOnly="${(readOnlyForPREQ) or not (fullEntryMode or amendmentEntry) or
-                    	displayCreditMemoFields or (lockB2BEntry and (displayRequisitionFields or
-                    	displayPurchaseOrderFields)) or not canEditVendorAddress}"
+                    	property="document.vendorCityName"
+                        readOnly="${displayCreditMemoFields
```
cah292: I would recommend changing these `readOnly` calculations back to their prior values, but with one small adjustment:

Replace this: `not (fullEntryMode or amendmentEntry)`

With this: `not (fullEntryMode or amendmentEntry or canEditVendorAddressForPREQ)`

src/main/java/edu/cornell/kfs/module/purap/document/authorization/CuPaymentRequestDocumentAuthorizer.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * The Kuali Financial System, a comprehensive financial management system for higher education.
```
cah292: We normally don't include these headers on Cornell-exclusive KFS files.

src/main/java/edu/cornell/kfs/module/purap/document/authorization/CuPaymentRequestDocumentAuthorizer.java
```diff
@@ -0,0 +1,73 @@
...
+
+        final PaymentRequestDocument preqDocument = (PaymentRequestDocument) document;
+        final WorkflowDocument workflowDocument = preqDocument.getDocumentHeader().getWorkflowDocument();
+
+        final boolean userCanEditVendorAddressInApReview = canEditVendorAddressInApReview(preqDocument);
+
+        if (!workflowDocument.isInitiated() && !workflowDocument.isSaved() && !userCanEditVendorAddressInApReview) {
+            documentActionsToReturn.remove(PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS);
+        } else if (userCanEditVendorAddressInApReview) {
+            documentActionsToReturn.add(PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS);
```
cah292: Can the `else if` block that adds the edit-vendor-address action be removed? The superclass should already be adding the edit-vendor-address action if the current user has the appropriate permission(s). Could a role/permission change be used instead of the `else if` portion here?

src/main/webapp/WEB-INF/tags/module/purap/vendor.tag
```diff
@@ -42,9 +42,12 @@
 <c:set var="canEditVendorAddress" value="${not isPaymentRequest or not empty
 KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}" />
 
+<%-- CU Customization: KFSPTS-38144 Add separate variable for PREQ vendor address editing. --%>
+<c:set var="canEditVendorAddressForPREQ" value="${not empty KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}" />
```
cah292: To ensure that this variable is only true for PREQs, please update the condition to also check for a true value from the `displayPaymentRequestFields` variable.

src/main/webapp/WEB-INF/tags/module/purap/vendor.tag
```diff
@@ -219,7 +222,9 @@ KualiForm.documentActions[PurapAuthorizationConstants.CAN_EDIT_VENDOR_ADDRESS]}"
                 <td class="datacell">
                    <kul:htmlControlAttribute 
                     	attributeEntry="${documentAttributes.vendorAttentionName}" property="document.vendorAttentionName" 
-                    	readOnly="${(readOnlyForPREQ) or not (fullEntryMode or amendmentEntry) or displayCreditMemoFields or (lockB2BEntry and (displayRequisitionFields or displayPurchaseOrderFields))}" tabindexOverride="${tabindexOverrideBase + 0}"/>
+                    	readOnly="${(readOnlyForPREQ) or not (fullEntryMode or amendmentEntry or canEditVendorAddressForPREQ) or
+                    	displayCreditMemoFields or (lockB2BEntry and (displayRequisitionFields or displayPurchaseOrderFields))}"
```
cah292, line 226: I just noticed that the version of this condition on the `vendorAttentionName` is a little different, where it omits the `or not canEditVendorAddress` check. (It's also set up that way in base code.) We'll need to make sure the functionals verify that this field's editability or non-editability is behaving as expected.

